PR #55, opened 2014-12-14
update version Atom

gutierri:


gutierri:
@ntwb The URL has the version of the release, by as much as do a nightly manifest?

ex: github.com/atom/atom/releases/download/ ***v0.165.0*** /atom-windows.zip

PR #72, opened 2015-02-28
Fix rust-nightly urls

deevus:
Urls were 404ing as the filenames were changed.


lukesampson:
Sweet, thanks :smile:

PR #76, opened 2015-03-13
Update pandoc

moigagoo:
I couldn't find the hash for this release, so I had to remove the hash field. According to the docs, it is optional.

### DIFF
--- a/pandoc.json
+++ b/pandoc.json
@@ -1,10 +1,9 @@
 {
 	"homepage": "http://johnmacfarlane.net/pandoc/",
-	"version": "1.12.2.1",
+	"version": "1.13.2",
 	"license": "GPL",
 	"depends": "latex",
-	"url": "https://pandoc.googlecode.com/files/pandoc-1.12.2.1.msi",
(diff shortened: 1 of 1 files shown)


lukesampson:
Scoop will tell you the hash if you test the install from this manifest (i.e. when the hash has been removed or left out). I know it's not as good as an official hash, but it's better than no hash because it will become obvious that something's wrong if the hash check starts failing.

You can quickly test an install with `scoop install local/path/to/manifest.json`, and it'll give you something like this:

```powershell
installing pandoc (1.13.2)
downloading https://github.com/jgm/pandoc/releases/download/1.13.2/pandoc-1.13.2
-windows.msi...done
warning: no hash in manifest. sha256 is:
4047ae56658b6a3b1cea2ab0c2469687cb2c89715dc136812ae33cde930ce32d
extracting...done
creating shim for pandoc
pandoc (1.13.2) was installed successfully!
```

So then you can use that SHA256 hash in the manifest.

Do you want to add that hash to the pull request?

moigagoo:
> Do you want to add that hash to the pull request?
Absolutely! Thanks for the tip!

I've added the hash.

lukesampson:
Sweet :+1:

PR #83, opened 2015-04-10
Update and correct conemu manifest

scottwillmoore:
Been playing around with conemu lately. If you havn't tried it as a replacement to the default windows powershell/cmd console I would *highly* recommend it. It has many features which I find increase my productivity heaps! It takes a bit of customising and tinkering to figure everything out.

- Easily resize windows
- Copy-paste support
- Highlighting text
- Tabs and split panes
- Borderless, maximised windows
- Better font rendering
- Easily hide and show with a keypress

conemu

### DIFF
--- a/conemu.json
+++ b/conemu.json
@@ -1,7 +1,11 @@
 {
-	"homepage": "https://code.google.com/p/conemu-maximus5",
-	"version": "131010",
-	"url": "https://conemu-maximus5.googlecode.com/files/ConEmuPack.131010.7z",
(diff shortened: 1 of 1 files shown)


ntwb:
This is freaky, was talking about just this thing a couple of hours ago...

scottwillmoore:
I was amazed when I started playing around with it. I don't know how I used to old console for so long! It also reminds me of my Quake style terminal that I enjoy using when on Linux, that drops down from the top for really quick console access.

PR #84, opened 2015-04-10
Add sublime text editor

scottwillmoore:
Been thinking of adding Sublime Text 3 to scoop for a while. It isn't exactly open source, however this bucket is meant for non open source apps. The beta as far as I am aware is free, but if you have a license you can enter it in which disables *free version* notifications that pop-up every time you save.

The app installs fine, but I wouldn't recommend using it through scoop if you already have it installed on your system. It is the portable version which has some implications. Your data (packages, etc) will be stored in the sublime-text folder and everytime that scoop is updated, your packages will be removed with the old version. I don't know whether it is possible to change this to using AppData. Perhaps later it can magically extract the non-portable executable and use that instead of the portable zip version.

Another issue that I havn't tested but I assume will occur is file association. I have nearly every text based file assoiciated with sublime text so that when I open that file it will open with sublime text. I assume if sublime is installed with scoop this may not work properly anymore. If it does work, then when it is updated it may break the association due to the movement of directory (or maybe not if it is referenced to the shim).

Anyway, thoughts? I probably won't use it unless some of these issues are resolved, but others might want to try it out, maybe they will have solutions? Would be nice to have in scoop. :smile:

### DIFF
--- a/sublime-text.json
+++ b/sublime-text.json
@@ -11,5 +11,6 @@
 			"hash": "382a1fb1c19869e2a6f8501fb4d03552252edaf2fb7917f8bd99bba8b9d32a90"
 		}
 	},
-	"bin": "subl.exe"
+	"bin": "subl.exe",
+	"notes": "This is a beta version of Sublime Text 3. For more information please see http://www.sublimetext.com/3"
 }
--- a/sublime-text.json
+++ b/sublime-text.json
@@ -1,0 +1,15 @@
+{
+	"homepage": "http://www.sublimetext.com/3",
+	"version": "b3083",
+	"architecture": {
+		"64bit": {
+			"url": "http://c758482.r82.cf2.rackcdn.com/Sublime%20Text%20Build%203083%20x64.zip",
+			"hash": "7b1665efcb64933530347244c99482e878a70bc90c59fe5ff0d0ec9862d5ed26"
+		},
+		"32bit": {
+			"url": "http://c758482.r82.cf2.rackcdn.com/Sublime%20Text%20Build%203083.zip",
+			"hash": "382a1fb1c19869e2a6f8501fb4d03552252edaf2fb7917f8bd99bba8b9d32a90"
+		}
+	},
+	"bin": "subl.exe"
+}


guillermooo:
I wouldn't add this without consent from Sublime HQ.

scottwillmoore:
Homebrew-cask which is an extension of homebrew for installing Mac applications has it included in their collection of apps. I doubt Sublime HQ would respond due the the slowed pace of updates lately for sublime text (also i'm lazy :smile:). As I said above I don't think I would ever use sublime text through scoop unless some of the issues above where adressed, so I think I close this pull request for now.

The manifest works if anyone has any interest in playing around with it, fixing some of the issues or using it privately.

guillermooo:
Asking wouldn't hurt, and I don't think ST dev can be said to be slow any more. Also, the fact that someone has done a similar thing wihout consent doesn't mean that Sublime HQ agrees (or disagrees) with it.

I once asked about their opinion on installing trial versions on CI servers for testing and their reply was quick and positive.

I sound grumpy, but I'm just writing fast and short. Just an opinion.

scottwillmoore:
Alright, I have sent off an email to the developers, so it will be interesting to see what we get back. For now I will keep this pull request closed until further notice. I was already hesistant about adding this to scoop, just wanted to see some opinions. :smile:

My opinion remains the same, it would be cool if we could use Sublime Text in scoop, but in its current state I don't think it is particularly advantages over the executable installer. I also believe we should get consent from Sublime HQ first as well, so we'll wait for that before progressing.

guillermooo:
Agreed. It'd be cool to be able to install ST from scoop, but an `--install-prefix` would be desirable in this case. I don't think scoop has/wants that?

scottwillmoore:
Could you please describe what an `--install-prefix` is? I cannot seem to find anything specific on the web.

ntwb:
Take a look at https://github.com/lukesampson/scoop/issues/126

scottwillmoore:
Alright I have emailed Sublime HQ and we have a response. :)

> 
> **Sales Team** <sales@sublimetext.com> (sent by kari@sublimetext.com)
>
> Hi Scott,
> 
> Thanks for your email.
> 
> Yes, you may use the free install of Sublime as part of your package. If you could credit us in some way on your install page that would be great.
> 
> Cheers
> 

Thoughts?

deevus:
@scottwillmoore I reckon just put the credit in pre_install or post_install

lukesampson:
Thanks. I think this is a worthwhile addition for now, although obviously there are issues with updates as with Ruby and NodeJS. I think [travelling directories](https://github.com/lukesampson/scoop/wiki/Travelling-directories) might eventually be the solution for this, although that hasn't been explored fully.

Some people seem to have a preference for control over where things get installed, whereas my preference is low or zero cognitive burden for installing an app—a good solution for me is one that just works, not one that can work if you know the right combination of command line arguments.

scottwillmoore:
@deevus Scoop also has a `"notes: "Useful information here."` which will be dsiplayed when the manifest is installed.

@lukesampson Yeah I believe a simple implementation of travelling directories, or using symlinks to link the directories to an external folder `scoop/apps/sublime_text/b3083/Packages ` -> `scoop/apps/sublime_text/Packages` which does not move when you install a new version would be a better idea.

If you want to have lots of control and adjust settings to install an app, scoop is not your app. :smile: That said, the scoop apps should be made so they work just as good as a manually installed app.

deevus:
@scottwillmoore The more you know :) You'd think I would considering my contributions of late haha

scottwillmoore:
@deevus It's like you havn't even read the wiki! I'm kidding. Also do you have any ideas about what the note should say? I'm not a legal guru so would it be something like: `Copyright Sublime HQ`, or a thank you note: `Thanks for Sublime HQ's permission to include Sublme Text`? I am not really sure.

lukesampson:
@scottwillmoore yeah if symlinks worked on Windows that'd be much easier. Unfortunately creating symlinks  (or directory junctions or whatever) requires Administrator elevation so that's a deal-breaker.

How about 
> This is a beta version of Sublime Text 3. For more information please see http://www.sublimetext.com/3

scottwillmoore:
Yep that note sounds good, I have added it to this pull-request. Ah I wasn't aware that you require adminstrator elavation. I guess travelling directories will work fine then. How about we implement and then we can find any issues that occur when using travelling directories?

deevus:
Gah I rebased and merged from command line and it didn't detect it as merged :(

PR #86, opened 2015-04-16
Git 2.3.5 Preview

myty:
https://github.com/git-for-windows/git/releases/tag/v2.3.5.windows.8

### DIFF
--- a/git-preview.json
+++ b/git-preview.json
@@ -1,0 +1,19 @@
+{
+	"homepage": "https://git-for-windows.github.io/",
+	"license": "GPL2",
+	"version": "2.3.5.8-dev-preview",
+	"architecture": {
+		"64bit": {
+			"url": "https://github.com/git-for-windows/git/releases/download/v2.3.5.windows.8/PortableGit-2.3.5.8-dev-preview-64-bit.7z.exe#/dl1.7z",
+			"hash": "850bca6a507c3d2114748c5527471c1bb7ccf209337ce7e0199e465c6c7badd3"
+		},
+		"32bit": {
+			"url": "https://github.com/git-for-windows/git/releases/download/v2.3.5.windows.8/PortableGit-2.3.5.8-dev-preview-32-bit.7z.exe#/dl1.7z",
+			"hash": "64a2f5e1f067565ed8c5c936965205fba151a78038f54cfc0afb8d462d99d8fe"
+		}
+	},
+	"bin": [ "cmd\\git.exe", "cmd\\gitk.exe", "cmd\\git-gui.exe" ],
+	"post_install": [
+		"git config --global credential.helper wincred"
+	]
+}


ntwb:
Nice, but, digging around looking for some info on this v2.x release and by the looks of things this should actually be in the main bucket and be an update to Git.

via https://github.com/msysgit/git/blob/master/README.md
> Please note that the Git for Windows project will soon switch to another development environment and that the upcoming Git for Windows 2.* will be maintained here.

Old Git for Windows https://github.com/msysgit/git
New Git for Windows https://github.com/git-for-windows/git (edit

myty:
Yea... I was wondering if I should do that or not. I know they were looking for people to test it, so I figured an official release might be near and once that happened, I'd just do a pull request to the main bucket.

With all that said, I could move it over the main bucket if you want. Just let me know.

deevus:
As long as it works I reckon move it to the main bucket.

myty:
Next question... Keep it as git-preview or just replace git?

deevus:
I think it should replace git. The current version in the main bucket is technically a preview build. Once you create the pull request I think a few of us can test it to see that all the major things are working, and then we'll merge it in.

ntwb:
Yeah, lets leave it as is for a few days, I'll uninstall Git v1.x and grab this v2.x from the extras bucket and test away, if we don't have any issues in a few days lets merge it in :)

myty:
Sounds good to me.

myty:
It's been a week.  How did this work out for everyone?  Anything that should prevent us from moving this to the main branch?

deevus:
I honestly forgot I was testing this! Which is good because I've had no issues. +1 here

lukesampson:
Something's not working quite right with SSH keys and ssh-agent—this new version of Git asks for my SSH password each time, even though ssh-agent is running and my key is added. It worked fine with Git 1.9.5.

deevus:
@lukesampson I'm not having this problem. What SSH package is it using? I've had problems before and the solution was to run `scoop reset openssh`

ntwb:
Likewise, no problems running this, it just works :+1: I've also not had any SSH issues :/

lukesampson:
I'm using `ssh-agent` from openssh. The ssh plugin from [pshazz](https://github.com/lukesampson/pshazz) is `ssh-add`ing my SSH key when Poweshell starts up, but I just tried starting `ssh-agent` manually and using `ssh-add` with no success. I just tried `scoop reset openssh` too and it didn't work. What are you guys using for SSH?

deevus:
Yeah I'm using `openssh`.

myty:
I was having the same issue as @lukesampson and I updated the git-preview manifest with the ssh files included in the git folders.  This essentially overrides all of the openssh files, but it works.  I created a new pull request #88 to test these changes, but I'm not sure if it's something we'd want to do or not...

lukesampson:
Thanks @myty, yeah #88 seems to fix it, even though it looks like the same version of SSH that's already installed with `openssh`. No idea how they managed to make that not work.

I'm not sure how I feel about this either.

@deevus and @ntwb: how are you guys avoiding having to type your SSH key passwords every time?

deevus:
@lukesampson I just checked and my git ssh key doesn't have a password, so that's probably why :D

lukesampson:
@deevus oh right, no worries. I just went to report this as an issue on the git-for-windows GitHub and saw [this](https://github.com/git-for-windows/git/issues/67) issue where the developer says

> please feel free to start contributing by installing the SDK, fixing the bugs you find, and opening Pull Requests. You followed the call of the *developers'* preview, therefore it is fair to treat you as a developer.

So, maybe we should leave this as `git-preview` for now, because I wouldn't even know where to start debugging their code. Thoughts?

deevus:
@lukesampson I think we leave this here for now. I'll continue using it since it's working fine for my use case but we can revisit in future. `git-2.3.6` is out so maybe it will be fixed with that if there's a Windows build soon.

lukesampson:
@deevus OK cool.

PR #88, opened 2015-04-24
Git for Windows 2.3.5 - SSH Fix

myty:
This fixes an issue with the openssh version of ssh-agent where you need to enter credentials every time you connect over ssh.

I'm not sure if this is something we really want to do or not.  If you want to go back to the openssh version, you can just `scoop reset openssh`


deevus:
Will this still be the case with the new version (2.3.7)?

PR #92, opened 2015-05-06
The notepad++ 6.6.9 url has seemingly changed

umaritimus:


lukesampson:
Thanks :smile:

deevus:
Perhaps we could just get it directly from the NPP github? https://github.com/donho/notepad-plus-plus/releases?after=v6.7.3

lukesampson:
That's a good idea.

PR #98, opened 2015-05-30
fix rust-nightly manifest

### LINKED ISSUE #97: Rust and cargo manifests need updating
There are some issues with the `rust-nightly` and `cargo-nightly` manifests. Because they generate shim exes, the `cargo` build tool has trouble shelling out to the `rustc` compiler (see https://github.com/rust-lang/cargo/issues/1661). Also, cargo is no longer a separate component, it is generally packaged together with rust so having a separate manifest for it is confusing.

I'll be submitting a PR for this momentarily. It's fixed the issue in my environment, but I'll loop in the original submitters of the `{rust,cargo}-nightly` manifest so they can take a look too.

Scoop-Extras should (and my PR will):

1. Remove the `cargo-nightly` manifest, since `rustc` and `cargo` are generally installed together from a single MSI
2. Change from using shims to `env_add_path` because otherwise rust-lang/cargo#1661 gets in the way. I'd still like to track down why the shim isn't working, since I prefer the shim mechanism, but for now this will fix the issue.

anurse:
* remove separate cargo manifest, cargo is built in now and it is expected to be installed alongside rust
* switch from EXE to MSI installer
* rust really doesn't like running through a shim, so use env_add_path instead (see rust-lang/cargo#1661)

Works On My Machine :).

Fixes #97 

/cc @deevus (original submitter of rust manifests)


lukesampson:
Thanks Andrew.

deevus:
@anurse Looks good to me. Thanks for pinging me.

PR #104, opened 2015-06-28
update to dart 1.11.0

guillermooo:


guillermooo:
I think this one needs to be re-run. Failure seems to be due to connectivity glitch.

lukesampson:
Oh that was just the SSL cert expiring on get.scoop.sh, I think the manifest is fine. I guess this is output from a script, but is there anyone to clean up the JSON formatting?

PR #107, opened 2015-07-08
Update vlc.json

nishanthkarthik:
Updated from 2.2.0 to 2.2.1. I used the SHA 256 hash values for `hash` property.


lukesampson:
Thanks :smile:

PR #112, opened 2015-08-08
Add Ragel 6.9

glebm:
Ragel compiles executable finite state machines from regular languages.
Read more at http://www.colm.net/open-source/ragel/

This is a pre-built binary from https://github.com/eloraiby/ragel-windows

### DIFF
--- a/ragel.json
+++ b/ragel.json
@@ -1,0 +1,10 @@
+{
+    "homepage": "https://github.com/eloraiby/ragel-windows",
+    "version": "6.9",
+    "license": "GPL",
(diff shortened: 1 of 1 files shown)


deevus:
Any particular reason that this shouldn't be in the main bucket?

glebm:
@deevus Not sure what the policy on this is. Put it here because latex and phantomjs are here. One other concern is that this binary is not official (there is no official one).

deevus:
You're right there does seem to be a precedence here, and I couldn't find anything that specifically states what is appropriate in the main bucket. 

@lukesampson Perhaps we should review some of the apps in extras and move them to main if deemed appropriate.

PR #142, opened 2015-12-30
Update eclipse.json due to broken links

umaritimus:
The links are broken and the version is super old


umaritimus:
never mind, something else is broken

PR #145, opened 2016-01-08
webstorm 11.0.3

sestegra:
I have 2 open points before merging my pull-request
- how to tag a proprietary license in the manifest
- the uninstallation gives following error:
```couldn't remove ~\appdata\local\scoop\apps\webstorm\11.0.3: it may be in use```
I don't know if the root cause is scoop or not

### DIFF
--- a/webstorm
+++ b/webstorm
@@ -1,7 +1,0 @@
-{
-  "version": "11.0.3",
-  "url": "https://download.jetbrains.com/webstorm/WebStorm-11.0.3.exe",
-  "hash": "678888c8e4cce921a88bcdb16db4bdc784196983b3147e854a759102ad14b037",
-  "installer": { "args": ["/S", "/D=$dir"] },
-  "uninstaller": { "file" : "bin\\Uninstall.exe", "args" : ["/S"] }
-}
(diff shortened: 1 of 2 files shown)


lukesampson:
* You can add a URL to a proprietary license if it's available, otherwise you can just leave it blank, since this is the extras bucket.
* The installer might have started services or opened file locks and didn't clean them up properly when it uninstalled. This is pretty typical for exe installers, where requiring restarts to properly uninstall isn't considered bad form. This is why scoop manifests usually try to circumvent installers so you know the install will just copy files and won't block or produce unknown side-effects like registry entries. If the installer is the only option then there's not much else you can do. In this case the .exe is extractable by 7-zip, so you could change the URL to `"https://download.jetbrains.com/webstorm/WebStorm-11.0.3.exe#dl.7z"` and that will make scoop extract the files. Then you can add shims for the exes and perform whatever other installation steps are necessary. It may be that the program isn't very portable and relies on registry entries or other secret moves to run, in which case it's okay to go back to the installer.

sestegra:
The manifest should be ok now.

lukesampson:
Cool, thanks :smile:

PR #147, opened 2016-01-10
Version updates for Atom and SourceTree

damnhandy:


damnhandy:
Also updating SourceTree as well

PR #152, opened 2016-01-21
Addition of WinMerge2011 (inofficial WinMerge successor)

engelhro:


lukesampson:
Thanks for all these!

PR #167, opened 2016-02-10
update to latest stable conemu

anurse:
Moves to the latest (stable) version of ConEmu: 150813g, also using the GitHub download site, since the main site (powered by fosshub, doesn't have an accessible download URL).


lukesampson:
Thanks.

PR #179, opened 2016-03-07
Fix hash for Atom

vidarkongsli:
Sorry about that :(


vidarkongsli:
My previous PR contained a wrong hash.

PR #182, opened 2016-03-11
Updated to 0.67

eugenesvk:
Messed up a bit with the previous PR by pulling your newer changes, so doing a fresh one with only this update with static links that won't change on update.

### DIFF
--- a/putty.json
+++ b/putty.json
@@ -1,15 +1,14 @@
 {
 	"homepage": "http://www.chiark.greenend.org.uk/~sgtatham/putty/",
-	"version": "0.64",
+	"version": "0.67",
 	"license": "http://www.chiark.greenend.org.uk/~sgtatham/putty/licence.html",
-	"url": "http://the.earth.li/~sgtatham/putty/0.64/x86/putty.zip",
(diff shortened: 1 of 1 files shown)


lukesampson:
Cool, thanks :)

eugenesvk:
By the way, maybe you can enlighten me on this git quirk — when I pulled all your other commits (without thinking too long, I'm behind, why not get updates), for some reason my name got attached to all of them and when I updated the **putty** file with the proper link my PR showed that **PLUS** all those other commits that your branch already had.
What's up with that? Why wouldn't it continue to show only the difference, which is the putty file?
Or maybe it's just the way I messed up trying to squash my two commits into one (though in the GUI window within SourceTree client I only saw exactly those two commits of mine when trying to combine them, never those other 6 of yours).

It's just the second time it happens when a very simple 1 commit-> 1 PR + a minor update per author's request  turns into too much confusion on my side.

If it's not enough info to go on, nevermind, next time I'll just not pull anything acccidentally and only focus on finishing updating my own little piece.

lukesampson:
I'm no git expert, but I think you can run `git rebase origin/master` when that happens. That should move your commits so that they're applied after the remote commits, and clean up the commit history.

If you remember (I never do), I think you can run `git pull --rebase` instead of just `git pull` to save you the extra rebase step.

PR #192, opened 2016-03-28
Add manifest for filezilla

neoeinstein:


lukesampson:
Thanks for all these! :smile:

PR #215, opened 2016-05-21
Add xmllint

zaps:
Hey there, here is a PR for [xmllint](http://xmlsoft.org/xmllint.html) 

Each of the 7z files that the manifest calls out have a bunch of extra supporting folders in them that aren't needed for xmllint.exe to run properly (`lib/`, `share/`, `include/`, etc). The only stuff that needs to be part of this are the dlls in the `bin/` folder in each 7z archive. 

I tried to exclude the unneeded stuff in the manifest via the `extract_dir` element but I couldn't get it to work. So as it is now, when you install you'll get all 23mb of stuff that's in these 7z's as opposed to the ~8 that's actually needed.

### DIFF
--- a/xmllint.json
+++ b/xmllint.json
@@ -1,17 +1,36 @@
 {
 	"homepage": "http://xmlsoft.org/",
 	"version": "2.9.3",
-	"url": [
-		"ftp://ftp.zlatkovic.com/libxml/64bit/iconv-1.14-win32-x86.7z",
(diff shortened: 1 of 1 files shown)


lukesampson:
That's cool – thanks.

To get `extract_dir` to work with more than one `url`, you need to make an array of the paths to extract for each `url`, e.g.

```
"extract_dir": [ "bin", "bin", "bin", "bin" ]
```

See [vim.json](https://github.com/lukesampson/scoop/blob/master/bucket/vim.json#L19) for a real example.

PR #254, opened 2016-08-19
Added sha256 hash to sysinternals

swyphcosmo:


lukesampson:
Thanks, but sysinternals executables get updated without using version specific URLs. So if we use hashes it will lead to unexpected installation failures down the track when they update the binary. I think it's better to have a hash warning then an installation failure.

swyphcosmo:
Gotcha. Thanks for letting me know. I'll cancel the PR.

PR #255, opened 2016-08-19
Fixed hash for whois.exe

swyphcosmo:
Previous hash for the zip download was wrong.


lukesampson:
Thanks. Same situation as #254—we'd probably be better not using a hash at all than having it fail when no one's paying attention.

PR #263, opened 2016-09-06
Update Notepad++

inosik:
I updated Notepad++ to the latest version (6.9.2) and also enabled it for `checkver.ps1`.

I also added a shortcut for it, tell me if I should remove it.


lukesampson:
Cool, thanks!

PR #266, opened 2016-09-09
Update VS Code to 1.5.1

inosik:
I think we need a better strategy for detecting available updates for VS Code, because the current one isn't working. Maybe we could check their GitHub releases page?


lukesampson:
Cool, thanks.

PR #274, opened 2016-09-22
Update ngrok to stable

moigagoo:
The download link always points to the latest stable version (today it's 2.1.14).

### DIFF
--- a/ngrok.json
+++ b/ngrok.json
@@ -1,15 +1,15 @@
 {
 	"homepage": "https://ngrok.com/",
-	"version": "2.0.19",
+	"version": "2.0.24",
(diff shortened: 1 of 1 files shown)


lukesampson:
Actually, using the generic 'stable' url the hash will break when they update the version. So it's probably better to use a url that contains the version. E.g. https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-2.0.24-windows-amd64.zip

moigagoo:
Yeah, I've been worried about it too. Well, the latest stable version is 2.1.14, so here's another PR :-) https://github.com/lukesampson/scoop-extras/pull/275

PR #275, opened 2016-09-23
Update ngrok to 2.1.14

moigagoo:


lukesampson:
Haha whoops, looks like you can download `https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-WHATEVER-windows-amd64.zip` and it still works. Might just have to keep an eye on it and see what happens when they update the .zip version.

PR #282, opened 2016-10-05
Update Notepad++ to 7.0

inosik:
Notepad++ now has a x64 version.

I simply copied the SHA1 hashes from the download page, as I'm too lazy to download the files myself. Ping me if I should change them to SHA256.


lukesampson:
SHA1 should be fine. Thanks!

PR #337, opened 2017-01-18
Update Oracle JDK to 1.8.0-u121

guylabs:
Hi

first of all thanks for this great tool! We use this to automatically setup the dev machines. I managed to update to the newest Oracle Java version and use the silent installer shipped within the executable. The only issue is that there is a UAC popup and I don't think we are able to accept that in any way.

As I am only a former Windows user I do not have the knowledge on how to install it without any UAC (or somehow extract it etc.). In the updated manifest the `oraclejdk.ps1` script is not used anymore to install the JDK as this was not working anymore.

If there is someone else who is able to fix it please go ahead. You can use this pull request as a base. Or @lukesampson you can accept it if the UAC is not an issue.

The rest is working properly, so you have Java installed without the public JRE.

Cheers,

Guy

### DIFF
--- a/oraclejdk.json
+++ b/oraclejdk.json
@@ -1,25 +1,21 @@
 {
     "homepage": "http://www.oracle.com/technetwork/java/javase/overview/index.html",
-    "version": "1.8.0-u74",
+    "version": "1.8.0-u121",
     "architecture": {
         "64bit": {
             "url": [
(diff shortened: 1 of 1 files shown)


lukesampson:
Thanks for this. The UAC isn't a deal-breaker. If someone knows how to do a non-admin install, please submit a PR.

guylabs:
Thanks @lukesampson for this quick merge!

PR #349, opened 2017-02-07
Update pandoc to 1.17.2

phitsc:


r15ch13:
Hi could you add autoupdate information? https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate

PR #352, opened 2017-02-13
Adds SpeedCrunch and HTML tidy to extras bucket

phitsc:
Merge only if you also merge my other pull request here (or get rid of the URL check as suggested by rrelmy):

https://github.com/lukesampson/scoop/pull/1333

as SpeedCrunch is hosted on bitbucket.

If you would rather not do that I can move HTML tidy onto a different branch and open a PR for that alone.


lukesampson:
Thanks for these 😄

PR #361, opened 2017-02-23
Adds CMake RC

berwyn:
Needed this to access newer features in CMake and thought it might be useful for others as well :)


lukesampson:
Thanks 👍

PR #363, opened 2017-03-16
Add SumatraPDF, Shadowsocks-windows, vswhere

Congee:


lukesampson:
Thanks for these.

Would it be worth adding a `shortcuts` (see [here](https://github.com/lukesampson/scoop/wiki/App-Manifests)) for SumatraPDF if it's a GUI program? That way it will show up in the start menu.

Congee:
Thanks for this tip. SumatraPDF tries to be a command line pdf viewer invoked by other apps. For example, you need an external pdf viewer when using latex. While shadowsocks-windows is a GUI app, should appear in the start menu. So, I will add shortcuts to shadowsocks-windows. ;-)

PR #373, opened 2017-03-29
Use official Emacs dependencies -> enables eww, image display etc

pcrama:
The Emacs project proposes a bundle of dependencies to install alongside emacs.  Using these instead of the gnutls zip file originally in the app manifest enables more functionality, e.g.

- eww (a web browser in Emacs) because libxml2 can now be found
- inline image display because the image libraries can now be found
- probably others

Note that the same binaries as those contained in the gnutls zip are also included, so that dependency/download was dropped.


r15ch13:
Sounds good

pcrama:
That was fast! Thanks a lot.

PR #376, opened 2017-04-03
Cloudfoundry CLI with autoupdate and multiarch.

se35710:
Maybe Cloudfoundry CLI could be included in extra?


lukesampson:
Thanks 👍

PR #379, opened 2017-04-12
All Eclipse Packaging projects plus SDK and platform

se35710:
A rather big commit with all Eclipse projects added.

The Eclipse project uses “release trains” for its packaging project, and since the Europa release (3.3), the packaging project is named after the train, replacing all version numbers. For the main Eclipse project, a drop structure is used, with build id in its path. There are a couple of sources of information needed to do the mapping:

1.	The Oomph product map (http://git.eclipse.org/c/oomph/org.eclipse.oomph.git/plain/setups/org.eclipse.products.setup)
2.	The Simultaneous Release Schedule (https://wiki.eclipse.org/Simultaneous_Release)
3.	The Eclipse Project’s archived releases (http://archive.eclipse.org/eclipse/downloads/)
4.	The Eclipse Packaging Project’s archived releases (https://wiki.eclipse.org/Older_Versions_Of_Eclipse)

To make this work with the new functionality introduced with [#1332](https://github.com/lukesampson/scoop/pull/1332) I had to create a translator that now is running on Azure. It maps the actual version to the correct download link, no matter if it’s a named release or a drop, and then redirects to the Eclipse mirroring network to fetch the files.

The main reason for all of this is that scoop is missing a regex replace feature with callbacks configurable in the manifest, which would need to be evaluated both when `checkver.ps1` is run, and when installing older versions with `scoop install <whatever>@<older version>`. If this is something that in the scope of scoop, I would gladly help to create such a feature, though PowerShell skills are not currently on my resume.

If the new Eclipse manifests are somewhat too complicated for scoop extras, I can put them in a different bucket, but I do think the Eclipse users would greatly benefit from using scoop to handle the different packages and versions needed (sometimes your most valued plugins does not work with the latest Eclipse version etc, etc).

### DIFF
--- a/eclipse-android.json
+++ b/eclipse-android.json
@@ -1,0 +1,48 @@
+{
+    "homepage": "https://www.eclipse.org",
+    "license": "https://www.eclipse.org/legal/epl-v10.html",
+    "version": "4.6.3",
+    "architecture": {
+        "64bit": {
+            "url": "https://version-to-train.azurewebsites.net/epp.package.android.zip?version=4.6.3&architecture=64bit",
+            "hash": "sha512:454682b9bc3358fdbd36ac7efd0f87b343a19475fab167f8a67a3d7c2c8cc31a79960d7c0cca42d9ef33bab95a624f49083eefb86e046d859e0dacdabfde2fd8"
+        },
+        "32bit": {
+            "url": "https://version-to-train.azurewebsites.net/epp.package.android.zip?version=4.6.3&architecture=32bit",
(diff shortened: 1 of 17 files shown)


lukesampson:
This looks like a huge amount of work—thank you for contributing this!

I don't think we've come across any other projects of this complexity that require regex callbacks. So I'm not sure if this would be used for anything else if we added it to Scoop. But if you're happy maintaining the Azure website to do the translation, that would be great.

se35710:
Thanks! And yes, I will continue to maintain the Azure website.

r15ch13:
Azure Source: [se35710/train-to-version](https://github.com/se35710/train-to-version) (for documentation purposes only 😃)

PR #385, opened 2017-04-19
Add Tomcat 6.0, 7.0 and 8.0, suggest server-jre or openjdk.

se35710:
Some applications require a specific release (major version) of Tomcat.


r15ch13:
Adding these to the [versions bucket](https://github.com/scoopinstaller/versions) might be better because the main and extras bucket mostly contain the latest version.

se35710:
Ok, I'll move them to versions.

PR #390, opened 2017-05-01
added protobuf v3.3.0

gageorsburn:
Not sure if manifest is as complete as it needs to be. Let me know if it's not.

### DIFF
--- a/protobuf.json
+++ b/protobuf.json
@@ -1,10 +1,14 @@
 ﻿{
-    "version":  "v3.3.0",
+    "version":  "3.3.0",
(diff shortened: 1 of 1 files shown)


r15ch13:
You can add checkver and autoupdate with this, but change the version from v3.3.0 to only 3.3.0 👍 
```
    "checkver": "github",
    "autoupdate": {
        "url": "https://github.com/google/protobuf/releases/download/v$version/protoc-$version-win32.zip"
    }
```

lukesampson:
Thanks 👍

PR #391, opened 2017-05-01
Changed atom.exe to atom.cmd

hannesdorn:
In atom.json I have changed atom.exe to atom.cmd. atom.exe blocks the command line, while atom.cmd does not.


lukesampson:
Thanks 👍

PR #399, opened 2017-05-08
Update VS Code to 1.12.1

inosik:
Maybe we need to find a better way to check for updates, the VS Code guys often forget to make a proper GitHub release.


r15ch13:
Changed it to use the update feed used by vscode itself 👍

PR #401, opened 2017-05-11
Added love, a 2D lua-based gaming engine

Guard13007:
I also added to the notes a link to a bucket that has alternate versions of the LÖVE engine. It's helpful when working on it to test code in multiple versions of the engine since there are significant changes between even patch releases.

I would submit those instead to the versions bucket, but as they are not popular-dev-tool related, I do not feel they are appropriate, and there is no scoop-extras-versions bucket.

If this is unacceptable, let me know and I will make another commit to remove the note.

### DIFF
--- a/love.json
+++ b/love.json
@@ -1,0 +1,38 @@
+{
+  "homepage": "https://love2d.org/",
+  "version": "0.10.2",
+  "architecture": {
+    "64bit": {
+      "url": "https://bitbucket.org/rude/love/downloads/love-0.10.2-win64.zip",
+      "hash": "a9603b1e1fb353bc0d5608629ea4c0d2e94df3598f3cc4c2e9e84723e6a5a2b9",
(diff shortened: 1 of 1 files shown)


lukesampson:
Thank you! FYI we have a [versions](https://github.com/scoopinstaller/versions/) bucket for alternate versions if you would like to submit there for greater visibility.

Guard13007:
@lukesampson I am aware of that, however, I felt that as it is not a core development tool, it was not appropriate for the versions bucket. I feel it would be more appropriate for an extras-versions.

If you disagree however, I am happy to make the needed pull request and update this manifest to mention that instead of my own bucket.

lukesampson:
Yes please, it would be great if you could submit to the versions bucket. The idea is to have the one bucket to collect alternate versions, no matter what bucket is used for the stable version.

PR #402, opened 2017-05-11
Added MoonScript, a language that compiles to Lua

Guard13007:
[The author does use GitHub releases](https://github.com/leafo/moonscript/releases), however there is a hash value (I believe a commit hash) used on the download URLs so I'm not sure how to properly set up autoupdate on this one.


Guard13007:
(Specifically, this is a compiler and a basic linter.)

lukesampson:
I think that's a good start. Thank you.

PR #403, opened 2017-05-11
KiTTY, a fork of PuTTY adding a whole lot of commonly requested features

Guard13007:


r15ch13:
Added autoupdate information: cbe4de6ec22cb492fc6d1d9b070ba91a69fabb84

Guard13007:
@r15ch13 Uncompressed version is not portable. :(

lukesampson:
What do you mean by "not portable"? It seems to install fine.

r15ch13:
It stores the setting in the Registry. The portable variant uses file based storage.

To run it in portable mode there has to be a `kitty.ini` with this settings
```
[KiTTY]
savemode=dir
```
But we have to make `kitty.ini` and `Sessions` directory persisten.

r15ch13:
Change it to the portable version with d30f72fc409133245c9de177bc917df833d58dfa

Guard13007:
@r15ch13 That fosshub link is a webpage. I tried using that in the first place and it doesn't work. :/

r15ch13:
By opening [FossHub](https://www.fosshub.com/KiTTY.html) it requests this URL (https://www.fosshub.com/gensLink/KiTTY/kitty_portable.exe/51f5253ee472ceb30900012c) to receive a generated download URL as a string. Then JavaScript is used to start the download with this generated URL.
I replicated this behavior with this commit https://github.com/lukesampson/scoop/commit/14a461296bebf18ee137046f43bf8a883cb39b4f so other programs hosted on FossHub can be downloaded.

Hm, http://kitty.9bis.net/ is down now 😕

Guard13007:
Awesome, thank you for explaining. And oh crap. :/ Hope that's a temporary bug or something.

PR #404, opened 2017-05-12
Update tomcat.json

Alorel:
fix: made URLs time-resistant

www.apache.org only seems to keep the latest version, so whenever there's an update we had to manually update the initial URL
archive.apache.org seems to keep all the versions.


Alorel:
On a side note, I find it funny that my PR is solving a 404 error and just happens to be the 404th PR :')

r15ch13:
Thanks. This closes the possible 1-hour gap until the automated update service kicks in 👍

PR #405, opened 2017-05-13
winmerge.json: Changed link from http to https

hannesdorn:


lukesampson:
Thank you. I fixed some formatting which was causing the tests to fail. Also added auto-updating.

PR #406, opened 2017-05-13
add some missing start menu shortcuts

rbnlffl:
While installing apps via Scoop, I noticed some missing start menu shortcuts.

I also changed `subl.exe` to `sublime_text.exe` to prevent the flashing cmd from opening before starting Sublime Text.


r15ch13:
Thanks 👍

PR #410, opened 2017-05-16
Context menu note @ vscode

Alorel:
One of the things I missed from the non-scoop installation of vscode is the "Open in VSCode" context menu in explorer.

Added a note to a guide on how to do it in case others are in the same situation.


lukesampson:
Okay thanks.

PR #411, opened 2017-05-16
Adding tortoise-svn

NaanProphet:
In order for the installer to run properly the UI must show at least in reduced/passive mode

### DIFF
--- a/tortoisesvn.json
+++ b/tortoisesvn.json
@@ -1,0 +1,23 @@
+{
+    "_comment": "Because of Explorer integration, this installer requires admin privileges to run.
+        Maintainers should update the MSI GUID for uninstaller by navigating to registry path, i.e.
+        for 64-bit installs HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows/CurrentVersion/Uninstall/
+
+        Sometimes, during the uninstall process, the user is prompted some applications should be
+        closed before continuing. Rather than have the MSI automatically close applications, please
+        choose the other option to reboot manually later. Otherwise, the uninstaller never returns
+        to scoop and not all directories are removed, though the actual program is uninstalled.",
+    "homepage": "https://tortoisesvn.net",
+    "version": "1.9.5",
+    "license": "GPL",
+    "url": "https://sourceforge.net/projects/tortoisesvn/files/1.9.5/Application/TortoiseSVN-1.9.5.27581-x64-svn-1.9.5.msi",
+    "hash": "dbb3e538fde2fac56781b4893ae6ab45b6cb331d2a3888ed4308a5ec47e85468",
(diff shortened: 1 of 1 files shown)


NaanProphet:
Suggesting this in scoop-extras because though SVN is still used, Git is more popular

lukesampson:
Firstly, congrats on finding a way to make that work. Unfortunately I assume it will require admin elevation to run the MSI. It also has unknown side effects (e.g. registry changes) and doesn't copy files to the user's directory.

There is old code in Scoop to do MSI installs (see [here](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L533)), which was deprecated because these kind of installs are at odds with Scoop's objectives.

I guess the problem is that there's no way to get Explorer integration for TortoiseSVN without making registry changes. And that requires admin elevation and is hard to reverse when it's uninstalled.

I can see you've put in some effort to get this working. If you could use adjust the `installer.args` to include `TARGETDIR` and `INSTALLDIR` (see [here](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L543-L547)) so it at least ends up in the user's directory then I would accept this.

tortoisesvn.json
```diff
@@ -0,0 +1,19 @@
...
+    "_comment": "Maintainers should update the MSI GUID for uninstaller by navigating to registry path for 64-bit installs, i.e. HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows/CurrentVersion/Uninstall/",
+    "homepage": "https://tortoisesvn.net",
+    "version": "1.9.5",
+    "license": "GPL",
+    "url": "https://sourceforge.net/projects/tortoisesvn/files/1.9.5/Application/TortoiseSVN-1.9.5.27581-x64-svn-1.9.5.msi",
+    "hash": "dbb3e538fde2fac56781b4893ae6ab45b6cb331d2a3888ed4308a5ec47e85468",
+    "installer": {
+        "args": "/qr"
+    },
+    "post_install": "iex \"cp /Windows/System32/msiexec.exe $dir\"",
```
deevus: Would this fail if scoop isn't installed on the Windows drive? eg. Windows is on C: and scoop is on D:

NaanProphet: Thanks for the feedback guys! First time contributing to scoop, so still trying to read all the documentation and get familiarized ;)

@lukesampson really appreciate the explanation. Ya, if it were not for the Explorer integration, admin escalation wouldn't be necessary... Will make the changes you suggest!

@deevus would that still be a concern if we install using `TARGETDIR=$dir INSTALLDIR=$dir` ?

lukesampson:
Thanks. I added a `C:` drive to `post_install` in case people are running from a different drive.

PR #415, opened 2017-05-17
Some useful graphics editors / tools

se35710:


lukesampson:
Looks good. Why is Inkscape 32 bit in a separate manifest though?

se35710:
The download url is unfortunately different depending on architecture, so I created a separate manifest so that autoupdate would work.

lukesampson:
@r15ch13 Is it possible to have a different autoupdate URL based on architecture?

r15ch13:
That's a tricky one. Currently `checkver` doesn't support architectures.

PR #416, opened 2017-05-20
Added R Studio

cholmquist:
Created a manifest for RStudio, an IDE for the R language.
post_install script creates an initial R Studio config file that connects the IDE to the R package which should already have been installed by scoop. The user can easily change which R package to use from within the IDE, but this post_installer provides a sensible default so that no further configuration is required to get going.
"scoop install rstudio && rstudio" should just work (tm)

### DIFF
--- a/rstudio.json
+++ b/rstudio.json
@@ -1,22 +1,22 @@
 {
-	"version": "1.0.143",
-	"license": "AGPL",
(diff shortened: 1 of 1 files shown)


lukesampson:
Thanks.

PR #418, opened 2017-05-22
Update slack to 2.6.2

talha131:


r15ch13:
Nice they haven't even updated their site 👍

PR #419, opened 2017-05-22
Add zeal doc viewer

talha131:


r15ch13:
Thanks! Will install it right away! 😄 

Fixed the checkver url and added autoupdate support.

PR #422, opened 2017-05-24
Update atom.json to correct 64bit download w/ hash

goudaonmixer:
The previous version of atom.json had the same download link and hash for both the 64bit and 32 architectures. This commit corrects this issues by updating the download link and adding its hash.

### DIFF
--- a/atom.json
+++ b/atom.json
@@ -27,6 +27,13 @@
         "github": "https://github.com/atom/atom"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


r15ch13:
Thanks 👍

goudaonmixer:
Oh goodness. That revert was a mistake. I meant to add one more commit to this PR that fixes an issue with how the 64bit version extracts. The fix is coming!

PR #430, opened 2017-05-25
Add Listary

wsw0108:


r15ch13:
Thanks, but I let Luke decide if we add this, because of the [pro version](http://www.listary.com/pro)

wsw0108:
OK. That's reasonable.

PR #431, opened 2017-05-25
Added PyCharm

cholmquist:
Added PyCharm

### DIFF
--- a/pycharm.json
+++ b/pycharm.json
@@ -1,0 +1,18 @@
+{
+    "version": "2017.1.3",
+    "homepage": "https://www.jetbrains.com/pycharm/",
+    "url": "https://download.jetbrains.com/python/pycharm-community-2017.1.3.zip",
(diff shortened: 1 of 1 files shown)


r15ch13:
Thanks. Added autoupdate and version checks.
Is it okay to use pycharm.bat to run the correct pycharm.exe?

cholmquist:
When using pycharm.bat for the shortcuts you don't get the nice looking icon.
To make things consistent I've removed usage of pycharm.bat altogether, which I think is the better option.

Is "env_add_path": "bin" necessary?

nueko:
[jetbrains-toolbox](https://github.com/lukesampson/scoop/blob/master/bucket/jetbrains-toolbox.json) already covered PyCharm

yeikel:
@nueko What if you only want to install PyCharm rather than the full toolbox?

nueko:
then you need to create your own manifest.json.
but why you don't want to install toolbox is that size matter?

yeikel:
My bad, I thought that the toolbox would download all of them. You can actually pick which one to download for the price of the little toolbox binary.

PR #436, opened 2017-05-31
updated the vlc version regex

jfmherokiller:
I updated the autoupdate regex for vlc to check the http://www.videolan.org/vlc/download-windows.html page instead of the previously used page


r15ch13:
Hm, okay they forgot to change the `last` directory. :smile:
But please add the https URL.

PR #441, opened 2017-06-04
Add android-sdk (need testing)

nueko:


r15ch13:
Still, needs some shims or has to be added to PATH.

nueko:
yes, it does.

edgardmessias:
Same file? https://github.com/lukesampson/scoop-extras/blob/master/bucket/android-sdk.json

rasa:
Appears to be a duplicate of https://github.com/lukesampson/scoop-extras/blob/master/bucket/android-sdk.json

PR #442, opened 2017-06-04
add android-studio

nueko:


android-studio
```diff
@@ -34,10 +34,14 @@
...
             "64bit": {
-                "url": "https://dl.google.com/dl/android/studio/ide-zips/$version/android-studio-ide-$matchBuild-windows.zip"            },
+                "url": "https://dl.google.com/dl/android/studio/ide-zips/$version/android-studio-ide-$matchBuild-windows.zip"
+            },
             "32bit": {
                 "url": "https://dl.google.com/dl/android/studio/ide-zips/$version/android-studio-ide-$matchBuild-windows32.zip"
             }
+        },
+        "hash": {
+            "url": "https://developer.android.com/studio/"
```
nueko: How this work?

r15ch13: There are some sha256 in the source of the page. Search for "COPY THIS FOR ARCHIVE PAGE" and you will find them. If they get removed, it will fallback to simply downloading the files :smile:

nueko: 🥇  I found that only when I download the source with curl.
`view-source:https://developer.android.com/studio/index.html` not showing that

PR #446, opened 2017-06-11
Remove Oracle 7 JDK since it is not publically available.

se35710:
After update 7u80, you can not access newer versions of Oracle 7 JDK without a support agreement with Oracle.

Update 7u80 is available for download only after signing in to Oracle.


dsbert:
General question - How does scoop handle package removal when it is already installed for a user?

PR #452, opened 2017-06-15
Create mremoteng.json

masaeedu:
Adds mRemoteNG remote desktop tool from https://mremoteng.org/

### DIFF
--- a/mremoteng.json
+++ b/mremoteng.json
@@ -11,11 +11,18 @@
             "mRemoteNG"
         ]
     ],
+    "persist": [
+        "confCons.xml",
+        "confCons.xml.backup",
+        "pnlLayout.xml",
+        "extApps.xml",
+        "Themes.xml"
+    ],
     "checkver": {
-        "url": "https://github.com/mRemoteNG/mRemoteNG/releases/latest",
-        "re": "/releases/download/(?<short>[^/]+)/mRemoteNG-Portable-(?<version>(?:\\d+\\.)+\\d+).zip"
+        "github": "https://github.com/mRemoteNG/mRemoteNG",
+        "re": "/releases/download/(?<short>[^/]+)/mRemoteNG-Portable-(?<version>[\\d.]+).zip"
     },
     "autoupdate": {
-        "url": "https://github.com/mRemoteNG/mRemoteNG/releases/download/$matchShort/mRemoteNG-Portable-$matchVersion.zip"
+        "url": "https://github.com/mRemoteNG/mRemoteNG/releases/download/$matchShort/mRemoteNG-Portable-$version.zip"
     }
 }
--- a/mremoteng.json
+++ b/mremoteng.json
@@ -1,28 +1,22 @@
 {
     "homepage": "https://mremoteng.org/",
-    "version": "1.75.7008.17935",
-    "hash": "172c0e4b2a305ae64db247d40f92adf4116bf46eb565b9528dfe0ecfd8cf539a",
+    "version": "1.75.7012.16814",
+    "hash": "86363f511785ce56acdb4440fa990de5c992fe40c55dcf0fc43fad03476c5011",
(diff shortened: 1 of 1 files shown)


mremoteng.json
```diff
@@ -11,11 +11,18 @@
...
+        "confCons.xml.backup",
+        "pnlLayout.xml",
+        "extApps.xml",
+        "Themes.xml"
+    ],
     "checkver": {
-        "url": "https://github.com/mRemoteNG/mRemoteNG/releases/latest",
-        "re": "/releases/download/(?<short>[^/]+)/mRemoteNG-Portable-(?<version>(?:\\d+\\.)+\\d+).zip"
+        "github": "https://github.com/mRemoteNG/mRemoteNG",
+        "re": "/releases/download/(?<short>[^/]+)/mRemoteNG-Portable-(?<version>[\\d.]+).zip"
```
masaeedu: Does a `.` inside a `[]` automatically get treated as a literal, or are you accidentally matching arbitrary characters? If the former, this won't work, because you have two `.` characters back to back at the end of the version group (`1234..zip`).

masaeedu: Nvm, just realized `[\d.]+` is an arbitrary mix of numbers and `.`.  Carry on 😅

r15ch13:
Hm, adding persistence to those files isn't that easy. We can't provide a default `confCons.xml` because the `Protected` attribute has to be generated by mRemoteNG and we can't create a link until it has created the file.

masaeedu:
@r15ch13 Does it choke if you just generate a blank file? What if you generate the symlink against a blank file and delete the file in `post_install`?

r15ch13:
Windows doesn't support symbolic links for files, which is why we use hard links (`mklink /h`).
The linked file will not change and still point to the "same" file if the source is deleted. So you end up with two different files. Already tried that... 😢 
A blank file crashes mRemoteNG. I also tried adding only the XML Doctype and Doctype + Root node.
mRemoteNG doesn't like that either...

masaeedu:
@r15ch13 It seems I'm able to create symlinks in Powershell using `New-Item -Type SymbolicLink -Target test.txt sym.txt` using PowerShell. If I open `sym.txt` I see the content of `test.txt`. I can also delete `test.txt` without `sym.txt` disappearing (although attempting to open it no longer does anything). If I then create a different `test.txt`, and try opening `sym.txt` again, I once again see the content of the new `test.txt`.

lukesampson:
@masaeedu As far as I know, administrator privileges are required to create symbolic links, except in recent versions of Windows 10 with Developer Mode enabled.

That's why we don't want to use symbolic links. If we did use them, most users would see admin elevation prompts and that ruins the experience and breaks unattended installs.

masaeedu:
@lukesampson Gotcha. After some digging, I've found you can configure the location in which confCons.xml will be stored/read using the following setting in `mRemoteNG.exe.config`:

```
            <setting name="LoadConsFromCustomLocation" serializeAs="String">
                <value>True</value>
            </setting>
            <setting name="CustomConsPath" serializeAs="String">
                <value>C:\Users\asaeeduddin\AppData\Roaming\mRemoteNG\confCons - Copy.xml</value>
            </setting>
```

So after the folder is extracted and before the program is run for the first time, that can be updated so that the `confCons.xml` file is stored directly in the persist folder.

masaeedu:
@r15ch13 

> This directory has a random string and will be created for each program version

I think that file is copied from `mRemoteNG.exe.config`, which is part of the archive. If you modify `mRemoteNG.exe.config`, I think the copied file (whatever random subfolder in localappdata it ends up), will have the modified settings.

ecornely:
What's the status of that pull request ?
I took back the file in my repo and evertyhing seems ok except a few warnings about persisted files that are not created yet...

PR #457, opened 2017-06-18
Update appengine-go to 1.9.54

rasa:


rasa:
Used updated https://github.com/rasa/scoops/blob/master/check_https.py

PR #462, opened 2017-06-23
Fix checkver in gifcam

rasa:


rasa:
On my system, the \s+ wasn't matching the UTF-8 non-breaking spaces (0xc2 0xa0).

PR #469, opened 2017-06-29
Use same approach for download URL as official Docker images for Tomcat

se35710:
Tomcat Archive does not always carry the latest versions of Tomcat. This commit reuses the official [Dockerfile](https://github.com/docker-library/tomcat/blob/5ac222d258dc70c77bb3a9a4fab81ea286c9abd1/8.5/jre8/Dockerfile#L49) way.


r15ch13:
`"find": "([a-fA-F0-9]+)"` and `"mode": "extract"` can be omitted, because they are the default.

se35710:
@r15ch13 I think `^([a-fA-F0-9]+)$` and `([a-fA-F0-9]+)\s+\*?(?:$basename)` is the default, but in this case, scoop assigns closer.cgi to `$basename`, and is therefor unable to find the hash value in the `.sha1`-file,

r15ch13:
You are absolutely right. I didn't think about that 👍

Issue #470, opened 2017-06-29
Autoupdating of Eclipse packages


se35710:
Version 4.7.0 (Oxygen) of Eclipse was released yesterday, hence [versions.txt](https://github.com/se35710/train-to-version/blob/master/versions.txt) was updated. Manually running `auto-pr` and/or `checkver` updates the manifests correctly.

Looking at the web server log files, I can see requests from what I presume is the auto-pr running in @r15ch13 Docker container, but the manifests are not pushed to this repo.

Also, the sha512 file is requested, but then the following request is for the zip file (I guess for calculating checksums?)

r15ch13:
I get `The remote server returned an error: (302) Found.` messages in my container. That is a problem with Powershell 6 on Unix...  HTTP 302 is handled as an error code instead of an actual redirect.

se35710:
Well isn't that a nice bug^H^H^Hfeature... It might be [this issue](https://github.com/PowerShell/PowerShell/issues/2896) though, because some redirects are from https to http. Future versions of Powershell should support `-allowInsecureRedirect`.

Until then, should I push them in manually by PR instead?

PR #473, opened 2017-06-30
Eclipse Oxygen (4.7.0)

se35710:
In reference to #470.


r15ch13:
Should we move the 4.6 to the versions bucket?

se35710:
`scoop install eclipse-<whatever>@<version>` works, so I don't see the need for it in the versions bucket.

PR #483, opened 2017-07-19
Fix extract_dir for oracle-(jre|jre-server) 8u141-b15

se35710:
Seems that auto-pr does not work with match groups. Should I create an issue?


r15ch13:
Strange, `auto-pr.ps1` simply calls `checkver.ps1`. If I run it locally it works. Maybe something strange with Powershell 6 on Linux again :grin:

PR #485, opened 2017-07-23
Simple DHCP server (for use with Hyper-V or other use case)

se35710:


r15ch13:
No `bin` or `persist` (for `dhcpsrv.ini`) properties?

se35710:
You are correct 👍. I've added `persist` and `shortcuts`. `bin` is not necessary.

PR #489, opened 2017-07-30
added a basic zeronet app manifest

jfmherokiller:


r15ch13:
Thanks. Interesting program 👍 

Too bad they don't use GitHub releases ... auto updating this will not work. :cry:

jfmherokiller:
yes, I tried to figure out how their upload setup works and it seems they have a repo containing a version of the program compiled for windows with all dependencies and have a link to the master.zip of that repo.

PR #500, opened 2017-08-17
Add hack ttf font version 1.2.0.

matthewjberger:
Adds the monospace programming ttf font "Hack"

### DIFF
--- a/hack-font.json
+++ b/hack-font.json
@@ -1,0 +1,32 @@
+{
+    "version":  "1.2.0",
+    "license":  "MIT",
+    "url":  "https://github.com/source-foundry/Hack-windows-installer/releases/download/1.2.0/HackWindowsInstaller.exe",
+    "homepage":  "http://sourcefoundry.org/hack/",
+    "checkver": {
+        "github" : "https://github.com/source-foundry/Hack-windows-installer"
+    },
+    "autoupdate": {
+        "url": "https://github.com/source-foundry/Hack-windows-installer/releases/download/$version/HackWindowsInstaller.exe"
+    },
+    "hash":  "d566623369d830afa7e4680ffdee13a2aac2b739fd58725a79ea793eabdb1428",
+    "installer": {
+        "file": "HackWindowsInstaller.exe",
+        "args": [
+            "/VERYSILENT",
+            "/DIR=$dir",
+            "/SUPPRESSMSGBOXES",
+            "/NORESTART",
+            "/CLOSEAPPLICATIONS",
+            "/NORESTARTAPPLICATIONS"
+        ]
+    },
+    "uninstaller": {
+        "file": "unins000.exe",
+        "args": [
+            "/VERYSILENT",
+            "/SUPPRESSMSGBOXES",
+            "/NORESTART"
+        ]
+    }
+}


r15ch13:
hm, maybe we could create an installer/uninstaller script for https://nerdfonts.com/ instead and provide them in an extra bucket?

r15ch13:
Maybe a bit complicated, because we need to keep track of the files in the windows fonts directory 😁 
For the time being, lets add the hack installer 👍

r15ch13:
Potential installer :grin:
```powershell
gci . -filter '*Windows Compatible*' | % {
    New-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Fonts' -Name $_.Name.Replace($_.Extension, ' (TrueType)') -Value $_.Name -Force | Out-Null
    Copy-Item $_ -destination "$env:windir\Fonts"
}
```

matthewjberger:
I'm putting together a bucket now based on your comments. Good ideas!

matthewjberger:
I started the bucket [here](https://github.com/matthewjberger/scoop-nerd-fonts) but it still has a long way to go. I need to prevent the prompts in the installer and create an uninstaller. Right now the manifest for hack is setup and hooked into an install script and uninstall script, and a template is available for generating the other font manifests at some point.

r15ch13:
Take a look at my second snippet above. It also shows the process dialog but it copies all files at once (it's also way faster)
You could try to get admin privileges and run the first snippet, which doesn't show the process dialog at all. :smile:

matthewjberger:
The bucket is set up now. I made a [script that will generate the font manifests](https://github.com/matthewjberger/scoop-nerd-fonts/blob/master/bin/generate-manifests.ps1). 

There is also a font [installation script](https://github.com/matthewjberger/scoop-nerd-fonts/blob/master/bin/install-nerd-fonts.ps1) and an [uninstallation script](https://github.com/matthewjberger/scoop-nerd-fonts/blob/master/bin/uninstall-nerd-fonts.ps1).

Right now the installation and uninstallation methods are silent, but require admin privileges (and Powershell 4.0+). As far as I've observed, uninstalling via removing the registry key for the font and the font files from the windows fonts folder doesn't seem to work (although it might with a restart).

I'll continue trying to get the installation and uninstallation scripts to function as expected, ideally silently but I'm not sure if it can be done without admin rights unfortunately.

The Hack font installer lists [many of the issues that can occur when installing fonts manually on windows](https://github.com/source-foundry/Hack-windows-installer/blob/master/FontInstallationIssues.md). The issues I encountered were that fonts aren't visible directly after installation, the font data in the registry and font files can be different, and that fonts cannot be updated without restarting.

It was worth a shot though, and I'll leave the repository up so that if anyone wants to try and create a script that can install and uninstall fonts successfully they won't need to recreate the scaffolding for the manifests. 

Thank you for the ideas!

matthewjberger:
The fonts in the [nerd-fonts bucket](https://github.com/matthewjberger/scoop-nerd-fonts) now install and uninstall properly. Would you be able to take a look when you get a chance @r15ch13?

The [script that generates the manifests](https://github.com/matthewjberger/scoop-nerd-fonts/blob/master/bin/generate-manifests.ps1) uses the `checkver.ps1` from scoop to update the url, version, and hash of the manifests.

I was able to verify that the registry key is properly added and that the fonts were copied properly to the `%windir%\fonts` folder upon installation.
After uninstalling, the registry key is properly removed and the fonts are removed from the system fonts folder. 

And finally I checked that the font was considered 'installed' by running the following command and ripgrepping for the font name ( I installed `monoid-nf` for testing):

```
[void] [System.Reflection.Assembly]::LoadWithPartialName("System.Drawing")
(New-Object System.Drawing.Text.InstalledFontCollection).Families | rg Monoid
```

The result after installing was: `Monoid NF`, and after uninstalling with scoop it was still present. It was not present after restarting, so I added a note saying that a restart after uninstalling is required.

Would this eventually be added to the `buckets.json` file in the main repo?

r15ch13:
Nice!
Is it possible to run the installer script as Admin? Running it as a user results in errors because of the registry stuff. 😢

r15ch13:
https://github.com/lukesampson/scoop/commit/f79cb5d9eb3b380495a3e7441cbdf7aa0d74d5ae 🎉

matthewjberger:
The scripts work properly with admin rights, but I couldn't find a way to uninstall  the fonts that didn't require admin rights. Also, the method that didn't require admin rights put dialogs up when the font files existed in the system fonts folder already. 😞 I figured that silent was better and having admin rights was sensible for something like installing fonts in the end.

I'm definitely always open to any ideas that will improve the bucket though!

PR #506, opened 2017-08-26
Add dungeon crawl stone soup tiles and console version.

matthewjberger:
This adds manifests for the ASCII version and graphical version of the roguelike [Dungeon Crawl Stone Soup](https://crawl.develz.org/).


matthewjberger:
Hold on, actually. I noticed running the exes creates two folders in the directory where it is run. I'll see if there is a way around this

PR #515, opened 2017-08-31
Add Foobar2000

54:
Adds Foobar2000 - a very popular and powerful music player.

Using Videhelp as a mirror because the official site uses a timed hash to prevent hotlinking.

### DIFF
--- a/foobar2000.json
+++ b/foobar2000.json
@@ -2,30 +2,29 @@
     "version": "1.3.16",
     "license": "Commercial",
     "homepage": "http://www.foobar2000.org/",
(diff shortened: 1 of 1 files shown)


r15ch13:
Both versions save all settings to the current directory, but only the portable version delivers theses files initially.
One problem I couldn't solve is with the `theme.fth` which can't be hardlinked. Foobar2000 isn't able to write changes to this file so they reflect back to the persistent directory.

54:
Yeah, it only seems to want to work with a symlink. Everything else appears to be good, though.

r15ch13:
Let's use the installer. All settings will be in `%appdata%` and we don't have to worry about them 😁

PR #518, opened 2017-08-31
Add Anaconda 3 version 4.4.0.

matthewjberger:
This adds the full Anaconda 3 distribution.


r15ch13:
For the sake of my servers traffic, I added hash extraction 😀

h404bi:
May add `uninstaller` property into the manifest.
https://github.com/h404bi/dorado/blob/master/anaconda3.json

PR #522, opened 2017-09-02
Update mpv.json

pulse-wave:
mpv.json hack that permits to create a shortcut to mpv.exe and have access to mpv cli at the same time, proposed by  r15ch13


r15ch13:
https://github.com/lukesampson/scoop/issues/1680

PR #525, opened 2017-09-06
add https://github.com/remind101/assume-role 

yunspace:
handy helper for AWS STS assume role commands.


r15ch13:
Fixed some things 😁

yunspace:
Nice one, thanks for fixing up!

PR #540, opened 2017-09-11
Update wiztree to version 3.10

brandon93s:
Update wiztree to version 3.10.


brandon93s:
Was playing around with auto-pr. Very useful!

PR #544, opened 2017-09-13
Fetch md5 from RSS feed instead of calculating

se35710:


se35710:
@r15ch13 Current hashes are incorrect again (maybe redirect issue with PS6?), hence this change.

PR #552, opened 2017-09-18
Add persist to notepadplusplus

rasa:


rasa:
Without this change, any plugins are lost after an upgrade

PR #553, opened 2017-09-21
Create gcloud.json for Google Cloud SDK

abejenaru:
Hi, I created a manifest for Google Cloud SDK.

### DIFF
--- a/gcloud.json
+++ b/gcloud.json
@@ -17,9 +17,7 @@
         "bin\\gsutil.cmd",
         "bin\\bq.cmd"
     ],
-    "post_install": [
-        "gcloud init"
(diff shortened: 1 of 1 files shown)


r15ch13:
`post_install` gets also executed after every update, so it will run `gcloud init` every time. Is this okay? (also `post_install` has to be a simple string.)

abejenaru:
I can change `post_install` to be a string, didn't catch that in the docs.
For me it's not a problem to run `gcloud init` every time, it just refreshes the environment, sometimes adding new configuration options; however, it is interactive, so if people expect to run unattended updates, it will not be the case.

Issue #554, opened 2017-09-22
Support for STS


javaboy79:
Add support for STS (Spring Tool Suite).  This is a very popular variation of eclipse designed for Spring Framework developers.  

https://spring.io/tools/sts

PR #556, opened 2017-09-22
Support for STS

javaboy79:
Implemented #554 

A few issues I could not figure out...

1. shortcut path is version dependent, so autoupdate isn't entirely automatic
2. shortcut created should set a target directory to the same dir as the .exe
3. uninstall seems to choke...  I suspect file paths are too long to handle gracefully, as explorer gives this warning also

### DIFF
--- a/hyper.json
+++ b/hyper.json
@@ -1,9 +1,9 @@
 {
-    "version": "1.4.6",
+    "version": "1.4.7",
     "license": "MIT",
-    "url": "https://github.com/zeit/hyper/releases/download/1.4.6/hyper-1.4.6-full.nupkg#/dl.7z",
+    "url": "https://github.com/zeit/hyper/releases/download/1.4.7/hyper-1.4.7-full.nupkg#/dl.7z",
     "homepage": "https://hyper.is",
-    "hash": "82522b17e11912121b2b51c99e7890b9f011674c1c52e85668fcdb407a1415ab",
+    "hash": "dd652039f6daf673bce7958b145c163b7761f2dc34d494b2517dbaf4fdb65429",
     "extract_dir": "lib\\net45",
     "shortcuts": [
         [
--- a/kakaotalk.json
+++ b/kakaotalk.json
@@ -1,9 +1,9 @@
 {
-    "version": "2.6.0.1591",
+    "version": "2.6.1.1618",
     "license": "Commercial",
     "url": "http://app.pc.kakao.com/talk/win32/KakaoTalk_Setup.exe#/dl.7z",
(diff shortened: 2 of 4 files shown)


r15ch13:
Is the stuff from `pivotal-tc-server-developer-3.2.6.RELEASE` not needed? If so, you could add the `extract_dir` property and set it to `sts-bundle\sts-3.9.0.RELEASE`  (in `autoupdate` set it to `sts-bundle\sts-$version.RELEASE`.

Also here is a improved version regex: 
```regex
spring-tool-suite-(?<version>[\d.]+).RELEASE-e(?<eclipse>(?<short>[\d.]+).[\d]+)-win32.zip
```

Result:
```
$version = 3.9.0
$matchEclipse = 4.7.0
$matchShort = 4.7
```

javaboy79:
@r15ch13  Good eye on the sha1, I didn't notice that.  This is fixed.

I plugged in your regex and it did not work exactly.  I could not see where you were going with that...  Wouldn't adding more matchers to the path be more error prone long term, requiring the script to be updated even if the url format changes slightly?  Especially when we don't really need those bits parsed?  I left the regex as is for now.

Also, I did not see a `formatjson.ps1` script in the bin dir.  It does seem to pass build now though.

As for setting the extract_dir..  Yes, as far as I know `pivotal-tc-server-developer-3.2.6.RELEASE` is required.  It is referenced from STS and I think without it STS is slightly crippled.  The whole bundle is what makes STS, well, STS.  If you tear it apart, you just end up with Eclipse + Spring plugin.  So I do not think extract_dir should be set.

Any comment on how to give the created shortcut a target directory?  Or the uninstall problem I mentioned?

rasa:
@javaboy79: @r15ch13 is referring to https://github.com/lukesampson/scoop/blob/master/bin/formatjson.ps1

r15ch13:
* Most download URLs are stable. IMHO if something changes it's better to take a look at it.
* The shortcut WorkingDirectory setting was discussed before (currently it's not supported)
   https://github.com/lukesampson/scoop/pull/1319
   https://github.com/lukesampson/scoop/pull/1320

* This fixes the long path problem and removes the version from the shortcut path:
  ```json
    "extract_dir": "sts-bundle",
    "pre_install": "mv \"$dir\\sts-$version.RELEASE\" \"$dir\\sts\"",
    "shortcuts": [
        [
            "sts\\STS.exe",
            "Spring Tool Suite"
        ]
    ],
   ```

javaboy79:
@r15ch13 I have incorporated all the recommend changes and also add a `bin` value.  Please review.

PR #558, opened 2017-09-26
Add PlantUML

niboan:
I've added [PlantUML](http://plantuml.com/), a tool to create UML diagrams, using a simple and human readable text description.


r15ch13:
Can't test it right now. SourceForge is currently in "static offline mode" ... 😞

PR #562, opened 2017-09-28
revouninstaller: Add version 2.0.3

lprc:
RevoUninstaller is a tool for uninstalling programs including a file scan for leftovers after uninstallation. Version 2.0.3 is a free version.
Normally the program displays an UAC prompt at start. In order to avoid that prompt the post_install script writes a value to the registry which let's the program start with flag "RunAsInvoker" (see https://technet.microsoft.com/en-us/library/dd638389(v=ws.10).aspx).
An alternative for writing into the registry would be creating a .bat-file in the pre_install-step and using that as bin (and for shortcut). It would contain:
`cmd /min /C "set __COMPAT_LAYER=RUNASINVOKER && start "" RevoUPort.exe"`
A pro for this approach is that it's automatically removed at uninstall. The contra is that the shortcut would have the default icon for .bat-files instead of the icon of RevoUPort.exe. Maybe the attribute "shortcut" might be even extended for the ability to set an icon path.


r15ch13:
What's wrong with an uninstaller showing the UAC prompt?

lprc:
Basically nothing's wrong with it, but in my case I don't have admin rights on the laptop at work ans thus I can't start RevoUninstaller without disabling th

joegreene:
Bit of a bump but I'm commenting here because I'd like to see Revo in `scoop-extras`. Why are `SumatraPDF.json` and `fscapture.json` added to this PR? Is it because this branch is out of date with `lukesampson:master`?

lprc:
@joegreene I think I was missing fscapture, revo and the shortcut for sumatrapdf in scoop so I added all of them but only submitted one PR. Apparently I didn't pay attention to do only one thing per PR, so my bad on this. Nevertheless I think the maintainers of scoop didn't really like the idea of avoiding the UAC for revo using a bat file or editing the registry. However that's just an assumption as I never got an answer as you can see above.

Long story short, you might want to pick the revo json file, remove the uac prompt evasion and create a PR on your own. I don't think this PR here will get any further recognition whatsoever (I myself had already forgotten about this).

PR #563, opened 2017-09-28
Added JPEGView

niboan:
Added [JPEGView](http://jpegview.sourceforge.net/), a lean, fast and highly configurable viewer/editor for JPEG, BMP, PNG, WEBP, TGA, GIF and TIFF images with a minimal GUI.

I had to include `extract_dir` in the autoupdate section, even though they are static. Otherwise, they are reset during auto update (bug?).

### DIFF
--- a/jpegview.json
+++ b/jpegview.json
@@ -1,0 +1,42 @@
+{
+    "homepage": "http://jpegview.sourceforge.net/",
+    "version": "1.0.35.1",
+    "license": "GPLv2",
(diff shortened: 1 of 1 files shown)


r15ch13:
Yeah, it was a bug 😁
Fixed it with https://github.com/lukesampson/scoop/commit/6380981b4c1fcefa4c9eb07474033036e356b1ff and removed the redundant properties.

PR #573, opened 2017-10-05
Add IntelliJ IDEA Community

SinisterBlade:
Added Community Edition of IntelliJ IDEA.

How about using the current idea.json for the Community edition instead. We can create an idea-ultimate.json for the Ultimate edition.


SinisterBlade:
Shall I push a commit swapping the file contents then?

SinisterBlade:
Done!

PR #569, opened 2017-10-03
WIP: Add Oracle JDK 9

ruifigueira:
Proposal to add Oracle JDK 9. I basically modified `oraclejdk` to install Oracle JDK 9+181, and passed Oracle JDK 8 into `oraclejdk8`.

Installation is working properly, but I didn't test `checkver` and `checkver`.

PR #576, opened 2017-10-05
Oracle Java 9

se35710:
Should be future proof with the new version naming model.

I am aware of the work from PR #569, but this also includes the other Java 9 manifests. Also, installer is not run for the JDK, it's just unzipped.

The version should be harmonized with [scoop](https://github.com/lukesampson/scoop) PR [#1727](https://github.com/lukesampson/scoop/pull/1727)

### DIFF
--- a/oraclejdk.json
+++ b/oraclejdk.json
@@ -1,46 +1,34 @@
 {
     "homepage": "http://www.oracle.com/technetwork/java/javase/overview/index.html",
-    "version": "8u144-b01",
+    "version": "9-181",
     "architecture": {
         "64bit": {
-            "url": "http://download.oracle.com/otn-pub/java/jdk/8u144-b01/090f390dda5b47b9b721c7dfaa008135/jdk-8u144-windows-x64.exe",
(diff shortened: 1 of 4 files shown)


yeikel:
@se35710 You need to resolve the conflicts before we can consider merging it

se35710:
@yeikel Once a decision is made regarding PR [#1727](https://github.com/lukesampson/scoop/pull/1727), then I will resolve the conficts. Oracle Java 8 still have releases, so it makes more sense to wait.

h404bi:
I think once we update JAVA to 9, we should move 8 to [versions](https://github.com/scoopinstaller/versions) bucket.

se35710:
@r15ch13 @lukesampson What do you say, should we try to make this happen? Either this or #569. I can remove the `oracle8.json` file and update to current versions...

rasa:
See https://github.com/BjoernPetersen/scoop-bucket/blob/master/oraclejdk9.json

se35710:
@rasa It runs the installer though. With Java 9 and Java 10 you can unzip the exe, and then unzip the tools.zip file.

se35710:
Replaced with #864.

PR #579, opened 2017-10-09
add rufus

winteryoung:


r15ch13:
Nice!
But it's not starting in portable mode because the shortcut doesn't have the working directory set.

winteryoung:
It seems currently scoop doesn't support setting working directory to shortcuts. Is that right?

winteryoung:
Or we may set the working directory to the app directory automatically?

winteryoung:
https://github.com/lukesampson/scoop/pull/1757
Addressing this on this PR.

PR #595, opened 2017-10-18
Update rabbitmqadmin to version 3.6.12

brandon93s:
Update to 3.6.12

Bin script is still valid.


brandon93s:
Held off on autoupdate for now. 3.7 is in RC and could require some other changes.

PR #596, opened 2017-10-18
updated to 3.9.1

doxicity:
Links/checksum from: http://dist.springsource.com/release/STS/index.html


r15ch13:
Nice. Can you try to fix the regex so it matches the eclipse version e4.7.1a? The 1a seems to be the problem here. :smile:

r15ch13:
👍

PR #601, opened 2017-10-21
Add Idea EAP

Hollow1838:
Hello, I'm using idea eap at work which is basically ultimate with additional features, free and usable for commercial uses. I personnaly think it's a great addition but I would understand if you don't think so.

Thanks for your work 👍


Hollow1838:
Might need to check the version check, I'm not too familiar with the tool finding the version. Here the build number is the version

r15ch13:
Added the ultimate eap version and fixed the community eap URL.

PR #619, opened 2017-10-30
pandoc: remove optional dependency on latex, add suggestion instead

oltolm:
Hello,

I tried installing pandoc 2.0 and scoop started installing miktex. MiKTeX is a huge download 191 MB and an *optional* dependency and I already have it installed.

I changed MiKTeX from a dependency to a suggestion.


r15ch13:
Good idea💡

PR #620, opened 2017-10-30
Update idea-ultimate-eap.json

Hollow1838:


Hollow1838:
Apparently, they make changes using the same build number...

r15ch13:
`a513f06a12842708e74dc1706079ce8ef810198e48f313ac08cd978bd0f72806` seems to be the right hash. Downloaded it with firefox and aria2c. Maybe your download failed?

Hollow1838:
Yes my bad, our network suck ^^". I didn't see a package corrupted message so I though it was only hash.

PR #621, opened 2017-11-01
Add Emacs HOME environment variable

cilq:
Adding this variable sets the location of the init file for emacs.
The value points to the users home directory which allows a
configuration through ~/.emacs and ~/.emacs.d like on unix machines.

### DIFF
--- a/emacs.json
+++ b/emacs.json
@@ -46,6 +46,9 @@
             }
         }
     },
+    "env_set": {
+        "HOME": "$HOME"
+    },
     "shortcuts": [
         [
             "bin\\runemacs.exe",


oltolm:
I don't set `%HOME%` and have no problems using emacs on Windows.  Emacs looks for `.emacs` in the `%APPDATA%` directory. Why is this a problem?

cilq:
You are right, not setting HOME does not prevent you from using Emacs. It really is an opinionated change because this behavior renders the ~ shorthand inside Emacs pretty useless and also trips up a lot of new users.

oltolm:
I'm sorry, I still don't understand what you mean. What is the problem with `~` and how does this patch solve it? I use `~` all the time in Emacs and don't see a problem. What is the value of `$HOME` in your patch?

cilq:
Ok, I will try to explain by walking you through the events that led to my decision of creating this patch. First off I have to say that I am usually a Unix user and that I'm still fairly new to Emacs, which might explain why I thought this patch would be a good idea :).

I wanted to give spacemacs a try and followed their installation instructions which didn't work. It took some time before I even considered the option that Emacs might behave differently on Windows and found numerous threads on the internet of people having the same issue.

So I moved the configuration to `%APPDATA%`, started Emacs and wanted to edit a file, let's say the scoop app-manifest of Emacs. In Powershell I can look at the file by doing something like this: `cat ~/scoop/buckets/extras/emacs.json` but opening `~/scoop/buckets/extras/emacs.json` in Emacs failed, because `~` does not point to `%USERPROFILE%` but `%APPDATA%`.

Of course I can always do something like `~/../../scoop...` but this becomes old quite fast. Ultimately having the configuration in `%APPDATA%` would be ok for me, but having `~` point there as well is really ill-conceived to me. But than again maybe because I lack some knowledge of Emacs I missed something?

By setting the `%HOME%` to `$HOME` which is identical to `%USERPROFILE%` [(Reference)](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-3.0) I created the behavior I expected in the first place.

oltolm:
Thanks for the explanation. I understand why you want this change, and I'm against it. Setting `%home%` to `%userprofile%` means that Emacs won't be able to find the configuration and everyone will either have to change the value of the `%home%` variable or move the configuration to `%userprofile%`.

In my opinion scoop should never force the user to move application data. Every application should work the same, whether it was installed through scoop or through the installer.

cilq:
Ok, I do understand your decision and will add this change to my personal configuration somehow.

PR #622, opened 2017-11-02
add normal nonportable Ditto

oltolm:
This commit adds a normal nonportable version of Ditto. There are multiple problems with the portable version:

* when you update Ditto you lose your settings and the clipboard history
* after installation the portable version does not use the settings and clipboard history of the nonportable Ditto that I had installed

### DIFF
--- a/ditto.json
+++ b/ditto.json
@@ -3,15 +3,22 @@
     "extract_dir": "Ditto",
     "architecture": {
         "32bit": {
-            "url": "https://sourceforge.net/projects/ditto-cp/files/Ditto/3.21.185.0/DittoPortable_3_21_185_0.zip",
-            "hash": "f438a687365a4db7738576cbcea6def65912f420ad5b48e255b85e7a220609e5"
+            "url": "https://sourceforge.net/projects/ditto-cp/files/Ditto/3.21.185.0/DittoSetup_3_21_185_0.exe",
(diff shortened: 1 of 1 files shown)


oltolm:
Sorry, there was a type. Fixed now.

r15ch13:
I improved the ditto manifest with https://github.com/lukesampson/scoop-extras/commit/382d6bcaf984c584e276d698d57f757098dd81a4. Does this work for you?

oltolm:
OK I have now tried the improved version of ditto. It now persists data across updates. Thank you. That is great, but it still does not use the data of the installed version. I now have to manually copy that data to the nonstandard directory `scoop\persist`. And if I stop using scoop some day, I will have to copy it back.

I just do not understand why scoop uses a nonstandard directory for application data and only for some apps. Others use the standard app directories. I don't see any advantages and I see some disadvantages. But I won't insist on changing things. If my manifest is not acceptable, I will create my own public bucket with nonportable apps.

oltolm:
Are you going to accept this pull request or do you want a separate bucket for normal apps that work with with standard application directories and allow the users to keep their application data?

r15ch13:
Let's merge these changes to the normal `ditto.json`. Those electron apps like slack/whatsapp/rambox and so on all use `%APPDATA%` for their settings. So why not let ditto use it's default settings path.

oltolm:
Great, I updated the pull request. I renamed `ditto-normal.json` to `ditto.json`.

PR #629, opened 2017-11-09
Added dbeaver and foxit-reader

ecornely:
Hello,

I was missing these two tools and plan on adding others that I'll need.

Let me know if I'm doing it correctly.

Kr,

Eric


rasa:
We're getting the error `The following files do not end with a newline` for both files

r15ch13:
Made some improvements. (Tests fail because of slack/gitkraken/whatsapp manifests being malformed. Fixed in another commit)

PR #650, opened 2017-11-17
add ImgBurn 2.5.8

oltolm:
ImgBurn is a lightweight CD / DVD / HD DVD / Blu-ray burning application.

### DIFF
--- a/imgburn.json
+++ b/imgburn.json
@@ -1,0 +1,20 @@
+{
+    "homepage": "http://www.imgburn.com/",
+    "url": "http://download.imgburn.com/SetupImgBurn_2.5.8.0.exe#/dl.7z",
+    "version": "2.5.8.0",
+    "hash": "49aa06eaffe431f05687109fee25f66781abbe1108f3f8ca78c79bdec8753420",
+    "bin": "ImgBurn.exe",
+    "shortcuts": [
+        [
+            "ImgBurn.exe",
+            "ImgBurn"
+        ]
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


brandon93s:
Looks like the file hashes are also available on the download page. Could extract those instead of downloading to compute the hash.

r15ch13:
The official setup contains AdWare or something. (See [Virustotal](https://www.virustotal.com/#/file/d7dea2819edc77bc44db637cd324e61942b54930cb3034f8f1a417b7dd27b514/detection))
All other download mirrors use URLs with time-based tokens.

I recommend https://etcher.io/
Will add it right now.

oltolm:
>The official setup contains AdWare or something. (See Virustotal)

Does it matter? We are not executing the installer, just unpacking it.

r15ch13:
The download will get interrupted by any AV listed it as bad on VirusTotal (I have ESET NOD32 installed)

oltolm:
I don't think that etcher offers the same functionality. It can't burn images to DVD for example. I'm still in favor of adding imgburn. Not everyone runs an AV that will block it.

r15ch13:
Copied the clean setup from the mirrors to my public keybase.  I removed the `checkver` and `autoupdate` properties since ImgBurn doesn't get updates anymore.
Etcher does is for burning images to USB. That's something completely different. My bad 😁

oltolm:
Thank you, please push to https://github.com/lukesampson/scoop-extras.

PR #652, opened 2017-11-18
Add PyPy manifest to extras bucket

jfmherokiller:


r15ch13:
Wouldn't it better to name it `pypy2.json`? Just in case `pypy3` for windows gets released. Also the zip file itself is called `pypy2` :smile:

Can you add the `hash` property to extract the checksum from the download page?
```json
    "autoupdate": {
        "url": "https://bitbucket.org/pypy/pypy/downloads/pypy2-v$version-win32.zip",
        "extract_dir": "pypy2-v$version-win32",
        "hash": {
            "url": "https://pypy.org/download.html"
        }
    }
```

jfmherokiller:
yes you are probably right i just thought naming it pypy would follow the convention used for python

r15ch13:
The build-in [regex ](https://github.com/lukesampson/scoop/blob/master/lib/autoupdate.ps1#L58) works fine because they include the checksums as standard `sha256sum` output format. :smile:

jfmherokiller:
Fixed and i just noticed in the commit i did before applying your suggestion i didnt even have the hash object so regardless of if i had the regex correct it would not have been applied.

r15ch13:
The `mode` property can be omitted because it's the default.

PR #669, opened 2017-11-28
resource-hacker: Add start menu shortcut

stkb:
Not sure if there was a reason why this was omitted or if it was just forgotten.


stkb:
Thanks!

PR #674, opened 2017-12-01
Eclipse jee

ecornely:
When I tried to install eclipse-jee download failed or something because no zip was downloaded it complained on sha512 so tried to fix the problem and just renaming to something ending with .zip worked

Do you think it's better to respect .zip instead of .7z ?


r15ch13:
Maybe the download failed. After you changed it to .zip it just redownloaded it and it worked. It works with .7z for me.

Also, please consider creating a new branch, commit your change to it and push it to GitHub. Then create a PR for the new branch. This way it will not include the other stuff from your master branch.

PR #677, opened 2017-12-05
Colorcop

ecornely:
Added colorcop tool


r15ch13:
Please use a branch and create a new PR. 😉👍

ecornely:
I'm not completely used to git. I mainly use it on projects I own and
develop alone.

How could I have a scoop bucket with things I test and that you might
reject and still easily propose pull requests ?

Do you have example of an empty bucket ? So that I could have a simple
bucket with only the few files you haven't yet integrated to scoop extra
but that we use with few friends. (like the mremoteng that I integrated in
my own bucket)

Kr,

Eric

r15ch13:
The best way would be to create a new bucket only for the programs you need and add with: `scoop bucket add https://github.com/ecornely/my-scoop-bucket`.
This way you can add as many programs as you like. For contributing a manifest to scoop, you can just copy it to a new branch of your scoop fork and create a PR for it.

PR #694, opened 2017-12-14
Add Rainmeter

niboan:
I've added Rainmeter, a desktop customization tool.

The package is an NSIS executable installer, and contains both 32 and 64 bit versions.
It's unpacked with 7Zip, and it works on my 64 bit Windows, but I don't know if it will work in a 32 bit environment.
If not, it should probably use the installer instead.

### DIFF
--- a/rainmeter.json
+++ b/rainmeter.json
@@ -2,8 +2,12 @@
     "homepage": "https://rainmeter.net/",
     "version": "4.1.0.2989",
     "license": "GPLv2",
-    "url": "https://github.com/rainmeter/rainmeter/releases/download/v4.1.0.2989/Rainmeter-4.1.exe#/Rainmeter.7z",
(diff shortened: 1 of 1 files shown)


r15ch13:
Since the installer does not offer any way to install only the 32bit version in an automated way, let's make it a 64bit only.

wangzq:
@niboan I cannot install it on my Windows 10 64-bit:
```
Linking ~\scoop\apps\rainmeter\current => ~\scoop\apps\rainmeter\4.1.0.2989
Creating shim for 'Rainmeter'.
Can't shim 'Rainmeter.exe': File doesn't exist.
```

It seems the unpacked files are having some non-ASCII characters. I have tried updating 7zip in scoop to ensure I am using the latest 7zip version to unpack but it still fails.

wangzq:
Never mind, it turns out I still have an older version 7z.exe in my PATH. Sorry for the spam. Thanks for adding rainmeter.

PR #705, opened 2017-12-21
Adding KeeWeb, a cross-platform password manager

Guard13007:


Guard13007:
I am a derp.

r15ch13:
Why are you a derp?

Guard13007:
@r15ch13 Because it's been a while since I added something to scoop and I tried to add an installer as an executable instead of however to tell it to run it and link to the place it saves or run it and specific to save within scoop or whatever.

It's an electron-based app.. not sure exactly how to put one of those within scoop..

PR #707, opened 2017-12-23
Fix hash in cpu-z

rasa:


rasa:
@r15ch13 Would you prefer hash/url fixes in one commit, or one per app?

r15ch13:
For users to track changes, single commits might be better. But I don't mind which way you choose to commit these changes. 😄

PR #721, opened 2017-12-31
Winscp 5.12 beta

yeikel:
I am not quite sure if you accept beta versions as well so please let me know

Edit : Build failed because it did not expect spaces in the name of the version.


r15ch13:
We try to avoid adding beta versions. WinSCP seems quite fast with its releases, so let's wait until 5.12 is complete.

PR #730, opened 2018-01-04
Fix the problem sourcetree can not open

c7i:


rasa:
Very weird that https://www.sourcetreeapp.com/ all refer to the 2.3.5 version. This suggests that version 2.4.7 is beta, but then https://blog.sourcetreeapp.com/2017/11/07/announcing-build-status-within-sourcetree-for-macos/ says that 2.7b1 is the beta version.

r15ch13:
When you use SourceTrees update function it updates itself to 2.4.7 😕

PR #733, opened 2018-01-06
sysinternals: Fix hash, add shortcuts, rename on x64 systems

rasa:
Depends on https://github.com/lukesampson/scoop/pull/1945


rasa:
Shortcuts need to be architecture aware.

PR #744, opened 2018-01-15
Update to emacs 25.3_1 for 64bit

ekroon:
See: http://ftp.gnu.org/gnu/emacs/windows/emacs-25.3_1.readme


r15ch13:
Can you change the `version` property back to `25.3` because the update service will revert it back.

ekroon:
And the link would stay the same?

r15ch13:
Yes leave the link to `25.3_1`. To install this fix one has to reinstall emacs.

PR #768, opened 2018-02-01
Persist Notepad++ configuration file

inosik:


inosik:
There are probably more files which should be persisted.

PR #774, opened 2018-02-03
Fix installer args, remove double quotes in foobar2000

akiakishitai:
The foobar2000 installer file don't work to enclose output path in double quotes.

This commit remove double quotes around `$dir` .


rasa:
Well crap. You’re right. See http://nsis.sourceforge.net/Docs/Chapter3.html

I broke it so I should fix it. What a drag.

PR #776, opened 2018-02-05
Python 3.7.0b1 added!

yigitemres:
Python 3.70b1 added for experimenting!
Installation is working but not sure for autoupdating!


r15ch13:
This is better suited for the versions bucket.

yigitemres:
Thx for advice! 
It is going to be added versions bucket.

PR #782, opened 2018-02-10
Update vlc to version 3.0.0

brandon93s:
Huge release for VLC!

https://www.videolan.org/vlc/releases/3.0.0.html


r15ch13:
Auto update got it an hour later (but uses other URLs)

PR #795, opened 2018-02-21
Add MusicBrainz Picard

dsbert:


r15ch13:
Made some small changes. When installing plugins they get copied to `%appdata%\MusicBrainz\Picard\plugins`, so persisting them is not need, I think.

PR #797, opened 2018-02-21
Add Paint.NET

stkb:
Not sure if the `.` in the name is a problem. If so it could be renamed `paintdotnet`.

Also it requires .Net 4.7.1 installed to run. Don't know if that's a problem.

Thanks for the help :)


r15ch13:
Neat!

PR #803, opened 2018-02-22
Hashfix for ScreenToGif v2.12.1

yigitemres:
Hashfix for ScreenToGif v2.12.1!


yigitemres:
There is a pull req. by @rasa .

PR #813, opened 2018-02-28
Added mitmproxy and pathod v3.0.3!

yigitemres:
Added mitmproxy and pathod v3.0.3!
Feel free to correct any mistakes!


yigitemres:
Forgetting Dev Setup! It will be fixed!

yigitemres:
mitmprroxy devsetup script (dev.ps1) don't take arguments from outside and it only needs for hacking mitmproxy itself. mitmproxy and pathod can be used without it!

PR #821, opened 2018-03-01
Qutebrowser updated to v1.1.2!

yigitemres:
Qutebrowser updated to v1.1.2!


yigitemres:
It is already updated by @r15ch13

r15ch13:
My update cronjob runs every full hour, so just be patient 😁

yigitemres:
And I wonder that how can be so fast! 😄 I hope that one day WSL can run Arch. 🙏

PR #833, opened 2018-03-08
Fixing kdiff3 SourceForge Address!

yigitemres:
Fixing kdiff3 SourceForge Address!


yigitemres:
@r15ch13 This is maybe not the place for asking this question but anyway...
Is there need to change all sourceforge links with downloads.sourceforge? If so I'm willing to do it!

PR #834, opened 2018-03-08
Fixing WinSCP SourceForge Address!

yigitemres:
Fixing WinSCP SourceForge Address! lukesampson/scoop#2071


r15ch13:
SourceForge problems seem to be solved

PR #835, opened 2018-03-08
Add Bazel

masaeedu:


masaeedu:
ping @r15ch13 😀

rasa:
It's a 64bit executable, but it's not wrapped in an `architecture` tag, so I'm guessing it's gonna be rejected, as the install will fail on 32bit systems.

masaeedu:
@rasa There's no build for x86 that they publish. Is the correct way to do this to make a `architecture` node and then only put x64 in there?

r15ch13:
Sorry for the late reply, could you add hash extraction because they provide `.sha256` files?

```json
"hash": { "url": "$url.sha256" }
```

masaeedu:
@r15ch13 Ok, done.

rasa:
Unless it has the `architecture` tag, 32bit users wont be told it's not available for their system. Instead, the exe will barf the CPU is wrong.

PR #839, opened 2018-03-12
Add tup

masaeedu:


r15ch13:
It's console only I guess. Let's put in the main bucket :smile:

masaeedu:
@r15ch13 Ok, opened lukesampson/scoop#2119.

PR #841, opened 2018-03-15
Fix https in neovim-qt, netbeans

rasa:


rasa:
My script flipped github back to http. I have to fix it before running it again.

PR #846, opened 2018-03-18
Add Skaffold

markspolakovs:


r15ch13:
Thanks.
Fixed a small bug with the update URL and removed the `mode` property because it can be omitted. Default mode is always `extract`

PR #851, opened 2018-03-20
Added XYplorer (free)

medoni:


medoni:
Please could you shortly explain why you use '"XYFreeHash-([\\d.]+).txt"' instead of the other regex?

r15ch13:
They seem to update this file on every update (look at the full version download page https://www.xyplorer.com/download.php)
I tend to use easiest possible regex for version extraction the minimize maintenance.
Filenames they mostly the same, unlike texts.

PR #853, opened 2018-03-20
add flux

s-i-v-r-a-j:


r15ch13:
This is missing the `*.json` file extension 😁

s-i-v-r-a-j:
oops! fixed. :laughing:

flux.json
```diff
@@ -0,0 +1,21 @@
+{
+  "homepage": "https://justgetflux.com/",
+  "description": "Makes the color of your computer's display adapt to the time of day, warm at night and like sunlight during the day",
+  "version": "4.66",
+  "url": "https://justgetflux.com/flux-setup.exe",
+  "hash": "79fc802d9225d98ee15a288393bb2d4708575315b9edad33dbfcf29d1af578b1",
+  "bin": "flux.exe",
+  "installer": {
+    "script": "cmd /c start /WAIT $dir\\flux-setup.exe /S /D=$dir"
```
r15ch13: Why do you start the setup via `cmd` instead of using the `file` property?

s-i-v-r-a-j: flux auto-starts immediately after installing, and so causes scoop installation to hang. not sure if there is an arg to pass to keep from auto-starting. Also not sure if there is a better way to handle this from command line.

s-i-v-r-a-j: you're meaning using something like 

```
  "installer": {
    "file": "flux-setup.exe",
    "args": ""
```

r15ch13: Good point with the blocking installer.
But could you add quotes around the setup exe? This prevents issues with whitespaces in `$dir` (`/D=$dir` doesn't need quotes)
`cmd /c start /WAIT \"$dir\\flux-setup.exe\" /S /D=$dir`

s-i-v-r-a-j: done! thanks

PR #859, opened 2018-03-22
Add Linux Live USB creator.

matthewjberger:
Adds lili version 2.4.9, a tool for creating bootable USB drives from Linux ISOs.

### DIFF
--- a/lili.json
+++ b/lili.json
@@ -1,0 +1,29 @@
+{
+    "version": "2.9.4",
+    "license": "GPL",
+    "url": "https://www.linuxliveusb.com/downloads/?version=stable-portable#/dl.7z",
+    "homepage": "https://www.linuxliveusb.com",
+    "extract_dir": "LinuxLive USB Creator 2.9.4",
+    "bin": [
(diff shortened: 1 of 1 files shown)


lili.json
```diff
@@ -0,0 +1,23 @@
+{
+    "version": "2.9.4",
+    "license": "GPL",
+    "url": "https://www.linuxliveusb.com/downloads/?version=stable-portable#/dl.7z",
+    "homepage": "https://www.linuxliveusb.com",
+    "extract_dir": "LinuxLive USB Creator 2.9.4",
+    "bin": "LiLi USB Creator.exe",
```
rasa: What about also adding an alias without spaces?:
```
"bin": [
  "LiLi USB Creator.exe",
  [
    "LiLi USB Creator.exe",
    "LiLiUSBCreator"
  ]
],
```

matthewjberger: Excellent idea. I added the alias as just 'lili' to keep it short and sweet. Thank you! :)

PR #860, opened 2018-03-22
Update Inkscape to 0.92.2 and remove autoupdate

niboan:
The old autoupdate didn't catch this version, as the site structure has changed, but the old page used in `checkver` is still available, listing the old version.

Autoupdate is not possible with the current site structure. The links to the .7z files can not be constructed from the version number alone, and the URL to the page where they can be parsed from, contains the version number.

I've updated both `inkscape` and `inkskape32bit`, but I've also added the 32 bit version to inkscape.
The architectures were split to make autoupdate work, but it won't work anyway now, so I think the manifest should follow best practice.

### DIFF
--- a/inkscape.json
+++ b/inkscape.json
@@ -1,12 +1,16 @@
 {
     "homepage": "https://inkscape.org/",
     "license": "https://inkscape.org/en/about/license/",
-    "version": "0.92.1",
+    "version": "0.92.2",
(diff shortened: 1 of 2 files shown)


r15ch13:
A good example, how not to create a download page ... 😭
Thanks for the fix.

PR #864, opened 2018-03-23
Oracle Java 10

se35710:
Closes #569

### DIFF
--- a/oraclejdk.json
+++ b/oraclejdk.json
@@ -1,47 +1,34 @@
 {
     "homepage": "http://www.oracle.com/technetwork/java/javase/overview/index.html",
-    "version": "8u162-b12",
+    "version": "10-46",
     "architecture": {
         "64bit": {
-            "url": "http://download.oracle.com/otn-pub/java/jdk/8u162-b12/0da788060d494f5095bf8624735fa2f1/jdk-8u162-windows-x64.exe",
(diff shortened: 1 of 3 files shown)


deki:
Great, I was just looking for that. Developers might need to install different JDK version in parallel (especially because JDK10 isn't LTS). How to achieve that?

se35710:
@deki The usual `scoop install oraclejdk@<version>` doesn't work with Oracle Java. Not sure if `scoop reset oraclejdk@<version>` works, but that would only be applicable for installed versions, which you could use manually by changing `JAVA_HOME`.

For other Java releases, including early access, I've set up my own [scoop bucket](https://github.com/se35710/scoop-java). Feel free to use it.

r15ch13:
@se35710 are you going to maintain your Java bucket and should we add it to the [bucket list](https://github.com/lukesampson/scoop/blob/master/buckets.json)?

se35710:
@r15ch13 Sure, it can be added to the bucket list. I've enabled auto-updates for it and will probably add new GA/EA versions as they are released, but I think the Java manifests should remain in the normal default/extra buckets (but maybe only LTS version, eg. 8 and then 11 (or its’ alias 18.9)).

That would render the Oracle manifests in the versions bucket useless (since the downloads would be behind a login page). Not sure what the retention policy for OpenJDK is.

r15ch13:
Replaced by https://github.com/lukesampson/scoop-extras/pull/898

PR #875, opened 2018-03-28
add goldendict 

specter119:
A feature-rich dictionary lookup program, supporting multiple dictionaries' formats, perfect article rendering with the complete markup, illustrations and other content retained, and allowing to type in words without any accents or correct case.

https://sourceforge.net/projects/goldendict/


specter119:
@rasa okay, but I have install goldendict with current json, i will change spaces to %20 later.

PR #878, opened 2018-03-30
add goldendict

specter119:


r15ch13:
You can push updates to your current pull request without creating a new one. Just use the same branch and keep commiting, then push. :smile:

PR #890, opened 2018-04-16
xmedia-recode 3.4.3.0: extract_dir changed

stkb:
Xmedia Recode v3.4.3.0 fails to install. The name of the inner dir to copy was changed.

Will have to see if this change is permanent or if they go back to the old scheme in the next version.


stkb:
Added back newline at end of file to fix build

PR #891, opened 2018-04-16
Fix for love extract_dir

leonidboykov:
Love fails to install.

Replaced `love-11.1-winXX` with `love-11.1.0-winXX`. Zip and directory don't follow the same version scheme in 11.1, for 11.0.0 both zip and dir have the same versions.


r15ch13:
Aah inconsistent versioning. I like that. 😁 
Thanks for the fix. 👍

PR #895, opened 2018-04-18
Added MSMPI runtime environment

SimplyKnownAsG:
The installer itself requires administrative permissions, so the script I wrote uses 7z to
extract the MSI, and then extract the platform-specific MSI.

### DIFF
--- a/msmpi.json
+++ b/msmpi.json
@@ -1,0 +1,37 @@
+{
+    "homepage": "https://www.microsoft.com/en-us/download/details.aspx?id=56727",
+    "version": "v9.0.1",
+    "license": "license.rtf",
+    "env_add_path": ".",
+    "url": "https://download.microsoft.com/download/4/A/6/4A6AAED8-200C-457C-AB86-37505DE4C90D/msmpisetup.exe",
+    "hash": "2fa35146e3d7dce6aba3d9cd81c1e16166a405bfe7094d1ae03c60fcdeb8e455",
+    "architecture": {
+        "64bit": {
+            "installer" : {
+                "script": "
+                cd $dir
+                7z x -t# msmpisetup.exe 4.msi
+                7z x 4.msi
+                del msmpires.dll
+                del msmpi.dll
+                move msmpires64.dll msmpires.dll
+                move msmpi64.dll msmpi.dll
+                del msmpisetup.exe
+                del *.msi
+                "
+            }
+        },
+        "32bit": {
+            "installer" : {
+                "script": "
+                cd $dir
+                7z x -t# msmpisetup.exe 2.msi
+                7z x 2.msi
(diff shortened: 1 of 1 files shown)


SimplyKnownAsG:
@lukesampson @r15ch13 is MSMPI appropriate for the scoop-extras? if not, is there some other recommended repo?

r15ch13:
Didn't know about the `-t#` switch. 👍 
Since it's a CLI tool it can be added to the main bucket. 

I made some changes to the manifest. 
- Remove `v` from version
- Add `bin` property instead of `env_add_path`
- Use fixed path with `$dir` instead of `cd $dir` because it doesn't change back after installing (`pushd` and `popd` would've been better)
- Split operations into `pre_install`, `installer` and `post_install`
- Use Powershell `remove-item` instead of `del`

```json
{
    "homepage": "https://www.microsoft.com/en-us/download/details.aspx?id=56727",
    "version": "9.0.1",
    "license": "license.rtf",
    "url": "https://download.microsoft.com/download/4/A/6/4A6AAED8-200C-457C-AB86-37505DE4C90D/msmpisetup.exe",
    "hash": "2fa35146e3d7dce6aba3d9cd81c1e16166a405bfe7094d1ae03c60fcdeb8e455",
    "bin": [
        "mpiexec.exe",
        "msmpilaunchsvc.exe",
        "smpd.exe"
    ],
    "pre_install": [
        "7z x -t# \"$dir\\msmpisetup.exe\" -o\"$dir\\tmp\""
    ],
    "architecture": {
        "64bit": {
            "installer": {
                "script": "
                    7z x \"$dir\\tmp\\4.msi\" -o\"$dir\"
                    Remove-Item \"$dir\\msmpires.dll\"
                    Remove-Item \"$dir\\msmpi.dll\"
                    Rename-Item \"$dir\\msmpires64.dll\" \"$dir\\msmpires.dll\"
                    Rename-Item \"$dir\\msmpi64.dll\" \"$dir\\msmpi.dll\""
            }
        },
        "32bit": {
            "installer": {
                "script": "7z x \"$dir\\tmp\\2.msi\" -o\"$dir\""
            }
        }
    },
    "post_install": [
        "Remove-Item \"$dir\\tmp\"",
        "Remove-Item \"$dir\\msmpisetup.exe\""
    ]
}
```

SimplyKnownAsG:
great edits, good stuff to know. Are you saying this should be moved to scoop?

Also, I used `env_add_path` because other applications using MPI (Python for example with mpi4py) need to access `msmpi.dll`. Is there another way to make the dll available?

r15ch13:
Okay if the `dll` is needed then `env_add_path` is the only way. Yes, please create a PR for the main bucket.

PR #897, opened 2018-04-19
add: boot-clj - http://boot-clj.com

netpyoung:
i made manifest for boot-clj.


netpyoung:
thanks, for correcting.

PR #898, opened 2018-04-19
Java bucket followup PR to #2209

r15ch13:
Java bucket followup PR to #2209


hermanjustnu:
@r15ch13 We forgot to remove [oraclejdk.ps1](https://github.com/lukesampson/scoop-extras/blob/master/scripts/oraclejdk.ps1). Should I create a PR?

PR #903, opened 2018-04-25
Update Spring Boot CLI to 2.0.1

andreysaksonov:
Auto-Update rolled it back to 1.5.12 because it was published after 2.0.1 https://github.com/spring-projects/spring-boot/releases

### DIFF
--- a/springboot.json
+++ b/springboot.json
@@ -14,7 +14,7 @@
     },
     "checkver": {
         "github": "https://github.com/spring-projects/spring-boot",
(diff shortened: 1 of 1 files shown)


r15ch13:
This seems to be a [big upgrade](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide), should we add 1.5 to the versions bucket?

andreysaksonov:
Tested manifests here: https://github.com/andreysaksonov/scoop-bucket

andreysaksonov:
https://github.com/lukesampson/scoop-extras/pull/903

PR #912, opened 2018-04-30
Update GIMP to version 2.10.0

niboan:
GIMP 2.10 is out. Updated the manifest to the new version.

It seems that there's only a 64bit build, so I've added an `architecture` tag. I don't know if that will cause problems when updating.


rasa:
Perhaps we create a gimp32 in versions, stuck at 2.8, or maybe even a new 32bit bucket for orphaned apps?

PR #914, opened 2018-04-30
Filezilla: removed hash url

hannesdorn:
I removed the hash url from the autoupdate section.


r15ch13:
What's the reason for this change?

hannesdorn:
I heard, that it is not necessary to provide an url to the hash values. But I may be wrong. 
Sorry, I am very new to this.

r15ch13:
It's not necessary, that's right but it should be preferred. It reduces traffic on my server and the download servers. Scoop auto-update server will download the files and calculate the hashes if no URL is provided.

PR #924, opened 2018-05-07
XMedia-Recode: extract_dir name was reverted

stkb:
Following on from #890, the naming scheme for the inner dir was reverted back again in v3.4.3.4


stkb:
That was fast :)

PR #927, opened 2018-05-13
Update serve to version 0.2.1

philippgille:


r15ch13:
You could also add the hash property since you serve `sha256` files. 👍 
```json
"autoupdate": {
        "url": "https://github.com/philippgille/serve/releases/download/v$version/serve_v$version_Windows_x64.zip",
        "hash": {
            "url": "$url.sha256"
        }
},
```

philippgille:
Ah, right, I think that's even the reason why I added those files to the release, but then forgot that this possibility existed - thanks!

But the syntax with `{}` seems wrong according to the AppVeyor check, and the [Wiki](https://github.com/lukesampson/scoop/wiki/App-Manifests) says it should be a string or array of strings, so I suppose the following would be correct:

`"hash": "$url.sha256"`

If you're not sure, don't waste time on this, I'll try it out in one of the next days in my test bucket and then get back to this PR.

r15ch13:
The `hash` property has to be a child of `autoupdate`. 😄

r15ch13:
Moved it to the main bucket
https://github.com/lukesampson/scoop-extras/commit/2b4bd6e23364980c3808da8e0310dd0f1e65bc6c > https://github.com/lukesampson/scoop/commit/2fbcb0327c78ba1a23cf9d3b89b1d383f50c1ae1

philippgille:
Ah now I got it, sorry :) Thank you for fixing it!

And I feel honored to have my application included in Scoop's main bucket! That's awesome!

rasa:
serve is a CLI so it should live in the main bucket.

PR #938, opened 2018-05-23
Add Bitmessage

0x326:
It would be preferable if this manifest exposed a binary such as `Bitmessage.exe` to the PATH instead of `Bitmessage_x86_0.6.3.2.exe`.  Do any of you know how to do this?

[Bitmessage](https://bitmessage.org/) does not have a 64-bit binary for the `0.6.3.2` release but they are stated they will recontinue 64-bit builds at some future release.  Right now, both 32-bit and 64-bit fields in the manifest point to the 32-bit build.

### DIFF
--- a/bitmessage.json
+++ b/bitmessage.json
@@ -1,0 +1,37 @@
+{
+    "homepage": "https://bitmessage.org/",
+    "version": "0.6.3.2",
+    "license": "https://github.com/Bitmessage/PyBitmessage/blob/master/LICENSE",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


Ash258:
You could use [this](https://github.com/lukesampson/scoop/wiki/App-Manifests#required-properties)

> To change the filename of the downloaded URL, you can append a URL fragment (starting with #) to URLs. For example,

So it would look like this: 
`"Bitmessage_x86_0.6.3.2.exe#/Bitmessage.exe"`

More examples: [IOGraph](https://github.com/lukesampson/scoop-extras/blob/736fa2ecba91833a1c465b1888b31e1ba9f51e61/iographica.json), [devdocs](https://github.com/lukesampson/scoop-extras/blob/5784ff7356ea0e984d3f3fe55f4fe1aa3fae8451/devdocs.json)

PR #941, opened 2018-05-25
 Added LibreOffice 6.0.4 

medoni:


r15ch13:
Checkver URL is wrong

PR #946, opened 2018-05-27
Added KiCad to extras bucket.

Ardelean-Calin:
Kicad is an open-source CAD program used for designing printed circuit boards and circuit schematics.
The program works from the extracted .exe file and we don't need to run the setup for it to install correctly. 

I have not encountered any issues thus far.


rasa:
Should the `$dir\\share\\kicad\\template`, `$dir\\share\\kicad\\modules\\packages3d` , and ` $dir\\share\\kicad\\modules` be added to a `persist` tag to be persisted across upgrades?

Ardelean-Calin:
@rasa Good question. Thing is, `packages3d` for example, could change from an upgrade to the next, since KiCad might get some new 3D models from one version to the next. Same goes for `modules` and `template`.

 I didn't do my homework on the `persist` tag yet, but does it only persist the files that didn't change? I.E. if new files are added it won't be a problem? If yes, I will gladly add that tag!

rasa:
Yes, new files will be copied over, while existing files will be left alone.

PR #948, opened 2018-05-29
Fixed persist tag having the `dir` variable

Ardelean-Calin:


Ardelean-Calin:
Please don't merge this yet. I am still testing if everything works all right after this change.

Ardelean-Calin:
Now it can be merged. Persisting those directories was a mistake, because the files inside them were not persisted, only the directory structure. Installing kicad just created empty directories with those names, the contents being in directories of the form `library.original`.

Maybe this is unintended behavior? If so, I could try and work for a fix.

rasa:
Already merged in #946 via https://github.com/lukesampson/scoop-extras/commit/68dd89a38d5cdd3fc928e6226d651b4a5a8f493b

Ardelean-Calin:
@rasa I did some changes. The `$dir` specifier in the persist tag messed everything up, since the paths in the persist tags are relative.

I have, however, removed the persist tag for the reasons specified above. (TL:DR the persistent directories were created but they were empty. They needed to have files in them which KiCad installs.)

Please re-open and merge this commit.

PR #956, opened 2018-06-02
WIP: Add Angry IP Scanner

0x326:
The `64bit` URL is for users that have 64-bit Java installed and the `32bit` URL is for users with the 32-bit version.  However, users with a 64-bit OS might have the 32-bit version of Java installed and be unable to use the binary at the `64bit` URL.

Java's architecture can be detected with the commands: `java -d64 -version` and `java -d32 -version`, and they should yield a non-zero return code when their respective versions are not installed.

Does any of you know how to add some simple logic to select the proper URL based on the user's Java architecture?

### DIFF
--- a/ipscan.json
+++ b/ipscan.json
@@ -1,0 +1,46 @@
+{
+    "homepage": "http://angryip.org/",
+    "version": "3.5.2",
+    "license": "GPL-2.0-or-later",
+    // 64bit URL needs 64bit Java
+    // and 32bit URL needs 32bit Java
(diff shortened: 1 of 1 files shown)


r15ch13:
If Java was installed via scoop then it will be most likely 64bit.
Adding the `notes` property with an advise to install ipscan using the `--arch` option might be enough.

0x326:
@r15ch13 How does that look?

r15ch13:
Looks great! 👍

PR #972, opened 2018-06-09
Add QuiteRSS

Paratsu:


r15ch13:
Cool, are there some config files to persist on update?

Paratsu:
None that I know of, I have read some threads on their forum asking where is the data stored, apparently on Linux it's stored in ~/.config/quiterss.
So I'm guessing on Windows it's stored in appdata. That shouldn't be a problem, right?

Paratsu:
Oh, turns out I was wrong, the info is stored on the program's folder, the files that should persist are 'feeds.db' and 'QuiteRss.in'.

Paratsu:
Is this ok? I've never used this property before :P

PR #976, opened 2018-06-11
Add manifest for Dns Jumper

benplm:
I wrote this app manifest for myself, but I think this little and useful software belongs here (I was quite surprised it wasn't here already, actually. Is there maybe already another app that makes this one redundant?).

Also, this is one of my first attempts in writing an app manifest. I believe it's fine, but please give me a feedback if something is wrong and/or can be improved.

### DIFF
--- a/dnsjumper.json
+++ b/dnsjumper.json
@@ -12,6 +12,7 @@
             "DNS Jumper"
         ]
     ],
+    "persist": "DnsJumper.ini",
     "checkver": "Dns Jumper v([\\d.]+)",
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice 👍
Made some small adjustments :smile:

benplm:
Thank you! However, I'm afraid that the author did not update the SHA1 on the website 😕

r15ch13:
@benplm dang it, one time I forget to check if the installation works and it bites me. 😁

PR #978, opened 2018-06-11
Add winpython manifest

benplm:
Not really sure if this belongs here, but since some people (me included) had trouble with the Python installer not being always handled properly by Scoop and/or not installing some .dll files and libraries even when installed manually (https://github.com/lukesampson/scoop/issues/504, https://github.com/lukesampson/scoop/issues/324, https://github.com/lukesampson/scoop/issues/2290, https://github.com/lukesampson/scoop/issues/856), I thought a 100% portable and working Python distribution could be useful for many.

I made it so that only Python is extracted, while the GUI tool developed by winpython is not, so no 3rd party software will be extracted (no launchers, no configuration tools other than Scoop).

I hope you'll find this useful.

### DIFF
--- a/winpython.json
+++ b/winpython.json
@@ -1,0 +1,44 @@
+{
+    "version": "3.6.5.0",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://netcologne.dl.sourceforge.net/project/winpython/WinPython_3.6/3.6.5.0/WinPython64-3.6.5.0Zero.exe#/dl.7z",
+            "hash": "sha1:4cb2e0144a182343a129519b837daa6de0b728dc",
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice 👍
Also made some adjustments :smile:

rasa:
Stoked that this has been added, but shouldn’t it live in the main bucket with python?

r15ch13:
@benplm can you make a PR for this manifest on the main repo to move it? 😁

benplm:
@r15ch13 sure, I'll do asap. I thought it would have been redundant there, that's why I did a PR here. 😅

rasa:
Not redundant at all. If anything, it's more a CLI than the standard python package.

PR #981, opened 2018-06-14
add autohotkey-installer

hed0n1st:
I use autohotkey a lot and was a bit frustrated by the original scoop install. After readings i propose this manifest as a new type of AHK installation using their installer and its command line options to get a more complete installation with shell menu entries.
Please consider adding it if you dont see too many mistakes, i am quite new in this.


hed0n1st:
@rasa Thanks for the advices

rasa:
@hed0n1st thanks for the contribution

PR #985, opened 2018-06-15
Update download urls for Emacs

kostrse:
Directory structure of the GNU server seems has been changed.


r15ch13:
Did they combine both zip file?

PR #989, opened 2018-06-20
Update winmerge2011 to version 0.2011.008.226

ranclo:
Maintainer also changed to CPL file format as the only download option, but it can still be extracted with 7-Zip. Details: https://bitbucket.org/jtuc/winmerge2011/downloads/Rationale_for_using_CPL_as_an_installer_format.txt

First contrib, so please let me know if I should do anything different in the future.


r15ch13:
Looks good 👍

PR #996, opened 2018-06-24
pandoc: Update mechanism and version

lahcim8:
Update to the newer github autoupdate mechanism.

Support new 64-bit pandoc releases.

Update to 2.2.1


r15ch13:
Nice 👍

PR #997, opened 2018-06-24
Add pandoc-crossref

lahcim8:
https://github.com/lierdakil/pandoc-crossref

"pandoc-crossref is a pandoc filter for numbering figures, equations,
tables and cross-references to them."

### DIFF
--- a/pandoc-crossref.json
+++ b/pandoc-crossref.json
@@ -10,7 +10,15 @@
     },
     "bin": "pandoc-crossref.exe",
     "checkver": {
-        "github": "https://github.com/lierdakil/pandoc-crossref/"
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice. Can you add auto update the information?

lahcim8:
I am afraid auto update is not possible with this package. The naming scheme is like this:

```
<platform>-<ghc-version>-<pandoc-version>
```

`platform` is always `windows` in our case, `ghc-version` is the used version of the compiler, `pandoc-version ` is the version of pandoc this is compiled against (?)

I don't think we can deduce the information for autoupdate.

lahcim8:
Thanks for the tip!

I used your suggestion, but changed the regex to match on the windows version and also not on latest release, because not every release has a windows zip available.

PR #1001, opened 2018-06-26
Add Mill, a next generation Scala build tool

nightscape:
Mill is a new Scala build tool which aims to be much simpler than SBT.
One question that came up while building the autoupdate/checkver part:
Is it possible to find the longest/highest number via Regex?
The [patch numbers are > 9 here](https://github.com/lihaoyi/mill/releases/tag/0.2.3), so I can't seem to locate the highest version `14`, instead it finds the lexicographically later version `2`.
I assume it is not, but maybe someone has an idea 😃

### DIFF
--- a/mill.json
+++ b/mill.json
@@ -1,0 +1,22 @@
+{
+    "version": "0.2.3",
+    "license": "MIT",
+    "url": "https://github.com/lihaoyi/mill/releases/download/0.2.3/0.2.3-14-c7c914#/mill.bat",
+    "bin": "mill.bat",
+    "homepage": "http://www.lihaoyi.com/mill/",
+    "hash": "611ff0ded19e798b308e1f974f3f1e7ca77bac2b9f350f0e895f25bc96117131",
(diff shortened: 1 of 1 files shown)


r15ch13:
That's a tough one ...
All files are listed in the correct order on the releases page, but on the tag page, they are ordered differently. Same with the API results ... 😢

r15ch13:
The [documentation states](http://www.lihaoyi.com/mill/#development-releases) that these are development builds. We can skip them because scoop should only provide stable versions :smile:
It may also be better suited for the main bucket (`SBT` lives there too)

This should be enough:
```json
"checkver": {
    "github": "https://github.com/lihaoyi/mill",
},
```

nightscape:
I can move it to main as well, I just thought Mill is rather new and not as established and frequently used as SBT.
Shall I create a fresh PR against the main bucket (using the stable version then)?

r15ch13:
Yes, please add the stable version there 👍

PR #1008, opened 2018-06-29
Add Yubikey tools

0x326:


r15ch13:
Nice 👍

rasa:
How about shortening/standardizing the names bit? Perhaps
```
yubikey-manager-qt > yubikey-man, 
yubikey-personalization-gui > yubikey-pers
yubikey-piv-manager > yubikey-piv
```
or
```
yubikey-manager-qt > yubikey-qt-man, 
yubikey-personalization-gui > yubikey-pers-gui
yubikey-piv-manager > yubikey-piv-man
```
?

0x326:
@rasa I'm all for adhering to any naming conventions this repo may have.  However, I think abbreviating package names will make it harder for the user to discover them.  Consider the following use case:

A Scoop user who wants to install these packages has probably learned of their existence from Yubico's developer website.  When he checks to see if an app manifest exists with `scoop search [query]`, he will likely search with the official Yubico name.  Take the Yubikey Personalization GUI for example.  The user will likely search `yubikey-personalization-gui` but may also search with `yubikey` or `yubikey-personalization`.  If the package name is `yubikey-pers-gui`, it won't show up for any of these search queries except for `yubikey`.  But since they are all reasonable queries, the user may not think to try different query variations when no matches are found and may mistakenly believe no manifest exists.

Maybe Scoop needs to support a `"name"` attribute, which would contain an app's full product name.  Then the manifest filename could be anything and the user could still find it by its official name

rasa:
I agree, let's leave current naming scheme alone, and a name attribute is a great idea.

HWataru:
How about adding path to "yubico-piv-tool.exe"  in yubico-piv-manager.json?

```json
    "bin": [
        "pivman.exe",
        "yubico-piv-tool.exe"
    ],
```
Or add yubico-piv-tool.json.

there is my personal use bucket
https://github.com/82p/scoop-yubico-bucket/blob/master/yubico-piv-tool.json

PR #1009, opened 2018-06-29
Removed duplicate standalone sysinternal suite tools

### LINKED ISSUE #935: accesschk is a duplicate
Access check is found in sysinternals suite, also available in the bucket.
Maybe this should be removed or does it warrant a standalone install?

[accesschk.json](https://github.com/lukesampson/scoop-extras/blob/master/accesschk.json)
[sysinternals.json](https://github.com/lukesampson/scoop-extras/blob/master/sysinternals.json)

gbursson:
fixes #935


rasa:
I’m cool with this, as it removes the burden of maintenance.

rasa:
We missed https://github.com/lukesampson/scoop/blob/master/bucket/du.json

PR #1012, opened 2018-06-30
changed json package names to all lowercase

gbursson:


rasa:
There are 2 packages that use underscores. Should we change those to dashes? Of course, users will have to uninstall and reinstall.

gbursson:
I would say so. Either dashes or just mangle words together.

r15ch13:
@rasa yes, why not

PR #1019, opened 2018-07-04
Adding foxe to extras bucket

rajmenon:
Firstobject's free XML editor for Windows is called foxe. 
It loads big files (multi-megabyte) fast and 
lets you format XML and edit HTML and any loosely formed XML or other markup. 

more info: http://www.firstobject.com/dn_editor.htm


r15ch13:
@brandon93s missing shortcut 😄

PR #1020, opened 2018-07-04
Adding posh-git to extras bucket

rajmenon:
posh-git is a PowerShell module which provides Git/PowerShell integration.

more info:  http://dahlbyk.github.io/posh-git/


rajmenon:
Thanks @r15ch3 . Will update.

PR #1021, opened 2018-07-04
Adding pixie to extras bucket

rajmenon:
oldie but goodie tiny color picker

http://www.nattyware.com/pixie.php


r15ch13:
@brandon93s missing shortcut 😄

brandon93s:
Whoops! I can add them.

PR #1031, opened 2018-07-07
changed url for new installer

ooploopl-:
Previous installer is broken due to an error. See below:
https://gitlab.gnome.org/GNOME/gimp/issues/1789


r15ch13:
Thanks, but the hash has to updated as well. :+1:

ooploopl-:
Here you go, also scoop may need to take special care of gimp they have an odd naming scheme for installers. They put simple integers at the end of the file for new installers of same version. AFAIK scoop only updates for major version bump.

r15ch13:
Should be fixed now. Thanks for the report 👍

PR #1032, opened 2018-07-07
Add potrace (v1.15)

lahcim8:
The manifest includes both 64 and 32 bit version as well as
checkver-autoupdate mechanism.

### DIFF
--- a/potrace.json
+++ b/potrace.json
@@ -1,0 +1,31 @@
+{
+    "homepage": "http://potrace.sourceforge.net/",
+    "version": "1.15",
+    "license": "GPL-2.0-or-later",
+    "architecture": {
+        "64bit": {
+            "url": "http://potrace.sourceforge.net/download/1.15/potrace-1.15.win64.zip",
+            "hash": "d7d4cdf7e61eec711522e7f4868d3ea4f1f95eeff0824f12c2bbf645b82f2614",
+            "extract_dir": "potrace-1.15.win64"
+        },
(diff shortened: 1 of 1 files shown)


r15ch13:
As a CLI tool, it might be better suited for the main bucket. Do you mind adding it there?
Can you change the `checkver` value to `potrace-([\\d.]+).win64.zip` to match the windows version?

rasa:
Shouldn't that be `potrace-([\\d.]+)\\.win64\\.zip`?

lahcim8:
@rasa Yes, I missed that, thanks.
@r15ch13 I edited the commit, and will also PR on main.

Also could you give me advice on what is the right workflow on editing PR based on suggestions? I feel like changing the commit (what I did now) isn't right but adding more to fix small mistakes seems worse.

r15ch13:
@rasa There are many more manifest using only a dot, but technically you're right :smile:
@lahcim8 just make a new commit to your `potrace` branch and push it. It will show up here.

lahcim8:
I felt like that would make the git history a bit ugly, because apart from this edit I also forgot the trailing end of line character, that would make 3 commits for a simple package add...

But now I edited the commit which was already reviewed by you and I don't feel this is quite right either.

r15ch13:
We use "squash and merge" feature of GitHub. The result is only a single commit in Scoops Git history.

rasa:
Added to main bucket

PR #1037, opened 2018-07-09
enable portable mode: https://code.visualstudio.com/docs/editor/portable

DavidBrockmeier:


rasa:
An exe with spaces... classic.

inosik:
I'd prefer if this feature was enabled in a new manifest instead of the existing one. I spend 10 minutes to copy my data and extensions to the new location and reverted it again because it wasn't working properly.

Ash258:
This PR is missing some stuff. 
Mainly VScode insiders release is each day (sometimes more than once a day). So there will be tons of hash check failures. And checkver will not be working (only with -f) when using `.name` for checking, which is not changing. But `.version` property is changing as it should.

Data folder is missimg TMP subfolder. Temporary data will be still stored inside some appdata folder.

r15ch13:
`vscode-insiders` should be changed to nightly.
At least adding a note about the portable mode should be added.
Users that updated vscode will be presented with a fresh and empty editor. 👎

Copying `%USERPROFILE%\.vscode\extensions` to `%USERPROFILE%\persist\vscode\data\extensions`
and `%APPDATA%\Code` to `%USERPROFILE%\scoop\persist\vscode\data\user-data` just worked fine for me.

Should we add additional manifest for the portable version?

r15ch13:
Made some changes with https://github.com/lukesampson/scoop-extras/commit/214eba81a2a6a913b3e58dafe7059a12c510119e and https://github.com/lukesampson/scoop-extras/commit/34ff30788f0a0acee20b760702a37812a7c83a43

rasa:
Sorry I accepted this without testing. The changes seemed pretty innocuous.

rasa:
So no checkver, even on the version attribute (git commit sha1)?

r15ch13:
No it's a different version for 64 and 32 bit

inosik:
> Should we add additional manifest for the portable version?

I don't think we should enable the portable mode at all. The idea behind it is to be able to put your extensions and editor state on a thumb drive and to take it with you. That's outside of Scoops scope, IMHO.

rasa:
Here’s a manifest keyed of the `version` property: https://github.com/Ash258/scoop-Ash258/blob/master/vscode-insiders.json

r15ch13:
The URL to the 32bit version fails: https://az764295.vo.msecnd.net/insider/adfa9ce9778a6e79197477fcc60cb5420a31979a/VSCode-win32-ia32-1.26.0-insider.zip

PR #1042, opened 2018-07-12
Disable portable mode of VS Code

inosik:
See discussion in #1037.

### DIFF
--- a/vscode-portable.json
+++ b/vscode-portable.json
@@ -1,0 +1,77 @@
+{
+    "homepage": "https://code.visualstudio.com/",
+    "version": "1.25.0",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://code.visualstudio.com/License/"
+    },
+    "notes": [
+        "VSCode now supports Portable Mode! Please move the following directories:",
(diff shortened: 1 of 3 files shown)


vscode-portable.json
```diff
@@ -0,0 +1,77 @@
...
+        "if(Test-Path(\"$dir\\vscode-install-context.reg\")) {",
+        "  $codepath = \"$dir\\Code.exe\".Replace('\\', '\\\\')",
+        "  $content = Get-Content \"$dir\\vscode-install-context.reg\"",
+        "  $content = $content.Replace('$code', $codepath)",
+        "  $content | Set-Content -Path \"$dir\\vscode-install-context.reg\"",
+        "}"
+    ],
+    "checkver": {
+        "url": "https://code.visualstudio.com/sha?build=stable",
+        "jp": "$.products[?(@.platform.os == 'win32-x64-archive')].name"
```
rasa: Are we sure we want to key off the `.name` attribute, and not the `.version` attribute?

r15ch13: `.version` is just a hash. `.productVersion` and `.name` are the same.

Ash258: `.version` is github commit, which was used to build binaries.

rasa: Hence why it should be in the checkver. Or am I missing something?

PR #1043, opened 2018-07-14
Added flutter to scoop-extras

Ardelean-Calin:
[Flutter](https://flutter.io/) is a new way to write native Android and iOS applications in a short time. This pull request adds flutter to the scoop-extras bucket.

A post_installation script is used to install the latest build-tools and a platform SDK so that the user can start developing right-away.


flutter.json
```diff
@@ -0,0 +1,40 @@
...
+    "license": "https://github.com/flutter/flutter/blob/master/LICENSE",
+    "url": [
+        "https://storage.googleapis.com/flutter_infra/releases/beta/windows/flutter_windows_v0.5.1-beta.zip",
+        "https://raw.githubusercontent.com/lukesampson/scoop-extras/master/scripts/flutter-dev-setup.ps1"
+    ],
+    "extract_dir": "flutter",
+    "hash": "1d6267358f90ad2ff46c23d53add0a503c10385efe6b5bb8c50ca4d670f9197a",
+    "depends": [
+        "android-sdk",
+        "https://raw.githubusercontent.com/se35710/scoop-java/master/oraclejdk8.json"
```
rasa: @r15ch13 Is a depends on a URL allowed? Shouldn't this be `java/oraclejdk8`, instead?

flutter.json
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    "description": "Flutter is Google’s mobile app SDK for crafting high-quality native interfaces on iOS and Android in record time. This is the beta version, since Flutter wasn't officially 'released' yet.",
+    "env_add_path": "bin\\cache\\dart-sdk",
+    "bin": [
+        "bin\\flutter.bat",
+        "flutter-dev-setup.ps1"
```
rasa: This is in `bin` & `post_install`? Why?

flutter.json
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    "post_install": [
+        "flutter-dev-setup.ps1",
+        "Write-Host Some licenses need to be accepted before developing. We recommend you do so by running 'flutter doctor --android-licenses'. -ForegroundColor Yellow",
+        "flutter doctor"
+    ],
+    "checkver": {
+        "url": "https://github.com/flutter/flutter/wiki/Changelog",
+        "re": "Changes in v(.{3,7})\\s.*beta.*"
```
rasa: Why limit the range of numbers?

flutter.json
Ardelean-Calin: Unfortunately, that doesn't work. URL depending works, I have tested it. Using `java/oraclejdk8` says it cannot find the manifest for `oraclejdk8`. The only viable solution was with URL. I will open an issue about that on the main scoop repository.

flutter.json
Ardelean-Calin: In case it fails when post_installing. So the user can just run it again without looking for where it is. Do you think I should remove this?

flutter.json
rasa: Did you add the `java` bucket first?

Ardelean-Calin: Yes, but that's not ideal, since the user **doesn't get a message** to know that he should add another bucket. The error message when using `java/oraclejdk8` without the `java` bucket being added is 
```
Couldn't find manifest for 'oraclejdk8' from 'java' bucket.
``` 
This message is weird in that it doesn't tell the user that he should add the `java` bucket. It tells the user that the `java` bucket does not have an `oraclejdk8`, which is false.

What would have made sense for me, since scoop already knows of the existence of the `java` bucket, is that the missing bucket would be added automatically or the manifest taken from it anyway since `oraclejdk8` is specified as a **dependency**.

That's why I used the URL, to force the installation of OracleJDK8 even if the user does not have the `java` bucket added.

rasa: I think scoop should automatically add a bucket if it’s depended on, unless `—noadd` is specified.

Ardelean-Calin: I agree, but it doesn't. Is this a bug with scoop?

rasa: No, a feature request.

flutter.json
rasa: No, it’s just an interesting implementation.

flutter.json
r15ch13: Lets change only the message to be more clear. It's just a small and easy fix.

r15ch13: Improved the message with https://github.com/lukesampson/scoop/commit/7a6218c58677170fe32cf1c2bfcfe7488e4c3655

Ardelean-Calin: @r15ch13 Nice! Adapted the pull request now. Removed the URL in the dependencies entry. Please also take a look at `flutter-dev-setup.ps1` when you have the time.

PR #1044, opened 2018-07-15
Fix Nomacs shim and shortcut binary path

Gargauth:
Nomacs seems to have changed folder structure. Executable now resides in bin subfolder upon extraction.

```powershell
PS > scoop install nomacs
Installing 'nomacs' (3.10.2) [64bit]
Loading nomacs-3.10.2.zip from cache
Checking hash of nomacs-3.10.2.zip... ok.
Extracting... done.
Linking ~\scoop\apps\nomacs\current => ~\scoop\apps\nomacs\3.10.2
Creating shim for 'nomacs'.
Can't shim 'nomacs.exe': File doesn't exist.
```

Both bin and shortcut needs to be changed accordingly.

### DIFF
--- a/nomacs.json
+++ b/nomacs.json
@@ -4,14 +4,14 @@
     "license": "GPL-3.0",
     "url": "http://download.nomacs.org/portable/nomacs-3.10.2.zip",
     "hash": "e789b6c712ddbd490e5a1353462b215b8721418a8af1dc03de34576bb47d33cb",
-    "extract_dir": "nomacs-3.10.2",
-    "bin": "bin/nomacs.exe",
+    "extract_dir": "nomacs-3.10.2\\bin",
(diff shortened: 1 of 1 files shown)


r15ch13:
Just downloaded, can't confirm:

image

Gargauth:
You downloaded 3.10.1 as per your screenshot. This is version 3.10.2

image

r15ch13:
Dang it, you're right ... 😁

r15ch13:
Changing the `extract_dir` option would be better. Because of the persistent `settings.ini`

`"extract_dir": "nomacs-3.10.2\\bin",`
and for autoupdate: `"extract_dir": "nomacs-$version\\bin"`

Gargauth:
Makes sense. Done.

PR #1049, opened 2018-07-16
Add Textadept version 9.6

Calinou:
This adds [Textadept](https://foicica.com/textadept/) to scoop-extras.

I couldn't figure out automatic updates here, since the download URL only has "LATEST" in it (and not the actual version number).

### DIFF
--- a/textadept.json
+++ b/textadept.json
@@ -4,7 +4,7 @@
     "version": "9.6",
     "license": "MIT",
     "extract_dir": "textadept_9.6.win32",
-    "url": "https://foicica.com/textadept/download/textadept_LATEST.win32.zip",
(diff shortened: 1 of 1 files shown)


Ash258:
Why not make something like: 
```json
"checkver": "Download v\\(((\\d+\\.?){2})\\)",
"autoupdate": {
    "url": "https://foicica.com/textadept/download/textadept_LATEST.win32.zip"
}
```
Link will be same, but hash will be recalculated on update

Ash258:
Also when you go [here](https://foicica.com/textadept/CHANGELOG.html) there are version numbers for files (etc: https://foicica.com/textadept/download/textadept_9.6.win32.zip)

PR #1050, opened 2018-07-16
Add Total Commander

Gargauth:
As title says...adds Total Commander file manager.

https://www.ghisler.com/

### DIFF
--- a/totalcommander.json
+++ b/totalcommander.json
@@ -1,0 +1,57 @@
+{
+    "version": "9.20",
+    "license": "Shareware",
+    "homepage": "https://www.ghisler.com",
+    "description": "Total Commander is a Shareware file manager for Windows® 95/98/ME/NT/2000/XP/Vista/7/8/8.1/10, and Windows® 3.1.",
+    "architecture": {
+        "64bit": {
+            "url": "http://totalcommander.ch/win/tcmd920x64.exe",
+            "hash": "1887c4cb3498572965ad813c6b8c7888514f515b9acba14a6618111db1a43260",
+            "bin": "TOTALCMD64.EXE",
+            "shortcuts": [
+                [
+                    "TOTALCMD64.EXE",
+                    "Total Commander"
+                ]
+            ]
+        },
+        "32bit": {
+            "url": "http://totalcommander.ch/win/tcmd920x32.exe",
+            "hash": "90b7f759df3a7e35c0511b1402508d265704b120622dcd4effb6aa74c238f2c6",
+            "bin": "TOTALCMD.EXE",
+            "shortcuts": [
+                [
+                    "TOTALCMD.EXE",
+                    "Total Commander"
+                ]
+            ]
+        }
+    },
+    "pre_install": [
+        "extract_7zip \"$dir\\$fname\" \"$dir\\_tmp\" $false",
+        "extract_7zip \"$dir\\_tmp\\INSTALL.CAB\" \"$dir\" $true",
+        "Move-Item \"$dir\\e\\*\" \"$dir\"",
+        "Remove-Item -Recurse \"$dir\\d\", \"$dir\\e\", \"$dir\\_tmp\", \"$dir\\$fname\""
+    ],
+    "persist": [
+        "DEFAULT.BAR",
+        "default.br2",
+        "wcx_ftp.ini",
+        "wincmd.ini",
+        "wincmd.key"
+    ],
+    "checkver": {
+        "url": "https://www.ghisler.com/download.htm",
+        "re": "Download version\\s+([\\d\\.]+) of Total Commander"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "http://totalcommander.ch/win/tcmd$cleanVersionx64.exe"
+            },
+            "32bit": {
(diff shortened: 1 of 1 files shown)


totalcommander.json
```diff
@@ -0,0 +1,57 @@
...
+            "bin": "TOTALCMD.EXE",
+            "shortcuts": [
+                [
+                    "TOTALCMD.EXE",
+                    "Total Commander"
+                ]
+            ]
+        }
+    },
+    "pre_install": [
```
Ash258: Preinstall could be easier when you rewrite installer in first way and then just extract, move.

https://github.com/Ash258/scoop-Ash258/blob/licences/TotalCommander.json#L32-L37

I am creating blank wincmd.key if its not already persisted, because persisting file, which does not exists in first place is quite unexpected. (Need to find time to investigate more).

Ash258: When you persist single file and the file does not exists it will lead to Empty persist folder. And when file is created (by program or manual) its not persisted.

Gargauth: Unfortunately, scoop won't create hardlinks and directory junctions unless the files/folders already reside in persist directory. 

There are some other issues I found with current implementation of persist, especially with programs that do not overwrite hardlinked files, but shuffle files by renaming original file by appending .bak (for example) and writing new file from scratch, which will effectively erase hardlink (examples of such program is nomacs, Everything or KeepassXC)

However, in this instance, we could pre-seed configuration files with empty files, and to activate portable mode by setting `UseIniInProgramDir=7` within program's ini file.

Like this:

```json
    "pre_install": [
        "extract_7zip \"$dir\\$fname\" \"$dir\\_tmp\" $false",
        "extract_7zip \"$dir\\_tmp\\INSTALL.CAB\" \"$dir\" $true",
        "Move-Item \"$dir\\e\\*\" \"$dir\"",
        "Remove-Item -Recurse \"$dir\\d\", \"$dir\\e\", \"$dir\\_tmp\", \"$dir\\$fname\"",
        "\"[Configuration]`nUseIniInProgramDir=7\" | Out-File \"$dir\\wincmd.ini\"",
        "ForEach ($emptyFile in $manifest.persist) {New-Item \"$dir\\$emptyFile\" -ItemType file -ErrorAction SilentlyContinue | Out-Null}"
    ],
```

I just tried it, and it doesn't seem to break any functionality of the program. It will overwrite configuration files with its own, as needed.

Ash258: When i am persisting single file which does not exists after installing I am just creating $null or "{}" (For JSOn files). Havent find some spare time to inspect and report whole persist flow. But this look nice.

Gargauth: It seems to work with either CR and NL characters. But I will update it to CR.

r15ch13:
Is this ready for merging? @Gargauth @Ash258

totalcommander.json
```diff
@@ -0,0 +1,59 @@
...
+        "extract_7zip \"$dir\\$fname\" \"$dir\\_tmp\" $false",
+        "extract_7zip \"$dir\\_tmp\\INSTALL.CAB\" \"$dir\" $true",
+        "Move-Item \"$dir\\e\\*\" \"$dir\"",
+        "Remove-Item -Recurse \"$dir\\d\", \"$dir\\e\", \"$dir\\_tmp\", \"$dir\\$fname\"",
```
Ash258: _tmp should exists after robocopy, fname could be removed when you use something like

Ash258:
Almost. Only 2 questions. 
- Would not `wincmd.ini` be overwriten?
     - `New-item -ea SC` is not overwriting file without `-Force` right?
- Why to copy `e` folder? There are 2 files which does not look like needed.

Gargauth:
Pre-install phase is before persistance, so any user's file (eg. `wincmd.ini` will not be overwritten, as the hardlinks will be created after installation.

`e` folder contains `.chm` files (help files) in English. `d` contains help files in German. If you use official installer, those files are copied to root folder (probably to help Total Commander to locate help files if you want to use help from within Total Commander. That's why I move those files - to mimic official installer, and then clean up by removing `e` and `d` directories.

@r15ch13 I use this myself, have not found any issues so far. I'd say it's ready.

totalcommander.json
Gargauth: That's what I used before I created this PR. Reason why I chose to extract installer to its own directory `_tmp` was because it's easier to clean up afterward. If you let scoop unpack the installer to default directory by rewriting filename, it will unpack other files along with `INSTALL.CAB` as well, which you would have to selectively remove afterward.

r15ch13:
`pre_install` will also be run on every update, so it will always overwrite `wincmd.ini` with `[Configuration]\rUseIniInProgramDir=7`. This should only be executed once, I guess.

r15ch13:
Hm, yeah it would simply rename the file to `wincmd.ini.original`

Gargauth:
Yes. I can't think of edge cases that would cause user's configuration data loss. If any program comes with its own default configuration file, there would always be same risk (same effect as `Out-File` in `pre-install` phase).

Implementing simple check if there's already configuration file present in `$dir` would be fairly straightforward and simple, but at this point I don't think it's necessary as it would be redundant.

PR #1052, opened 2018-07-18
Add persist to audacity, mpv, bleachbit, and ccleaner. Fix Everything config

nickbudi:
Makes audacity and mpv portable since they will read/write settings from persistent directory instead of %AppData%.
See:
+ https://wiki.audacityteam.org/wiki/Portable_Audacity
+ https://mpv.io/manual/master/#files-on-windows

### DIFF
--- a/audacity.json
+++ b/audacity.json
@@ -10,6 +10,7 @@
             "Audacity"
         ]
     ],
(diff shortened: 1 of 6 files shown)


r15ch13:
Nice! Thanks 👍

PR #1055, opened 2018-07-19
Cleanup wox directory and update sqlitebrowser to use portable release

nickbudi:
I don't believe electron apps are able to auto update with the previous manifest format right? So add extract_dir since we don't need the other nupkg files.


r15ch13:
Thanks for these changes and additions! 👍 
Scoop shows the commit messages as a changelog while updating, therefore I recommend creating separate pull-requests :smile:

PR #1071, opened 2018-07-22
Add JDownloader

Gargauth:
This proved to be a bit tricky to implement in scoop, and I'm not sure if my approach is the most viable one.

Due to self-installer that by itself is dependent on java, it ended up being more complex that I would have liked.

With my approach, scoop will wait until jdownloader downloads icon file in `pre_install` phase, then it proceeds with installation, creates dummy shortcut (so that scoop can track and remove in case of uninstall), and then manually modify icon and target of the shortcut in `post_install` phase.

Given the nature of this manifest, whether you want to merge this or not, is up for your consideration. I just thought I'd share my manifests that I built over the past few months, maybe some folks using brilliant scoop would find this handy.

### DIFF
--- a/jdownloader.json
+++ b/jdownloader.json
@@ -1,0 +1,33 @@
+{
+    "homepage": "http://jdownloader.org/",
+    "description": "JDownloader is a free, open-source download management tool",
+    "license": "GNU GPLv3",
+    "version": "2",
+    "url": "http://installer.jdownloader.org/JDownloader.jar",
+    "depends": "java/oraclejre",
+    "persist": [
+        "cfg"
+    ],
+    "bin": "JDownloader.jar",
+    "pre_install": [
+        "Start-Process $(which javaw.exe) -ArgumentList '-jar', \"$dir\\JDownloader.jar\", '-?'",
+        "while (!(Test-Path $dir\\themes\\standard\\org\\jdownloader\\images\\logo\\icon.ico)) { Start-Sleep 1 }"
+    ],
+    "shortcuts": [
+        [
+            "..\\..\\..\\shims\\JDownloader.cmd",
+            "JDownloader 2"
+        ]
+    ],
+    "post_install": [
+        "$wsShell = New-Object -ComObject WScript.Shell",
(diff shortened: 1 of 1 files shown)


Gargauth:
Web installer version contains adware. Downloading .jar and letting it download through their update servers is cleanest way to get only the software you want. Second reason, it installs embedded JRE, taking up additional space as well as taking away control on version of JRE you want to run (when I already have JRE installed anyway).

Ash258:
So [this file](https://mega.nz/#!mctCnIhC!fNqx1UOf06cVB7h9Cis1Kk5J_tAGMTFsSQcuQ2FPfYw) which is at their download page (Other platform) is containing adware? On official page?

> Downloading .jar and letting it download through their update servers is cleanest way to get only the software you want

Thats exactly what that other platform do. Its single file which after executing as `javaw.exe -Xmx192m -jar .\JDownloader.jar` will lead to downloading from their servers.

Ash258:
`pre_install` could be converted for "installer" :{"script": ""}. It would be much cleaner

Gargauth:
Looks better now, thanks @Ash258 . Only downside is that due to shortcut being .bat it now flashes a new cmd line window quickly. But that's a nitpick, and users can of course create their own shortcuts with direct targets.

If shortcuts parameters could evoke commands (much like pre_install for example can) it would be much easier to do shortcut properly without relying on .bat.

r15ch13:
Or try using a powershell script instead of cmd: https://github.com/lukesampson/scoop/blob/master/bucket/neofetch.json

Ash258:
Make it `Start-Process javaw.exe -ArgumentList "-jar", ".\JDownloader.jar" -wait` It will be waiting until process exit (No need for while)

Gargauth:
Looking much better now than first commit. Thank you for suggestions.

One more thing I would like to add would be autoupdate/checkver. but sadly it seems developers don't publish changelog for this file, they simply just update `JDownloader.jar`. This file seem to be updated only occasionally (current version was first submitted to VirusTotal on 2018-03-24). Hopefully it wouldn't cause too much manual maintenance over time, with updating hashes when the files change.

Alternative would be to omit hash, but then users see warning that there are no hashes. 🤷‍♂️

PR #1073, opened 2018-07-22
Add MSI Afterburner with RTSS

Gargauth:
> MSI Afterburner is the world's most recognized and widely used graphics card overclocking utility which gives you full control of your graphics cards. Works with all graphics cards vendors.

https://www.msi.com/page/afterburner


r15ch13:
Interesting how all these download pages use only HTTP...

Gargauth:
Yes, it is...There's alternative download server over at guru3d.com but they don't work with scoop. I tried it and if I recall correctly there was issue with Referrer header that you couldn't set in scoop.

PR #1079, opened 2018-07-24
firefox: extract instead of running installer

shoogle:
Also adds license and description.

### DIFF
--- a/firefox-beta.json
+++ b/firefox-beta.json
@@ -1,16 +1,18 @@
 {
+    "description": "Beta builds of Firefox: the popular open source web browser.",
     "homepage": "https://www.mozilla.org/en-US/firefox/beta/",
     "version": "nightly",
+    "license": "MPL-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win64&lang=en-US#/firefox-setup.exe"
+            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win64&lang=en-US#/firefox.7z"
         },
         "32bit": {
-            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win&lang=en-US#/firefox-setup.exe"
+            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win&lang=en-US#/firefox.7z"
         }
     },
+    "extract_dir": "core",
     "bin": [
-        "firefox.exe",
         [
             "firefox.exe",
             "firefox-beta"
@@ -21,14 +23,5 @@
             "firefox.exe",
             "Firefox Beta"
         ]
-    ],
-    "pre_install": "echo \"[Install]`nInstallDirectoryPath=$dir`nQuickLaunchShortcut=false`nDesktopShortcut=false`nStartMenuShortcuts=false\" | sc \"$dir\\install.ini\"",
-    "installer": {
-        "file": "firefox-setup.exe",
-        "args": "/INI=$dir\\install.ini"
-    },
-    "uninstaller": {
-        "file": "uninstall\\helper.exe",
-        "args": "/S"
-    }
+    ]
 }
--- a/firefox-developer.json
+++ b/firefox-developer.json
@@ -1,16 +1,18 @@
 {
+    "description": "Developer builds of Firefox: the popular open source web browser.",
     "homepage": "https://www.mozilla.org/en-US/firefox/developer/",
     "version": "nightly",
+    "license": "MPL-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://download.mozilla.org/?product=firefox-devedition-latest-ssl&os=win64&lang=en-US#/firefox-setup.exe"
+            "url": "https://download.mozilla.org/?product=firefox-devedition-latest-ssl&os=win64&lang=en-US#/firefox.7z"
         },
         "32bit": {
-            "url": "https://download.mozilla.org/?product=firefox-devedition-latest-ssl&os=win&lang=en-US#/firefox-setup.exe"
+            "url": "https://download.mozilla.org/?product=firefox-devedition-latest-ssl&os=win&lang=en-US#/firefox.7z"
         }
     },
+    "extract_dir": "core",
     "bin": [
-        "firefox.exe",
         [
             "firefox.exe",
             "firefox-dev"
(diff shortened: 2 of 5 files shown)


r15ch13:
Now I ask myself why I didn't come up with this solution ... 😁
Can you change add these changes to all other Firefox versions as well?

shoogle:
@r15ch13, probably because someone else wrote the [original script](https://github.com/shoogle/scoop-extras/commit/92a4cf30faa292f312dc15b1704223e48d1be113#diff-fed8398e0cf4638d316ac91d4ed9b614) and you followed their approach.

Sure I can fix the other scripts, but since some of them are for development builds, I have a few questions/comments.

### 1) What's the deal with development builds (nightlies, betas, etc.) of scoop apps?

I noticed there are a few (but not many) in the extras bucket. (Firefox is one of the few.)

### 2) Does it really make sense to have them in the extras bucket?

Most people will not be interested in them, so it seems like a waste of disk space and bandwidth to have them with the main apps. It would make more sense to store them in a separate bucket (perhaps that's what [this repository](https://github.com/scoopinstaller/nightlies) will be used for eventually?).

### 3) Does it make sense to have a centralised bucket for development builds at all?

Even users who want to install a development build probably only want to do so for a single app. Why not have upstream developers maintain their own buckets for nightly builds, or even just [individual manifests](https://gist.github.com/shoogle/bdba57b9dcb186fab07377916f2552cd) that users can install like this:

```
scoop install https://gist.githubusercontent.com/shoogle/bdba57b9dcb186fab07377916f2552cd/raw/musescore-nightly-master.json
```

### 3) Does it make sense to update manifests for development builds manually?

[MuseScore](/musescore/MuseScore) releases 2-3 development builds *per day*. Consider these points:

1. Users are usually only interested in the latest development build.
2. Builds get superseded very rapidly.
3. Multiple builds will have been missed by the time somebody gets around to updating the manifest.

Scoop manifest already contain update information, so, **specifically for development builds**, it might make sense to **always install the latest available version** rather than the version in the manifest. Perhaps there doesn't need to be a version in the manifest at all (the update information is enough). This is the only way that a centralised nightlies bucket can be useful, otherwise the manifest will either always be out of date, or it will constantly be updated with new manifests for apps people don't use.

The alternative is for upstream developers to host the development manifests/buckets themselves, and for them to be updated automatically by the build scripts that are used to generate each development build.

### 4) What about past versions of development builds?

There are some occasions users *do* want to install a specific version of a development build (e.g. to look back to see when a bug first appeared). For regular scoop apps, it is possible to get a list of previous versions by looking back at the history of the manifest file. However, for development versions, this may not be the case:

- if the manifest is not updated quickly enough some versions will be skipped
- development builds are only stored for a limited time, so past versions no longer be available

This means that, **specifically for development builds**, either:

- the manifest file should contain **all currently available versions**, or
- the update information should return **all currently available versions** (e.g. by extracting them from a page like [this one](https://ftp.osuosl.org/pub/musescore-nightlies/windows/))

---

P.S. Sorry if this is not the appropriate place to discuss this. I'm happy to copy this into an issue if required.

shoogle:
### TLDR

For scoop to be useful for development builds, either:

1. App manifests for development builds would be hosted and maintained centrally in a special bucket
    - the bucket would only contain development builds (no dev builds in main or extras buckets)
    - manifests for dev builds wouldn't contain a current version (no need to update manifest with each new version)
    - update information would report all development versions currently available on the upstream website

2. App manifests for development builds must be hosted and maintained by upstream developers
    - manifests would contain all currently available versions of the development build
    - manifests would be updated automatically by the upstream build scripts used to generate development builds

shoogle:
@r15ch13, I've updated manifests for the other Firefox builds, and I've also fixed an issue I found with the "checkver" regex in `firefox-esr.json`.

The old regex was looking on the [ESR page](https://www.mozilla.org/en-US/firefox/organizations/) for a version string like this:

```html
<html  data-esr-versions="52.9.0"
```

However, that would break at certain times of year (like now), when **two** ESR versions are available:

```html
<html  data-esr-versions="52.9.0 60.1.0"
```

I fixed the regex so that it can handle this situation too. __My fix chooses the version that appears first__, which looks to be __the older of the two versions__. Justification is below.

### Firefox ESR versions

As explained in more detail (and with nice diagrams) [here](https://www.mozilla.org/en-US/firefox/organizations/) and [here](https://wiki.mozilla.org/Enterprise/Firefox/ExtendedSupport:Proposal#Proposal)...

- Each Extended Support Release (ESR) is based on a specific version of Firefox Desktop.
- The ESR version receives extended support for a period of 54 weeks.
- 12 weeks before the end of that period, the next ESR version is released.

During the 12 week transition period, the incoming and outgoing ESR are *both* supported. Both ESRs continue to receive security updates which increase their MINOR and PATCH version numbers.

### Whether to offer the old or new ESR version?

It's tempting to think that scoop should offer the newest version as soon as it becomes available, but that ignores the whole point of extended support releases. ESRs are mainly targeted at organisations rather than individuals, and __organisations care about stability rather than new features__. The 12 week transition period is there to allow organisations to test and certify deployment of the new ESR version before upgrading all their systems, so I think __it makes sense for scoop to continue to offer the old ESR during the transition period__.

The alternative would be to offer two manifests, or build some kind of `fastlane`/`slowlane` facility, to allow scoop users to choose for themselves...

firefox-beta.json
```diff
@@ -1,16 +1,18 @@
...
+            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win64&lang=en-US#/firefox.7z"
         },
         "32bit": {
-            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win&lang=en-US#/firefox-setup.exe"
+            "url": "https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win&lang=en-US#/firefox.7z"
         }
     },
+    "extract_dir": "core",
     "bin": [
-        "firefox.exe",
```
shoogle: @r15ch13, I removed "firefox.exe" from `bin` for all manifests except [firefox.json](#diff-fed8398e0cf4638d316ac91d4ed9b614) to prevent conflicts if a user tries to install more than one build. Other builds are still available in `%PATH%` via their aliases (e.g. `firefox-beta`).

r15ch13:
Awesome work! Thank you! 👍
Removing `firefox.exe` is a good idea.
Using older ESR version is the right thing to do.

PR #1096, opened 2018-08-03
Add persist to OBS

nickbudi:
Makes OBS portable

### DIFF
--- a/obs-studio-small.json
+++ b/obs-studio-small.json
@@ -14,6 +14,12 @@
             "OBS Studio (32bit)"
         ]
     ],
+    "persist": "config",
+    "post_install": [
+        "if(!(test-path \"$dir\\portable_mode\")) {",
+        "   $null > \"$dir\\portable_mode\"",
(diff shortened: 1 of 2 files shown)


obs-studio-small.json
```diff
@@ -14,6 +14,12 @@
             "OBS Studio (32bit)"
         ]
     ],
+    "persist": "config",
+    "post_install": [
```
rasa: If you don't create the `config` (file or directory?) in the `post_install`, it won't persist (or does it already exist?)

If you don't list the `portable_mode` file in the `persist` tag, it won't persist across upgrades.

nickbudi: config is a directory, which isn't created until OBS is run in portable mode. I forgot about adding portable_mode to persist.

rasa: If the config directory doesn’t already exist, you’ll need to create it in the post_install, os it will persist across upgrades.

PR #1101, opened 2018-08-05
Add squashfs-tools and node-compiler

pcktm:


rasa:
Are these GUI or CLI apps? If CLI, perhaps they belong in scoop’s main bucket?

pcktm:
I don't think that they fit the criteria of the main bucket well. I don't know if they are so widely used.

PR #1104, opened 2018-08-06
Multiple fixes for manifests

se35710:


boostnote.json
```diff
@@ -1,9 +1,9 @@
 {
     "homepage": "https://boostnote.io/",
     "license": "GPL-3.0-or-later",
-    "version": "0.11.7",
-    "url": "https://github.com/BoostIO/boost-releases/releases/download/v0.11.7/BoostnoteSetup.exe#/dl.7z",
-    "hash": "d4df317aabca17cdc62c56deaf934ddc223735f777d5d3db6d1b68f55fd5b6f0",
+    "version": "0.11.8",
+    "url": "https://github.com/BoostIO/boost-releases/releases/download/v0.11.8/BoostnoteSetup.exe",
```
rasa: I think we need to leave the `#/dl.7z` on the URL

coretemp.json
```diff
@@ -20,5 +16,22 @@
             "Core Temp.exe",
             "Core Temp"
         ]
-    ]
+    ],
+    "persist": [
+        "CoreTemp.ini"
```
rasa: If `CoreTemp.ini` doesn't already exist in the .zip file, it needs to be created in a `post_install`

boostnote.json
se35710: Ahh, yes. I missed that the autoupdate didn't include the `#/dl.7z` part. I'll push a fix in a few seconds.

coretemp.json
```diff
@@ -20,5 +16,23 @@
...
+    ],
+    "checkver": {
+        "url": "https://www.alcpu.com/CoreTemp/history.html",
+        "re": "Version ([\\d.]+)"
+    },
+    "autoupdate": {
+        "architecture": {
+            "32bit": {
+                "url": "http://www.alcpu.com/CoreTemp/php/download.php?id=2#/coretemp32.zip"
```
rasa: The checkver is https, but homepage, and the download URLs are http?

yatqa.json
```diff
@@ -16,7 +16,7 @@
     },
     "checkver": {
         "url": "http://31.172.95.122/yatqa/.update",
```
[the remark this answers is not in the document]
r15ch13: Yes, this is the official update URL. Don't ask me why they use a hardcoded IP.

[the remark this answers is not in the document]
se35710: I trippled checked all urls, including reverse DNS lookups, but only using the IP address works.

Issue #1121, opened 2018-08-10
Add shadowsocksrr/shadowsocksr-csharp


Retia-Adolf:
[shadowsocksrr/shadowsocksr-csharp/releases](https://github.com/shadowsocksrr/shadowsocksr-csharp/releases)

ShadowsocksR is an Internet censorship circumventing tunnel tool, forked by Breakwa11 from shadowsocks, licensed under GNU GPLv3, maintained now by succeeding developers in team [shadowsocksrr](https://github.com/shadowsocksrr).

It is different from shadowsocks by concentrating on characteristics-emphasizing obfuscation, rather than heavily characteristics-weakening randomization, which is another way that original shadowsocks project is now working on.

No admin privilege is needed to run.

PR #1130, opened 2018-08-13
Add shadowsocksr-csharp

Retia-Adolf:
#1121


r15ch13:
Looks good, it's just missing the `.json` file extension :smile:

Ash258:
Will not work because: `"extract_dir": "ShadowsocksR-win-$version"`. Move it into autoupdate block and in main block change it to actual version

Issue #1123, opened 2018-08-10
Add Psiphon


Retia-Adolf:
Psiphon is a centrally managed, geographically diverse network of thousands of proxy servers, using a performance-oriented, single hop architecture, it's for Internet censorship circumvention purpose, seems currently developing only by Psiphon Inc. inside, but licensed under GNU GPLv3.

Open-sourced at:
```
https://github.com/Psiphon-Labs/psiphon-tunnel-core
https://bitbucket.org/psiphon/psiphon-circumvention-system
```

But released at:
```
https://psiphon.ca/
https://s3.amazonaws.com/psiphon/web/60l3-nnss-6gsn/index.html
```

P.S. They claim that:

> User IP addresses are not collected by Psiphon servers in the normal course of operation. Psiphon does not require user accounts, so, by default, there is no collection of email addresses, usernames, or passwords. Event logs include timestamps, region codes (country and city), and non-identifying attributes including sponsor ID (determined by which Psiphon client build is used), client version, and protocol type. Page views are aggregated by time and/or session before being logged.

PR #1133, opened 2018-08-14
Add Psiphon3

Retia-Adolf:
#1123


r15ch13:
I guess you're using "https://s3.amazonaws.com/psiphon/web/60l3-nnss-6gsn" instead of https://www.psiphon3.com/ because it prevents blocking? :grin:

Retia-Adolf:
@r15ch13 Yeah, and actually I'm not able to access psiphon3.com now.😅

PR #1135, opened 2018-08-14
Add Monero

Retia-Adolf:
[Monero](https://getmonero.org/) is a private, secure, untraceable, decentralised cryptocurrency.

### DIFF
--- a/monero.json
+++ b/monero.json
@@ -1,7 +1,7 @@
 {
     "homepage": "https://getmonero.org/",
     "description": "The secure, private, untraceable cryptocurrency",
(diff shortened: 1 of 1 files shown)


r15ch13:
The `bin` property only requires multidimensional arrays when you want to rename files (like `start-low-graphics-mode.bat`) or add parameters.
If you use `checkver.github` then you can omit the `/releases/latest` part at the end.
Hash extraction provides the filename via the `$basename` variable.

See: https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate

PR #1151, opened 2018-08-19
Add FastCopy version 3.54

pakonda:


rasa:
I added the persist file, and put quotes around `$dir`

PR #1153, opened 2018-08-22
Create mkcert.json

filippobuletto:
Add mkcert from https://github.com/FiloSottile/mkcert

It's already on chocolatey but I'd like to get it through scoop

See #1154

### DIFF
--- a/mkcert.json
+++ b/mkcert.json
@@ -1,0 +1,15 @@
+{
+    "version": "1.1.1",
+    "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice 👍

By adding `#/mkcert.exe` to end of the URL and autoupdate URL the file will get renamed automatically.
The `pre_install` script can be removed. :smile:

filippobuletto:
Good to know! I'm updating the descriptor

filippobuletto:
@r15ch13 Richard do you think it is acceptable?

PR #1159, opened 2018-08-23
Add CUETools 2.1.6.

spongbros:


Ash258:
Would be useful to run some post_install like `"Remove-Item \"$dir\\user_profiles_enabled\""`:
> If you want to run it as a portable application from a USB stick, remove the file called user_profiles_enabled so that settings will be saved to the same folder instead of your user profile folder. (If you already have non-portable settings you want to use, copy the subfolders named CUE Tools or CUERipper from the %appdata% folder to your portable install.)

PR #1169, opened 2018-08-25
Add Xtreme Download Manager

Retia-Adolf:


r15ch13:
Made some changes:
- use `msi` because the `jar` version is outdated
- remove `post_install` because `jre` is provided by `msi`
- simplified version regex (don't make them so strict)

I recommend installing `pester` and running tests locally before creating a PR :smile:
And at least 5 successful installs/uninstalls of the program to test it :grin:

PR #1170, opened 2018-08-25
Add Portacle

Retia-Adolf:
😂


r15ch13:
Wrong file extension :smile:

r15ch13:
Made some changes:
- version regex has to allow letters because of the hotfix version
- add `#/dl.7z` for automatic extraction

But there is one problem I see with persisting a directory like `all`.
Scoop doesn't overwrite/merge directories on an update. Which means you will always use the old contents. (e.g. `all\ssl\ca-bundle.crt` or `all\dictionaries` will not get updated)

rasa:
Could we persist specific files in `all`?

Retia-Adolf:
Maybe no, ok to fess, I don't know..

PR #1177, opened 2018-08-26
Add listen.moe Windows Client

Nikey646:
It's a simple portable app that allows you to listen to https://listen.moe without using your browser.

First time contributing something like this, so please tell me if there is something wrong.

### DIFF
--- a/listenmoeclient.json
+++ b/listenmoeclient.json
@@ -3,8 +3,8 @@
     "license": "MIT",
     "version": "2.0.0",
     "url": "https://github.com/LISTEN-moe/windows-app/releases/download/2.0.0/ListenMoeClient.exe",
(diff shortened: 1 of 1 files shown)


r15ch13:
Please run `.\bin\checkver.ps1 listenmoeclient -f` one time. This will guarantee that the auto update works and fix formatting issues.
Also, I would recommend adding the part after the newline of the description to the `notes` property.

Nikey646:
Running `.\bin\checkver.ps1 listenmoeclient -f` doesn't actually output anything for me :s I'm not sure if that is the intended result or not...

I saw the AppVeyor errors and manually fixed them (Just some simple tabs) and seperated the description into the notes property like you suggested, since it did make more sense.

r15ch13:
Looks good. 👍 
You have to run `.\bin\checkver.ps1` from the root of your forked repo.

PR #1178, opened 2018-08-26
Add Ocenaudio

Melvinkooi:
This pull request adds Ocenaudio into the scoop-extras bucket. It's a free single-track audio editor where you can record and edit audio or just view the waveform of an audio track. There's also a spectrum viewer so you can analyze if a lossless audio file is truely lossless.

I use this program frequently so i want this program in the bucket so i don't have to update manually from time to time.

This is also my first app manifesto. I've tested the auto-update feature which works. It also install and uninstall the application without any errors.

### DIFF
--- a/ocenaudio.json
+++ b/ocenaudio.json
@@ -3,11 +3,11 @@
     "version": "3.4.5",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice 👍

Issue #1164, opened 2018-08-24
Request: unetbootin


Guard13007:
A utility for installing operating system live disks to USB: https://unetbootin.github.io/

PR #1187, opened 2018-08-28
Add UNetbootin version 661

pakonda:
#1164


r15ch13:
Nice 👍 
But I would change `homepage` to https://unetbootin.github.io/ and add the GitHub URL to checkver. 😄

PR #1190, opened 2018-08-29
XX-Net: fix shortcuts

Retia-Adolf:
Due to shortcuts limit, only able to set target inside `$dir`, unable to call `wscript.exe`


Retia-Adolf:
Thx! Forgot to delete unwanted code before committing.

PR #1207, opened 2018-09-03
fix kid3 bin and shim paths

Guard13007:
Don't know if there is a better way to fix this or not (such as removing the extra folder).


Ash258:
Better fix would be add: `"extract_dir":"kid3-3.6.2-win32"` and `"autoupdate"ˇ: {"extract_dir":"kid3-$version-win32" }`

PR #1209, opened 2018-09-04
thunderbird: Add x64 version

ustsv:
Add x64 version,  en-US locale only.


Ash258:
Installing is unnecessary. It could be renamed into dl.7z and set extract_dir to "core"

r15ch13:
Like [firefox.json](https://github.com/lukesampson/scoop-extras/blob/master/firefox.json) was recently changed in https://github.com/lukesampson/scoop-extras/pull/1079

PR #1214, opened 2018-09-08
Corrected autoupdate URL

Ardelean-Calin:
Corrected autoupdate URL to use version provided by checkver.

### DIFF
--- a/flutter.json
+++ b/flutter.json
@@ -1,13 +1,16 @@
 {
     "homepage": "https://flutter.io/",
-    "version": "0.5.1",
+    "version": "0.7.3",
     "license": "https://github.com/flutter/flutter/blob/master/LICENSE",
     "url": [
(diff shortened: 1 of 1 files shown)


Ardelean-Calin:
Note to reviewers: A new Flutter version is out but checkver doesn't seem to detect that. Hopefully this change fixes it? What do you think?

r15ch13:
`$version` and `$matchVersion` are the same. This was not the problem.
`url` and `hash` didn't have the same types. If one is a string the other has to be one too. This resulted in an `ERROR Unable to index into an object of type System.String.`.

Fixed this by changing the `hash` property to an array and added the checksum for `flutter-dev-setup.ps1`. Updated it to `0.7.3` and changed `checkver` to our default format `([\d.]+)`.

PR #1222, opened 2018-09-12
Eclipse 2018-09 release

se35710:
Removes the need for a [version converter](https://github.com/se35710/scoop-eclipse/blob/master/versions.txt), since version numbers are now numbers and not names.


se35710:
2018-09 was released, so I have updated the manifests. including minor filename issue (R is not in the filename anymore), and ran a `checkver -u` before commit and push.

Ready to be merged now.

r15ch13:
Just wanted to ask if `eclipse-automotive` and `eclipse-android` are discontinued?
What is the reason for deleting `checkver` and `autoupdate`?

se35710:
@r15ch13 I think so, they were incubator projects that never became "real" projects.

r15ch13:
Okay 👍

PR #1223, opened 2018-09-12
Add Gogs - A painless self-hosted Git service

se35710:


gogs.json
```diff
@@ -0,0 +1,60 @@
...
+            "[database]",
+            "DB_TYPE  = sqlite3",
+            "PATH     = $persist_dir\\data\\gogs.db",
+            "",
+            "[repository]",
+            "ROOT = $persist_dir\\repositories",
+            "",
+            "[log]",
+            "ROOT_PATH = $persist_dir\\log",
+            "\"@ | % {$_ -replace \"\\\\\",\"/\"} | out-file -en oem $dir\\custom\\conf\\app.ini }"
```
Ash258: You could go with more readable `-replace '\\\\', '/'`.
No quotes in app.ini around $dir and $persist_dir?? What about paths with spaces?

r15ch13: It can also be a bit more compact like [`gitea.json`](https://github.com/lukesampson/scoop-extras/blob/master/gitea.json)

r15ch13: @Ash258 gitea works without quotes. Since gitea is a fork of gogs, my guess is that it works too.

image

se35710: Missed the $dir quotes, thanks. But I can not find any $persist_dir quotes missing, or do you mean inside the .ini file?

Anyway, replaced the install script with the one from [`gitea.json`](https://github.com/lukesampson/scoop-extras/blob/master/gitea.json). I kept the `persist` in the manifest though, so you can run gogs without the -c.

PR #1233, opened 2018-09-16
Fix extract_dir for auto updates

xeijin:
extract_dir was at the root level rather than inside the respective architecture elements, and was therefore not being updated to the correct value upon a checkver update, preventing the installation from completing successfully


Ash258:
Its same for both architectures == Irrelevant to make it both.

Just make it: 

```json
"extract_dir": "knime_3.6.1",
"autoupdate":{
    "architecture" {
         ....
    }
    "extract_dir": "knime_$version"
```

PR #1241, opened 2018-09-18
Fix libreoffice fresh.

medoni:
version urls has changed.

also changed checkver regex. there is a libreoffice-6.1.1.2.tar.xz but no corresponding 6.1.1.2.msi file

### DIFF
--- a/libreoffice-fresh.json
+++ b/libreoffice-fresh.json
@@ -1,14 +1,14 @@
 {
     "homepage": "https://libreoffice.org/",
     "license": "MPL-2.0",
-    "version": "6.1.1.2",
+    "version": "6.1.1",
     "architecture": {
         "64bit": {
-            "url": "https://downloadarchive.documentfoundation.org/libreoffice/old/6.1.1.2/win/x86_64/LibreOffice_6.1.1.2_Win_x64.msi",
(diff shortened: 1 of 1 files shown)


r15ch13:
The old URL and `checkver` are still working. Why the change?

medoni:
yesterday these urls didn't work here in cologne. now it works.

medoni:
scoop doesn't follow http 302 (temporary moved) links.
here is the output of scoop update libreoffice-fresh
```
Updating 'libreoffice-fresh' (6.1.1 -> 6.1.1.2)
Uninstalling 'libreoffice-fresh' (6.1.1)
Unlinking ~\scoop\apps\libreoffice-fresh\current
Installing 'libreoffice-fresh' (6.1.1.2) [64bit]
The remote server returned an error: (302) Found.
URL https://downloadarchive.documentfoundation.org/libreoffice/old/6.1.1.2/win/x86_64/LibreOffice_6.1.1.2_Win_x64.msi is not valid
```

r15ch13:
Ah, they redirect from HTTPS to HTTP. PowerShell 6 doesn't like that. 😁
Changing the URLs to HTTP should be enough.

PR #1249, opened 2018-09-19
Add eclipse-mat version 1.8.0.20180604

TheCjw:


eclipse-mat.json
```diff
@@ -0,0 +1,37 @@
...
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.eclipse.org/mat/downloads.php",
+        "re": "MemoryAnalyzer-([\\d.]+)-win32"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "http://download.eclipse.org/mat/1.8/rcp/MemoryAnalyzer-$version-win32.win32.x86_64.zip"
```
Ash258: That 1.8 inside URL is changing with Version? 
I suggest to use `$majorVersion.$minorVersion/rcp`

TheCjw:
@Ash258 FIXED.

TheCjw:
@se35710 Great suggestions!

Ash258:
Could be hash universal like:
```json
"architecture":{
},
"hash": {
    "url": "https://www.eclipse.org/downloads/sums.php?file=/mat/$majorVersion.$minorVersion/rcp/$basename&type=sha512",
    "find": "([a-fA-F0-9]{128})"
}
```

se35710:
@Ash258 There is an issue with `$basename`, it will use download.php as the filename, so the match will fail.

@TheCjw You don't need to add `#/dl.7z`, scoop will handle the extraction anyway.

TheCjw:
@se35710 `#/dl.7z` has been removed.

PR #1250, opened 2018-09-20
Add watchman-nightly.json

jas99:


Ash258:
Missing Hash, license, description, ...

This should be more complex using checkver as github facebook/watchmen and some URL sticking using $matchVars to get proper version URL.

or

make it nightly.

PR #1271, opened 2018-09-25
Add steamcmd

Lomeli12:


vaguiners:
fix this pls

r15ch13:
Made some improvements:
- ut now downloads the newest steamcmd instead of the old bootstrapper
- persist appcache, config, steamapps and userdata
- added autoupdate and checkver

vaguiners:
This could be added to "scoop" repo since steamcmd is cmd only?

PR #1291, opened 2018-09-30
Add lightbulb@1.6.4

Melvinkooi:
LightBulb is an opensource alternative to f.lux. It's a bit more customizable than f.lux and should also be more "gaming-friendly" as the author of the program explains it here: https://old.reddit.com/r/osugame/comments/5rpq1q/lightbulb_gamingfriendly_flux_alternative/

The app manifest is based on the portable version of the progam. I've added notes that the program needs .Net framework, Run a registry key if changing the gamma doesnt work correctly (the readme file advises to run it at least once) and that you need to run the bat file inside the folder of the program to enable portable mode, which the created shortcut already does.

I'm not sure if the notes can be shorter or anything can be removed from it. Otherwise the app manifest works perfectly for me. It also allows to install older versions of the program.

Please leave feedback if necessary.

### DIFF
--- a/lightbulb.json
+++ b/lightbulb.json
@@ -14,17 +14,22 @@
     "url": "https://github.com/Tyrrrz/LightBulb/releases/download/1.6.4/LightBulb-Portable.zip",
     "hash": "847053924ab64a872e01d70070d2a6ece816fe0e276003038ba9148c69d9896c",
     "pre_install": [
-        "   #Create placeholder file",
+        "#Create placeholder file",
         "function CreateFile([String] $file, [String] $content = $null) {",
         "    if(!(Test-Path \"$persist_dir\\$file\")) {",
-        "        Write-Host \"Create placeholder for shadowsocksr-csharp: $file\"",
+        "        Write-Host \"Create placeholder for $file\"",
         "        New-Item -Force -Path \"$persist_dir\\$file\" -ItemType file -Value $content | Out-Null",
         "    }",
         "}",
         "CreateFile 'Configuration.dat' '{}'"
     ],
     "persist": "Configuration.dat",
-    "bin": "LightBulb.exe",
+    "bin": [
+        [
+            "LightBulb_Portable.bat",
+            "LightBulb"
+        ]
+    ],
     "shortcuts": [
         [
             "LightBulb_Portable.bat",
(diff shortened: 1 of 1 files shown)


Ash258:
Change notes into multi string array (One sentence one string), also bins, persist, could be only String

Melvinkooi:
Updated the app manifest with the changes you've suggested. Any more feedback before a possible merge?

lightbulb.json
```diff
@@ -0,0 +1,40 @@
...
+        "function CreateFile([String] $file, [String] $content = $null) {",
+        "    if(!(Test-Path \"$persist_dir\\$file\")) {",
+        "        Write-Host \"Create placeholder for shadowsocksr-csharp: $file\"",
+        "        New-Item -Force -Path \"$persist_dir\\$file\" -ItemType file -Value $content | Out-Null",
+        "    }",
+        "}",
+        "CreateFile 'Configuration.dat' '{}'"
+    ],
+    "persist": "Configuration.dat",
+    "bin": "LightBulb.exe",
```
r15ch13: Wouldn't it be better to use the `LightBulb_Portable.bat`?
```
"bin": [
    [
      "LightBulb_Portable.bat"
      "LightBulb"
    ]
]
```

Melvinkooi: It was my intention at first, but i changed to to the exe file if you want to do something in the command line. Changed it back with your code suggestion.

lightbulb.json
```diff
@@ -0,0 +1,40 @@
...
+        "If the gamma doesn't change under 6600K, run the Manual_Gamma_Registry_Fix.reg registry key in the program's directory (scoop/apps/lightbulb/current).",
+        "Don't run LightBulb.exe when you're in that folder. Use the file LightBulb_Portable.bat or the created shortcut to run this program."
+    ],
+    "url": "https://github.com/Tyrrrz/LightBulb/releases/download/1.6.4/LightBulb-Portable.zip",
+    "hash": "847053924ab64a872e01d70070d2a6ece816fe0e276003038ba9148c69d9896c",
+    "pre_install": [
+        "   #Create placeholder file",
+        "function CreateFile([String] $file, [String] $content = $null) {",
+        "    if(!(Test-Path \"$persist_dir\\$file\")) {",
+        "        Write-Host \"Create placeholder for shadowsocksr-csharp: $file\"",
```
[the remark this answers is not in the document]
Melvinkooi: Done. Didn't noticed that.

PR #1308, opened 2018-10-08
Create datagrip.json

lllvantis:
Can you please adde JetBrains DataGrip?


r15ch13:
It's recommended to use the [JetBrains Toolbox](https://github.com/lukesampson/scoop/blob/master/bucket/jetbrains-toolbox.json)

PR #1309, opened 2018-10-08
Add jitsi-meet-electron.json

jas99:
Hey guys,

What do you recommend for auto updating hash, when the publisher is not publishing one. 

Thx

### DIFF
--- a/jitsi-meet-electron.json
+++ b/jitsi-meet-electron.json
@@ -1,0 +1,18 @@
+﻿{
+    "version": "1.1.0",
+    "license": "Apache License 2.0",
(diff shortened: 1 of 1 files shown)


r15ch13:
Scoop will download the file and calculates it itself. No problem 😄

jas99:
@r15ch13 
Thx for your help. 
I fixed mentioned corrections.

Additional question:
Author using prefix v before version tag
```
 "autoupdate": {
        "url": "https://github.com/jitsi/jitsi-meet-electron/releases/download/v$version/jitsi-meet.exe"
    }
```
So my $version (returned by checkver set to github) will be with or without v prefix?

Plus I guess we should not include v prefix in version of manifest.

Thx

r15ch13:
Correct, the `v` prefix should only be in the update URL.

PR #1322, opened 2018-10-15
gpu-z: Update to version 2.12.0

ustsv:
update to version 2.12.0
correct dot in checkver re
add bin


r15ch13:
It updated to 2.13.0 just recently https://github.com/lukesampson/scoop-extras/commit/08328eae33553403000c8be21ab6eade1fd29a61
What was the issue with the version?

Adding the `bin` property is a good idea. 👍

ustsv:
TechPowerUp GPU-Z v2.11.0  ok
TechPowerUp GPU-Z v.2.12.0 add dot after v => checkver .2.12.0, download error
TechPowerUp GPU-Z v2.13.0  ok
Regexp correct in all cases

PR #1324, opened 2018-10-15
JDownloader: ensure config directory exists at install time

Gargauth:
On clean installation, the configuration directory did not get created, which resulted in persistence not working.


Gargauth:
Disregard, it's working as intended.

Ash258:
This is needed only when persisting files. Folders works.

PR #1325, opened 2018-10-17
wsltty: move to dir, before run install.bat.

nozaki87:
Installing wsltty was failed in my environment without any error message.
 ( Win 10 Pro 1809 17763.55 64-bit, scoop 83cff3bb )
I think someone may have same problem.
Downloading and extraction was OK, but any files weren't copied to "user\AppData" folder.
Moreover, it couldn't add the start menu entry in "uesr\AppData\Roaming\Microsoft\Windows\Start Menu\Programs" .
If this is intended behavior, please reject PR.

To fix this problem, it needs to move into the folder including extracted files and other scripts making start menu entries, before executing install.bat.

### DIFF
--- a/wsltty.json
+++ b/wsltty.json
@@ -11,7 +11,7 @@
         "url": "https://github.com/mintty/wsltty/releases/download/$version/wsltty-$version-install.exe#/dl.7z"
     },
     "installer": {
-        "file": "install.bat",
+        "script": ["pushd \"$dir\"", ".\\install.bat", "popd"],
         "keep": true
     },
     "uninstaller": {


nozaki87:
@Ash258 Hi, thanks so much for your comment.
I think context is not set to $dir.
That "is_in_dir" in the code only ensures that installer file is inside of $dir.
To change current directory in PowerShell script, "cd" or "Push-Location" or "Set-Location" is necessary like below.
https://github.com/lukesampson/scoop/blob/757d83036cfb382c004d68f5ddfe981025fba95b/bin/scoop.ps1#L15
However, I couldn't find any of such function call in source code of scoop.

If it is suitable that scoop should set its context to $dir, I will report as a issue in scoop repository not scoop-extras.
Any advice is welcome.

Ash258:
It's manually loading program name using `$dir` prefix with coalesce of `installer.file` or downlouaded filename
`$prog = "$dir\$(coalesce $installer.file "$fname")"`

Ash258:
It works [everywhere](https://github.com/lukesampson/scoop-extras/search?q=installer"%3A%7B%5Cs%2B"file"&unscoped_q=installer"%3A%7B%5Cs%2B"file"), so it's problem on your end.

nozaki87:
I caused you a misunderstanding, and sorry for my poor explanation.
Downloading and extraction was OK. AND ALSO, EXECUTION of installer.file worked as you wrote.
However, `installer.file` (`install.bat`)  tries to copy files and execute files in installer packages, using relative path.
If the folder `installer.file` runs is not in $dir, it fails.
I think most of installers in bucket don't depend on which current folder is.

So, it might also be problem of wsltty installer...
I've come to think that better solution is to fix wsltty installer.
I will check that first.

Ash258:
Install.bat is using custom variables, so it will always be installed into where they want to be and never to `$dir`.

```bat
set refinstalldir=%%LOCALAPPDATA%%\wsltty
set installdir=%LOCALAPPDATA%\wsltty
set refconfigdir=%%APPDATA%%\wsltty
set configdir=%APPDATA%\wsltty
set oldroot=%installdir%
set oldhomedir=%installdir%\home\%USERNAME%
set oldconfigdir=%oldhomedir%\.config\mintty
if not "%1" == "" set refinstalldir=%1 && set installdir=%1
if not "%2" == "" set refconfigdir=%2 && set configdir=%2
```

But fast few search through that installer I mentioned it's only copying and renaming of files and dealing with version changes. So you can just write custom powershell script to install, how you want it.

nozaki87:
Where they copy to would be not bad.
`install.bat` fails at following lines, because it couldn't find the files it copy from.
Current directory is where `scoop install` runs by user, but the files it will copy is in $dir.

```bat
copy LICENSE.mintty "%installdir%"
copy LICENSE.wslbridge "%installdir%"

copy "add to context menu.lnk" "%installdir%"
copy "add default to context menu.lnk" "%installdir%"
copy "remove from context menu.lnk" "%installdir%"
copy "configure WSL shortcuts.lnk" "%installdir%"
copy "WSL Terminal.lnk" "%installdir%"
copy "WSL Terminal %%.lnk" "%installdir%"
copy wsl.ico "%installdir%"
copy config-distros.sh "%installdir%"
copy mkshortcut.vbs "%installdir%"
```
```bat
copy "wsltty home & help.url" "%smf%"
copy "add to context menu.lnk" "%smf%"
copy "add default to context menu.lnk" "%smf%"
copy "remove from context menu.lnk" "%smf%"
copy "configure WSL shortcuts.lnk" "%smf%"
copy "WSL Terminal.lnk" "%smf%"
copy "WSL Terminal %%.lnk" "%smf%"
```

Ash258:
Then it should be enough to make: 

```json
"installer": {
    "script": [ 
        "Push-Location \"$dir\"",
        "& .\\install.bat \"$dir\" \"$dir\\config\"",
        "Pop-Location"
    ]
}
```

Ash258:
Maybe adding `extract_to` directory, and running installer from that directory. For eliminating some conficts. (If some could occured)

nozaki87:
Thanks for your suggestion! I will try it.

rasa:
Can this PR be closed without merging?

nozaki87:
Yes, because I cannot take time to test it soon.
When I need it, I will do another PR from latest master branch.

PR #1331, opened 2018-10-20
add Library\bin to PATH and remove envs from persist for miniconda3

specter119:
Adding `Library\bin` is necessary if the working environment is just the `base` environment of miniconda3.
> I think Anaconda should have the similar request.

Also persist `envs` will be less useful (it may relate to my issue <https://github.com/lukesampson/scoop/issues/1919> ), people who often use conda can manually add `%UserProfile%/.condarc` with the following contents:

```yaml
envs_dirs:
  - ~\.conda\envs
pkgs_dirs:
  - ~\.conda\pkgs
  - ~\scoop\apps\miniconda3\current\pkgs
```
This will let conda maintain envs in `%UserProfile%\.conda\envs` and store pkgs in `%UserProfile%\.conda\pkgs`. Therefore people can keep their custom environments and update the `base` environment in the fastest way.

### DIFF
--- a/bucket/cacher.json
+++ b/bucket/cacher.json
@@ -1,9 +1,9 @@
 {
     "homepage": "https://www.cacher.io/",
-    "version": "2.10.0",
+    "version": "2.10.2",
     "description": "Code snippet organizer for pro developers.",
-    "url": "https://s3.amazonaws.com/download.cacher.io/Cacher%20Setup%202.10.0.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


specter119:
@r15ch13 , is there something I should do? for example, a more detailed explanation?

r15ch13:
Was hoping for some input from miniconda users. Does this break current installations after an update?

specter119:
@r15ch13 this will no break current installation in my opinion.

the difference is simple, the "Library\\bin" folder contains DLL files, add this to $PATH can let us use the `base` environment directly(no need to `source base`), and will not bring other binaries.

while, persisting `envs` will help the people use `base` as their working environment, but I think the when the minor python version  updates, the persist env may facing some incompatibility. Therefore, I think custom `%UserProfile%/.condarc` will be better.

PR #1335, opened 2018-10-21
zerobrane-studio: Add version 1.80

SinisterBlade:
Adds [ZeroBrane Studio](https://studio.zerobrane.com/), a popular IDE for Lua.


SinisterBlade:
System-wide configuration is stored in `cfg/user.lua` and system-wide packages are in the `packages` directory, which I believe are the only files that need to be persisted.

I'm facing an issue adding `packages` to the persist list. On subsequent installations, scoop copies the default `packages` folder to `persist/zerobrane-studio/packages/packages` and begins failing eventually. Can you help?

PR #1345, opened 2018-10-23
Add ALTools version 1.0

dzampino:
The Account Lockout and Management Tools are a small suite of utilities from Microsoft for troubleshooting account lockouts. Microsoft has not released any updates in six years, so I did not include a version check. I was also unable to locate a license for these utilities.


r15ch13:
Replaced by https://github.com/lukesampson/scoop-extras/pull/1355

PR #1346, opened 2018-10-24
Add Lightshot version 5.4.0.35

fabiojb:
Lightshot is a utility tool for taking screenshots. See more at https://prnt.sc

This PR contains the Lightshot app manifest. The manifest has 2 principal configurations:
- Application autostart on install
- Application autostart on Windows startup

### DIFF
--- a/lightshot.json
+++ b/lightshot.json
@@ -5,8 +5,7 @@
         "url": "https://app.prntscr.com/en/license.html"
     },
     "url": "https://app.prntscr.com/build/setup-lightshot.exe",
-    "hash": "455b17124a474bfa512580ba9bcd275dc8e1119482ad604b83b3cb5611a6f73f",
-    "version": "5.4.0.35",
+    "version": "nightly",
     "extract_dir": "lightshot",
     "innosetup": true,
     "bin": "Lightshot.exe",
--- a/lightshot.json
+++ b/lightshot.json
@@ -1,0 +1,19 @@
+{
+    "homepage": "https://prnt.sc",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://app.prntscr.com/en/license.html"
+    },
+    "url": "https://app.prntscr.com/build/setup-lightshot.exe",
+    "hash": "455b17124a474bfa512580ba9bcd275dc8e1119482ad604b83b3cb5611a6f73f",
+    "version": "5.4.0.35",
+    "extract_dir": "lightshot",
+    "innosetup": true,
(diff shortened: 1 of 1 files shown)


Ash258:
I dont aggree with that auto start things. 

Lots of users use scoop because they dont want that installers bad behaviour. If someone want autostart, they will enable it in application / windows settings.

fabiojb:
> I dont aggree with that auto start things.
The both, or one of them in specific?

> Lots of users use scoop because it's non-invasive and does not inherit installers bad behavior of writing some junk where, it's not supposed to. If someone want auto-start, they will enable it in application / windows settings.
Another possibility would be to have a install param like `--enable-auto-start` so that the default behavior would be to do not configure auto-start, but you have the option to add the auto-start configuration if you want. What do you think about it?

> Also deleting innounp log is bad idea in overall. That log help users to detect problems. Why would you delete them?
I was thinking about cleaning up the installation. I will remove the innounp log delete instruction.

> For optimizing if you want it in your bucket or something like this. You dont need to create that link manually. It's post_install block, that means it's runned after shortcut creating. You can just copy shortcut from `StartMenu\Scoop Apps`
Nice tip

r15ch13:
I would recommend adding it without the autostart script.
```json
{
    "homepage": "https://prnt.sc",
    "license": {
        "identifier": "Freeware",
        "url": "https://app.prntscr.com/en/license.html"
    },
    "url": "https://app.prntscr.com/build/setup-lightshot.exe",
    "hash": "455b17124a474bfa512580ba9bcd275dc8e1119482ad604b83b3cb5611a6f73f",
    "version": "5.4.0.35",
    "extract_dir": "lightshot",
    "innosetup": true,
    "bin": "Lightshot.exe",
    "shortcuts": [
        [
            "Lightshot.exe",
            "Lightshot"
        ]
    ]
}
```

Ash258:
As there is no checkver and autoupdate, making it nightly would be smarter, since installer is only one which is replaced on each release.

fabiojb:
@Ash258 @r15ch13 is there any documentation about nightly build on the wiki?

Ash258:
What informations do you need??

Ash258:
If version is nightly, hash check is skipped. That's all.

r15ch13:
@fabiojb nope, only references in the code https://github.com/lukesampson/scoop/search?q=nightly&type=Code

fabiojb:
So I think this is a candidate to enhancement in the documentation. I'll open a issue for this in the scoop repository.

PR #1353, opened 2018-10-25
ccleaner: Fix download link

ustsv:
Not download correct file with / without aria2, html page only
1 redirect http 301
2 redirect JavaScript
correct direct link


Ash258:
Change that irrelevant file rename to old `dl.7z`. No need to have it named using version variables.

PR #1354, opened 2018-10-25
Add glogg version 1.1.4 (closes #1341)

### LINKED ISSUE #1341: Request: glogg
Would be nice to have in extras.
https://glogg.bonnefon.org/

> glogg - the fast, smart log explorer
> 
> glogg is a multi-platform GUI application to browse and search through long or complex log files. It is designed with programmers and system administrators in mind. glogg can be seen as a graphical, interactive combination of grep and less.

dzampino:
Add glogg v1.1.4. It has been 64-bit only since version 1.0.3, so I ignored x86. It uses NSIS to install by default, but seems to merely require administrative permission to write to Program Files. I have opted to extract the contents instead since it seems quite happy running as a portable app.

### DIFF
--- a/glogg.json
+++ b/glogg.json
@@ -5,7 +5,7 @@
     "license": "GPL-3.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://glogg.bonnefon.org/files/glogg-latest-x86_64-setup.exe#dl.7z",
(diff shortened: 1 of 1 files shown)


r15ch13:
Nice 👍
Please use `#/dl.7z` as the URL fragment just to be consistent.

PR #1390, opened 2018-11-06
Steam: Tweaks

Ash258:
- Remove my fingerprints / signatures
- Simplify bin
- Don't create folders in installer script, since persisted directories are created when does not exists.


linsui:
Why uninstaller is needed?

PR #1402, opened 2018-11-09
Added CodeTrack

medoni:


Ash258:
- What is ENV file?
- There is extract_dir property inside autoupdate, but none in manifest root and in zip file.
- Change regex capture group to `([\\d\\.]+)`

PR #1408, opened 2018-11-10
Add Scallion v2.1

MCOfficer:
A PR a day keeps the inactivity at bay ^^


MCOfficer:
that's what you get for committing old stuff without testing it once more. thanks for looking over it.

as for the config files: I honestly don't know what they could be, nor how you even found out they exist. so far, I've always used scallion portable, without noticing any difference.

PR #1409, opened 2018-11-10
Winrar: Tweaks

Ash258:
- Persist license keys
- Remove redundant `extract_dir`
- No reason to name file with portable prefix


Ash258:
Pipeline failed because of grafana. Not Winrar

r15ch13:
Ups!

PR #1413, opened 2018-11-10
Fix autoupdate.hash in kdenlive

rasa:


Ash258:
There was already update since last edit 😂. You can push it in this PR.

PR #1425, opened 2018-11-11
Add BeatDrop v1.0.0.0

MCOfficer:
Note: There's no autoupdate (yet) since there's only been 1 release so far.


Ash258:
I think you could assume, that it will be `https://github.com/mvsoft74/BeatDrop/releases/download/v$version/BeatDrop-bin-$version.zip`.

MCOfficer:
i can assume that, but i'll leave the decision to the maintainers who would have to fix it if our prediction is wrong

Ash258:
It would require other maintener to fix it if it's wrong, but if it's correct, it would be automatically, unattended updated and no one will touch it.

PR #1431, opened 2018-11-15
Add manifest to install Archi

FilipDeVos:
Archi is an archimate enterprise architecture tool that can be
downloaded from www.archimatetool.com.

### DIFF
--- a/archi.json
+++ b/archi.json
@@ -41,10 +41,5 @@
             "Archi"
         ]
     ],
-    "bin": [
-        [
-            "archi.exe",
-            "archi"
-        ]
-    ]
+    "bin": "archi.exe"
 }
--- a/archi.json
+++ b/archi.json
@@ -32,7 +32,7 @@
         },
         "hash": {
             "url": "$url.MD5",
-            "find": "MD5\\s+\\(.*\\)\\s+=\\s+([a-fA-F0-9]{32})"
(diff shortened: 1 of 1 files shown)


archi.json
```diff
@@ -0,0 +1,50 @@
...
+            "64bit": {
+                "url": "https://www.archimatetool.com/downloads/root$cleanVersion/Archi-Win64-$version.zip"
+            },
+            "32bit": {
+                "url": "https://www.archimatetool.com/downloads/root$cleanVersion/Archi-Win32-$version.zip"
+            }
+        },
+        "hash": {
+            "url": "$url.MD5",
+            "find": "MD5\\s+\\(.*\\)\\s+=\\s+([a-fA-F0-9]{32})"
```
Ash258: Find could: `MD5 \\($basename\\) = ([A-Fa-f\\d]{32})`

archi.json
```diff
@@ -0,0 +1,50 @@
...
+        }
+    },
+    "shortcuts": [
+        [
+            "archi.exe",
+            "Archi"
+        ]
+    ],
+    "bin": [
```
Ash258: Change to `"bin": "archi.exe"`

archi.json
FilipDeVos: I updated the regex as suggested.

archi.json
FilipDeVos: updated bin call as suggested

PR #1443, opened 2018-11-17
Add picasa & picasa starter

raisercostin:


picasa.json
```diff
@@ -0,0 +1,23 @@
+{
+    "homepage": "https://en.wikipedia.org/wiki/Picasa",
+    "version": "3.9.141.Build-259",
+    "description": "Picasa is a discontinued image organizer and image viewer for organizing and editing digital photos, plus an integrated photo-sharing website by Google",
+    "license": "Freeware",
+    "hash": "482c1a547d8d3aa25ee446d30ea986de63ef8c8d68b8d1109dd3d9b714e73e08",
+    "url": "https://github.com/raisercostin/raiser-scoop-bucket/releases/download/picasa-3.9.141.259/picasa39-setup-3.9.141.259.exe#dl.7z",
```
brandon93s: Why is this pointing to a random release in your repo? Does this need to be in scoop-extras if it's already in your personal bucket?

picasa-starter.json
```diff
@@ -0,0 +1,25 @@
...
+    "shortcuts": [
+        [
+            "PicasaStarter.exe",
+            "PicasaStarter"
+        ]
+    ],
+    "pre_install":[
+      "If(!(Test-Path \"$persist_dir\\PicasaStarterConfiguration.xml\")) { Set-Content \"$persist_dir\\PicasaStarterConfiguration.xml\" -Value $null }",
+      "If(!(Test-Path \"$persist_dir\\PicasaStarterSettings.xml\")) { Set-Content \"$persist_dir\\PicasaStarterSettings.xml\" -Value $null }",
+      "cmd /c mklink /H $dir\\PicasaStarterConfiguration.xml $persist_dir\\PicasaStarterConfiguration.xml",
```
Ash258: Why custom code persist links?? 
What's wrong with 

```json
"persist": ["PicasaStarterConfiguration.xml", "PicasaStarterSettings.xml"]
```

picasa-starter.json
```diff
@@ -0,0 +1,25 @@
+{
+    "homepage": "https://sites.google.com/site/picasastartersite/",
+    "version": "2.0",
+    "description": "Picasa is a discontinued image organizer and image viewer for organizing and editing digital photos, plus an integrated photo-sharing website by Google",
+    "license": "Freeware",
+    "hash": "9188940457742451bfe100bf33df3ddf4ff0fa02ce2b242c599f0632a5a9cfbd",
+    "url": "https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/picasastarter/PicasaStarter%20V2.0.zip",
+    "bin": [
+      ["PicasaStarter.exe","picasaStarter"],
```
Ash258: No need for this rewrite.

picasa.json
```diff
@@ -0,0 +1,23 @@
...
+    "bin": [
+      ["picasa3.exe","picasa"],
+      ["picasa3.exe","picasa3"],
```
Ash258: Irrelevant rewrite

picasa.json
```diff
@@ -0,0 +1,23 @@
...
+      "picasaPhotoViewer.exe"
+    ],
+    "shortcuts": [
+        [
+            "Picasa3.exe",
+            "picasa"
```
Ash258: Shortcuts does not look good with first small letter.

No vendors use this capitalization.

Also one shortcut is enough. (Why there is second totally same?)

picasa.json
raisercostin: The problem is that the picasa is discontinued and the binary is not to be found easily on internet.
I was thinking that there are still people that still use it.
I uploaded a version that shouldn't contain any malware.

picasa-starter.json
raisercostin: PicasaStarter.exe assumes that the configuration files are in the same folder. I saw that the folders are seen locally to executable but the files not.

picasa-starter.json
raisercostin: Agree. I'll remove it.

picasa-starter.json
Ash258: > I saw that the folders are seen locally to executable but the files not.

I don't understand what you said 😶

persist_data function is doing completely same thing: https://github.com/lukesampson/scoop/blob/f284dba5d75ab336d968f6bad719be74261179d2/lib/install.ps1#L1252

picasa.json
raisercostin: agree. fixed

picasa.json
raisercostin: agree. fixed

Ash258: Mark conversations as resolved.

rasa:
I don’t think we should host dead applications in extras, especially if the developer has yanked the executables. Maybe host them in versions, or another bucket named unsupported, or something like that.

moosad:
> 
> 
> I don’t think we should host dead applications in extras, especially if the developer has yanked the executables. Maybe host them in versions, or another bucket named unsupported, or something like that.

It's okay, but how about `peerblock` then? Is it an exception?

picasa-starter.json
raisercostin: fixed. thank your for advices.

PR #1444, opened 2018-11-18
Updated download path for IrfanView Plugins

pxl-th:
They've changed parameters from [path] to [query] in the download link.


brandon93s:
I don't think the `dwl` links will work. Scoop will try to download the webpage instead of the resulting download. 

More info: https://github.com/lukesampson/scoop-extras/issues/1304

rasa:
Fixed in #1449

PR #1454, opened 2018-11-21
kitty: Updated to version 0.70.0.6 and replaced fosshub.com with portableapps.com

moosad:
Fixed broken fosshub.com urls by replacing with portableapps.com links and updated version with one move.

### DIFF
--- a/kitty.json
+++ b/kitty.json
@@ -1,9 +1,9 @@
 {
     "homepage": "http://kitty.9bis.net",
     "license": "MIT",
-    "version": "0.70.0.5",
-    "url": "https://www.fosshub.com/KiTTY.html/kitty_portable.exe",
-    "hash": "7872424693ce23fac1d47f01c0cf09ddde98435de2997e64015ae15143fe9cef",
+    "version": "0.70.0.6",
+    "url": "https://portableapps.com/redirect/?a=KiTTYPortable&s=s&d=pa&f=KiTTYPortable_0.70.0.6_English.paf.exe#/KiTTY_portable.exe",
+    "hash": "b173fb0cbaafc64baf0c60d9656737672a6971994813898745fade1a0b1eefc1",
     "bin": [
         [
             "kitty_portable.exe",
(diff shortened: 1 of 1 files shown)


rasa:
We could download the files directly, as https://chocolatey.org/packages/kitty/0.70.0.6 says;
```
The files are downloaded from the following locations:
 
1) genpass.exe http://www.9bis.net/kitty/?file=genpass.exe
2) kageant.exe http://www.9bis.net/kitty/?file=kageant.exe
3) kitty.exe http://www.9bis.net/kitty/?file=kitty.exe
4) klink.exe http://www.9bis.net/kitty/?file=klink.exe
5) kscp.exe http://www.9bis.net/kitty/?file=kscp.exe
6) ksftp.exe http://www.9bis.net/kitty/?file=ksftp.exe

moosad:
Yes, it definitely seems better. Thanks! :slightly_smiling_face:

moosad:
Sorry, but I seem to have some problem with your suggestion:

- [Chocolatey package](https://chocolatey.org/packages/kitty/0.70.0.6) is non-portable, they set it to behave portable via `kitty.ini`, but it isn't packed to one file anyway.
- The virustotal result of [chocolatey/kitty.exe](https://www.virustotal.com/en/file/a9b122312a9867494f62eb7ce45ea1d00b64161096ce7e57625112d577c03f88/analysis/) is 1 / 67 while the result of [portableapps/KiTTYPortable_0.70.0.6_English.paf.exe](https://www.virustotal.com/#/file/b173fb0cbaafc64baf0c60d9656737672a6971994813898745fade1a0b1eefc1/detection) is 0 / 67.

rasa:
> * [Chocolatey package](https://chocolatey.org/packages/kitty/0.70.0.6) is non-portable, they set it to behave portable via `kitty.ini`, but it isn't packed to one file anyway.

For a kitty install to work, users will need access to all 6 executables, right?

> * The virustotal result of [chocolatey/kitty.exe](https://www.virustotal.com/en/file/a9b122312a9867494f62eb7ce45ea1d00b64161096ce7e57625112d577c03f88/analysis/) is 1 / 67 while the result of [portableapps/KiTTYPortable_0.70.0.6_English.paf.exe](https://www.virustotal.com/#/file/b173fb0cbaafc64baf0c60d9656737672a6971994813898745fade1a0b1eefc1/detection) is 0 / 67.

A 1/67 score is presumably a false positive.

rasa:
This PR doesn't install kitty, it installs the portableapps installer.

moosad:
Yes, but there would be further configuration needed to give access all those executables. It would also be needed to check the current configuration and modify if necessary. 
So why don't we keep using one executable packed as portable?

PR #1462, opened 2018-11-25
winmerge: add x64 version, fix url & extract_dir

ustsv:


winmerge.json
```diff
@@ -14,8 +22,15 @@
     ],
     "checkver": "WinMerge ([\\d\\.]+) - latest stable version",
     "autoupdate": {
-        "url": "https://downloads.sourceforge.net/project/winmerge/stable/$version/WinMerge-$version-exe.zip",
-        "extract_dir": "WinMerge-$version-exe"
+        "extract_dir": "WinMerge",
```
Ash258: Irrelevant extract_dir. There is no varible to be changed on update.

rasa: A constant extract_dir in autoupdate serves no purpose. Can it be removed, please?

ustsv: Ok, but constant var parsing is valid. Left for changes in the zip, only if to simplify json.

Issue #1433, opened 2018-11-16
Add ToDoList


cnzgray:
https://abstractspoon.weebly.com/

This free (for life) Windows desktop application offers an effective and flexible way to stay on top of your daily tasks

Ash258:
Fixed in 443c21c5cfb8f331f7a8594bffb4da3b9f572794 (#1465 )

PR #1465, opened 2018-11-27
todolist: Add app manifest

moosad:
#1433


moosad:
I would need some help with checkver. On the homepage there are only major and minor versions. The only place where I could find the full version number is the [forum](https://groups.google.com/forum/#!forum/abstractspoon-todolist-support), but I could use multi-line regex only.

Ash258:
You can't use forum, because Google forums are heavily generated after execution.

See resulted file of `curl https://groups.google.com/forum/#!forum/abstractspoon-todolist-support -o cosi.html`

You will only see something like: `You need to enable javascript to load this page` message in bunch of javascript ugly code.

Ash258:
Since it's single file URL which does not change with version, you could make it nightly.

moosad:
I found two up-to-date sites, [filehippo](https://filehippo.com/download_todolist) and [techworld](https://www.techworld.com/download/office-business/todolist-7150-3249271/) where I can check the version, so I implemented checkver. Also gave the same autoupdate url as actual url just to compute the latest hash.
I hope it's okay now. 🙂

PR #1468, opened 2018-11-27
Add awareness.

matthewjberger:
Awareness is an application that sits in your taskbar and makes a tibetan singing bowl sound every hour of continuous computer use, and then the sound stops when you take a 5 minute break.

Awareness is a pretty old application so there are no updates and it has no version listed, but it's been useful lately at work


matthewjberger:
going to open a separate issue, sorry

Issue #1312, opened 2018-10-11
Everything settings don't persist


joserebelo:
After opening Everything for the first time, the Everything.db and Everything.ini hardlinks get broken, turning into normal files. I am not sure why this is happening, but I'm guessing Everything deletes and recreates them.

If it turns out to be Everything's fault, one option might be saving the settings to APPDATA, using [the -app-data switch](https://www.voidtools.com/support/everything/command_line_options/#-app-data), or even using the [installer](https://www.voidtools.com/forum/viewtopic.php?f=5&t=5673&p=15546), as it was done with Ditto

rasa:
Closing as this issue is stale. Please reopen if this is still an issue.

PR #1478, opened 2018-11-29
everything: Move ini to a folder to prevent app from breaking hardlinks

moosad:
#1312 
There is two file to persist:
- Everything.ini : it contains the actual settings
- Everything.db : it contains an index of the filesystem

I had the idea to move them to a directory, so the app won't break the hardlinks by recreating the files. And it works, but I had to tell everything.exe to use ones from the directory.
As you can see [here](https://www.voidtools.com/support/everything/command_line_options/#-config), there is a command line option to pass the .ini path at startup. [There's one](https://www.voidtools.com/support/everything/command_line_options/#-db) for the .db as well, but it works in read-only mode only.
So I saw two options: to save data to user/appdata or save just .ini and ignore the index db.

I chose to not to persist db, because I think it's not a big deal after an update to re-index the filesystem. I takes max 4 seconds on my average pc, and the [doc](https://www.voidtools.com/support/everything/using_everything/) also claims it's just a few second.

### DIFF
--- a/everything.json
+++ b/everything.json
@@ -12,15 +12,25 @@
             "hash": "8d3d85888ccd01612993f89a9d79df11edb2f06de78972fa71656f74920ce342"
         }
     },
-    "bin": "Everything.exe",
+    "bin": [
+        [
+            "Everything.exe",
+            "everything",
+            "-config data/Everything.ini"
+        ]
+    ],
     "checkver": "Download Everything ([\\d.]+)",
     "persist": [
-        "Everything.ini",
-        "Everything.db"
+        "data"
     ],
     "pre_install": [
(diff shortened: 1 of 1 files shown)


rasa:
Looks good. Should we see if the ini file already exists in the persist dir and move it to data?

Ash258:
This is nice feature, but it has one disadvantage.

It work only when you run it via bin or from start menu shortcut, but when you run executable on your own other way you don't have these settings.

I am experiencing same with qBittorrent. I have cusom manifest which, use `profile` directory, passed via argument. Everything works as expected until you run qBittorrent within browser (magnet link). Browser will call default executable without that parameter.

moosad:
So what do you think? Could you review/merge if you find it appropriate? :)

PR #1488, opened 2018-12-05
Add the manifest file for the application Hosts File Editor

beppegg:
At work, I frequently use the application [Hosts File Editor](https://hostsfileeditor.com/) for handling my hosts file when I VPN into customers' networks.

I thought it would be useful to contribute the App Manifest back into the main bucket!


beppegg:
Aren't we already in the extras subproject? :open_mouth:

PR #1497, opened 2018-12-08
pidgin: Add version 2.14.8

raisercostin:


pidgin.json
```diff
@@ -0,0 +1,32 @@
...
+            "Pidgin IM"
+        ]
+    ],
+    "checkver": {
+        "url": "https://sourceforge.net/projects/pidgin/files/Pidgin/",
+        "re": ">([\\d.]+)<"
+    },
+    "autoupdate": {
+        "url": "https://downloads.sourceforge.net/project/pidgin/Pidgin/$version/pidgin-$version-offline.exe#/dl.7z",
+        "extract_dir": "$version"
```
Ash258: There is `extract_dir` in autoupdate, but none in root of manifest.

r15ch13:
The skype addon should be added with a separate manifest.
This should only install pidgin just to keep it simple.

raisercostin:
I'm not sure how I could do an addon. There is other way than making two installers?

r15ch13:
@raisercostin It's possible to create a manifest that just downloads and copies the plugin files to the pidgin directory.

- To ensure pidgin is installed it must have the `depends` property set
- The `installer` should contain a script that copies the files to `(appdir pidgin)\\current\\.purple\\plugins` or `$persist_dir\\..\\pidgin\\.purple\\plugins` (same thing)
- Every plugin has it's own `checkver` and `autoupdate` information and can be updated via scoops update command (e.g. `scoop update pidgin-libfacebook`)

Examples for manifests that do cross app stuff:
https://github.com/lukesampson/scoop/blob/master/bucket/lynx.json
https://github.com/lukesampson/scoop/blob/master/bucket/cacert.json
https://github.com/lukesampson/scoop-extras/blob/master/php-xdebug.json

raisercostin:
Wow. Thank you for pointing me in the right direction. I tried to find examples in other buckets or wiki but didn't know how to find.

issaclin32:
I updated the manifest myself, because the PR has been sit for a while.

**NOTES**:
* I deleted the `pidgin-full` (which includes third-party plugins) in the PR. As the comments above stated, the plugins should be put in seperate packages. Otherwise that will be hard to maintain and error-prone (for example, one of the packages drop their old files, causing download to fail)

rashil2000:
On this page https://pidgin.im/install/#windows

image

They have a portable zip - https://sourceforge.net/projects/pidgin/files/Pidgin/2.14.8/pidgin-2.14.8-win32-bin.zip

issaclin32:
> On this page https://pidgin.im/install/#windows
> 
> image
> 
> They have a portable zip - https://sourceforge.net/projects/pidgin/files/Pidgin/2.14.8/pidgin-2.14.8-win32-bin.zip

The problem is **GTK+** is not bundled in the portable version, which will cause `pidgin.exe` fail to start.

rashil2000:
I see. Any idea why `/verify` fails?

Otherwise LGTM

issaclin32:
> I see. Any idea why `/verify` fails?

Turns out that the manifest was not in the `bucket` folder.

**related code**
https://github.com/ScoopInstaller/Scoop-GithubActions/blob/main/src/Action/PR.psm1#L335

logs:

**pidgin**
https://github.com/ScoopInstaller/Extras/runs/4504360754?check_suite_focus=true#step:3:213
```
Log of Github Request:

    Uri     https://api.github.com/repos/ScoopInstaller/Extras/pulls/1497/files 
    Method  Get 
    Headers {Authorization}

INFO: Request Body

VERBOSE: GET with 0-byte payload
VERBOSE: received 2018-byte response of content type application/json
PR Changed Files: 

INFO: Only Changed Manifests

INFO: Checked manifests

Log of Invalids:

    

INFO: No compatible files in PR

Number of Github Requests: 2
```

**free-manga-downloader2** (the normal one)
https://github.com/ScoopInstaller/Extras/runs/4503153160#step:3:213
```
Log of Github Request:

    Headers {Authorization} 
    Method  Get 
    Uri     https://api.github.com/repos/ScoopInstaller/Extras/pulls/4357/files

INFO: Request Body

VERBOSE: GET with 0-byte payload
VERBOSE: received 3044-byte response of content type application/json
PR Changed Files: 

Only Changed Manifests: 

INFO: Starting bucket/webcopy.json checks

Manifest: C:\Users\runneradmin\AppData\Local\Temp\forked_repository\bucket\webcopy.json

```

PR #1508, opened 2018-12-11
Add command box .json

TL-pruddiman:
Added commandbox coldfusion development tool:

"CommandBox is a standalone, native tool for Windows, Mac, and Linux that will provide you with a Command Line Interface (CLI) for developer productivity, tool interaction, package management, embedded CFML server, application scaffolding, and some sweet ASCII art."


TL-pruddiman:
Ah! I was thinking one with the version and auto update would be best. Thanks!

PR #1511, opened 2018-12-12
Add uget

### LINKED ISSUE #1510: Add uGet
uGet is an open source cross-platform download manager. https://ugetdm.com/

There's a 7z archive : https://ugetdm.com/downloads/windows/

It also uses `aria2`, so it could be added as an optional dependency.

Ash258:
- Closes #1510 

For some reason autoupdate cannot extract hash using sourceforge mode.

hash

### DIFF
--- a/uget-portable.json
+++ b/uget-portable.json
@@ -1,0 +1,24 @@
+{
+    "version": "2.2.1",
+    "description": "Open source download manager.",
+    "homepage": "https://ugetdm.com/",
+    "license": "LGPL-2.1-only",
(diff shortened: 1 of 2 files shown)


rasa:
@Ash258: Why provide two versions? uget and uget-portable only differ by one line: 
`"persist": "config",` is in uget-portable.
Perhaps an install switch that ignores the persist tag would make more sense?

Ash258:
Everyone don't want to store settings inside application directory. (Mutliple users, some AD restrictions, ...)

See: palemoon, vscode

rasa:
Sorry, I missed that line.

PR #1519, opened 2018-12-14
add SwitchHosts!

batkiz:


r15ch13:
@rasa Checkver can be changed to use the GitHub feature and the $majorVersion in autoupdate can be replaced with $version.

rasa:
Ok, will do.

PR #1554, opened 2018-12-21
Add LDPlayer

Retia-Adolf:
An Android emulator optimized for mobile gaming with lower resource consumption


Retia-Adolf:
I'm downloading. No sure what these versions for other regions look like, they seems to have different tactics to push advertisements for different regions (or not).

PR #1555, opened 2018-12-21
Add Portable-VirtualBox

Retia-Adolf:
Due to this launcher's disability to download integrated installer, I added them into the manifest both.


janpio:
Great work of adding Portable-VirtualBox.

Can you explain why this also needs to download Virtualbox and the extension pack? Are we supposed to select them in the UI of Portable-Virtualbox?

(I figured out the files end up as `VirtualBox.exe` and `Extension` (no file extension)  in `C:\Users\...\scoop\apps\portable-virtualbox\current`)

Also, you maybe can take a look at #1677 where I describe a possible problem with the version naming of Portable-VirtualBox.

PR #1566, opened 2018-12-25
Added TortoiseGit

nitrocode:
https://github.com/lukesampson/scoop-extras/issues/1435

### DIFF
--- a/tortoisegit.json
+++ b/tortoisegit.json
@@ -1,0 +1,56 @@
+{
+    "homepage": "https://tortoisegit.org",
+    "description": "a Windows Shell Interface to Git and based on TortoiseSVN",
+    "version": "2.7.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://download.tortoisegit.org/tgit/2.7.0.0/TortoiseGit-2.7.0.0-64bit.msi",
+            "hash": "ea4d68423484143de7c2afe75d28eed29ecb695e3b4ad41c47d8365fe70c76bd"
+        },
+        "32bit": {
+            "url": "https://download.tortoisegit.org/tgit/2.7.0.0/TortoiseGit-2.7.0.0-32bit.msi",
+            "hash": "5c9c734cd5b354f5fdd485793677430a067a629f5019410d65a31e3306a81470"
+        }
+    },
+    "extract_dir": "Program Files\\TortoiseGit",
+    "bin": [
+        "bin\\TortoiseGitBlame.exe",
+        "bin\\TortoiseGitIDiff.exe",
+        "bin\\TortoiseGitMerge.exe",
+        "bin\\TortoiseGitPlink.exe",
+        "bin\\TortoiseGitProc.exe"
+    ],
+    "shortcuts": [
+        [
+            "bin\\TortoiseGitBlame.exe",
+            "TortoiseGitBlame"
+        ],
+        [
+            "bin\\TortoiseGitIDiff.exe",
+            "TortoiseGitIDiff"
+        ],
+        [
+            "bin\\TortoiseGitMerge.exe",
+            "TortoiseGitMerge"
+        ],
+        [
+            "bin\\TortoiseGitPlink.exe",
+            "TortoiseGitPlink"
+        ],
+        [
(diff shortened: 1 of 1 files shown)


Ash258:
You are missing the main feature, why is everyone installing Tortoise applications. Explorer addons need to be setupped manually using registry manipulation. Something like [this](https://github.com/lukesampson/scoop-extras/tree/master/scripts/tortoisesvn)

nitrocode:
Hi @Ash258, thanks for commenting. I didn't realize you were already working on. Is the main feature the shell extensions? I was trying to practice my scoop packaging skills for fun so I didn't understand how essential the file associations are.

I'm not sure I fully understand the TortoiseSVN registry scripts. I can see what the reg is supposed to do but I don't understand how, it's uncommented it. Also when I run the reg after `scoop install tortoisesvn`, I do not have additional options in my context menu...

Is there any reason a .reg file was chosen instead of a powershell command to associate file types with binaries? and any reason for it to be external to the json?

Ash258:
Notes

Tortoise applications are used because they provide contex menus for folders, files, which are under specific source control, so you do not have to use command line. Without [icon overlays](https://tortoisegit.org/docs/tortoisegit/tgit-dug.html#tgit-dug-general-icons) you have only new command line tool for git.

All needed registry keys could be found [here](https://github.com/TortoiseGit/TortoiseGit/blob/master/src/TortoiseGitSetup/StructureFragment.wxi), but I don't have time and sanity to dump them into adequate reg file

> could scoop be modified to allow for file associations to be configured directly 

In this case definitely no. There are lots of keys, which need to be set correctly. 

On general side of registering. Manifests should not change something like default associations on installing. Somebody don't want to change it. You can't just force user to change defaults and make him set it back (and over and over while updating). If you want some custom associations, make custom bucket / remote gist for it.

What would be acceptable if scoop could register some applications, into windows settings, so user could easily set it (lukesampson/scoop#2651)

nitrocode:
Thanks for the screenshot... I see what you mean regarding the associations for this application. As for the tortoisesvn I can find the registry files and run them but when I run them it doesn't do what it intends to do for me on Windows 10 which is to add the associations. It seems broken for me.

That's why I was thinking maybe we can run a powershell script that would add the file associations similar to the link above. There's no reason it _has_ to be a `.reg` file right?

    $fileType = (cmd /c "assoc $extension") 
    $fileType = $fileType.Split("=")[-1]  
    cmd /c "ftype $fileType=""$OpenAppPath"" ""%1"""

I can see your doubts about adding the file association in the manifest itself. I imagine that feature could be optional so I added my thoughts to this [thread](https://github.com/lukesampson/scoop/issues/2567#issuecomment-449863657).

Ash258:
It's reg file, because:

1. Everyone can run it by double clicking, without some terminal opening, copying commands.
1. It's easily editable using powershell (`(Get-Content reg_template) -replace '%%someSpecificValueToReplace', "RealValueOfPathToTgit" | Set-Content 'full_easily_runnable.reg'`)

I haven't installed TortoiseSVN, because i dont want to mess my registry with junk values just for testing, but it's possible it's not working. It's pretty old file, maybe they changed something.

nitrocode:
@Ash258 okay, I'll leave it to you then. You've made more progress than me on this anyway. 😄

Ash258:
Feel free to finish. I am not going to finish it in near future. I just made investigation and was pretty upset about their implemntation and maybe they finally release some portable version or some better way of doing these manipulations (https://gitlab.com/tortoisegit/tortoisegit/issues/726, https://stackoverflow.com/questions/53708161/how-to-manually-register-context-menus-using-command-line).

Everything excepts that registry files is ready. Feel free to reuse my implementation and write that registry templates.
You could just combine [svn](https://github.com/lukesampson/scoop-extras/blob/6bcf53e0b42a8c726c171efe2175ec46601cdc84/tortoisesvn.json#L59-L64) and [git](https://github.com/Ash258/scoop-Ash258/blob/master/TODO/tortoisegit.json)

haussmann:
Thank you for your discussions, these helped to create a non-portable manifest: please take a look at `nonportable/tortoisegit-np`.

PR #1574, opened 2018-12-29
Add Lantern-client

Retia-Adolf:


r15ch13:
Adding `#/dl.7z` to the end of the URL let's scoop handle the extraction. No need for doing that in the installer script.

r15ch13:
@Retia-Adolf ping

PR #1520, opened 2018-12-14
Add Jetbrains Rider

moosad:
On request  #1434

PR #1583, opened 2019-01-04
Added `rider.json` to add JetBrains Rider to Scoop Extras

hoest:
See https://www.jetbrains.com/rider/

### DIFF
--- a/rider.json
+++ b/rider.json
@@ -6,26 +6,13 @@
     "bin": "bin\\rider.bat",
     "env_add_path": "bin",
     "notes": "Please restart your command line for changes to take effect.",
-    "architecture": {
-        "64bit": {
-            "bin": "bin\\rider64.exe",
-            "shortcuts": [
-                [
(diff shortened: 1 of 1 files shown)


Ash258:
There is no 32bit version for rider. https://rider-support.jetbrains.com/hc/en-us/articles/207876035-Rider-system-requirements

Also see: lukesampson/scoop-extras#1520

rasa:
Dup of #1520.  See https://github.com/lukesampson/scoop-extras/pull/1520#issuecomment-447385484 . Is that an acceptable alternative?

hoest:
> Dup of #1520. See [#1520 (comment)](https://github.com/lukesampson/scoop-extras/pull/1520#issuecomment-447385484) . Is that an acceptable alternative?

OK, so it's recommended to use the Jetbrains Toolbox, that's OK with me. Maybe it's a good idea to remove other Jetbrains tools from the bucket, like Webstorm or PyCharm?

Issue #1573, opened 2018-12-28
musescore@3.0: download via aria2 failed


panathos:
ERROR Download failed! (Error 3) Resource was not found
DEBUG https://ftp.osuosl.org/pub/musescore/releases/MuseScore-3.0/MuseScore-3.0.msi
    referer=https://ftp.osuosl.org/pub/musescore/releases/MuseScore-3.0/
    dir=D:\Scoop\cache
    out=musescore#3.0#https_ftp.osuosl.org_pub_musescore_releases_MuseScore-3.0_MuseScore-3.0.msi

DEBUG & 'D:\Scoop\apps\aria2\current\aria2c.exe' --input-file='D:\Scoop\cache\musescore.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/5.1 (Windows NT 10.0; Win64; x64; Desktop)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn --enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=2988 --continue

ruifigueira:
The URL for download is now `https://download.musescore.com/releases/MuseScore-3.0.0/MuseScore-3.0.0.msi` and not `https://ftp.osuosl.org/pub/musescore/releases/MuseScore-3.0/MuseScore-3.0.msi`.

I created a PR, #1584, to fix it.

riaanvn:
Is there a reason why the Version variable was not brought in in line with the URL? e.g. I believe
"version": "3.0"
should be
"version": "3.0.0"

Looking at the release notes

riaanvn:
Hi Rui

Where did you hear that you should be using
https://download.musescore.com/releases
instead of 
https://ftp.osuosl.org/pub/musescore/releases
(the new URL just looks a lot nicer. Our Download Page, https://musescore.org/en/download, is still using the old URL and I need them to get it changed.

ruifigueira:
Hi @riaanvn,

I got it from the direct link at https://musescore.org/en/download/musescore.msi:

![image](https://user-images.githubusercontent.com/1374559/50917802-6b26b880-1436-11e9-85f5-0939fb20af95.png)

riaanvn:
[...]

PR #1584, opened 2019-01-04
Fix musescore URL

ruifigueira:
Fixes: #1573

### DIFF
--- a/musescore.json
+++ b/musescore.json
@@ -3,19 +3,19 @@
     "homepage": "https://musescore.org/",
     "version": "3.0",
     "license": "GPL-2.0-only",
-    "url": "https://ftp.osuosl.org/pub/musescore/releases/MuseScore-3.0/MuseScore-3.0.msi",
+    "url": "https://download.musescore.com/releases/MuseScore-3.0.0/MuseScore-3.0.0.msi",
     "hash": "53c1bc584035978515f83eca371105a22e98251c0e7023875e32f567f8b40cc4",
-    "extract_dir": "MuseScore 2",
+    "extract_dir": "MuseScore 3",
     "bin": [
-        "bin\\MuseScore.exe",
+        "bin\\MuseScore3.exe",
         [
-            "bin\\MuseScore.exe",
+            "bin\\MuseScore3.exe",
             "mscore"
         ]
     ],
     "shortcuts": [
         [
-            "bin\\MuseScore.exe",
+            "bin\\MuseScore3.exe",
             "MuseScore"
         ]
     ],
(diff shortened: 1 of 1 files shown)


Ash258:
Add `"extract_dir": "MuseScore $majorVersion"` into autoupdate property.

Ash258:
Also fix checkver to match `releases/MuseScore-([\\d\\.])/` from `https://musescore.org/en/download/musescore.msi` to get correct version without manually adding 0.

riaanvn:
Hi ruifigueira

(apologies for the duplicate post, still finding my feet)

Is there a reason why the Version variable was not brought in in line with the URL? e.g. I believe
"version": "3.0"
should be
"version": "3.0.0"

The next version is expected to be 3.0.1

shoogle:
Thanks for updating the script. I just wanted to point out a few consequences of these changes:

1. Renaming `MuseScore3.exe` to `MuseScore.exe` means that it is not possible to run two versions at the same time (see [this](https://answers.microsoft.com/en-us/windows/forum/windows_7-windows_programs/how-can-i-run-multiple-versions-of-a-program/d61d608b-a2c0-4a5e-b47c-a549f329af1d) and this). This is why `MuseScore.exe` is versioned, but if versioning interferes with scoop's functionality then it is of course up to you guys to decide which functionality you wish to prioritise.

2. Checking the version from `https://musescore.org/en/download/musescore.msi` rather than from `https://musescore.org/en/download` will detect a new version before the MuseScore developers are ready to publish it. Usually they wait a few days after uploading the new version to allow people to test it before publishing it in the download page, and if problems are found they will upload a new version to the same URL. I have contacted the MuseScore team to request that they provide a stable API to the current version that exactly matches the version in the URL.

r15ch13:
1. Hm, for Firefox this is true, but other programs like 7-Zip and MuseScore don't care about this. :smile:
image
image



2. 

> if problems are found they will upload a new version to the same URL

If they keep the version (3.0.0) and just update the, then this will break the manifest anyway.

Musescore itself uses this [appcast.xml](https://sparkle.musescore.org/stable/3/win/appcast.xml) to check for a new version including a build number. Using the full version.build would be the best way.

Version 2 used https://update.musescore.org/update_win_stable.xml, so I guess the URL will change for version 4 anyway. 😁

r15ch13:
https://github.com/lukesampson/scoop-extras/commit/a1f1d02141cf350a8dee071449e43b3b9e408caa

shoogle:
Thanks for fixing! Good idea to use the sparkle appcast!

> If they keep the version (3.0.0) and just update the file, then this will break the manifest anyway.

Indeed, but they wouldn't do this after changing the "current stable version is" text on the download page, which is what my regex was originally checking. However, the appcast is a better solution because there is a small possibility that they do a Windows-only release, e.g. to get the existing program working on a new "version" of Windows (Windows 10 is supposed to be the last version of Windows but you know what I mean - after a big update).

> Hm, for Firefox this is true, but other programs like 7-Zip and MuseScore don't care about this.

Interesting. This is probably because MuseScore 3 uses a different version of Qt to MuseScore 2. Probably whoever changed the executable name was going from prior knowledge that names *sometimes* cause conflicts and didn't test it in this specific case, or tested it but decided not to take the risk.

shoogle:
So I spoke to the MuseScore developers again and they gave me some more reasons for versioning the `MuseScore3.exe` executable. These include:

- Conflicts in the Windows Registry
- Easy distinguishability in the Open With list
- To allow MuseScore 2 to remain available in the Windows Store

None of these reasons are relevant to Scoop.

PR #1596, opened 2019-01-09
Ditto: Fix manifest

niheaven:
Current manifest of Ditto is a broken one, it's not portable and the `persist` is wrong.

The fixes are:

- Change `url` from installer to portable zip
- `homepage` move to new one
- Add `autoupdate.hash`
- Fix `persist` by some slightly complicated hardcode (start ditto.exe, wait until it created ditto.db, kill ditto.exe)

Now it works fine, at least for me.

### DIFF
--- a/ditto.json
+++ b/ditto.json
@@ -1,5 +1,9 @@
 {
     "version": "3.22.20.0",
+    "description": [
+        "Clipboard Manager that saves each item placed on the clipboard",
+        "allowing you access to any of those items at a later time."
+    ],
     "homepage": "https://ditto-cp.sourceforge.io/",
     "license": "GPL-3.0-only",
     "architecture": {
@@ -33,7 +37,7 @@
         "$file = 'Ditto.Settings'",
         "if (!(Test-Path \"$persist_dir\\$file\")) {",
         "    Write-Host 'File' $file 'does not exists. Creating.' -f Yellow",
-        "    $CONT = @('[Ditto]\r\nDBPath3=Ditto.db')",
+        "    $CONT = @('[Ditto]', 'DBPath3=Ditto.db')",
         "    Set-Content \"$dir\\$file\" ($CONT -join \"`r`n\") -Encoding Ascii",
         "}"
     ],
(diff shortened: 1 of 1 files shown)


ditto.json
```diff
@@ -1,38 +1,59 @@
...
+        "    while (!(Test-Path \"$dir\\Ditto.db\")) {",
+        "        Start-Sleep -Milliseconds 500",
+        "    }",
+        "    Start-Sleep 1",
+        "    Stop-Process -Name Ditto",
+        "}",
+        "$file = 'Ditto.Settings'",
+        "if (!(Test-Path \"$persist_dir\\$file\")) {",
+        "    Write-Host 'File' $file 'does not exists. Creating.' -f Yellow",
+        "    $CONT = @('[Ditto]\r\nDBPath3=Ditto.db')",
```
Ash258:
```text
[Ditto]\r\nDBPath...
```

Will be in final file. 

Make it `@('[Ditto]', 'DBPath3=Ditto.db')` appending line ending is done in L37

ditto.json
```diff
@@ -1,38 +1,59 @@
...
             "32bit": {
-                "url": "https://downloads.sourceforge.net/project/ditto-cp/Ditto/$version/DittoSetup_$underscoreVersion.exe"
+                "url": "https://downloads.sourceforge.net/project/ditto-cp/Ditto/$version/DittoPortable_$underscoreVersion.zip"
             },
             "64bit": {
-                "url": "https://downloads.sourceforge.net/project/ditto-cp/Ditto/$version/DittoSetup_64bit_$underscoreVersion.exe"
+                "url": "https://downloads.sourceforge.net/project/ditto-cp/Ditto/$version/DittoPortable_64bit_$underscoreVersion.zip"
             }
+        },
+        "hash": {
```
Ash258: It's downloding from sourceforge, so hash extraction is done automatically using sourceforge mode.

ditto.json
```diff
@@ -1,38 +1,59 @@
 {
     "version": "3.22.20.0",
-    "extract_dir": "Ditto",
+    "homepage": "https://ditto-cp.sourceforge.io/",
```
Ash258: No description

ditto.json
niheaven: Fixed.

ditto.json
niheaven: This mode is not mentioned in the wiki... Fixed.

ditto.json
niheaven: Added description.

PR #1601, opened 2019-01-10
Tweak tcpipmanager

Ash258:
- Add:
    - license
    - description
    - checkver
    - autoupdate
- Persist bin folder
- Use direct sourceforge url
    - For automatic hash extracion

### DIFF
--- a/tcpipmanager.json
+++ b/tcpipmanager.json
@@ -1,20 +1,48 @@
 {
+    "version": "4.1.1.29",
+    "description": "Keep track of network configuration in different locations.",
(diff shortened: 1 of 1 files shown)


rasa:
Dump question: should we be using each app’s RSS feed for checkver as the feed contains an md5 hash?
See 
https://github.com/lukesampson/scoop-extras/search?q=sourceforge+%2Frss&unscoped_q=sourceforge+%2Frss
and https://sourceforge.net/projects/tcpipmanager/rss?path=/
This potentially could even be built into scoop, say
`"checkver": {"sourceforge": "regex"}`

PR #1629, opened 2019-01-14
pdf2djvu: Fix extract_dir

jwilk:


jwilk:
Disclaimer: this is untested and I have no idea what I'm doing.

PR #1633, opened 2019-01-15
Fix IrfanView Hashes

MCOfficer:
Yes, mea culpa, mea maxima culpa (#1632). In my defense, you didn't notice it, either ;)


r15ch13:
Oops 😁

PR #1634, opened 2019-01-15
Use non mirror apache url

edgardmessias:
See tomcat for example

### DIFF
--- a/bucket/netbeans.json
+++ b/bucket/netbeans.json
@@ -1,7 +1,7 @@
 {
     "version": "10.0",
     "license": "CDDL-1.0|GPL-2.0-only",
-    "url": "https://www-eu.apache.org/dist/incubator/netbeans/incubating-netbeans/incubating-10.0/incubating-netbeans-10.0-bin.zip",
+    "url": "https://www.apache.org/dist/incubator/netbeans/incubating-netbeans/incubating-10.0/incubating-netbeans-10.0-bin.zip",
     "homepage": "https://netbeans.org",
     "hash": "sha512:ba83575f42c1d5515e2a5336a621bc2b4087b2e0bcacb6edb76f376f8272555609bdd4eefde8beae8ffc6c1a7db2fb721b844638ce27933c3dd78f71cbb41ad8",
     "extract_dir": "netbeans",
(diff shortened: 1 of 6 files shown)


edgardmessias:
Ping @Ash258

r15ch13:
This will cause issues because PowerShell doesn't like redirects from HTTPS to HTTP.

edgardmessias:
@r15ch13 , maybe change all URLs to use `https://www.apache.org/dist/`?

edgardmessias:
@r15ch13 @Ash258, solved problem with redirect from HTTPS to HTTP, now, only redirect from HTTP to HTTP or HTTP to HTTPS

Ash258:
I am boucing off this. Someone else review it.

- I don't use any of this
- I don't like http urls, when https are available, so this PR is totally unmergable for me.

r15ch13:
I agree on this @Ash258, downgrading to http just because some mirrors don't support https or a dynamic mirror redirection that doesn't favour https shouldn't be an option.
It would be better to use a fixed mirror with https or try to catch the error on powershell 6 and tell the user that there was an attempt to redirect to http.

_Sent from my OnePlus5T using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

edgardmessias:
Changed to use default apache url

bucket/zeppelin.json
```diff
@@ -17,10 +17,11 @@
         "re": "zeppelin-([\\d.]+)-bin-all.tgz"
     },
     "autoupdate": {
-        "url": "https://archive.apache.org/dist/zeppelin/zeppelin-$version/zeppelin-$version-bin-all.tgz#/zeppelin.tar.gz",
+        "url": "https://www.apache.org/dist/zeppelin/zeppelin-$version/zeppelin-$version-bin-all.tgz#/zeppelin.tar.gz",
         "extract_dir": "zeppelin-$version-bin-all",
         "hash": {
-            "url": "$url.sha512"
+            "url": "$url.sha512",
+            "find": "^($sha512)"
```
r15ch13: `"find": "^($sha512)"` is not required, scoop knows the file format and extracts the checksum automatically.

bucket/netbeans.json
```diff
@@ -28,9 +28,10 @@
         "re": "Apache NetBeans ([\\d.]+)</a></li>"
     },
     "autoupdate": {
-        "url": "https://www-eu.apache.org/dist/incubator/netbeans/incubating-netbeans/incubating-$version/incubating-netbeans-$version-bin.zip",
+        "url": "https://www.apache.org/dist/incubator/netbeans/incubating-netbeans/incubating-$version/incubating-netbeans-$version-bin.zip",
         "hash": {
-            "url": "$url.sha512"
+            "url": "$url.sha512",
+            "find": "^($sha512)"
```
r15ch13: `"find": "^($sha512)"` is not required, scoop knows the file format and extracts the checksum automatically.

PR #1635, opened 2019-01-16
Add qTox 1.16.3

Ash258:
- Closes lukesampson/scoop#2959


linsui:
Just a question, should we change the qtox.ini when installing to make qtox be portable? I want to know what the rule is, thanks.

PR #1647, opened 2019-01-18
atom: Modify autoupdate.hash

niheaven:
- Using nupkg instead of zip
- Using `$version` in regex to find hash in github `RELEASE` file.(https://github.com/lukesampson/scoop/commit/3a22526554d368edbd4f7aa177540ee749c32e72)


niheaven:
Remove all `#dl.7z` in url.

PR #1656, opened 2019-01-23
Add universal-ctags

deevus:
Originally opened https://github.com/scoopinstaller/nightlies/pull/3 but as discussed that repo is dead and here is a better choice.

### DIFF
--- a/bucket/universal-ctags.json
+++ b/bucket/universal-ctags.json
@@ -1,0 +1,33 @@
+{
+    "version": "2019-01-18",
+    "license": "",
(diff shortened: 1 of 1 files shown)


bucket/universal-ctags.json
```diff
@@ -0,0 +1,33 @@
+{
+    "version": "2019-01-18",
+    "license": "",
```
r15ch13:
```suggestion
    "license": "GPL-2.0-or-later",
```

deevus:
@r15ch13 Updated

PR #1667, opened 2019-01-26
add pycharm-professional

Wangrong23:


Ash258:
1. No need to append bin to path
1. It's not 'IDEA Pycharm', but 'PyCharm'
1. Wrong autoupdate url. (missing dash before $version)
1. NO description

For extras bucket it's recommended to use JetBrains Toolbox app. (See https://github.com/lukesampson/scoop-extras/pull/1520#issuecomment-447385484)

I could recommend you https://github.com/Ash258/Scoop-JetBrains

PR #1672, opened 2019-01-28
Add PhpStorm to the bucket

astappiev:


Ash258:
https://github.com/lukesampson/scoop-extras/pull/1667#issuecomment-457822759

PR #1673, opened 2019-01-28
Add Regkeys for BeyondCompare Context Menus

f0ff886f:
This adds registry keys to enable Beyond Compare in the context menus in Windows (things like "Select as Left File for Comare"), which is one of the key features of BComp.

Note, I have the paths set to my fork, they need to be changed to your tree before merge, should I do that on my branch?

Also, one other note. When you remove the context option with the regkey, you need to restart explorer to remove the app folder. This is unfortunate, but I feel this downside far outweighs the upside (could capture this in notes if they are available during uninstall).


Ash258:
1. Do not download scripts. Copy it from bucketsdir (Especially from your forked respository)
1. use folder structure inside scripts folder (as tortoisesvn)
1. DRY. Same if for 2 files. Make it function or directly foreach through options

```powershell
@('install32.reg', 'install64.reg') | ForEach-Object { $content ... }
```

See:
https://github.com/lukesampson/scoop-extras/blob/82d9509c70fd2dead1bca9e5881832973db9445a/bucket/tortoisesvn.json#L59-L64

Ash258:
This cannot be done in manifest directly it's need to be option. In that case you force all users to enable this, and not everyone want this inside their registry (organizations, users who like their registry clean)

This needs extrenal, simple applied script for all users (no need to know powershell or anything), .reg files serves this greatly. Only valid point is that system wide thing.

Fixable using only simple condition I believe:

`if ($global) { replace HKCU with HKLM }`

7wells:
Hello! I am curious to know if this is still work in progress. Thank you!

phanirithvij2000:
not stale

rashil2000:
The remote repository for this PR is deleted. @f0ff886f please open a new PR after resolving conflicts.

PR #1681, opened 2019-01-31
Add KeeAgent - KeePass 2.x plugin

ebugusey:
https://lechnology.com/software/keeagent/

KeePass doesn't have plugin update system like foobar2000 or notepad++. Package management system is useful for this task.


ebugusey:
@r15ch13 Is there some problem with this manifest? Is it not appropriate for this repo?

Issue #1375, opened 2018-11-04
ConEmu update overwrites configuration


niboan:
The latest changes to conemu.json (52485ee) breaks the ConEmu configuration when updating.

The `ConEmu.xml` file is overwritten with the default on every update.

Also, it doesn't seem that persisting the `ConEmu\ConEmu.xml` file works.
I'm not sure how persisting a single file should work, but after installation, there are two separate `ConEmu.xml` files in `persist\conemu`and `scoop/apps/conemu/current/ConEmu`.

r15ch13:
Should be fixed now.

niboan:
Previously, the default configuration location was `%APPDATA%\Roaming\ConEmu.xml` (which makes persisting the file redundant). On update, a new default configuration will be created in `scoop/apps/conemu/current/ConEmu`. ConEmu will now use the newly created default configuration instead of the old one.
I think it would be better to let ConEmu handle creation of the default configuration. It locates existing configuration in all supported locations, so existing configurations will not be overwritten.
ConEmu also handles discovery of a Far Manager installation, and configures it as the default console, if found.

rasa:
@niboan Did this change work for you? If not, can you describe how to let ConEmu create the config? I don’t use the tool. Thanks.

niboan:
> @niboan Did this change work for you?

Not really. If ConEmu was installed before the change that creates a default configuration, the configuration would probably be in `%APPDATA%\Roaming\ConEmu.xml`.
When updating, this file will be ignored, and a new default configuration will be created.
To restore your old configuration, you would have to manually copy your old `ConEmu.xml` to `scoop/apps/conemu/current/ConEmu/`. Copying to `scoop/persist/conemu/` seems to break the hardlink on the file.

> If not, can you describe how to let ConEmu create the config?

If no configuration exists, ConEmu will prompt the user to create a default configuration when started for the first time.

niboan:
On a side note, there is no reason to persist the `plugins`folder. According to the documentation (https://conemu.github.io/en/ConEmuFolders.html), it's not used to install ConEmu plugins, but contains only a Far Manager plugin.

rasa:
Closing per 
https://github.com/lukesampson/scoop-extras/pull/1689#issuecomment-459870330

PR #1689, opened 2019-02-01
conemu: Remove plugins from persisted data

ebugusey:
According to https://conemu.github.io/en/ConEmuFolders.html, plugins directory
is needed for Far Manager only. It's added as plugins source when Far is run from
ConEmu, therefore it should be updated with ConEmu and shouldn't be persisted.

### DIFF
--- a/bucket/conemu.json
+++ b/bucket/conemu.json
@@ -33,8 +33,7 @@
         ]
     ],
     "persist": [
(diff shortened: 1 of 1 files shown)


ebugusey:
On a side note. I have tested claims in #1375. Persisting `ConEmu.xml` works as intended. Issue should be closed.

PR #1690, opened 2019-02-02
Add Database.NET

wzv5:


bucket/database.net.json
```diff
@@ -0,0 +1,25 @@
...
+            "Database.NET"
+        ]
+    ],
+    "persist": "Database_Files",
+    "checkver": {
+        "url": "https://fishcodelib.com/version.xml",
+        "re": "<ver name=\"database4\" md5=\"[a-zA-Z0-9]+\" ver=\"([\\d.]+)\""
+    },
+    "autoupdate": {
+        "url": "https://fishcodelib.com/files/DatabaseNet4.zip"
```
Ash258:
```suggestion
        "url": "https://fishcodelib.com/files/DatabaseNet4.zip",
        "hash": {
            "url": "https://fishcodelib.com/version.xml",
            "regex": "database4\"\\s*md5=\"([A-Za-z\\d]{32})\""
        }
```

Ash258: It's one time bug or temporary "solution"?

wzv5: I don't know. But now all md5 in the xml is incorrect...
So maybe we shouldn't use it.

Ash258: There is no problem to have it. If it's one time issue next update will be fine and extracted automatically. If it's persist, hash extraction will fail and file will be downloaded (No problem) and developers could be contacted about this.

r15ch13: Hm, I don't think this is a one time issue:

- database5pro is missing a zero at position 1
- database4 is missing a zero at position 13
- dbmigration is correct
- dictionary is wrong
- convert is missing a zero at position 15

Maybe this should report this to fish.

PR #1692, opened 2019-02-03
Create portable_settings.txt before install to persist as file, not dir

f0ff886f:
Currently if you install heidisql, the persist happens as a dir and it causes errors in the app (it will crash trying to write to portable_settings.txt).

### DIFF
--- a/bucket/heidisql.json
+++ b/bucket/heidisql.json
@@ -19,6 +19,7 @@
             "HeidiSQL"
         ]
     ],
+    "pre_install": "New-Item \"$dir/portable_settings.txt\" > $null",
     "persist": "portable_settings.txt",
     "checkver": {
         "url": "https://www.heidisql.com/download.php",


Ash258:
Use check for persist_dir rather, then always creating that file

https://github.com/lukesampson/scoop-extras/blob/a84b257fd9636d02295b48c3fd32826487ca9bd3/bucket/aida64extreme.json#L10-L15

f0ff886f:
Is there a PR or a motion to improve some of the tooling out of the box with Scoop? I ask because there are too many undocumented "correct" ways to do things.

If we want to persist a file, isn't it better to just provide an option to the package to persist a file explicitly?

There is no mention here that persisting by default creates dirs: https://github.com/lukesampson/scoop/wiki/Persistent-data

Then when you run into the "dir" issue, you think OK, lets check the source: https://github.com/lukesampson/scoop/blob/master/lib/install.ps1 which has a [nice comment](https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L1218-L1221):

```
# we don't have neither source nor target data! we need to crate an empty target,
# but we can't make a judgement that the data should be a file or directory...
# so we create a directory by default. to avoid this, use pre_install
# to create the source file before persisting (DON'T use post_install)
```

Then you see scripts like https://github.com/lukesampson/scoop-extras/blob/master/bucket/wiztree.json do it a way, so ok, you go that way.

If we had more sane tooling, we could enable more people to create higher quality packages from the start instead of a back and forth each time.

Anyways, I will make the change, just wanted to say that tooling in a package manager gives power to the people :)

bucket/heidisql.json
```diff
@@ -19,7 +19,15 @@
             "HeidiSQL"
         ]
     ],
-    "pre_install": "New-Item \"$dir/portable_settings.txt\" > $null",
+    "installer": {
+        "script": [
+            "$FILE = 'portable_settings.txt'",
+            "if (!(Test-Path \"$persist_dir\\$FILE\")) {",
+            "    Write-Host 'No' $FILE 'exists. Creating.' -f Yellow",
+            "    New-Item \"$dir\\$FILE\" -Type File",
```
Ash258:
```suggestion
            "    New-Item \"$dir\\$FILE\" -Type File | Out-Null",
```

h404bi:
Actually, you can leave it there, the persist logic will check if there is already `target` in `persist_dir`. And it will override the `source` when the app updates.

Ash258:
But if you do not check for it and create all the time, it will be duplicated. If you have large files / folders, you take much more space than you need.

For example. I experienced this while persisting folder, which is about 8GB on default. It's isos files on enterprise E2B build, and when you persist it you will take these 8GB for original and different size (persisted from before, which is usually bigger as you add more ISOs, now about ~100GB for all windows ISOs and partition images). So persist check is needed to save space (in this case deleting not creating, if there are already persited files).

PR #1694, opened 2019-02-04
Fix broken dnSpy with the latest

parthopdas:


Ash258:
1. Do not create shortcut for console.
1. Move it to correct folder in repository (bucket)
1. Use architecture for 32bit and 64bit
1. Add description
1. Fix autoupdate aswell for both architectures
 
Arch

PR #1699, opened 2019-02-05
Add sccache

Jake-Shadle:
https://github.com/mozilla/sccache


r15ch13:
Seems to be a console only program, please reopen PR on [main bucket](https://github.com/lukesampson/scoop/).

Jake-Shadle:
Ahh ok, I didn't think it qualified for the "a reasonably well-known and widely used developer tool" criteria, but will do.

PR #1701, opened 2019-02-05
Add Kombucha v0.1.0

Kiedtl:
Add [Kombucha](https://github.com/lptstr/kombucha), a simple package manager for PowerShell.
Oh, and I didn't use any datetime hack 😜

### DIFF
--- a/bucket/kombucha.json
+++ b/bucket/kombucha.json
@@ -1,0 +1,13 @@
+{
+    "homepage": "https://github.com/lptstr/kombucha",
+    "version": "0.1.0",
+    "url": "https://codeload.github.com/lptstr/kombucha/zip/v0.1.0",
+    "extract_dir": "kombucha-0.1.0",
+    "hash": "051dd37864b9eac7262ccd9eec500f49b849e59d42e476341e36acb91dba041f",
+    "bin": "bin\\kombucha.ps1",
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://codeload.github.com/lptstr/kombucha/zip/v$version",
+        "extract_dir": "kombucha-$version"
+    }
+}
--- a/bucket/kombucha.json
+++ b/bucket/kombucha.json
@@ -1,5 +1,6 @@
 {
     "homepage": "https://github.com/lptstr/kombucha",
+    "license": "MIT",
     "version": "0.1.0",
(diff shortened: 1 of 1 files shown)


bucket/kombucha.json
```diff
@@ -0,0 +1,13 @@
+{
+    "homepage": "https://github.com/lptstr/kombucha",
+    "version": "0.1.0",
+    "url": "https://codeload.github.com/lptstr/kombucha/zip/v0.1.0#/dl.7z",
```
rasa: Should this be .zip? Or does 7zip not care it’s misnamed?

Kiedtl: From what I've seen of other manifests, 7zip doesn't seem to care whether its extracting ZIP files or 7Z files. I also know that Scoop uses 7zip to extract itself during updates.

rasa: Cool. Shoulda checked that myself.

rasa: Sorry, please add a license.

r15ch13:
Off topic question: Are these escaped strings working on Windows (ConEmu/Cmder/Hyper)? PowerShell 5/6? Currently my Windows machine is dead, so I can't test it :grin:

```powershell
# Escape character, needed for ANSI terminal sequences
[char]$global:E = [char]27

"$E[38;2;150;150;150mkombucha version $E[1m$E[38;2;110;140;253mv$E[4m1.0.0$E[0m$E[24m"
```
https://github.com/lptstr/kombucha/blob/74746d409c913e901cbfb5f62ac186d8300143d6/bin/kombucha.ps1#L10
https://github.com/lptstr/kombucha/blob/74746d409c913e901cbfb5f62ac186d8300143d6/bin/kombucha.ps1#L28

PS: Please add the `license` field, then it will be merged. :smile:

Kiedtl:
> Off topic question: Are these escaped strings working on Windows (ConEmu/Cmder/Hyper)? PowerShell 5/6?

As of February 2019, only the venerable Windows Console seems to fully support these ANSI escape strings with True Color support. See [this page] (https://github.com/lptstr/pixur#terminal-support). All other terminals resolve RGB colors to 16-bit colors. 😦 And yes, it does work with PowerShell 5+.

r15ch13:
@Kiedtl was just curious if we could drop `Write-Host` for colored output. So it would be possible to pipe scoops output. :smile:

PR #1703, opened 2019-02-06
Rename and fix sqlopsstudio (azure data studio)

Ash258:
Microsoft/sqlopsstudio redirects to azure data


r15ch13:
Exe, cmd, repo and zip are call `azuredatastudio`. Please rename to `azuredatastudio.json` :smile:

PR #1707, opened 2019-02-08
Add cherrytree 0.38.7

Ash258:


rasa:
So long https://github.com/rasa/scoops/blob/master/cherrytree.json !

PR #1722, opened 2019-02-11
Add APPLIST.md

Kiedtl:
I've added a huge markdown file which users can view to see which apps are in this bucket. The file also lists the homepage and the version of each app in a table.

The file can be automatically updated with the `bin/get-applist.ps1` file daily or weekly. Just run the script in the terminal and the script will do the rest.

I'd be happy to be the one updating the APPLIST, just in case the rest find that step unbearable.

The goal is to have a file which users can go to see what apps are in the bucket and what versions they are.


Kiedtl:
Oh, and can we add a link to the APPLIST.md file from the README?

PR #1738, opened 2019-02-15
Add marp package manifest v0.0.14

bgianfo:


bgianfo:
FYI @yhatt

yhatt:
As an app author, I don't recommend to add because *it is no longer maintained*.

See also the similar decision in awesome-list: https://github.com/sindresorhus/awesome-electron/pull/555

bgianfo:
@yhatt, sure I didn't realize it wasn't maintained.

PR #1741, opened 2019-02-16
Update arduino-portable.json

YenForYang:
Persist these custom files, used to override/add settings:
 - https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#platformlocaltxt
 - https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#boardslocaltxt
 (should also change arduino.json as well, probably)

### DIFF
--- a/bucket/arduino-portable.json
+++ b/bucket/arduino-portable.json
@@ -13,7 +13,12 @@
             "Arduino"
         ]
     ],
-    "persist": "portable",
+    "persist": [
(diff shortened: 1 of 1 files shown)


YenForYang:
Just realized that the files must be created beforehand in the persist directory, otherwise they will be installed as directory symlinks.

PR #1752, opened 2019-02-20
MEGAsync: Update to 4.0.2

niheaven:
When megasync's hash changes, one should check if its version number has changed... MEGAsync's sourcecode in github may change after its binary changes.


niheaven:
@r15ch13 The autoupdate server change version back to 4.0.1.0... So let it be, or may we delete `autoupdate` and bump the version when download hash check failed?

PR #1763, opened 2019-02-22
Add vcredist2013 to p4merge dependencies

### LINKED ISSUE #1762: Package p4merge broken on install
After installing, p4merge exits, without doing anything, immediately after being called. When called from the GUI, an error is received:

    ---------------------------
    p4merge.exe - System Error
    ---------------------------
    The code execution cannot proceed because MSVCR120.dll was not found. Reinstalling the program may fix this problem. 
    ---------------------------
    OK   
    ---------------------------

It looks like a dependency on the Visual C++ Runtime may not be met?

bgshacklett:
Fixes #1762

### DIFF
--- a/bucket/p4merge.json
+++ b/bucket/p4merge.json
@@ -16,5 +16,7 @@
         }
     },
     "bin": "Perforce\\Perforce\\p4merge.exe",
-    "depends": "extras/vcredist2013"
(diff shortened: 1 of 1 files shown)


Ash258:
Do not depend on vcredists. Make it only suggest.

PR #1769, opened 2019-02-24
Update irfanview.json

Befod:
persist "Toolbars"

### DIFF
--- a/bucket/foxit-reader.json
+++ b/bucket/foxit-reader.json
@@ -11,6 +11,7 @@
             "Foxit Reader"
         ]
     ],
+    "persist": "lang",
     "checkver": {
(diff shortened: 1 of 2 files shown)


r15ch13:
Files inside this directory will never get updated, because scoop doesn't support merging the new directory with the current one, yet.
The same goes for https://github.com/lukesampson/scoop-extras/issues/1766

rasa:
@r15ch13 That’s true for current users, but new users will benefit, yes?

r15ch13:
This will be the case on the next update for every user

rasa:
@r15ch13 so you’re advising rejecting this PR as ineffective?

r15ch13:
@rasa yes

PR #1787, opened 2019-02-28
Add painter v1.0.0

Kiedtl:
Add [painter](https://github.com/lptstr/painter), a tool to manage the Windows desktop wallpaper from the terminal.

### DIFF
--- a/painter.json
+++ b/painter.json
@@ -1,0 +1,11 @@
+{
+    "homepage": "https://github.com/lptstr/painter",
+    "version": "1.0.0",
+    "url": "https://github.com/lptstr/painter/releases/download/v1.0.0/painter.exe",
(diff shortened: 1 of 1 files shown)


Kiedtl:
OK, I added the license.

h404bi:
Also add description?

Kiedtl:
> Also add description?

Good point.

gregmac:
Note: This file was put into the root instead of `bucket` directory

Kiedtl:
> Note: This file was put into the root instead of bucket directory

You're right! @r15ch13 might need to fix this :frowning:

rasa:
Fixed in https://github.com/lukesampson/scoop-extras/commit/f92c7d93b9f73c38cd9ecfef81299f775b9f0f78

PR #1791, opened 2019-02-28
updated hash

gbursson:


gbursson:
added some descriptions as well

PR #1800, opened 2019-03-05
AppBuster, RegEditor: Add some O&O Freetools

niheaven:
- Add [AppBuster](https://www.oo-software.com/en/ooappbuster), [RegEditor](https://www.oo-software.com/en/ooregeditor) by O&O Software
- Fix ShutUp10's persist file creating.

### DIFF
--- a/bucket/regeditor.json
+++ b/bucket/regeditor.json
@@ -1,0 +1,39 @@
+{
+    "version": "12.0.2172",
+    "description": "Freeware for editing the Windows Registry.",
+    "homepage": "https://www.oo-software.com/en/ooregeditor",
+    "license": {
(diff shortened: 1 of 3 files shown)


bucket/regeditor.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": "12.0.2172",
+    "description": "Freeware for editing the Windows Registry.",
```
Ash258: No need to specify license in description.

```suggestion
    "description": "Windows Registry editor.",
```

niheaven: Thanks, the description is copy from the website without change, and it is fixed now.

PR #1805, opened 2019-03-05
Add Google Chrome 72.0.3626.121

Spiraster:
Uses third-party site to check version and get link, but autoupdates directly from Google.

https://github.com/Spiraster/scoop-bucket/commits/master/chrome.json


Ash258:
Why no 32 bit version? See my scoop-Ash258 buchet for implementation

Ash258:
Also no extension of file and wrong directory.

PR #1813, opened 2019-03-06
Add Workspace Utilities version 1.0.0.6

niboan:
Workspace Utilities provides hotkey access to custom window sizes and layouts.

The utility is on SourceForge with source, but no license is specified anywhere.
Also, I wasn't able to find the version number anywhere, so no autoupdate :-(


Ash258:
Last update was 2016-12-23, so there is small chance to new version.

PR #1835, opened 2019-03-09
Add Winaero Tweaker 0.14

### LINKED ISSUE #1581: [Request] Add Winaero Tweaker
In my opinion, [Winaero Tweaker](https://winaero.com/comment.php?comment.news.1836) is a must-have when installing windows on a new machine.
It allows you to customize the appearance and behavior of windows with the help of a GUI.
It has a portable mode.

gregmac:
Fixes #1581

### DIFF
--- a/bucket/winaero-tweaker.json
+++ b/bucket/winaero-tweaker.json
@@ -1,0 +1,28 @@
+{
+    "homepage": "https://winaero.com/comment.php?comment.news.1836",
+    "license": "Freeware",
+    "description": "Tweaks and hidden settings for power-users for Windows 7/8/8.1/10",
+    "version": "0.14",
+    "url": "https://winaero.com/winaerotweaker.zip",
+    "hash": "a73ddf39cc92213106573f974ac032b396b6ff793a48156484416a1f45aa4a98",
+    "installer": {
+        "file": "WinaeroTweaker-0.14.0.0-setup.exe",
+        "args": [
+            "/SP-",
+            "/VERYSILENT",
+            "/PORTABLE",
+            "/DIR=\"$original_dir\""
+        ]
+    },
+    "post_install": [
+        "if (Test-Path \"$dir\\SilentSetup.cmd\") { Remove-Item \"$dir\\SilentSetup.cmd\" }"
+    ],
+    "bin": "WinaeroTweaker.exe",
+    "shortcuts": [
+        [
+            "WinaeroTweaker.exe",
+            "Winaero Tweaker"
+        ]
+    ],
+    "checkver": "change log\\s*</h2>\\s*([\\d\\.]+)"
+}
--- a/bucket/winaero-tweaker.json
+++ b/bucket/winaero-tweaker.json
@@ -6,7 +6,7 @@
     "url": "https://winaero.com/winaerotweaker.zip",
     "hash": "a73ddf39cc92213106573f974ac032b396b6ff793a48156484416a1f45aa4a98",
     "pre_install": [
-        "Rename-Item (Get-ChildItem $original_dir *.exe).FullName setup.exe"
+        "Rename-Item (Get-ChildItem \"$original_dir\" *.exe).FullName setup.exe"
     ],
     "installer": {
         "file": "setup.exe",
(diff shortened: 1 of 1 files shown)


bucket/winaero-tweaker.json
```diff
@@ -0,0 +1,28 @@
...
+    "version": "0.14",
+    "url": "https://winaero.com/winaerotweaker.zip",
+    "hash": "a73ddf39cc92213106573f974ac032b396b6ff793a48156484416a1f45aa4a98",
+    "installer": {
+        "file": "WinaeroTweaker-0.14.0.0-setup.exe",
+        "args": [
+            "/SP-",
+            "/VERYSILENT",
+            "/PORTABLE",
+            "/DIR=\"$original_dir\""
```
rasa: Are we using `original_dir` as we’re running an installer extracted from the download?

bucket/winaero-tweaker.json
```diff
@@ -0,0 +1,28 @@
...
+    ],
+    "bin": "WinaeroTweaker.exe",
+    "shortcuts": [
+        [
+            "WinaeroTweaker.exe",
+            "Winaero Tweaker"
+        ]
+    ],
+    "checkver": "change log\\s*</h2>\\s*([\\d\\.]+)"
+}
```
rasa: Can we add an `autoupdate` section?

rasa:
Another option is done here: 
https://github.com/lukesampson/scoop-extras/blob/master/bucket/edex-ui.json#L10

Ash258:
Definitely better without installing would ve better if its possible. Have not downloaded it to check

bucket/winaero-tweaker.json
gregmac: Felt more semantically correct to be explicit and install to the "0.14" directory as opposed to "current" directory.

bucket/winaero-tweaker.json
```diff
@@ -5,8 +5,11 @@
     "version": "0.14",
     "url": "https://winaero.com/winaerotweaker.zip",
     "hash": "a73ddf39cc92213106573f974ac032b396b6ff793a48156484416a1f45aa4a98",
+    "pre_install": [
+        "Rename-Item (Get-ChildItem $original_dir *.exe).FullName setup.exe"
```
Ash258: - Why original dir? 
- No quotes

gregmac: During an upgrade, *.exe would conflict with other .exe's in the "current" directory, whereas $original_dir is only going to contain the extracted files.

I can quote the constants to be strict about string usages. Actually I guess a lot of scoop scripts even add quotes when not necessary (eg, `\"$dir\" here https://github.com/lukesampson/scoop-extras/blob/master/bucket/edex-ui.json#L11`) -- is that just the expected style? `$dir` is just a powershell variable (as opposed to something being string-replaced prior to execution), isn't it?

Ash258: For more info: https://stackoverflow.com/questions/5571579/escaping-splatted-variables-in-powershell-invoke-expression

TLDR: when using `Invoke-Expression` powershell is "replacing" variables with pure value, so commands like `Get-Child item $dir` will become `Get-ChildItem C:\Whatever with spaces\cosi`, which ends with error.

Ash258: I am still not sure about that $original_dir. Will test when arrive home, but i think normal $dir should be enough for everything. /cc @r15ch13

gregmac: I guess if $dir (current) is pointed at the currently-being-installed directory by then it would be fine? That requires knowing more about the internals of scoop than I currently know (which, IMHO, should be unnecessary for building a mainfest file). It also isn't clear to me what benefit `$dir` would have over `$original_dir`?

I personally think `$original_dir` (the `0.14` dir)  makes more sense semantically, if not technically, than `$dir` (`current`) but don't have a strong enough opinion either way.

r15ch13: `$dir` is used consistently in manifests.
`$dir` contains the versioned path at the time the installer gets executed and gets changed later to `current`

https://github.com/lukesampson/scoop/blob/3830d9f5f2cb544c9758c5ef3b591c9b0255df6a/lib/install.ps1#L39-L54

bucket/winaero-tweaker.json
gregmac: Changed to $dir per discussion below

PR #1839, opened 2019-03-09
goland: Add version 2018.3.5

rzpio:
Hey, I have added GoLand in version _2018.3.5_. The application wasn't in this bucket before.


Ash258:
https://github.com/lukesampson/scoop-extras/pull/1520#issuecomment-447385484

For extras it's reccomended to use toolbox app. You could use https://www.github.com/Ash258/Scoop-JetBrains with all IDEs in all versions available.

rzpio:
Ah, all right. Thanks. 😃

PR #1842, opened 2019-03-10
standardnotes: Simplify manifest

niheaven:
Maybe a standard NSIS extractor is prefered.

### DIFF
--- a/bucket/standardnotes.json
+++ b/bucket/standardnotes.json
@@ -2,27 +2,27 @@
     "homepage": "https://standardnotes.org/",
     "version": "3.0.6",
     "description": "A safe place for your notes, thoughts, and life's work.",
-    "license": {
-        "identifier": "AGPL-3.0",
-        "url": "https://github.com/standardnotes/desktop/blob/master/LICENSE"
-    },
+    "license": "AGPL-3.0",
     "url": "https://github.com/standardnotes/desktop/releases/download/v3.0.6/standard-notes-setup-3.0.6.exe#/dl.7z",
     "hash": "9143f26be822009d9fec72dc1faae227667b68f8831b13951e560cb7bf45053b",
-    "extract_dir": "\\$PLUGINSDIR",
-    "pre_install": "Get-ChildItem \"$dir\" -Exclude 'app-64.7z', 'app-32.7z' | Remove-Item -Force -Recurse",
     "architecture": {
         "64bit": {
             "installer": {
-                "script": "extract_7zip \"$dir\\app-64.7z\" \"$dir\""
+                "script": [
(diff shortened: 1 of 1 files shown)


Ash258:
There is much junk in pluginsdir (dlls needed for installer to work properly), which is not needed to have on user end after installation. That's why it's done original way. You just get only that you need (7zips with program data) without code duplication inside manifest.

niheaven:
The two methods are equivalent except for a left "Uninstall Standard Notes.exe" that in this modified one. And one step for extraction, another for removing unused `$PLUGINSDIR` is more simpler than the original one.

What I mean above is there are many packages that use a same install script for NSIS, maybe a uniform script that shipped with scoop's `install.ps1` or `decompress.ps1` is prefered (like `innosetup`).

rzpio:
> The two methods are equivalent except for a left "Uninstall Standard Notes.exe" that in this modified one. 

Why change the manifest when they are equivalent? 😄 

> And one step for extraction, another for removing unused $PLUGINSDIR is more simpler than the original one.

I think that's a matter of opinion. I had oriented myself to the manifest of [rambox-pro](https://github.com/lukesampson/scoop-extras/blob/master/bucket/rambox-pro.json). Basically, you're just moving the lines. I don't see any simplification in it. 😅 

The removal of the link to the license I also do not quite understand. `$baseurl` is actually a simplification. Thanks, I didn't think about the variable.

niheaven:
That is, just two lines in `installer.script`, no `pre-install`, no `post-install`, for better readability. 😄 

> The removal of the link to the license I also do not quite understand. 

Please see manifest [help](https://github.com/lukesampson/scoop/wiki/App-Manifests#optional-properties) for license identifier usage: 

> For well-known licenses, please use the identifier found at [spdx.org/licenses](https://spdx.org/licenses/). For other licenses, use the URL of the license, if available. Otherwise, use “Freeware”, “Proprietary”, “Public Domain”, or “Shareware”, as appropriate.

So for `AGPL-3.0`, you should use only identifier and that will give a url (https://spdx.org/licenses/AGPL-3.0.html) when call `scoop info standardnotes`.

For `Uninstall xxx.exe`, it is certainly needless, and that is a shortcoming of this two line code.

Finally, while there is no UNIQUE NSIS unpacking function in scoop, the two methods of extraction are equivalent, and are just about opinion. So, what I mentioned here is a unique unpacking function is neccecary, for simplification.

rzpio:
> So for AGPL-3.0, you should use only identifier and that will give a url (spdx.org/licenses/AGPL-3.0.html) when call scoop info standardnotes.

Ah, okay, thank you. 😄 

> My key point here is a unique unpacking function is neccecary, for simplification.

I agree. The process is always the same and the manifests are redundant.

niheaven:
Waiting for NSIS common installation.

PR #1844, opened 2019-03-10
LOVE: Fix shortcut and checkver

niheaven:
- Fix shortcut to "LOVE" to avoid shortcut creating error
- Change checkver to download page to ensure windows version is released


rasa:
So the `Ö` character causes errors? That’s nuts.

niheaven:
Yes，the Ö stop me from creating shortcut... Maybe related to codepage

PR #1859, opened 2019-03-14
Add Sysinternals Process Explorer v16.22

573:
Sophos antivirus software marks pskill and sorts as suspicious software preventing SysinternalsSuite installation.
Thus this separate package.

### DIFF
--- a/bucket/procexp.json
+++ b/bucket/procexp.json
@@ -1,0 +1,44 @@
+{
+	"description": "Process Explorer shows you information about which handles and DLLs processes have opened or loaded.",
+	"homepage": "https://technet.microsoft.com/sysinternals/bb896655",
+	"version": "16.22",
+	"hash": "d393f062091c4fcf720ce0cad56520a920ffcc9412cdc7941150e3bb3fc4fefa",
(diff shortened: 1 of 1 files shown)


rasa:
I would suggest you use https://github.com/lukesampson/scoop-extras/blob/master/bucket/sysinternals.json or install it from https://github.com/h404bi/dorado

I would prefer we not have to maintain both the sysinternals package, and then all its individual apps, as well. The same goes for nirsoft, with its 250+ apps.

rasa:
Another solution would be for someone to create a bucket of all the sysinternal apps, a la https://github.com/MCOfficer/scoop-nirsoft

573:
> 
> 
> I would suggest you use https://github.com/lukesampson/scoop-extras/blob/master/bucket/sysinternals.json or install it from https://github.com/h404bi/dorado
> 
> I would prefer we not have to maintain both the sysinternals package, and then all its individual apps, as well. The same goes for nirsoft, with its 250+ apps.

Well I didn't know of dorado yet, but that would be sufficient for me, so I'm closing this.

PR #1863, opened 2019-03-14
Add coursier (an artifact downloader like Ivy)

nightscape:
I haven't found an example where multiple `autoupdate` `url`s are used, so I'm assuming I have to hard-code the link to the second file?

### DIFF
--- a/bucket/coursier.json
+++ b/bucket/coursier.json
@@ -1,0 +1,22 @@
+{
+    "homepage": "https://github.com/coursier/coursier",
+    "description": "Pure Scala Artifact Fetching",
+    "license": "Apache-2.0",
+    "version": "1.1.0-M13-1",
+    "url": [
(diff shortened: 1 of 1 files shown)


bucket/coursier.json
```diff
@@ -0,0 +1,22 @@
+{
+    "homepage": "https://github.com/coursier/coursier",
```
Ash258:
```suggestion
    "homepage": "https://get-coursier.io/",
```

Ash258:
Also this could be moved to main bucket, since it's only CLI

nightscape:
Ah, that's the distinction? I never knew 😄 
Ok, I'll make a new PR to the main bucket then.

nightscape:
Superceded by https://github.com/lukesampson/scoop/pull/3220

PR #1877, opened 2019-03-19
archwsl: Optimize installation

yuk7:
Change for prevent dangerous registry hacks in ArchWSL script.
lxuid is buggy and should not be used.


r15ch13:
@kuinaein @rase any thought on this?

bucket/archwsl.json
```diff
@@ -43,30 +43,19 @@
...
-        "    `$lx_uid = & '$dir\\Arch.exe' get --lxuid",
-        "    if(!`$lx_uid) { error 'ArchWsl is not installed!'; return }",
-        "    Set-ItemProperty -Path ('HKCU:Software\\Microsoft\\Windows\\CurrentVersion\\Lxss\\' + `$lx_uid) -Name BasePath -Value '$persist_dir\\data'",
-        "    if (Test-Path ('$persist_dir' + '\\data\\rootfs')) {",
-        "        Remove-Item -Recurse -Path '$dir\\rootfs'",
-        "    } else {",
-        "        Move-Item -Path '$dir\\rootfs' -Destination '$persist_dir\\data\\rootfs'",
-        "    }",
-        "    Remove-Item -Recurse -Path '$dir\\temp'",
+        "    \"Copy-Item '$dir\\Arch.exe' '$persist_dir\\data\\Arch.exe'",
```
r15ch13: This will not work because `$dir` and `$persist_dir` can't be resolved when `'` is used instead of `"`

bucket/archwsl.json
```diff
@@ -43,30 +43,19 @@
...
+        "    `& '$persist_dir\\data\\Arch.exe' install `\"$dir\\rootfs.tar.gz`\";`$res=`$?",
```
r15ch13: What does `$res=$?` do?

bucket/archwsl.json
```diff
@@ -43,30 +43,19 @@
...
     ],
     "uninstaller": {
         "script": [
-            "$lx_uid = & \"$dir\\Arch.exe\" get --lxuid",
-            "if(!$lx_uid) { error 'ArchWsl is not registered!'; return }",
-            "Set-ItemProperty -Path \"HKCU:Software\\Microsoft\\Windows\\CurrentVersion\\Lxss\\$lx_uid\" -Name BasePath -Value $dir",
-            "New-Item -ItemType Directory -Path \"$dir\\rootfs\" | Out-Null",
-            "'y' | & \"$dir\\Arch.exe\" clean"
+            "& \"$dir\\Arch.exe\" isregd",
+            "if(!$?) { error 'ArchWsl is not registered!'; return }",
```
r15ch13: `!$?` what's the meaning of this?

yuk7: `Arch.exe isregd` is `True` if registered in the system, otherwise it returns `False`.
`$?` stores the return value (True / False) of the last command.
`!$?` Is a logical inversion of it.

bucket/archwsl.json
yuk7: Store the return value in `$res`.
Stores Installation successful:`True` /Failed:`False`

r15ch13: Ah like `$lastexitcode` but it returns only a boolean, nice. Didn't know about that :smile:

bucket/archwsl.json
yuk7: Ok,changed in 91cfd36a
Thanks.

r15ch13:
I commented some changes on the original PR.
https://github.com/lukesampson/scoop-extras/pull/1179#issuecomment-416246332
Maybe the installation can be simplified.

yuk7:
Installing with administrator privileges can break the permissions.
I think it is safer to install in user mode.

yuk7:
If installed in a directory to which you do not have full access permissions, you may not be able to start with user permissions.

kuinaein:
@r15ch13 @yuk7
Works great for me. Thanks a lot!
The hack is only for data persistence ( https://github.com/lukesampson/scoop-extras/pull/1179#issuecomment-416269681 ) and Arch Linux filesystem is still persisted.

bucket/archwsl.json
```diff
@@ -43,30 +43,19 @@
...
-        "    } else {",
-        "        Move-Item -Path '$dir\\rootfs' -Destination '$persist_dir\\data\\rootfs'",
-        "    }",
-        "    Remove-Item -Recurse -Path '$dir\\temp'",
+        "    \"Copy-Item `\"$dir\\Arch.exe`\" `\"$persist_dir\\data\\Arch.exe`\"",
+        "    `& '$persist_dir\\data\\Arch.exe' install `\"$dir\\rootfs.tar.gz`\";`$res=`$?",
+        "    Remove-Item ('$persist_dir'+'\\data\\Arch.exe') -Force",
+        "    if(!`$res) { error 'ArchWsl installation failed!'; return }",
         "    \" | Set-Content \"$dir\\install.ps1\"",
         "    & \"$dir\\install.ps1\"",
```
Ash258: Why are you creating install file and then executing it instead of just running chained commands?

yuk7: This is because rootfs is placed in the executable file directory according to the API specification.

Ash258: What?

What is the difference when you run this in sequence and you put it in file and then execute that script?

```powershell
Copy-Item "$dir\Arch.exe" "$persist_dir\data\Arch.exe"
& "$persist_dir\data\Arch.exe" install "$dir\rootfs.tar.gz"
$res=$?
Remove-Item "$persist_dir\data\Arch.exe" -Force
if(!$res) { error 'ArchWsl installation failed!' }
```

kuinaein: It may no longer be needed. 
The file was running with admin privileges but now it's running in user mode. (https://github.com/lukesampson/scoop-extras/pull/1877#issuecomment-476328832)

yuk7: I understood.
I'm sorry for my poor English

tcztzy:
Any updates about this package? This package is not compatible with WSL2 right now.

bucket/archwsl.json
yuk7: I pushed the commit that changed the point you pointed out now.
I'm sorry I'm too late

yuk7:
@tcztzy 
I remember that the WSL2 bug was fixed in the Windows update.
Still not working?
Please see https://github.com/yuk7/ArchWSL/issues/89 .

tcztzy:
@yuk7
ArchWSL is working fine, but this scoop package is not working if you set default version 2.
```bash
if (Test-Path ('$persist_dir' + '\\data\\rootfs')) {
        Remove-Item -Recurse -Path '$dir\\rootfs'
    } else {
        Move-Item -Path '$dir\\rootfs' -Destination '$persist_dir\\data\\rootfs'
    }
```
Since the rootfs directory doesn't exist in WSL2, the installation will occur error.

yuk7:
@tcztzy understood.  Once this PR is merged, it should be resolved.

yuk7:
@r15ch13 @kuinaein @Ash258 
Sorry for too late
I have pushed the requested changes.
Could you review it?
This change is also required for WSL2 support.

yuk7:
ok.

Ash258:
Did you test it?

yuk7:
Yes. tested

yuk7:
Should I resolve conflicts by following the latest master?

Ash258:
Someone please test this. I am not gonna mess my environments with this.

kidonng:
I've tested this version many times in the past few days.

tcztzy:
I used to install ArchWSL by applying this patch to the `archwsl.json` file.

TheWiresharkGuy:
This commit breaks ArchWSL update to [20.3.26.0](https://github.com/lukesampson/scoop-extras/commit/ba023c7d94824836ee083554858b4722d3dac678) for me ...

PR #1886, opened 2019-03-20
Add Basilisk browser 2019.03.08

Retia-Adolf:


Ash258:
Autoupdate is missing architecture property with URLs

Retia-Adolf:
Well, I thought it will not be updated so it could be omitted..

PR #1901, opened 2019-03-22
InnoSetup: Add innosetup and innoscriptstudio

niheaven:
Add Inno Setup (with encryption module) and Inno Script Studio


Ash258:
Rename setup studio with dashes as divider (As in homepage)

dont you need admin right to create new registry keys?

niheaven:
Will Inno Setup use dashes?

In my testing, creating registry entry in HKCU don't need admin right. (While in HKLM need)

PR #1924, opened 2019-03-26
fsviewer: Fix persistence

hoilc:
Fix #1918 (Placeholder files were created in `persist` directory by mistake.)


Ash258:
Duplicate of #1918

Ash258:
That is not problem. Persist will still work. There is no difference if you create new file in `$dir` or `$persist_dir`

hoilc:
It seems `$persist_dir` has not been created when `pre_install` script is running, so placeholders won't be created also.

PR #1925, opened 2019-03-26
CarotDAV: Add version 1.15.5

hoilc:


bucket/carotdav.json
```diff
@@ -0,0 +1,30 @@
+{
+    "homepage": "http://rei.to/carotdav_en.html",
+    "description": "A simple WebDAV / FTP / SFTP / Online Storages client for Windows OS",
+    "license": {
+        "url": "http://rei.to/carotdav_en.html#license"
```
Ash258:
```suggestion
        "identifier": "Freeware",
        "url": "http://rei.to/carotdav_en.html#license"
```

bucket/carotdav.json
```diff
@@ -0,0 +1,30 @@
...
+        "CarotDAVC.exe"
+    ],
+    "persist": "Setting.xml",
+    "pre_install": "if(!(Test-Path \"$persist_dir\\Setting.xml\")) {Set-Content -Path (\"$dir\\Setting.xml\") -Value ('<?xml version=\"1.0\" encoding=\"utf-8\"?><WebDAVClientSetting xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\"></WebDAVClientSetting>')}",
+    "checkver": {
+        "url": "http://rei.to/carotdav.version.txt",
+        "regex": "([\\d\\.]+)"
+    },
+    "autoupdate": {
+        "url": "http://rei.to/CarotDAV$version.portable.zip"
```
Ash258: Add hash extraction
```suggestion
        "url": "http://rei.to/CarotDAV$version.portable.zip",
        "hash": {
            "url": "http://rei.to/carotdav_en.html#download",
            "regex": "SHA-1: $sha1"
        }
```

bucket/carotdav.json
```diff
@@ -0,0 +1,30 @@
...
+    "version": "1.15.5",
+    "url": "http://rei.to/CarotDAV1.15.5.portable.zip",
+    "hash": "21b0b68ec45223c747d67b0e21ed87e8c195d52e5fc651b30b631a2f328bcd01",
+    "extract_dir": "CarotDAV",
+    "bin": [
+        "CarotDAV.exe",
+        "CarotDAVC.exe"
+    ],
+    "persist": "Setting.xml",
+    "pre_install": "if(!(Test-Path \"$persist_dir\\Setting.xml\")) {Set-Content -Path (\"$dir\\Setting.xml\") -Value ('<?xml version=\"1.0\" encoding=\"utf-8\"?><WebDAVClientSetting xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\"></WebDAVClientSetting>')}",
```
Ash258: Why is value array with only 1 string?

bucket/carotdav.json
hoilc: The first matched hash is sha1 of installer version. 
I can't found the way to capture second matched hash which belongs to portable version.
It could be nice if `hash` has the `reverse: true` option.

Ash258: `$basename">Ver. $version SHA-1: $sha1` Should do the trick

Ash258: Edited suggestion to reflect the change

Ash258: image

hoilc: Nice job!

bucket/carotdav.json
hoilc: It's the base config for CarotDAV and line break is unnecessary in XML.

Ash258: But then you do not need array. It could be just string
```suggestion
    "pre_install": "if(!(Test-Path \"$persist_dir\\Setting.xml\")) {Set-Content -Path \"$dir\\Setting.xml\" -Value '<?xml version=\"1.0\" encoding=\"utf-8\"?><WebDAVClientSetting xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\"></WebDAVClientSetting>'}",
```

hoilc: 😂

niheaven:
Please make sure `Set-Content -encoding UTF8` which created UTF8 w/ BOM in Powershell 5 works. `ASCII` is preferred.

hoilc:
> Please make sure `Set-Content -encoding UTF8` which created UTF8 w/ BOM in Powershell 5 works. `ASCII` is preferred.

I tested both UTF-8 and ASCII works well in my Powershell 5.

bucket/carotdav.json
```diff
@@ -0,0 +1,35 @@
+{
+    "homepage": "http://rei.to/carotdav_en.html",
+    "description": "A simple WebDAV / FTP / SFTP / Online Storages client for Windows OS",
+    "license": {
+        "identifier": "Freeware",
+        "url": "http://rei.to/carotdav_en.html#license"
+    },
+    "version": "1.15.5",
+    "url": "http://rei.to/CarotDAV1.15.5.portable.zip",
+    "hash": "21b0b68ec45223c747d67b0e21ed87e8c195d52e5fc651b30b631a2f328bcd01",
```
niheaven:
```suggestion
    "hash": "sha1:45a3859de58bde9c12fd4792fa54d48fb0a30184",
```

`hash` gotten by `checkver`

PR #1929, opened 2019-03-28
Add license file

0x326:
Here's the `LICENSE` file from the main Scoop bucket. I figured it'd be good to keep a copy of it in this repository as well


Jimimaku:
> Here's the `LICENSE` file from the main Scoop bucket. I figured it'd be good to keep a copy of it in this repository as well

PR #1960, opened 2019-04-02
Access exe of inno-setup from command line

rgozim:
Add `env_add_path` option to allow for access to `iscc.exe` from the command line.


Ash258:
Why appending path for single binary?? Just add `"bin": "iscc.exe"`

rgozim:
Sorry, didn't realise that was an option. I'll make the change you suggested.

PR #1961, opened 2019-04-02
MPC-BE: Persist

dodorz:
Add persist of ini-file.

### DIFF
--- a/bucket/mpc-be.json
+++ b/bucket/mpc-be.json
@@ -32,6 +32,7 @@
             ]
         }
     },
+    "persist": "mpc-be64.ini",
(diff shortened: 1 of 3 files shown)


Ash258:
I doubt, there is only 64bit files persist. What about 32bit systems? Do they have config files?

Config files does not exists in downloaded file, so they need to be created before persisting. (At least for MPC-BE)

So implement next:

1. Check for already persisted config files
1. Create them if not installed
1. Persist 32bit versions too

https://github.com/Ash258/scoop-Ash258/blob/master/bucket/PotPlayer.json#L64

PR #1966, opened 2019-04-04
GIMP: Using installer.script instead of original installer

### LINKED ISSUE #1369: scoop-extras/gimp.json requires admin to install
It would be better to use a portable version.

niheaven:
Close #1369 

Please consider merging https://github.com/lukesampson/scoop/pull/3204, current `unpack_inno` crashed occasionally (due to `Move-Item` for huge amount of files).

### DIFF
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -2,9 +2,9 @@
     "homepage": "https://www.gimp.org/",
     "license": "GPL-3.0-only",
     "description": "GNU Image Manipulation Program",
-    "version": "2.10.12-2",
-    "url": "https://download.gimp.org/mirror/pub/gimp/v2.10/windows/gimp-2.10.12-setup-2.exe",
-    "hash": "6cd67baf381c7950ec0a46f6ea9b9a6c9191e29fff09a575ce7c7af6d223b613",
+    "version": "2.10.12-3",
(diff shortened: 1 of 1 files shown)


Ash258:
- Unquoted `$dir` in `Push-Location`
- Why both architecture scripts? 

There could be one universal which could look like this:

```powershell
Push-Location $dir
Get-ChildItem -Filter '*.debug' -Recurse | Remove-Item -Recurse
if ($architecture -eq '32') {
   Rename ,1 to twain
   Remove ,2
....
}
```

niheaven:
- `$dir` is quoted.
- The installer script has similar sections, but files manipulation and config setting are differ a lots:
  - `64bit`
    - files have names as `*,2*`, `*,3*`
    - has a `32` dir, but need to copy two extra dirs into it.
    - python related files need to copy to `32`
    - has a 32bit `twain.exe` (for 32bit plugins supports)
    - `PATH` config set to `bin` and `32\bin`
  - `32bit`
    - files have names as `*,1*`
    - move `32` over base dir
    - `PATH` config set to `bin`

So two scripts is prefered, otherwise it is too complex.

niheaven:
Or using `pre_install`, `arch.installer.script`, `post_install`.

`pre_install` for deleting `.debug`
`post_install` for set config
`arch.install.er.script` for others

PR #1977, opened 2019-04-07
add beaker

linsui:
beaker browser is 64-bit only. Should `"architecture"` be used in `"autoupdate"` ?

### DIFF
--- a/bucket/beaker.json
+++ b/bucket/beaker.json
@@ -1,0 +1,34 @@
+{
+	"version": "0.8.8",
+	"description": "An experimental peer-to-peer Web browser",
+	"homepage": "https://beakerbrowser.com/",
+	"license": "MIT",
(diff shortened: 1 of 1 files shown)


bucket/beaker.json
```diff
@@ -0,0 +1,34 @@
...
+		]
+	},
+	"checkver": {
+		"github": "https://github.com/beakerbrowser/beaker"
+	},
+	"autoupdate": {
+		"url": "https://github.com/beakerbrowser/beaker/releases/download/$version/beaker-browser-setup-$version.exe#/dl.7z",
+		"hash": {
+			"url": "$baseurl/latest.yml",
+			"regex": "sha512: $sha512"
```
Ash258: It is not sha512. It is base64 encoded sha512.

```suggestion
			"regex": "sha512:\\s+(.*)"
```

r15ch13:
Please use spaces instead of tabs

PR #1993, opened 2019-04-09
regeditor: Add ItemType to preinstall entry

Dabombber:
Otherwise the user is prompted to specify the type.

### DIFF
--- a/bucket/regeditor.json
+++ b/bucket/regeditor.json
@@ -27,7 +27,7 @@
         }
     },
     "pre_install": [
-        "if (!(Test-Path $persist_dir\\OORegEdt.ini)) {",
+        "if (!(Test-Path \"$persist_dir\\OORegEdt.ini\")) {",
         "   New-Item \"$dir\\OORegEdt.ini\" -ItemType File  | Out-Null",
         "}"
     ],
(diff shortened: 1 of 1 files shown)


bucket/regeditor.json
```diff
@@ -28,7 +28,7 @@
     },
     "pre_install": [
         "if (!(Test-Path $persist_dir\\OORegEdt.ini)) {",
-        "   New-Item $dir\\OORegEdt.ini | Out-Null",
+        "   New-Item -ItemType file $dir\\OORegEdt.ini | Out-Null",
```
Ash258: Positional parameters should go first.
```suggestion
        "   New-Item \"$dir\\OORegEdt.ini\" -ItemType File  | Out-Null",
```

bucket/regeditor.json
```diff
@@ -28,7 +28,7 @@
     },
     "pre_install": [
         "if (!(Test-Path $persist_dir\\OORegEdt.ini)) {",
```
Ash258:
```suggestion
        "if (!(Test-Path \"$persist_dir\\OORegEdt.ini\")) {",
```

niheaven:
What's your Powershell version? In PS 5 and 6, `New-Item` will create file by default.

Dabombber:
> 
> 
> What's your Powershell version? In PS 5 and 6, `New-Item` will create file by default.

It was 3, which was the minimum recommended. I guess it might not be enough any more lukesampson/scoop#2402.

PR #2003, opened 2019-04-11
winfetch: update deps (again)

kiedtl:
Updated dependencies for Winfetch (with v1.2, deps curl and pwsh no longer needed).


rikuben:
Can you change imagemagick to suggests too? Because it's just an optional dependencies right?

kiedtl:
> Can you change imagemagick to suggests too? Because it's just an optional dependencies right?

Sounds like a good idea. I'll try that tomorrow.

PR #2019, opened 2019-04-17
Add krita

jeremy886:
a famous KDE drawing tool. first time contribution.


Ash258:
https://github.com/lukesampson/scoop-extras/blob/master/bucket/krita.json

jeremy886:
Oops, I think I mispelled it so I didn't find it. Thanks!!

PR #2021, opened 2019-04-17
Add rainmeter-portable: feedback needed

Skrity:
There was a problem with unpacking different architectures in https://github.com/lukesampson/scoop-extras/pull/694. Looking for input on this solution: it works, persistence isn't trivial here, but to my knowledge I managed to persist everything.

`"if((Get-Content \"$dir\\Rainmeter.ini\") -eq $Null) { Copy-Item \"$dir\\Defaults\\Layouts\\illustro default\\Rainmeter.ini\", \"$dir\\Defaults\\Skins\", \"$dir\\Defaults\\Layouts\" \"$persist_dir\" -Force -Recurse | Out-Null }",` makes default welcome skins appear on new installation.

`"if(Test-Path \"$dir\\Plugins.original\") { Copy-Item \"$dir\\Plugins.original\\*\" \"$persist_dir\\Plugins\" -Force -Recurse; Remove-Item \"$dir\\Plugins.original\" -Force -Recurse | Out-Null }",` updates default plugins in case of an update.

### DIFF
--- a/bucket/rainmeter-portable.json
+++ b/bucket/rainmeter-portable.json
@@ -32,9 +32,7 @@
         "Plugins"
     ],
     "post_install": [
-        "if(!(Test-Path \"$persist_dir\\Layouts\\*\")) { Copy-Item \"$dir\\Defaults\\Layouts\\*\" \"$dir\\Layouts\" -Force -Recurse | Out-Null }",
-        "if(!(Test-Path \"$persist_dir\\Skins\\*\")) { Copy-Item \"$dir\\Defaults\\Skins\\*\" \"$dir\\Skins\" -Force -Recurse | Out-Null }",
-        "if((Get-Content \"$dir\\Rainmeter.ini\") -eq $Null) { Copy-Item \"$dir\\Defaults\\Layouts\\illustro default\\Rainmeter.ini\" \"$persist_dir\" -Force -Recurse | Out-Null }",
+        "if((Get-Content \"$dir\\Rainmeter.ini\") -eq $Null) { Copy-Item \"$dir\\Defaults\\Layouts\\illustro default\\Rainmeter.ini\", \"$dir\\Defaults\\Skins\", \"$dir\\Defaults\\Layouts\" \"$persist_dir\" -Force -Recurse | Out-Null }",
         "if(Test-Path \"$dir\\Plugins.original\") { Copy-Item \"$dir\\Plugins.original\\*\" \"$persist_dir\\Plugins\" -Force -Recurse; Remove-Item \"$dir\\Plugins.original\" -Force -Recurse | Out-Null }",
         "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst.exe.nsis\", \"$dir\\inst.exe\" -Force -Recurse"
     ],
--- a/bucket/rainmeter-portable.json
+++ b/bucket/rainmeter-portable.json
@@ -1,0 +1,57 @@
+{
+    "homepage": "https://rainmeter.net/",
+    "description": "Rainmeter is a desktop customization tool for Windows.",
+    "version": "4.3.0.3298",
+    "license": "GPL-2.0-only",
+    "url": "https://github.com/rainmeter/rainmeter/releases/download/v4.3.0.3298/Rainmeter-4.3.exe#/inst.exe",
+    "hash": "1bb57129f6871c7e5a835ded406de7096aebbe53196006ca0d00073948fbb16b",
+    "depends": "7zip",
+    "architecture": {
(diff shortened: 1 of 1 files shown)


bucket/rainmeter-portable.json
```diff
@@ -0,0 +1,55 @@
...
+                "script": "7z x \"$dir\\inst.exe\" -o\"$dir\" -aoa"
+            }
+        },
+        "32bit": {
+            "installer": {
+                "script": "7z x \"$dir\\inst.exe\" -o\"$dir\" -aos"
+            }
+        }
+    },
+    "pre_install": [
```
Ash258:
```suggestion
    "pre_install": [
        "@('Rainmeter.ini', 'Rainmeter.data', 'Rainmeter.stats') | ForEach-Object {",
        "    if(!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" -Type File | Out-Null }",
        "}"
    ],
```

r15ch13:
Using the overwrite mode is a neat trick. :smile:

bucket/rainmeter-portable.json
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    "persist": [
+        "Rainmeter.ini",
+        "Rainmeter.data",
+        "Rainmeter.stats",
+        "Layouts",
+        "Skins",
+        "Plugins"
+    ],
+    "post_install": [
```
Ash258: Tried to make it as readable as possible and added comments for other maintainers.
Array usage is better for these operations since you do not need to repeat $dir\....
```suggestion
    "post_install": [
        "    # Makes default welcome skins appear on new installation.",
        "if(!(Get-Content \"$dir\\Rainmeter.ini\")) {",
        "    @('Layouts\\illustro default\\Rainmeter.ini', 'Skins', 'Layouts') | ForEach-Object {",
        "        Copy-Item \"$dir\\Defaults\\$_\" \"$persist_dir\" -Force -Recurse | Out-Null",
        "    }",
        "}",
        "    # If there was update for plugin, copy default plugins after persisting.",
        "if (Test-Path \"$dir\\Plugins.original\") {",
        "    Copy-Item \"$dir\\Plugins.original\\*\" \"$persist_dir\\Plugins\" -Force -Recurse",
        "    Remove-Item \"$dir\\Plugins.original\" -Force -Recurse | Out-Null",
        "}",
        "@('$PLUGINSDIR', 'uninst.exe.nsis', 'inst.exe') | ForEach-Object { Remove-Item \"$dir\\$_\" -Force -Recurse }"
    ],
```

Ash258:
And for that installation steps. You assume that first file is for 32bit and second file is for 64bit right? (If they have same name)

F

bucket/rainmeter-portable.json
```diff
@@ -0,0 +1,55 @@
...
+        "if((Get-Content \"$dir\\Rainmeter.ini\") -eq $Null) { Copy-Item \"$dir\\Defaults\\Layouts\\illustro default\\Rainmeter.ini\", \"$dir\\Defaults\\Skins\", \"$dir\\Defaults\\Layouts\" \"$persist_dir\" -Force -Recurse | Out-Null }",
+        "if(Test-Path \"$dir\\Plugins.original\") { Copy-Item \"$dir\\Plugins.original\\*\" \"$persist_dir\\Plugins\" -Force -Recurse; Remove-Item \"$dir\\Plugins.original\" -Force -Recurse | Out-Null }",
+        "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst.exe.nsis\", \"$dir\\inst.exe\" -Force -Recurse"
+    ],
+    "bin": [
+        "Rainmeter.exe",
+        "SkinInstaller.exe"
```
Ash258: Is it really neede to shim SkinInstaller?? Does this have some command line handler?

Skrity: You can pass skin file to it, that's as far as I know, I just iterated on existing rainmeter.json.

Skrity:
Yeah, the second portion of binaries is the "64-bit mode" option in installer. The existing rainmeter.json does this as well, just the default mode is "always overwrite".
Also maybe we can somehow pass `-ao` parameter via `extract_7zip`, that will eliminate the need for explicit 7zip dependency? But It seems that will require a patch to scoop.

Skrity:
How would I go about naming the shortcut: "Rainmeter (portable)" will allow for two shortcuts form `rainmeter` and `rainmeter-portable` to coexist, or is it a non-supported scenario?

niheaven:
No need to `#/inst.exe`, you could use `7z x "$dir\$fname"

PR #2029, opened 2019-04-19
Convert sourceforge urls to use downloads mirror

Ash258:
- Closes lukesampson/scoop#2199


r15ch13:
The URL should be `https://downloads.sourceforge.net/project/` and not `https://downloads.sourceforge.net/projects/`. 😢
Will make a fix for that.

PR #2033, opened 2019-04-20
Add Intelligent Standby List Cleaner (ISLC) version 1.0.1.1

Gargauth:


bucket/islc.json
```diff
@@ -0,0 +1,24 @@
...
+            "Intelligent Standby List Cleaner (ISLC)"
+        ]
+    ],
+    "persist": "Intelligent standby list cleaner ISLC.exe.Config",
+    "checkver": {
+        "url": "https://www.wagnardsoft.com/",
+        "re": "Intelligent standby list cleaner [Vv]([\\d\\.]*) Released"
+    },
+    "autoupdate": {
+        "url": "https://www.wagnardsoft.com/ISLC/ISLC%20v$version.exe#/ISLC-$version.7z",
```
Ash258: Do not rename with versoin variable. It is redundant and do not help it anything.

```suggestion
        "url": "https://www.wagnardsoft.com/ISLC/ISLC%20v$version.exe#/ISLC.7z",
```

Gargauth:
Why not let checkver calculate hash from actual downloaded file? Wouldn't it be more reliable and prone to human error (eg. author forgetting to update sha hash when the utility gets updated)?

Ash258:
When the URL will not exists, then it will be downloaded and calculated. But when it exists, it is better to use provided hashes for security reasons.

bucket/islc.json
```diff
@@ -0,0 +1,28 @@
...
+        "extract_dir": "ISLC v$version",
+        "hash": {
+            "url": "https://www.wagnardsoft.com/content/intelligent-standby-list-cleaner-v1000-released",
```
Ash258:
```suggestion
            "url": "https://www.wagnardsoft.com/content/intelligent-standby-list-cleaner-v$cleanVersion-released",
```

PR #2035, opened 2019-04-21
add windowsspyblocker

linsui:
Block spying and tracking on Windows


rasa:
Replaces https://github.com/rasa/scoops/blob/master/wsb.json

PR #2053, opened 2019-04-24
Calibre: Change URL to fosshub

niheaven:
FossHub is more stabler and faster than GitHub.

### DIFF
--- a/bucket/calibre-normal.json
+++ b/bucket/calibre-normal.json
@@ -5,13 +5,13 @@
     "license": "GPL-3.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/kovidgoyal/calibre/releases/download/v3.42.0/calibre-64bit-3.42.0.msi",
-            "hash": "sha512:25a37b93d155ab9fd4730fe6f2a388605ceac1a12782540c194c288c2baa25cef87730fe94bd012f1ca6cf2dcb3a83f65306b1577db3792cb847e66ca61ae232",
+            "url": "https://www.fosshub.com/Calibre.html/calibre-64bit-3.41.3.msi",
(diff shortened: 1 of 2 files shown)


Ash258:
Why to remove hash and checver from github? Files should be identical or not?

niheaven:
FossHub link could use `fosshub` mode to get hash, and checkver from fosshub to avoid inconsistency of release time between github and fosshub.

niheaven:
Files are identical from all releasing sources.

Ash258:
Could be fosshub hash extract changed to sha1 or md5? This would distinguish fosshub hashmode at first for review. Without specifying `hash.mode`

My point is that lots of new contributors do not know about checkver and make everything manually (url, hashes, extracts,..) and when you see that there is sha1 or md5 hash inside manifest you know, that user ran checkver and you do not need to download that file and check it while review. (Same as in sourceforge). When you review sourceforge and you see sha256 hash you know, something is wrong.

niheaven:
Yes, it could, the hash extraction regex is https://github.com/lukesampson/scoop/blob/master/lib/autoupdate.ps1#L217 (`$Matches.filename+'.*?"sha256":"([a-fA-F0-9]{64})"'`), just change it to sth. likes `"sha1":"([a-fA-F0-9]{40})"` or `"md5":"([a-fA-F0-9]{32})"`.

The hash in html is as follow:

```
{"n":"calibre-3.41.3.msi","r":"5cba8e83d0a6f00ad7fd0fb5","hash":{"md5":"ec913c8ea7527078c55a5012bf0bf49e","sha1":"08a624eaaccc17168a6545a55758e96dd17f426c","sha256":"727d57b5096ca236113af68a9b69a253af40676985ca63b2f2d1816800aaa022"},"s":false}
```

niheaven:
If sha1 is preferred for review, I could submit a PR to change it. For daily use, the safety of sha1 and sha256 should be identical.

PR #2063, opened 2019-04-25
obs-studio: Update to version 23.1.0

raulfpl:
The automatic update using `checkver` will fail because the new version doesn't have the `.0` in the file name in the [download URLs](https://github.com/obsproject/obs-studio/releases/download/23.1.0/OBS-Studio-23.1-Full-x64.zip) from [the Github releases page](https://github.com/obsproject/obs-studio/releases).

### DIFF
--- a/bucket/obs-studio.json
+++ b/bucket/obs-studio.json
@@ -1,11 +1,11 @@
 {
     "homepage": "https://obsproject.com/",
     "license": "GPL-2.0-only",
-    "version": "23.0.2",
+    "version": "23.1.0",
     "architecture": {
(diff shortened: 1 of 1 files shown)


Ash258:
Make it manually is not solutions..

Just do something like this:

1. Add `checkver.regex` with `/download/([\\d\\.]+)/OBS-Studio-(?<short>[\\d\\.]+)-Full`
1. Replace `autoupdate.urls` `$version` inside filenames with `$matchShort`
1. Run checkver -f

raulfpl:
Done

PR #2067, opened 2019-04-28
Inno-Setup: Fix url

niheaven:
Wait until https://github.com/lukesampson/scoop/issues/3407 is fixed.


niheaven:
Could be merged, but useless until innounp updated.

r15ch13:
Dang it, yeah it won't work if `innounp` doesn't support it ...
That was a total brainfart from me ... :grin:

r15ch13:
Locked it to version 5 with https://github.com/lukesampson/scoop-extras/commit/7a8a8438515ebca830f96333b028f1ac7605b0bb

PR #2070, opened 2019-04-29
HashCheck: Add version 2.4.0

niheaven:
Close https://github.com/lukesampson/scoop/issues/3038.

Please MERGE AFTER https://github.com/lukesampson/scoop/pull/3411

### DIFF
--- a/bucket/hashcheck.json
+++ b/bucket/hashcheck.json
@@ -32,5 +32,8 @@
         "hash": {
             "url": "$url.sha256"
         }
-    }
+    },
+    "notes": [
+        "Before uninstallation, please close Explorer (or other files management tool, e.g., Total Commander) windows where you've used HashCheck from."
+    ]
 }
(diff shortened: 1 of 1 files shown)


indrakaw:
HashCheck required administrator to modify registry for context menu options. Instead having to `sudo` on installation, why don't you do like `VSCodium` done?

https://github.com/lukesampson/scoop-extras/blob/master/bucket/vscodium.json

niheaven:
- Property page is added by register DLL.
- Context menu is the only entrance of HashCheck, no need to ask user run reg or sth. else themselves.
- Why use this app without proper privilege? There are many cli checksum tools...

indrakaw:
I haven't see the executable binnary yet. Would you please add it to `bin` so it available on cmd?

niheaven:
There is no binary for HashCheck, only `Property` page dll. If you prefer cli tools, please consider `dirhash` or `shasum`.

PR #2076, opened 2019-04-30
fix: foobar2000-encoders, update checkver re and encoders folders 

ShuguangSun:
update foobar2000-encoders version, chan checkver's re; fix the Copy-item issue, and make the folder of encoders

### DIFF
--- a/bucket/foobar2000-encoders.json
+++ b/bucket/foobar2000-encoders.json
@@ -1,12 +1,12 @@
 {
-    "version": "2018-10-19",
+    "version": "2019-04-13",
     "license": {
         "identifier": "Freeware",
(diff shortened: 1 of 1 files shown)


Ash258:
Why to keep installer when you added extract_to property and changed exe to 7z.

ShuguangSun:
If it is not changed from exe to 7z, it can't installed in my machine. Don't know why.

Ash258:
Then just delete `installer` property if it could be just extracted

ShuguangSun:
Thanks to point it out.

The rational I change exe to 7z is because the exe installer require administrator rights.
Windows 10.

PR #2086, opened 2019-05-01
freecommander: Persist Settings folder

hpsbranco:


Ash258:
https://freecommander.com/en/faq-freecommander/#change-settings-location

You need to first set settings folder location inside config file and then you can persist settings folder.
If you set this, you will break setting for all users, since settings will not be copied from AppData to local folder.

Ash258:
> The fcStart.ini file make the FreeCommander portable: the settings are saved in subfolder of the installation folder.
If you already have the fcStart.ini in your target folder, then you should not overwrite the file.

Ash258:
So revert this change and create new manifest `freecommander-portable` with `installer.script` with `New-Item \"$dir\\fcStart.ini\" -Type File | Out-Null`

hpsbranco:
Actually, on a fresh unzipped install, the "FreeCommander Install Folder" is the default option, and it automatically creates the Settings folder if it doesn't exist. That's why we can get away by just creating the junction.

PR #2057, opened 2019-04-25
Fix issue #2039

TheCjw:
Fixed by renaming downloaded file to `*.7z`.

Details: https://github.com/lukesampson/scoop-extras/issues/2039#issuecomment-486510768

PR #2095, opened 2019-05-03
fix: solve shadowsocksr-csharp install error

waderchan:
fix: solve shadowsocksr-csharp install error on ExtractToFile because of the archive entry is compressed using an unsupported compression method.


Ash258:
Duplicate of #2057

PR #2112, opened 2019-05-07
Add WizNote 4.11.21

Retia-Adolf:


Ash258:
- You cannot simply set content for enhanced character map. 
- Do not edit persisted files inside post installl phase
- Autoupdate URL is not following normal URL format

ALl chinese characters will be replaced

image

Ash258:
Since homepage is not translated to english i would not recommend to merge this into extras.

Retia-Adolf:
gave up

PR #2116, opened 2019-05-09
Add syncthing-discosrv 1.0.1

linsui:


bucket/syncthing-discosrv.json
```diff
@@ -0,0 +1,46 @@
...
+        }
+    },
+    "bin": "stdiscosrv.exe",
+    "persist": [
+        "discovery.db",
+        "cert.pem",
+        "key.pem"
+    ],
+    "pre_install": [
+        "foreach ($name in @('cert.pem', 'key.pem')) {",
```
rasa: As `discovery.db` is not in the .zip, it should be created here, otherwise, it won’t be properly persisted.

linsui: It's a directory. It can be created automatically. Should I creat it manully?

linsui:
There is a problem that the stdiscosrv will generate key in the run directory. So if run `stdiscosry` directly it will generate a different key in user directory instead of use the key in scoop directory.

linsui:
Move to main bucket

PR #2127, opened 2019-05-10
cryptomator: add settings and logs to persist

raulfpl:
as suggested by @porti on 938df7a

### DIFF
--- a/bucket/cryptomator.json
+++ b/bucket/cryptomator.json
@@ -25,6 +25,6 @@
     "persist": "data",
     "post_install": [
         "# change appdata settings folder",
-        "(Get-Content \"$dir\\app\\Cryptomator.cfg\").replace('~/AppData/Roaming/Cryptomator', '../data').replace('\\', '/') | Set-Content \"$dir\\app\\Cryptomator.cfg\""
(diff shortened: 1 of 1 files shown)


bucket/cryptomator.json
```diff
@@ -21,5 +21,10 @@
         "hash": {
             "url": "https://dl.bintray.com/cryptomator/cryptomator/$version/Cryptomator-$version-x64.exe.sha256"
         }
-    }
+    },
+    "persist": "data",
+    "post_install": [
+        "# change appdata settings folder",
+        "(Get-Content \"$dir\\app\\Cryptomator.cfg\").replace('~/AppData/Roaming/Cryptomator', '../data').replace('\\', '/') | Set-Content \"$dir\\app\\Cryptomator.cfg\""
```
Ash258: - Why to replace `\` with `/`?
- No encoding
    - Use `ASCII`

raulfpl: I really don't see why replace \ with /.
Removed that part and added ASCII encoding

raulfpl:
Any other changes needed?

PR #2130, opened 2019-05-11
Add DriverStore Explorer version 0.10.15

DavidSHwang:


r15ch13:
Please run `.\bin\checkver.ps1 driverstoreexplorer -f` and push the changes. No need to make a new Pull-request.
https://github.com/lukesampson/scoop-extras/pull/2129

Ash258:
There is some pdb file. Should it be persisted?

r15ch13:
@Ash258 pdb files contain .NET debugger stuff and can be ignored

DavidSHwang:
Seems that `Rapr.exe.config` should be persisted. Working on it.

Ash258:
There is usually saved only files which are required to run exe files (dlls, ...)
Is there any special config option, which is related?

DavidSHwang:
The config file memorises the language option and the location of the window. Are these necessary to be kept?

r15ch13:
Nope, that is also a .NET file. It just describes the configuration. The settings will actually be stored in `%appdata%`. Please remove the `persist` property.

DavidSHwang:
Done.

Issue #2126, opened 2019-05-10
[Request] Add LibreCAD.


ssfjhh:
LibreCAD is a cross-platform 2D CAD program written in C++11 using the Qt framework. It can read DXF and DWG files and can write DXF, PDF and SVG files. The user interface is highly customizable and has dozens of translations. <http://librecad.org/>

PR #2145, opened 2019-05-14
Add librecad version 2.1.3

linsui:
#2126

### DIFF
--- a/bucket/librecad.json
+++ b/bucket/librecad.json
@@ -1,0 +1,21 @@
+{
+    "version": "2.1.3",
+    "description": "LibreCAD is a cross-platform 2D CAD program written in C++11 using the Qt framework. It can read DXF and DWG files and can write DXF, PDF and SVG files. The user interface is highly customizable, and has dozens of translations.",
+    "homepage": "https://librecad.org/",
+    "license": "GPL-2.0",
+    "url": "https://github.com/LibreCAD/LibreCAD/releases/download/2.1.3/LibreCAD-Installer-2.1.3.exe#/dl.7z",
+    "hash": "4c0855316934fbb9af712df2bdbc7c11e46b4569af2e6f2d4d7e6574d13f05cb",
+    "bin": "LibreCAD.exe",
+    "shortcuts": [
+        [
+            "LibreCAD.exe",
+            "LibreCAD"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/LibreCAD/LibreCAD"
+    },
+    "autoupdate": {
+        "url": "https://github.com/LibreCAD/LibreCAD/releases/download/$version/LibreCAD-Installer-$version.exe#/dl.7z"
+    }
+}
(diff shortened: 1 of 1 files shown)


ssfjhh:
It works very well. Thank you.

Ash258:
<details>
<summary>Optimized manifest</summary>

- Added hash extraction
- Corrected license identifier
- Simplier description.
    - No need to feed users with informations about language or framework

```json
{
    "version": "2.1.3",
    "description": "Cross-platform 2D CAD program.",
    "homepage": "https://librecad.org/",
    "license": "GPL-2.0-only",
    "url": "https://github.com/LibreCAD/LibreCAD/releases/download/2.1.3/LibreCAD-Installer-2.1.3.exe#/dl.7z",
    "hash": "sha1:03ea2ea9634294608c26f4a7835c17e372bac0d4",
    "bin": "LibreCAD.exe",
    "shortcuts": [
        [
            "LibreCAD.exe",
            "LibreCAD"
        ]
    ],
    "checkver": {
        "github": "https://github.com/LibreCAD/LibreCAD/"
    },
    "autoupdate": {
        "url": "https://github.com/LibreCAD/LibreCAD/releases/download/$version/LibreCAD-Installer-$version.exe#/dl.7z",
        "hash": {
            "url": "https://sourceforge.net/projects/librecad/files/Windows/$version",
            "regex": "$basename\":.+?\"sha1\":\\s\"$sha1"
        }
    }
}
```

</details>

linsui:
@Ash258 Thanks!

Ash258:
Keep it sha1. This regex is used for sourceforge mode.
That one you presented is too greedy. If there will be more files it will get incorrect one.

https://github.com/lukesampson/scoop/blob/6141e46d6ae74b3ccf65e02a1c3fc92e1b4d3e7a/lib/autoupdate.ps1#L252

linsui:
Oh, I get. Thanks. But could you tell me more about it? :confused: And will there be a variable to extract hash from sourceforge?

Ash258:
Tell you more about what?

The hash would be extracted automatically if you would set download url to sourceforge, but sourceforge is not much reliable. Github is more stable.

linsui:
@Ash258 So the regex here is just for reliability and without it scoop can still extract hash? And why don't use sha256? Is there any documents? Thanks.

Ash258:
Without this regex, scoop will download that file and calculate hash from this downloaded file. If you specify this regex there will not be needed to download the full installer file.

`sha256` hash is completely OK. But the regex you presented will potentionally match some other hash (hash for other file, if there will ever be file listed before installer). As you used `$sha256.+$basename`
In worst case, it could ended like this: https://regex101.com/r/2jwjDo/1/

linsui:
@Ash258 Thanks! I'd like to use `$sha256\/analysis\/">[A-Za-z\s]+$basename` first but it's ugly. Maybe `$sha256.{35}$basename` is OK?

Ash258:
Keep it as is. It is working and tested in lots of manifests.

linsui:
@Ash258 Just a question.  :joy: Thanks.

PR #2179, opened 2019-05-20
msiafterburner: add vcredist2008 suggestion

Francesco149:
msi afterburner requires vcredist2008 otherwise it won't start, complaining about side-by-side configuration if started as admin and silently crashing if started as user

### DIFF
--- a/bucket/msiafterburner.json
+++ b/bucket/msiafterburner.json
@@ -5,6 +5,9 @@
     "version": "4.6.0",
     "url": "http://download.msi.com/uti_exe/vga/MSIAfterburnerSetup.zip#/MSIAfterburnerSetup460.7z",
     "hash": "d186cdf533727df267e0cb7cccfdb5673fbb8cb443ab009de023cae07345e31a",
+    "suggest": {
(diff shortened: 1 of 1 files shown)


bucket/msiafterburner.json
```diff
@@ -5,6 +5,7 @@
     "version": "4.6.0",
     "url": "http://download.msi.com/uti_exe/vga/MSIAfterburnerSetup.zip#/MSIAfterburnerSetup460.7z",
     "hash": "d186cdf533727df267e0cb7cccfdb5673fbb8cb443ab009de023cae07345e31a",
+    "depends": "extras/vcredist2008",
```
Ash258: Never depends on vcredists.

Users normally uninstall vcredist packages after installation (remove installatino files) so you would install them twice.

Use:

```suggestion
    "suggest": {
        "Visual C++ Redist 2008": "extras/vcredist2008"
    },
```

r15ch13:
Please add vcredist only as a suggestion.

Francesco149:
ah i see, makes sense. it's a suggestion now

PR #2180, opened 2019-05-20
Add ssh-agent-wsl version 2.2

JanPokorny:
Hello, I am new to scoop, so I would appreciate if someone could check this manifest.

- Is the auto-update part correct?
- Is there a way to make the notes more user-friendly, eg. by a script echoing the correct path instead of the hardcoded ones?

### DIFF
--- a/bucket/ssh-agent-wsl.json
+++ b/bucket/ssh-agent-wsl.json
@@ -9,10 +9,13 @@
     "autoupdate": {
         "url": "https://github.com/rupor-github/ssh-agent-wsl/releases/download/v$version/ssh-agent-wsl.7z"
     },
-    "notes": [
-        "Please add ssh-agent-wsl to your WSL ~/.bashrc using the full path.",
-        "Usually:",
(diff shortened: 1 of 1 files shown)


Ash258:
For notes.

Cannot really test because I do not want to install WSL, but if wsl inherit user environment variables, then you could use something like this: `eval $($SCOOP/apps/ssh-agent-wsl/current/ssh-agent-wsl -r)` or `eval $($SCOOP_GLOBAL/..... -r)`

JanPokorny:
@Ash258 Env isn't shared to WSL, you have to explicitly enable it case-by-case, which is more work for the user. Besides, it looks like those variables aren't even set on my machine -- `Get-ChildItem env:` doesn't list them.

Can post-install script echo stuff? That would be nice, as then `wslpath -a '...path'` could be used to transform the path.

Ash258:
You could do everything you want inside pre|post_install blocks.

JanPokorny:
@Ash258 I think it's ready to be merged now.

JanPokorny:
What should I do to help this get merged?

PR #2188, opened 2019-05-23
Added conan io package manager

rgozim:
Please note, the `autoupdate` block is not entirely correct. The actual version should replace `.` with `_` in the url strings. Could someone help with this?


Ash258:
ScoopInstaller/Main#83

PR #2205, opened 2019-05-28
rapidee: Enable portable mode and better version extraction

### LINKED ISSUE #2202: Implement (found) fix for RapidEE to be fully portable
https://www.rapidee.com/en/portable-mode

> ...To do this just create an empty `rapidee.ini` file in the app directory (if it doesn't already exist).

Adding an empty file named `rapidee.ini` in the same (root) directory as `rapidee.exe` does the job.

hoilc:
fix #2202

### DIFF
--- a/bucket/rapidee.json
+++ b/bucket/rapidee.json
@@ -1,5 +1,7 @@
 {
-    "version": "937",
+    "version": "9.2.937",
+    "description": "Environment variables editor.",
+    "homepage": "https://www.rapidee.com/en/about",
     "license": {
         "identifier": "Freeware",
         "url": "https://www.rapidee.com/en/license"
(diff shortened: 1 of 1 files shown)


bucket/rapidee.json
```diff
@@ -16,6 +16,8 @@
     },
     "homepage": "https://www.rapidee.com/en/about",
     "bin": "rapidee.exe",
+    "persist": "rapidee.ini",
+    "pre_install": "if(!(Test-Path \"$persist_dir\\rapidee.ini\")) {Set-Content -Encoding ASCII -Path \"$dir\\rapidee.ini\" -Value ''}",
```
Ash258: No need to open that file and insert empty string. Just create it to save opening and saving.

```suggestion
    "pre_install": "if(!(Test-Path \"$persist_dir\\rapidee.ini\")) { New-Item \"$dir\\rapidee.ini\" -Type File }",
```

hoilc: It works. It's strange that I tried `New-Item` and rapidee didn't recognize the generated empty file. 😂

PR #2210, opened 2019-05-28
Add tinode-mysql 0.15.14

linsui:
Should I also add tinode-rethinkdb?


r15ch13:
`templ` and `static` should probably not be persistent. But it will create an `uploads` directory for file attachments.

linsui:
@r15ch13  `static` and `templ` may be changed if users want to customize the website.

PR #2217, opened 2019-05-30
Add Shiba version 1.2.1

blyoa:
I added a markdown live preview app.

I did not use `extract_dir` because the working directory of the binary file Shiba.exe must be an actual directory instead of a symbolic link.

### DIFF
--- a/bucket/shiba.json
+++ b/bucket/shiba.json
@@ -38,7 +38,7 @@
                 "url": "https://github.com/rhysd/Shiba/releases/download/v$version/Shiba-win32-x64.zip"
             },
             "32bit": {
-                "url": "https://github.com/rhysd/Shiba/releases/download/$$version/Shiba-win32-ia32.zip"
+                "url": "https://github.com/rhysd/Shiba/releases/download/v$version/Shiba-win32-ia32.zip"
             }
         }
     }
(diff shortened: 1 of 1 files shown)


bucket/shiba.json
```diff
@@ -0,0 +1,45 @@
...
+        }
+    },
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/rhysd/Shiba/releases/download/v$version/Shiba-win32-x64.zip"
+            },
+            "32bit": {
+                "url": "https://github.com/rhysd/Shiba/releases/download/$$version/Shiba-win32-ia32.zip"
```
Ash258:
```suggestion
                "url": "https://github.com/rhysd/Shiba/releases/download/v$version/Shiba-win32-ia32.zip"
```

blyoa: Thank you for the review. I updated the PR according to your review.

Ash258:
> I did not use extract_dir because the working directory of the binary file Shiba.exe must be an actual directory instead of a symbolic link.

@r15ch13 Have you tried installation and executing from current / PATH

r15ch13:
@Ash258 yes, I did test it and removed the `extract_dir` again. My fault.

Opened an issue for the symlink problem: https://github.com/rhysd/Shiba/issues/52

PR #2224, opened 2019-05-31
filezilla: persist fzdefaults.xml and improvments

vmvmpvm:
Various settings can be configured in fzdefaults.xml, it would be convenient if this config file is persisted across version upgrades.


vmvmpvm:
thanks @Ash258 for fixed manifest, committed it.

PR #2229, opened 2019-06-01
Add qBittorrent-Enhanced-Edition 4.1.6.1

linsui:


Ash258:
Adding fork of repository, which is actively maintananced is nonsense. 

Will you add new qbittorrent manifest for each feature, which is WIP or not priority for developers??

This is rather candidate for custom bucket or custom gist

linsui:
:ok:

PR #2243, opened 2019-06-04
tor-browser: update to version 8.5.1

navrudh:
changed checkver regex url


navrudh:
There seems to be a pull request for tor-browser that's under consideration already.

PR #2245, opened 2019-06-05
grafana: Fix download hash

felixsteghofer:
Hi, 
the hash you provided is for the .msi installer but the standalone binaries are downloaded.
See [https://grafana.com/grafana/download?platform=windows](https://grafana.com/grafana/download?platform=windows)


Ash258:
Then change `autoupdate.hash` into `Binaries.+$sha256`.
This will automatically get hash for correct file on next autoupdate

PR #2247, opened 2019-06-05
add devart codecompare

vmvmpvm:


vmvmpvm:
> How do official buckets stand for trial software?

code compare has a free edition too, when it prompts to register we can select Free edition.

But please hold off on this, I just noticed an error when starting this app after installing with scoop. Hopefull I will it or cancel this PR.

vmvmpvm:
Not able to get portable install working, logged below issue with Devart, hopefully they will provide an option.

https://devart.uservoice.com/forums/93035-code-compare/suggestions/37839019-non-admin-single-user-portable-install-please

PR #2248, opened 2019-06-05
add microsoft rdcman

vmvmpvm:
Courtesy https://github.com/jfut/scoop-jfut/blob/master/rdcman.json

### DIFF
--- a/bucket/rdcman.json
+++ b/bucket/rdcman.json
@@ -1,0 +1,19 @@
+{
+    "homepage": "https://www.microsoft.com/en-us/download/details.aspx?id=44989",
+    "version": "2.7.1406.0",
+    "license": "Freeware",
+    "url": "https://download.microsoft.com/download/A/F/0/AF0071F3-B198-4A35-AA90-C68D103BDCCF/rdcman.msi",
+    "hash": "6e29f25c5e7f5ec587d07642a344b05429472a75b4f73177cd6aac63be61a455",
+    "extract_dir": "Program Files\\Microsoft\\Remote Desktop Connection Manager",
+    "bin": "RDCMan.exe",
+    "shortcuts": [
+        [
+            "RDCMan.exe",
+            "Remote Desktop Connection Manager"
+        ]
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


Ash258:
Stealing someone elses work?? How do author feel about this PR?? Do you even ask for permission to post his work to this bucket?

vmvmpvm:
well I can create the rdcman, it is not a big deal to create this, I had to mention because I already saw it is there and I used that json to install use it locally.

I mentioned the link here, so I assume you mean the author is not automatically notified here.

Mentioning author here @jfut

jfut:
If you like, I will send it as my new PR.

vmvmpvm:
> If you like, I will send it as my new PR.

yes, thanks @jfut, please do, I am closing this PR.

rasa:
It’s not stealing as it’s in the public domain, and therefore everyones, per the Unlicense: “A license with no conditions whatsoever which dedicates works to the public domain. Unlicensed works, modifications, and larger works may be distributed under different terms and without source code.”

vmvmpvm:
Thanks @rasa, that makes me feel better.

Ash258:
@rasa Yeah that is true, but still you can inform author about his manifest is being used in other bucket. You do not need permission from him under Unlicensed, but it is pretty dishonest to just copy manifest and leave credits to self.

vmvmpvm:
@Ash258 I agree with that as well, in my case I did not try to leave the credit to myself, I mentioned the URL of where I got it from, and it was not intentional that I did not mention author, I added author when I realized that. If the intent was to steal credit there was no need to mention the URL itself.

I hope you get the point, I will leave it here because further discussion is not worth all our time.

rasa:
@Ash258 I respect your position, but he did say it was “courtesy” of @jfut, so he was giving credit to the author. The Unlicense does not require credit, and it does not include an attribution clause, as most other licenses do, except for things like CC0, WTFPL, etc.
See https://en.m.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses#General_comparison

PR #2258, opened 2019-06-07
Update VSCodium binary name

jingyu9575:
The binary name of VSCodium has [changed](https://github.com/VSCodium/vscodium/pull/176).

Also, the hash check failed.


Ash258:
#2253

PR #2262, opened 2019-06-08
Add bug.n: dynamic tiling window manager

ertwro:
The only available window manager, hotkey daemon for Windows. Free, open source.

### DIFF
--- a/bucket/bugn.json
+++ b/bucket/bugn.json
@@ -1,0 +1,21 @@
+{
+    "homepage": "https://github.com/fuhsjr00/bug.n",
+    "description": "Dynamic Tiling Window Manager written in AutoHotkey and based on suckless dwm",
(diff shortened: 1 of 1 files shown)


bucket/bugn.json
```diff
@@ -0,0 +1,31 @@
+{
+    "homepage": "https://github.com/fuhsjr00/bug.n",
+    "description": "bug.n is a tiling window manager add-on for the Explorer shell of Microsoft Windows. It is written in the scripting language AutoHotkey",
+    "license": {
```
Ash258: - Wrong SPDX identifier.
- Do not use URL when it is SPDX license

```suggestion
    "license": "GPL-3.0-or-later",
```

ertwro: Thanks. Done.

PR #2144, opened 2019-05-14
Chromes: Add hash extraction and all chrome versions

Ash258:
- Closes #1984 

- Use direct google url.
- Add hash extraction

Thx @h404bi for website ❤😍

This should  be merged when lukesampson/scoop#3458 is in master

TODO: 

- [ ] Simplify installer.script with `Expand-7zipArchive \"$dir\\chrome.7z\" -ExtractDir 'Chrome-bin' -Removal` (When lukesampson/scoop#3466 is in master)

PR #2291, opened 2019-06-17
chrome: Update to 75.0.3770.90 (manually)

rasa:
The [excavator](https://scoop.r15.ch/extras/mud-20190617-210001.log) continues to fail with:
````
Downloading 75.0.3770.90_chrome_installer.exe to compute hashes!
The SSL connection could not be established, see inner exception. Authentication failed, see inner exception.
URL https://redirector.gvt1.com/edgedl/release2/chrome/AN_q_RfOOOCt_75.0.3770.90/75.0.3770.90_chrome_installer.exe#/cosi.7z is not valid
Could not find hash!
ERROR Could not update chrome 32bit

### DIFF
--- a/bucket/chrome.json
+++ b/bucket/chrome.json
@@ -1,16 +1,16 @@
 {
-    "version": "74.0.3729.131",
+    "version": "75.0.3770.90",
     "description": "Fast, secure, and free web browser, built for the modern web.",
     "homepage": "https://www.google.com/chrome/",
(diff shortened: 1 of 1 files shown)


niheaven:
Ref: https://github.com/lukesampson/scoop-extras/pull/2144

rasa:
> Ref: #2144

Ok, let’s hold off until #2144 is merged in.

PR #2294, opened 2019-06-17
lunacy: Update to 4.0.1 (manually)

rasa:
Tested and working. .exe is 32-bit.


niheaven:
Suggest using `>([\\d.]+)<\/h2>` as checkver's regex.

PR #2298, opened 2019-06-18
openscad-dev: Update to 2019.05.19 (manually)

rasa:


bucket/openscad-dev.json
```diff
@@ -1,50 +1,40 @@
...
-            "openscad.exe",
-            "OpenSCAD Snapshot"
-        ]
-    ],
     "post_install": [
         "# OpenSCAD can't be started from a symlinked directory. See: https://github.com/openscad/openscad/issues/1309",
-        "startmenu_shortcut \"$original_dir\\openscad.exe\" \"OpenSCAD Snapshot\"",
-        "shim \"$original_dir\\openscad.exe\" $false \"openscad-dev\""
+        "startmenu_shortcut \"$original_dir/openscad.exe\" 'OpenSCAD Snapshot'",
+        "shim \"$original_dir/openscad.exe\" $false 'openscad-dev'"
```
Ash258: Why do you add inconsistency to manifests?

Almost every manifest and other codebase use windows backslashes for paths. Why to change it?

rasa: I personally think forward-slashes are better, but I think consistency is more important. I will stop making this change unless it's decided to change it *everywhere* to forward-slash.

PR #2317, opened 2019-06-19
uniextract2: Add version 2.0.0-rc.2

issaclin32:
Official homepage for UniExtract2: https://github.com/Bioruebe/UniExtract2

### DIFF
--- a/bucket/uniextract2.json
+++ b/bucket/uniextract2.json
@@ -19,7 +19,7 @@
     ],
     "checkver": {
         "url": "https://github.com/Bioruebe/UniExtract2/releases",
-        "regex": "\\/Bioruebe\\/UniExtract2\\/releases\\/download\\/v(?<version>[\\w.-]+)\\/(?<filename>\\w+\\.zip)"
+        "regex": "download/v([\\w.-]+)/(?<filename>\\w+\\.zip)"
     },
     "autoupdate": {
         "url": "https://github.com/Bioruebe/UniExtract2/releases/download/v$version/$matchFilename"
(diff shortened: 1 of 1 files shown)


bucket/uniextract2.json
```diff
@@ -0,0 +1,27 @@
...
+    ],
+    "shortcuts": [
+        [
+            "UniExtract.exe",
+            "Universal Extractor"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/Bioruebe/UniExtract2/releases",
+        "regex": "\\/Bioruebe\\/UniExtract2\\/releases\\/download\\/v(?<version>[\\w.-]+)\\/(?<filename>\\w+\\.zip)"
```
Ash258: No need to be so strict.

```suggestion
        "regex": "download/v([\\w.-]+)/(?<filename>\\w+\\.zip)"
```

Ash258: A

bucket/uniextract2.json
```diff
@@ -0,0 +1,27 @@
+{
+    "version": "2.0.0-rc.2",
+    "description": "A program designed to decompress and extract files from any type of archive or installer.",
```
Ash258:
```suggestion
    "description": "Decompress and extract files from any type of archive or installer.",
```

rasa:
In general, do we want to allow `-rc` builds? If so, perhaps we should modify scoop, instead of processing them manually?

niheaven:
For UniExtract2, we need RC build, since its stable version (maybe 2.0.0) is still not within the foreseeable future.

PR #2322, opened 2019-06-19
fiddler: Extract instead of installation

Ash258:
There are two installers inside `$PLUGINSDIR`, not sure how to handle these, opening this PR for some feedback, suggestions
A

### DIFF
--- a/bucket/fiddler.json
+++ b/bucket/fiddler.json
@@ -1,26 +1,27 @@
 {
-    "license": "Freeware",
-    "homepage": "https://www.telerik.com/fiddler",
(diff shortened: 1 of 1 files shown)


Ash258:
When extraction will be handled manually there is only need to investigate what is purpose of two installers with same name.

`Expand-7zipArchive $dir\$fname -Overwrite Rename -Removal`

Then i guess one will be for 32 bit and second for 64bit

`architecture.64bit.installer.script = Expand-7zipArchive $dir\Installer_1 -Removal`

Issue #2326, opened 2019-06-20
defraggler@2.22.995: hash check failed


tnotstar:
While trying to install **defraggler**, check of the hash code from downloaded file has beed failed.  After investigation the download url has beed changed from **www.piriform.com** to **www.ccleaner.com**.

tnotstar:
This looks as duplicate of issue #2272

rasa:
The hash 2edfa974fc2e8e34c0b8c152077da9c3a4ab28cad81a4ced463f0cf6634a0f01 in the manifest (the “expected” hash above) is correct. Please try again.

PR #2327, opened 2019-06-20
defraggler: changed domain name from urls (#2326)

tnotstar:
* upstream has changed its domain name from `www.piriform.com`
  to `www.ccleaner.com`, so this fix urls up

### DIFF
--- a/bucket/defraggler.json
+++ b/bucket/defraggler.json
@@ -1,7 +1,7 @@
 {
     "homepage": "https://www.ccleaner.com/defraggler",
     "version": "2.22.995",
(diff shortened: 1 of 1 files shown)


Ash258:
piriform url works too. There will be just redirect to ccleaner.com, which is completely fine

image

tnotstar:
You're right. It could be a transient condition, to me the redirection also works now.

rasa:
If it removes a 302, I’d say let’s merge it in. Agreed?

tnotstar:
The issue #2326 looks as duplicated of #2272. Both are related to #2329.

Ash258:
Supressed by #2351

tnotstar:
With #2351 this PR should be closed.

PR #2339, opened 2019-06-23
TranslucentTB: Add version 2018.2

### LINKED ISSUE #2155: [request] Add TranslucentTB.
A lightweight (uses a few MB of RAM and almost no CPU) utility that makes the Windows taskbar translucent/transparent on Windows 10.
website:<https://github.com/TranslucentTB/TranslucentTB>

issaclin32:
close #2155
Homepage of TranslucentTB: https://github.com/TranslucentTB/TranslucentTB


linsui:
You could use `"innosetup": true,` instead of un/installer script.

issaclin32:
> You could use `"innosetup": true,` instead of un/installer script.

Thanks for the suggestion.
I will still keep the "uninstaller" option, since TranslucentTB only show as a tray icon (instead of a window).
This avoids error when user is tries to uninstall TranslucentTB while it's still running in the background.

linsui:
Maybe you don't need to worry about that, I don't know...
It seems the portable mode have not been released. And a `portable` file is needed to enable portable mode.
See https://github.com/TranslucentTB/TranslucentTB/issues/206.
So the `persist` is not useful.

issaclin32:
Thank you for the information.

I am actually not sure what to do. 

Is it better to just cancel the PR and wait for future releases? Or just add comments (like what I just did), and wait for someone to see it and modify the manifest after the "portable ZIP version" released.

PR #2348, opened 2019-06-24
SciTE: Add version 4.1.7 and Sc1

niheaven:
https://www.scintilla.org/SciTE.html

### DIFF
--- a/bucket/sc1.json
+++ b/bucket/sc1.json
@@ -1,20 +1,24 @@
 {
     "homepage": "https://www.scintilla.org/SciTE.html",
-    "version": "4.1.7",
+    "version": "4.2.0",
     "license": {
         "identifier": "Freeware",
         "url": "https://www.scintilla.org/License.txt"
     },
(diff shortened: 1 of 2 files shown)


Ash258:
What is different between these two manifests?? Both have same descriptions, same version, but one is simple exe and second is zip file

niheaven:
>  Windows Executables
> 
> Windows executables only support Windows XP and later. It may be possible to build with older compilers for Windows 2000 but that is no longer tested.
> 
> A full download (1550K) includes the SciTE executable, any required DLLs, configuration files and documentation. After downloading the file, unzip it, and run SciTE.EXE. The files required to run SciTE are SciTE.EXE, SciLexer.DLL, and SciTEGlobal.properties.
> 
> A single file executable called Sc1 (1000K) does not need any DLL or properties files as these are linked into the executable. You may still create properties files if you wish. Sc1.exe has been compressed with the UPX compressor so that it is a fast download. It does not need to be decompressed to be used.

PR #2350, opened 2019-06-24
Using http instead of https allows 302 redirects to work for Eclipse

nightscape:
Currently, scoop [cannot follow `302` redirects from an `https` URL to an `http` one](https://github.com/lukesampson/scoop/issues/3155).
Unfortunately, some of the Eclipse mirrors are still using `http`, so the redirect does not work, depending on which mirror gets chosen.

### DIFF
--- a/bucket/eclipse-automotive.json
+++ b/bucket/eclipse-automotive.json
@@ -4,11 +4,11 @@
     "version": "4.5.2",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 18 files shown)


rasa:
Dup of https://github.com/lukesampson/scoop-extras/pull/2557 . Sorry we didn’t use your work, but thanks for the contribution!

nightscape:
No worries 😄

PR #2352, opened 2019-06-24
grepWin: Add version 1.9.1

issaclin32:
Official repo of grepWin: https://github.com/stefankueng/grepWin

I added some Powershell scripts in the manifest, and put registry files in the `/script` folder (similar to the solution from `VSCode` manifest),

so that users can add grepWin to the right-click context menu by **explicitly** typing **$dir\grepwin-install-context.reg**


bucket/grepwin.json
```diff
@@ -0,0 +1,82 @@
...
+        "32bit": {
+            "url": "https://github.com/stefankueng/grepWin/releases/download/1.9.1/grepWin-1.9.1_portable.exe#/grepWin_portable.exe",
+            "hash": "015a9427d29be42dfa0defc73aa613a44d6eb3b0f5b76152f988d941fb8d9868"
+        },
+        "64bit": {
+            "url": "https://github.com/stefankueng/grepWin/releases/download/1.9.1/grepWin-x64-1.9.1_portable.exe#/grepWin_portable.exe",
+            "hash": "95335b4ad1e42149a8e5e3ce4ce7114ba8cc50d5d671209a917c4aa6b8eca653"
+        }
+    },
+    "installer": {
```
Ash258: I would rather see more maintainable approach similar to vscode / tortoisesvn

issaclin32: The solution from `VScode` looks much cleaner. I'll modify the manifest later.

issaclin32: all done.

issaclin32:
I have set `core.autocrlf` to "input", and set core.eol to "crlf".
Not sure why it is failing. :confused:

未命名 - 3
未命名 - 2

bucket/grepwin.json
```diff
@@ -0,0 +1,81 @@
...
+            ]
+        }
+    },
+    "installer": {
+        "script": [
+           "$app_path = \"$dir\\grepWin_portable.exe\".Replace('\\', '\\\\')",
+           "$reg_content = Get-Content \"$dir\\grepwin-install-context.reg\"",
+           "$reg_content = $reg_content.Replace('$app_path', $app_path)",
+           "$reg_content | Set-Content -Path \"$dir\\grepwin-install-context.reg\"",
+           "$script_content = @(\"",
```
Ash258: You do not need to save it into ps1 scripts. 

These files are easily clickable from terminal

issaclin32: I'll remove the PS1 scripts, and prompt users to run **reg import $dir\grepwin-install-context.reg** (as it doesn't require admin rights)

Ash258:
There is https://github.com/lukesampson/scoop/blob/master/.gitattributes for default eol. No need to change anything to your git config.

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
...
+                "4e99b868f8078fad8faee220c214dbb3288848109ae515fd50eca205b44252c5"
+            ]
+        }
+    },
+    "installer": {
+        "script": [
+           "$app_path = \"$dir\\grepWin_portable.exe\".Replace('\\', '\\\\')",
+           "$reg_content = Get-Content \"$dir\\grepwin-install-context.reg\"",
+           "$reg_content = $reg_content.Replace('$app_path', $app_path)",
+           "$reg_content | Set-Content -Path \"$dir\\grepwin-install-context.reg\""
```
Ash258:
```suggestion
           "Set-Content \"$dir\\grepwin-install-context.reg\" $reg_content -Encoding ASCII"
```

bucket/grepwin.json
Ash258: Resolve conversations!

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
+{
+    "version": "1.9.1",
+    "description": "Regular expression search and replace tool for Windows.",
+    "homepage": "https://tools.stefankueng.com/grepWin.html",
+    "license": "GPL-3.0-only",
+    "notes": "Run 'reg import $dir\\grepwin-install-context.reg' to add grepWin to right-click context menu. \n(This does not require admin rights)",
```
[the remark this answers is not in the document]
issaclin32: Why removing `reg import`?

Opening the `.reg` file directly with regedit will require admin rights.
`reg import xxx.reg` does not require admin rights. (when only registries under `HKCU` are modified)
This is useful when the user does not have an admin account (e.g. using a PC at the workplace)

Also, opening `.reg` file directly will show up a message box. (even with admin rights)
If you are running all the `scoop` installs with a simple batch file, this will interrupt the whole process.
I think it will be better to not showing the message box at all.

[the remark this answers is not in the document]
Ash258: If you are running in batch there is no difference to execute it from explorer or opening new terminal to run this command.

issaclin32:
> There is https://github.com/lukesampson/scoop/blob/master/.gitattributes for default eol. No need to change anything to your git config.

Thank you for the information.
I didn't change any settings at the first place. I am still not sure what went wrong.

bucket/grepwin.json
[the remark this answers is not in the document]
issaclin32: okay, I will remove the "reg import", for the sake of simplicity.

issaclin32:
> > There is https://github.com/lukesampson/scoop/blob/master/.gitattributes for default eol. No need to change anything to your git config.
> 
The registry file was encoded `UCS-2`, re-encode it in `UTF-8` solved the problem.

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
...
+           "$app_path = \"$dir\\grepWin_portable.exe\".Replace('\\', '\\\\')",
+           "$reg_content = (Get-Content \"$dir\\grepwin-install-context.reg\") -replace '$app_path', $app_path",
+           "$reg_content = $reg_content.Replace('$app_path', $app_path)",
+           "Set-Content \"$dir\\grepwin-install-context.reg\" $reg_content -Encoding ASCII"
+        ]
+    },
+    "uninstaller": {
+        "script": "reg import grepwin-uninstall-context.reg"
+    },
+    "bin": "grepWin_portable.exe",
```
Ash258: _portable? Why? 

Can you image you are using `cat file.txt | grepWin_portable`??

Ash258: Rename all urls into `grepWin.exe`

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
...
+                "a8fcb0a817564696699bead7055e5a3b453cc2030b9a37714ff5a58db687e773",
+                "602386d2fbabf1ef29bf32fc601c4b575d65c0fa0d8377e075f6821afecae74c"
+            ]
+        }
+    },
+    "installer": {
+        "script": [
+           "$app_path = \"$dir\\grepWin_portable.exe\".Replace('\\', '\\\\')",
+           "$reg_content = (Get-Content \"$dir\\grepwin-install-context.reg\") -replace '$app_path', $app_path",
+           "$reg_content = $reg_content.Replace('$app_path', $app_path)",
```
Ash258: Mixing two syntaxes for no reason

```suggestion
           "$reg_content = $reg_content -replace '$app_path', $app_path",
```

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
...
+           "Set-Content \"$dir\\grepwin-install-context.reg\" $reg_content -Encoding ASCII"
+        ]
+    },
+    "uninstaller": {
+        "script": "reg import grepwin-uninstall-context.reg"
```
Ash258: This need $dir

```suggestion
        "script": "reg import \"$dir\\grepwin-uninstall-context.reg\""
```

bucket/grepwin.json
```diff
@@ -0,0 +1,65 @@
...
+           "$reg_content = (Get-Content \"$dir\\grepwin-install-context.reg\") -replace '$app_path', $app_path",
+           "$reg_content = $reg_content -replace '$app_path', $app_path",
+           "Set-Content \"$dir\\grepwin-install-context.reg\" $reg_content -Encoding ASCII"
+        ]
+    },
+    "uninstaller": {
+        "script": "reg import \"$dir\\grepwin-uninstall-context.reg\""
+    },
+    "bin": "grepWin_portable.exe",
+    "persist": "grepwin.ini",
```
Ash258: Also ini file do not exists after downloading. You need to create it inside installer.script so it is not created as directory

issaclin32: I have added the content of *grepwin.ini* in the installer script, and tested to make sure it's working properly.

bucket/grepwin.json
```diff
@@ -0,0 +1,66 @@
...
+            ]
+        }
+    },
+    "installer": {
+        "script": [
+           "$app_path = \"$dir\\grepWin.exe\".Replace('\\', '\\\\')",
+           "$reg_content = (Get-Content \"$dir\\grepwin-install-context.reg\") -replace '$app_path', $app_path",
+           "$reg_content = $reg_content -replace '$app_path', $app_path",
+           "Set-Content \"$dir\\grepwin-install-context.reg\" $reg_content -Encoding ASCII",
+           "Set-Content \"$dir\\grepwin.ini\" \"[global]`n[Software\\grepWin\\History]`n\""
```
Ash258:
```suggestion
          "if (-not (Test-Path \"$persist_dir\\grepwin.ini\")) {",
           "    Set-Content \"$dir\\grepwin.ini\" (@('[global]', '[Software\\grepWin\\History]') -join \"`r`n\") -Encoding ASCII",
           "}"
```

bucket/grepwin.json
```diff
@@ -0,0 +1,68 @@
...
+                "a8fcb0a817564696699bead7055e5a3b453cc2030b9a37714ff5a58db687e773",
+                "602386d2fbabf1ef29bf32fc601c4b575d65c0fa0d8377e075f6821afecae74c"
+            ]
+        }
+    },
+    "installer": {
+        "script": [
+           "$app_path = \"$dir\\grepWin.exe\".Replace('\\', '\\\\')",
+           "$reg_content = (Get-Content \"$dir\\grepwin-install-context.reg\") -replace '$app_path', $app_path",
+           "$reg_content = $reg_content -replace '$app_path', $app_path",
```
Ash258: Why is this happening twice??

```suggestion
```

issaclin32: I forgot to remove the duplicated codes when applying suggestions. Sorry about that.

PR #2354, opened 2019-06-24
tor-expert: Fix checkver.url, add description

rasa:
Fixes:
tor-expert: couldn't match 'torbrowser/(?<browser>[\d.]+)/tor-win32-(?<version>[\d.]+)\.zip' in https://www.torproject.org/download/download.html.en
in https://scoop.r15.ch/extras/mud-20190624-210001.log

### DIFF
--- a/bucket/tor-expert.json
+++ b/bucket/tor-expert.json
@@ -1,5 +1,6 @@
 {
     "version": "0.3.5.8",
+    "description": "Enables anonymous communication over the onion network (export mode).",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


rasa:
Installs file locally.
Both .exes are console apps, so app should really live in the main bucket.

bucket/tor-expert.json
```diff
@@ -18,5 +19,8 @@
             "url": "https://dist.torproject.org/torbrowser/$matchBrowser/sha256sums-signed-build.txt"
```
Ash258:
```suggestion
            "url": "$baseurl/sha256sums-signed-build.txt"
```

PR #2383, opened 2019-06-29
MeshLab: Add version 2016.12

issaclin32:
MeshLab is an open source tool for processing and editing 3D triangular meshes.
Homepage: http://www.meshlab.net

### DIFF
--- a/bucket/meshlab.json
+++ b/bucket/meshlab.json
@@ -1,0 +1,21 @@
+{
+    "homepage": "http://www.meshlab.net/",
+    "description": "Open source system for processing and editing 3D triangular meshes.",
+    "version": "2016.12",
+    "license": "GPL-3.0-or-later",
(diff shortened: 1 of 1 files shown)


bucket/meshlab.json
```diff
@@ -0,0 +1,21 @@
+{
+    "homepage": "http://www.meshlab.net/",
+    "description": "Open source system for processing and editing 3D triangular meshes.",
+    "version": "2016.12",
+    "license": "GPL-3.0-or-later",
```
linsui:
```suggestion
    "license": "GPL-3.0-only",
```

issaclin32: I suppose it means `GPL-3.0-or-later`. (If I misunderstood it, please tell me)

lic

linsui: Yeah, you are right :joy:

PR #2385, opened 2019-06-29
Add Notepad3 5.19.108.1602

linsui:
Close https://github.com/lukesampson/scoop-extras/issues/2384

This is a portable version.


linsui:
Maybe we can move Notepad2 and Notepad2-mod into Version?

bucket/notepad3.json
```diff
@@ -0,0 +1,47 @@
...
+        "Notepad3.exe",
+        "minipath.exe"
+    ],
+    "shortcuts": [
+        [
+            "Notepad3.exe",
+            "Notepad3"
+        ],
+        [
+            "minipath.exe",
```
Ash258: What is this shortcut for?
This sounds like helper command rather than full gui app

bucket/notepad3.json
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    "persist": [
+        "Favorites",
+        "minipath.ini",
+        "Notepad3.ini"
+    ],
+    "architecture": {
+        "64bit": {
+            "installer": {
+                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x64.zip\" \"$dir\""
```
Ash258:
```suggestion
                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x64.zip\""
```

bucket/notepad3.json
```diff
@@ -0,0 +1,47 @@
...
+            }
+        },
+        "32bit": {
+            "installer": {
+                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x86.zip\" \"$dir\""
```
Ash258:
```suggestion
                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x86.zip\""
```

bucket/notepad3.json
linsui: I don't know what it is, but it has a gui

linsui:
Since Notepad2-mod has been archived, I remove auto update.

linsui:
@Ash258 Please review.
I wonder what if there is space in `$dir`

```json
"    $notepad3 = \"$dir\\Notepad3.exe\".Replace('\\', '\\\\')",
```

bucket/notepad3.json
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+    },
+    "post_install": [
+        "Remove-Item \"$dir\\Notepad3_${version}_x*.zip\"",
+        "if (Test-Path(\"$dir\\replace-windows-notepad.reg\")) {",
+        "    $notepad3 = \"$dir\\Notepad3.exe\".Replace('\\', '\\\\')",
+        "    $content = Get-Content \"$dir\\replace-windows-notepad.reg\"",
+        "    $content = $content.Replace('$notepad3', $notepad3)",
+        "    $content | Set-Content -Path \"$dir\\replace-windows-notepad.reg\"",
```
Ash258: - No encoding
- Do not use pipe for this easy expressions

```suggestion
        "    Set-Content \"$dir\\replace-windows-notepad.reg\" $content -Encoding ASCII",
```

bucket/notepad3.json
```diff
@@ -0,0 +1,63 @@
...
+                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x86.zip\""
+            }
+        }
+    },
+    "post_install": [
+        "Remove-Item \"$dir\\Notepad3_${version}_x*.zip\"",
+        "if (Test-Path(\"$dir\\replace-windows-notepad.reg\")) {",
+        "    $notepad3 = \"$dir\\Notepad3.exe\".Replace('\\', '\\\\')",
+        "    $content = Get-Content \"$dir\\replace-windows-notepad.reg\"",
+        "    $content = $content.Replace('$notepad3', $notepad3)",
```
Ash258:
```suggestion
        "    $content = $content -replace '$notepad3', $notepad3",
```

bucket/notepad3.json
```diff
@@ -0,0 +1,63 @@
...
+        "32bit": {
+            "installer": {
+                "script": "Expand-7zipArchive \"$dir\\Notepad3_${version}_x86.zip\""
+            }
+        }
+    },
+    "post_install": [
+        "Remove-Item \"$dir\\Notepad3_${version}_x*.zip\"",
+        "if (Test-Path(\"$dir\\replace-windows-notepad.reg\")) {",
+        "    $notepad3 = \"$dir\\Notepad3.exe\".Replace('\\', '\\\\')",
```
Ash258: Use powershell syntax

```suggestion
        "    $notepad3 = \"$dir\\Notepad3.exe\" -replace '\\', '\\\\'",
```

Luckz:
FWIW there's a newer version, https://github.com/rizonesoft/Notepad3/releases/tag/RELEASE_5.19.630.2381

linsui:
Sorry, I forgot to update hash...

Ash258:
Why aren't you running checkver or let excavator do it?

linsui:
And there is another bug. It seems that `-replace` is different from `.Replace()` and accept a regex as the second arg. So `\\\\` doesn't work.

linsui:
In first arg of `-replace`, `\` need to be escape because it's a regex. But first arg of `.Replace()` is a string.

Luckz:
> This is a portable version.

Would it make sense to split this into two, a `notepad3.json` that indeed just adds a "portable" application, and a `notepad3-replace.json` that also runs the .reg files (which will require admin permissions)?
Scoop does not offer any way of making this optional with a single JSON, right?
Another option would be downloading but not executing the .reg files. That's less automation, but the lost automation requires administrator permissions anyway.

Also, the path set for the .reg should perhaps really not be the "actual" scoop app folder but the exe inside the \current\ symlink folder which always points at the newest executable. Otherwise the .reg would have to be set again on every update, which again makes little sense on paper.

linsui:
The reg file won't be run unless you run it manully. The path is set to current now.

PR #2387, opened 2019-06-30
DBVis: Add suggest java

linsui:
Close https://github.com/lukesampson/scoop-extras/issues/2244


bucket/dbvis.json
```diff
@@ -1,6 +1,8 @@
 {
-    "homepage": "https://www.dbvis.com/",
     "version": "10.0.21",
+    "homepage": "https://www.dbvis.com/",
+    "description": "A universal database tool for developers, DBAs and analysts.",
+    "license": "Proprietary",
```
Ash258: No url for proprietary license

linsui: Could you tell me what do `Proprietary` and `Freeware` mean? I can't find an explanation in doc.

bucket/dbvis.json
```diff
@@ -1,6 +1,11 @@
 {
-    "homepage": "https://www.dbvis.com/",
     "version": "10.0.21",
+    "homepage": "https://www.dbvis.com/",
+    "description": "A universal database tool for developers, DBAs and analysts.",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://www.dbvis.com/purchase/"
```
Ash258:
```suggestion
        "url": "https://www.dbvis.com/eula/"
```

bucket/dbvis.json
Ash258: https://en.wikipedia.org/wiki/Freeware

> There is no agreed-upon set of rights, license, or EULA that defines freeware unambiguously

https://en.wikipedia.org/wiki/Software_license#Proprietary_software_licenses

> The hallmark of proprietary software licenses is that the software publisher grants the use of one or more copies of software under the end-user license agreement (EULA), but ownership of those copies remains with the software publisher (hence use of the term "proprietary"). This feature of proprietary software licenses means that certain rights regarding the software are reserved by the software publisher. Therefore, it is typical of EULAs to include terms which define the uses of the software, such as the number of installations allowed or the terms of distribution.

linsui: I got. Freeware does not means free in freedom. Thanks!

PR #2395, opened 2019-07-01
euler: Add version 2019-06-27

niheaven:
Its External Editor needs Java

### DIFF
--- a/bucket/euler.json
+++ b/bucket/euler.json
@@ -15,22 +15,22 @@
                     "   New-Item -Path \"$dir\\.euler.profile\" -ItemType 'File' | Out-Null",
                     "}"
                 ]
-            }
+            },
+            "bin": [
+                "euler64.exe",
+                [
+                    "euler64.exe",
+                    "euler"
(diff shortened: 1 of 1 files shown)


bucket/euler.json
```diff
@@ -0,0 +1,53 @@
...
+                    "Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir '{sys}' -Removal",
+                    "if (!(Test-Path \"$persist_dir\\.euler.profile\")) {",
+                    "   New-Item -Path \"$dir\\.euler.profile\" -ItemType 'File' | Out-Null",
+                    "}"
+                ]
+            }
+        }
+    },
+    "bin": [
+        "euler64.exe",
```
Ash258: You have architecture property and still 64bit shortcuts and binaries are out of it.

niheaven: Okay, move into arch.

The behavior should be the same, since 32bit platform will be blocked by 64bit only url.

PR #2402, opened 2019-07-03
Notepad2-mod: persist settings

run-ze:


bucket/notepad2-mod.json
```diff
@@ -19,6 +19,9 @@
             "Notepad2-mod"
         ]
     ],
+    "persist": [
```
Ash258: Does this file exists in downloaded archive??
```suggestion
    "persist": "Notepad2.ini",
```

run-ze: yes
image

PR #2328, opened 2019-06-20
filezilla: more portable

wzv5:

PR #2425, opened 2019-07-06
filezilla: Configure and persist config directory

xfournet:
- Closes #3059

The way filezilla configuration is persisted is not working out of the box because the provided fzdefault.xml template has flaws:
* configuration directory is set to an undefined location because usage of %SOMEDIR% variable
* filezilla resources cache (icons) is done into user "Documents" directory which is weird
* hardlinking of editable files like fzdefault.xml is very fragile: copy/paste, VCS and many text editor can break the hardlink, leading to 2 separate file and defeating the scoop persist approach. This should be avoided when possible.

I'm also not convainced the way #2328 change it:
* cache directory is a very internal thing of filezilla and doesn't deserve to be in persist directory by default
* so you don't need the extra config directory under persist\filezilla
* fzdefault.xml is still hardlinked and no more needed into persist\filezilla since now there is the persist/filezilla/config/filezilla.xml file that is able to persist all the configuration

In order to avoid all theses problems, I propose to configure fzdefault.xml only in the app directory with just the configuration location set to the persit/filezilla directory and disable autoupdate since it's should be done by scoop
* filezilla is able to create the directory when not existing and will populate it with a default filezilla.xml file and non-configuration files (layout.xml, queues.sqlite3)
* by default cache directory is nicely set into %USERPROFILE%\AppData\Local\FileZilla . If you really want to change it you can edit it in persist/filezilla/filezilla.xml, no need to edit the fzdefault.xml for that (but really i don't why this would be needed in first place)


Ash258:
#2328

bucket/filezilla.json
```diff
@@ -27,7 +32,6 @@
             "FileZilla"
         ]
     ],
-    "persist": "fzdefaults.xml",
```
Ash258: You do not persist anything. Which is nonsense.

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
-        "}"
+        "@\"",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\" ?>",
+        "<FileZilla3>",
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">$persist_dir</Setting>",
+        "    <Setting name=\"Disable update check\">1</Setting>",
+        "  </Settings>",
+        "</FileZilla3>",
+        "\"@ | Out-File -Encoding utf8 \"$dir\\fzdefaults.xml\""
```
Ash258: UTF8 is UTF8-with-bom in powershell 5

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
     "pre_install": [
-        "$file = 'fzdefaults.xml'",
-        "if (-not (Test-Path \"$persist_dir\\$file\")) {",
-        "    Copy-Item \"$dir\\docs\\fzdefaults.xml.example\" \"$dir\\fzdefaults.xml\"",
-        "}"
+        "@\"",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\" ?>",
+        "<FileZilla3>",
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">$persist_dir</Setting>",
```
Ash258: Create subfolder and then persist this folder

bucket/filezilla.json
xfournet: ok, i added the config directory

bucket/filezilla.json
xfournet: thanks i was not aware of that. Filezilla support UTF8 with BOM so it was not a problem. Anyway I changed to ascii (no UTF-8 with no BOM in PowerShell 5) since now Config location is assured to be ascii only, and so the entire fzdefault.xml file

bucket/filezilla.json
xfournet: Yes, done

xfournet:
Changed to introduce the extra folder as required

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
     "pre_install": [
-        "$file = 'fzdefaults.xml'",
-        "if (-not (Test-Path \"$persist_dir\\$file\")) {",
-        "    Copy-Item \"$dir\\docs\\fzdefaults.xml.example\" \"$dir\\fzdefaults.xml\"",
-        "}"
+        "@\"",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\" ?>",
+        "<FileZilla3>",
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">config</Setting>",
```
Ash258: Why it is not fullpath?

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
     },
     "extract_dir": "FileZilla-3.47.1",
     "pre_install": [
-        "$file = 'fzdefaults.xml'",
-        "if (-not (Test-Path \"$persist_dir\\$file\")) {",
-        "    Copy-Item \"$dir\\docs\\fzdefaults.xml.example\" \"$dir\\fzdefaults.xml\"",
-        "}"
+        "@\"",
```
Ash258: You do not need double quotes when no variable is used

bucket/filezilla.json
```diff
@@ -27,7 +32,7 @@
             "FileZilla"
         ]
     ],
-    "persist": "fzdefaults.xml",
+    "persist": "config",
```
Ash258: Can user change fzdrfaults.xml from UI?

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
-        "}"
+        "@\"",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\" ?>",
+        "<FileZilla3>",
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">config</Setting>",
+        "    <Setting name=\"Disable update check\">1</Setting>",
+        "  </Settings>",
+        "</FileZilla3>",
+        "\"@ | Out-File -Encoding ascii \"$dir\\fzdefaults.xml\""
```
Ash258:
```suggestion
        "\"@ | Out-File \"$dir\\fzdefaults.xml\" -Encoding Ascii"
``

File: bucket/filezilla.json (hunk 6 above)
Comment (Ash258) [reply bucket/filezilla.json]: What will config directory contain?

File: bucket/filezilla.json (hunk 4 above)
Comment (xfournet) [reply bucket/filezilla.json]: full path is not needed, it's relative to the installation directory. Could add it if you want ?

File: bucket/filezilla.json (hunk 5 above)
Comment (xfournet) [reply bucket/filezilla.json]: it depend if i need to add the fullpath from the previous comment. let me know.

File: bucket/filezilla.json (hunk 6 above)
Comment (xfournet) [reply bucket/filezilla.json]: it contains the configuration files of filezilla (filezilla.xml, layout.xml, etc..). It's initialized by filezilla when no files exists at the first start.

Comment (xfournet) [reply bucket/filezilla.json]: fzdefault.xml is not modifiable from the UI, the UI modify filezilla.xml file.

**Comment by xfournet** [comment]
#3025 is about filezilla-server which is not the same than filezilla which a client
#3059 look like much complicated and it let two redundant configuration file in the persistent directory, one which is not modifiable by the user in anyway + the one that can be changed by the user

IMHO my PR is more aligned in the way Filezilla configuration is designed (unmodifiable default in app dir, config in persist dir)

File: bucket/filezilla.json (hunk 4 above)
Comment (xfournet) [reply bucket/filezilla.json]: fullpath added

File: bucket/filezilla.json (hunk 5 above)
Comment (xfournet) [reply bucket/filezilla.json]: since i added the fullpath, the $dir variable is now used

Comment (Ash258) [inline bucket/filezilla.json]:
```suggestion
        "@'",
```

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
     "pre_install": [
-        "$file = 'fzdefaults.xml'",
-        "if (-not (Test-Path \"$persist_dir\\$file\")) {",
-        "    Copy-Item \"$dir\\docs\\fzdefaults.xml.example\" \"$dir\\fzdefaults.xml\"",
-        "}"
+        "@\"",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\" ?>",
+        "<FileZilla3>",
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">$dir\\config</Setting>",
```
Ash258: I do not understand your mind flow at all ... You say it is not needed to use fullpath, but you still change it...

```suggestion
        "    <Setting name=\"Config Location\">config</Setting>",
```

bucket/filezilla.json
```diff
@@ -15,10 +15,15 @@
...
+        "    <Setting name=\"Disable update check\">1</Setting>",
+        "  </Settings>",
+        "</FileZilla3>",
+        "\"@ | Out-File \"$dir\\fzdefaults.xml\" -Encoding Ascii"
```
Ash258:
```suggestion
        "'@ | Out-File \"$dir\\fzdefaults.xml\" -Encoding Ascii"
```

bucket/filezilla.json
xfournet: Both are working. I was thinking you were preferring the absolute path. Sorry for the confusion

Ash258:
This will not be backward compatible right now. There is no handling of copying current configs into new `config` dir. So as soon as users update they will loose all settings and they will start with empty `config` directory

Where is config located with normal manifest installation? Move all files from curernt config into new `config` dir in pre_install

xfournet:
Since current manifest configuration is broken (it references an undefined variable called "SOMEDIR") the configuration is located in a filezilla directory in the drive root of the persist directory (ie with non customized scoop installation it will be in C:\filezilla directory).

I added the script to move the configuration files in the persist directory. In case the fzdefault.xml file was modified by hand, it's renamed to avoid data loss.

bucket/filezilla.json
```diff
@@ -15,9 +15,31 @@
...
+        "  <Settings>",
+        "    <Setting name=\"Config Location\">config</Setting>",
+        "    <Setting name=\"Disable update check\">1</Setting>",
+        "  </Settings>",
+        "</FileZilla3>",
+        "'@ | Out-File \"$dir\\fzdefaults.xml\" -Encoding Ascii",
+        "",
+        "$oldfzdefault = \"$persist_dir\\fzdefaults.xml\"",
+        "if (Test-Path \"$oldfzdefault\") {",
+        "    Write-Host 'Fixing existing configuration...' -f Yellow",
```
Ash258:
```suggestion
        "    Write-Host 'Adopting new persist configuration...' -f Yellow",
```

bucket/filezilla.json
```diff
@@ -15,9 +15,31 @@
...
+        "    $oldConfigDir = \"$((Get-Item \"$persist_dir\").PSDrive.Root)filezilla\"",
```
Ash258: This cannot be correct

My scoop folder is `S:\SCOOP`, filezilla installed in `S:\SCOOP\apps\filezilla`, persist in `S:\SCOOP\persist`

Filezilla configuration from old config is in C:\filezilla

$env:SystemDrive

bucket/filezilla.json
```diff
@@ -15,9 +15,31 @@
...
+        "    $newConfigDir = \"$persist_dir\\config\"",
+        "    if (Test-Path \"$oldConfigDir\") {",
+        "        Write-Host \"Moving existing configuration from '$oldConfigDir' to '$newConfigDir'\" -f Yellow",
+        "        Move-Item \"$oldConfigDir\" \"$newConfigDir\"",
```
Ash258: This will move folder and put it in folder.

Resulting in `..persist\filezilla\config\filezilla\<configuration files>`

```suggestion
        "        Move-Item \"$oldConfigDir\\*\" \"$newConfigDir\"",
```

bucket/filezilla.json
```diff
@@ -15,9 +15,31 @@
...
+        "    }",
+        "    if (Compare-Object $(Get-Content \"$oldfzdefault\") $(Get-Content \"$dir\\docs\\fzdefaults.xml.example\")) {",
```
Ash258: - Why do you need `compare-object`?
	- Only to make sure that user did not changed it?
- Subshells are not needed

bucket/filezilla.json
xfournet: If you customize the scoop directory the configuration is under the corresponding drive. Eg mine is customized to D:\scoop so the old configuration is under D:\filezilla. For you the old filezilla configuration should be under S:\filezilla. So we need to relate to the drive of the persist dir (since it's relative to the old fzdefault.xml which was under persist).
I tested it and it's correct.

bucket/filezilla.json
xfournet: The destination directory should not exist, in that case the move is correct (i tested it...)
The proposed fix don't work since this destination directory doesn't exists.
In case the destination directory already exists (eg a user has created it manually the 'config' by it self, whatever the reason) i will add a check to handle that

bucket/filezilla.json
xfournet: Yes just to be sure to not delete a file that would have been manually edited by the user. I will look to remove the subshells.

Ash258: Then do not delete the file in any use case.

Just rename it all the time

PR #2429, opened 2019-07-09
vlc: Update download URLs.

nirvdrum:
I don't track VLC's mirror system, so I don't know if there was a wider change that didn't get picked up by Scoop or if it's just something that changed with newer releases of VLC. But, without this patch I'm unable to install VLC.

### DIFF
--- a/bucket/vlc.json
+++ b/bucket/vlc.json
@@ -11,12 +11,12 @@
     ],
     "architecture": {
         "64bit": {
-            "url": "https://download.videolan.org/pub/vlc/3.0.7.1/win64/vlc-3.0.7.1-win64.7z",
+            "url": "https://get.videolan.org/vlc/3.0.7.1/win64/vlc-3.0.7.1-win64.7z",
             "hash": "4860310edad634fa058c40633a5b80310085d2ab13aad91e485885a45c90890c",
             "extract_dir": "vlc-3.0.7.1"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


Ash258:
I can use old urls without any problem.

A

B

nirvdrum:
Hmm... thanks for that info. I guess I have a DNS issue somewhere then. When I saw there was an issue, I went to VLC's site and saw all the URLs switched to the "get" host and the new host worked, so I didn't think to debug other network services.

Ash258:
It could be switched there is no problem if it is used in official webpage.

Do you know when they started using these new URLs (approximately)?

nirvdrum:
Sorry, I have no idea. It's never anything I used to pay attention to. It only became problematic for me when trying to upgrade VLC. Scoop had successfully installed an older version for me, but couldn't upgrade. I'm happy to attribute this to an environment error though. I'll have to try a different DNS config this evening and see if the problem clears up.

Ash258:
I tried it from 4 location which I have remote access right now and everywhere it works. (Switzerland, Germany, Czech and Romania)

I will approve this change as it is working with message to other collaborators to test before merging and waiting if there will be some issues opened regarding this.

Ash258:
There is only one problem with new URLs. They are just URLs which leads to mirror. This could potentionally lead into https => http redirects and break installation as it is not possible in powershell.

B

A

PR #2431, opened 2019-07-10
prime95: Fix checkver & au.*

rasa:
Tested and working locally.
Fixes:
```
URL https://www.mersenne.org/ftp_root/gimps/p95v298b5.win32.zip is not valid
Could not find hash!
ERROR Could not update prime95 32bit

### DIFF
--- a/bucket/prime95.json
+++ b/bucket/prime95.json
@@ -25,20 +25,24 @@
     ],
     "checkver": {
         "url": "https://www.mersenne.org/download/",
-        "regex": "Windows:\\s+64-bit</td><td>([^<]+)</td>"
+        "regex": "Windows:\\s+64-bit\\D+(?<version>[\\w.]+).*href=\"https?://www.mersenne.org/ftp_root/gimps/(?<file64>[^.]+).win64.zip\".*Windows:\\s+32-bit\\D+(?<version32>[\\w.]+).*href=\"https?://www.mersenne.org/ftp_root/gimps/(?<file32>[^.]+).win32.zip\""
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://www.mersenne.org/ftp_root/gimps/p95v$cleanVersion.win64.zip"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by niheaven** [comment]
Is it necessary to use different version for 64bit and 32bit? The stable version here should be 0.9.3, right?

**Comment by r15ch13** [comment]
@niheaven because they are release with separate versions (https://www.mersenne.org/download/)
@rasa please change the update URLs back to HTTPS.

**Comment by niheaven** [comment]
I mean why not wait for developer update all build to 29.8b3+? If 29.8b5 of 64bit is just bug-fixed, a temporal manual fix is enough. The website said latest version is 29.8b3 and all excepts windows 64bit is 29.8b3, I don't think this is a normal case.

File: bucket/prime95.json [hunk 1]
```diff
@@ -25,20 +25,24 @@
...
+        "regex": "Windows:\\s+64-bit\\D+(?<version>[\\w.]+).*href=\"https?://www.mersenne.org/ftp_root/gimps/(?<file64>[^.]+).win64.zip\".*Windows:\\s+32-bit\\D+(?<version32>[\\w.]+).*href=\"https?://www.mersenne.org/ftp_root/gimps/(?<file32>[^.]+).win32.zip\""
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://www.mersenne.org/ftp_root/gimps/p95v$cleanVersion.win64.zip"
+                "url": "https://www.mersenne.org/ftp_root/gimps/$matchFile64.win64.zip",
+                "hash": {
+                    "url": "https://www.mersenne.org/download/",
+                    "find": "\\.win64\\.zip[^:]+SHA256:<br>$sha256"
```
niheaven:
```suggestion
                    "find": "$basename.*?$sha256"
```

rasa:
@niheaven It’s a good question. While we can fix it to work in the way you suggest, I think we need to defer to the publisher, and use whatever they show as the latest (stable) version, as we can’t assume the previous version (in this case: 29.8b3 64bit) will remain downloadable.

PR #2441, opened 2019-07-12
Slack: checkver

linsui:
Slack'version is changing because the `RELEASES` file is changing. 
https://github.com/lukesampson/scoop-extras/commits/f933635c661b882ef73d471041f3407dff0f9a03/bucket/slack.json


bucket/slack.json
```diff
@@ -20,11 +21,7 @@
             "Slack"
         ]
     ],
-    "checkver": {
-        "url": "https://downloads.slack-edge.com/releases_x64/RELEASES",
-        "re": "slack-([\\d.]+)-full.nupkg",
-        "reverse": true
-    },
+    "checkver": "Version ([\\d.]+)",
```
Ash258: I would rather see getting it from https://slack.com/intl/en/release-notes/windows with `Slack ([\d.]+)`

linsui: Ok.

PR #2446, opened 2019-07-12
Updated setting.json in .vscode folder

greg-ruminski:
Issue with formatting on save in VSCode fixed upstream


Ash258:
It was fixed, but since then there was no release. So remove it is not good idea until new versino of extension is released.

Issue #2308, opened 2019-06-19
Outstanding Excavator issues


rasa:
Here's a list of the outstanding Excavator issues found in https://scoop.r15.ch/extras/mud-20200203-010001.log at [https://scoop.r15.ch/extras/](https://scoop.r15.ch/extras/?sort=time&order=desc):

## Needing triage (newest to oldest):

None.

## Fixable:

1. [darktable](../blob/master/bucket/darktable.json): The [.exe](https://github.com/darktable-org/darktable/releases/download/release-2.6.2/darktable-2.6.2-win64.exe) appears to be an NSIS archive, but running it, causes a UAC popup. Extracting it with 7zip, gives: `Open ERROR: Can not open the file as [Nsis] archive`. See https://github.com/lukesampson/scoop-extras/issues/1831

## Being fixed:

1. qbittorrent-portable: The operation has timed out. URL https://www.qbittorrent.org/download.php is not valid. Fixed in #3558. Thanks @rasa!
1. pureref: 1.10.0 (scoop version is 1.9.2) autoupdate available. ERROR Could not update pureref 32bit. Fixed in #3556. Thanks @rasa!
1. lxrunoffline: 3.4.1 (scoop version is 3.4.0) autoupdate available. Fixed in #3555. Thanks @rasa!
1. gb-studio: 1.2.1 (scoop version is 1.2.0) autoupdate available. ERROR Could not update gb-studio 32bit. Fixed in #3554. Thanks @rasa!
1. vidi: 0.2.5.6 (scoop version is 0.2.5.5). Fixed in #3553. Thanks @rasa!
1. gnucash: 3.8 (scoop version is 3.7) autoupdate available. Fixed in #3552. Thanks @rasa!
1. [chromium-dev](../blob/master/bucket/chromium-dev.json): couldn't match `'v([\d.]+-r(?:\d+))-win64'` in https://chromium.woolyss.com/api/v3/?os=windows&bit=64&type=dev-codecs-sync&out=string. See https://github.com/lukesampson/scoop-extras/pull/2759#issue-311417571. See also #2885. Fixed in #3476. Thanks @linsui!
[...]

PR #2455, opened 2019-07-14
praat: Fix checkver and autoupdate

issaclin32:
mentioned in **Outstanding Excavator issues** (#2308)

The file name for praat version 6.1 is **praat6100_win64.zip** rather than **praat61_win64.zip**.
This fixes the problem by matching file name using regex.

### DIFF
--- a/bucket/praat.json
+++ b/bucket/praat.json
@@ -21,7 +21,7 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/praat/praat/releases",
+        "github": "https://github.com/praat/praat",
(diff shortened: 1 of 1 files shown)


bucket/praat.json
```diff
@@ -21,15 +21,16 @@
         ]
     ],
     "checkver": {
-        "github": "https://github.com/praat/praat"
+        "url": "https://github.com/praat/praat/releases",
```
Ash258: I do not see a reason to prefer url of all releaes instead of latest

```suggestion
        "github": "https://github.com/praat/praat",
```

PR #2456, opened 2019-07-14
onefetch: Fix checkver

issaclin32:
mentioned in Outstanding Excavator issues (#2308).

The problem is caused by that only some of the release versions has a Windows version of it (v1.5.4 has Windows version, but v1.5.5 does not) (see: https://github.com/o2sh/onefetch/releases)

This fixes the problem by manually matching "release files with Windows version" with regex.


Ash258:
What is the reason why there are no windows binaries in some releases?

PR #2458, opened 2019-07-15
uget: Update to 2.2.2 , and add au.hash

linsui:
The official websit haven't update

Note: There is a locale package. After https://github.com/lukesampson/scoop/pull/3518 is merged we could add it.

### DIFF
--- a/bucket/uget-portable.json
+++ b/bucket/uget-portable.json
@@ -1,10 +1,10 @@
 {
-    "version": "2.2.1",
+    "version": "2.2.2",
     "description": "Open source download manager.",
     "homepage": "https://ugetdm.com/",
     "license": "LGPL-2.1-only",
-    "url": "https://downloads.sourceforge.net/project/urlget/uget%20%28stable%29/2.2.1/uget-2.2.1-win32%2Bgtk3.7z",
-    "hash": "a722bd02a6d77ac24b8bb7d1b3a6c7f37ca4dbab40859264b1e50dedbcda584a",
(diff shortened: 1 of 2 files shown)


bucket/uget-portable.json
```diff
@@ -15,10 +15,14 @@
...
     "checkver": {
-        "url": "https://ugetdm.com/downloads/windows/",
-        "regex": "Latest Stable Version: ([\\d\\.]+)"
+        "url": "https://sourceforge.net/projects/urlget/rss?path=/uget%20%28stable%29",
+        "regex": "/uget-([\\d.]+)-win32\\+gtk3\\.7z"
     },
     "autoupdate": {
-        "url": "https://downloads.sourceforge.net/project/urlget/uget%20%28stable%29/$version/uget-$version-win32%2Bgtk3.7z"
+        "url": "https://downloads.sourceforge.net/project/urlget/uget%20%28stable%29/$version/uget-$version-win32%2Bgtk3.7z",
+        "hash": {
```
Ash258: What is wrong with normal sourceforge mode?

linsui: I guess because the `+` in filename, `$basename` can't work. By the way, why not use `+` directly in url?

niheaven: Please try it, should work. (sf is blocked in my office...)

Ash258: Because `+` is reserved entity. It is used to concat string.

`https://....?scriptparams_=-k+PRTGAIH>` will result in `scriptsparams_` property contains string `-k PRTGAIH`

linsui: @Ash258 Thanks. 
@niheaven It does not work, as you can see, the hash is sha256 before.

PR #2461, opened 2019-07-15
mp3tag: fix hash

hsiktas:
Seems like the devs uploaded a new file without changing the version.


TheZoker:
@r15ch13 can you please merge this? The package can't be installed right now

PR #2468, opened 2019-07-16
Wire: Add 3.9.2928

linsui:


bucket/wire.json
```diff
@@ -0,0 +1,26 @@
...
+    "extract_dir": "lib\\net45",
+    "bin": "Wire.exe",
+    "shortcuts": [
+        [
+            "Wire.exe",
+            "Wire"
+        ]
+    ],
+    "checkver": {
+        "url": "https://wire-app.wire.com/win/prod/RELEASES",
```
Ash258: Are you sure you want to check version fro this? I would try to avoid this after recent Slack updates

linsui: Well... I thought it's not so bad. Ok, I'll change it to github.

PR #2471, opened 2019-07-17
Tor-expert: Move to Main

linsui:
https://github.com/ScoopInstaller/Main/pull/249


rasa:
Waiting on https://github.com/ScoopInstaller/Main/pull/249 to be unblocked.

PR #2481, opened 2019-07-19
smartsynchronize: Fix checkver and force update to 4.0.0

niheaven:


bucket/smartsynchronize.json
```diff
@@ -27,16 +27,16 @@
     },
     "persist": ".settings",
     "checkver": {
-        "url": "https://www.syntevo.com/smartsynchronize/changelog.txt",
-        "regex": "([\\d.]+)"
+        "url": "https://www.syntevo.com/smartsynchronize/download/",
+        "regex": "smartsynchronize-portable-([\\d.]+).7z"
```
Ash258:
```suggestion
        "regex": "smartsynchronize-portable-([\\d.]+)\\.7z"
```

niheaven: In this situation, `.` is enough, there should be one character `.`, so regex `.` will match it. Of couse, `\.` could be used, too.

Ash258: There is no need to expect anything when you know there will always be dot literally.

r15ch13:
`(?sm)` is another neat regex trick I didn't know about. Thanks :smile:

PR #2487, opened 2019-07-20
Freecad: remove 32-bit, use portable version

issaclin32:
mentioned in #2308 (Outstanding Excavator Issues).

**FreeCAD** does not provide 32-bit binaries in newer versions, therefore causing the issue.

Notes:
* Remove 32-bit binaries.
* Use **portable version**(.7z) instead of extracting files from the installer(.msi).
* FreeCAD provides checksum files ([link](https://github.com/FreeCAD/FreeCAD/releases/download/0.18.3/FreeCAD-0.18.16131.3129ae4-WIN-x64-portable.7z-SHA256.txt)), but there are **line breaks**(\r\n) in the middle of the hash string.

### DIFF
--- a/bucket/freecad.json
+++ b/bucket/freecad.json
@@ -1,16 +1,12 @@
 {
-    "version": "0.18.2",
+    "version": "0.18.3",
     "description": "Free and open-source general-purpose parametric 3D CAD modeler and a building information modeling software with finite-element-method support.",
     "homepage": "https://www.freecadweb.org/",
     "license": "LGPL-2.0-or-later",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


niheaven:
Please wait for their 32bit installer, and don't just remove `32bit`. They just don't provide 32bit portable build. Even their 0.19_pre has 32bit build, so just wait.

https://github.com/FreeCAD/FreeCAD/tree/0.18.3

> NOTE: binaries will be posted when available
> Windows
> Installers are available.
> The Win x64 installer uses Conda as a package manager, The binaries use PY3.6, QT5, and VS2015+
> The Win x32 installer uses Libpack 11.11. The binaries use PY2.7, QT4, and VS2013

issaclin32:
> Please wait for their 32bit installer, and don't just remove `32bit`. They just don't provide 32bit portable build. Even their 0.19_pre has 32bit build, so just wait.

Thanks for the information. Closing PR now.

PR #2489, opened 2019-07-20
Thunderbird portable: Add version 60.8.0

linsui:


niheaven:
Making current `thunderbird` portable should be better than adding new one.

Ash258:
@niheaven I do not agree. Since it is set by parameter there is possible that thunderbird will start without this parameter (executed from browser for example), which then will use normal profile from appdata

Ash258:
General configuration is done from current folder, where is simple executable without parameter. No one configure this from start menu / shims folder.

linsui:
@niheaven I don't make current thunderbird portable because only the profile can be portable. Thunderbird has a profile manager and the arg just surpass the manager. The standard version can have several profile but portable version can only use one profile.

rasa:
Question for all: should both packages have a note alerting the user to the other package, and the difference between them?

linsui:
Do you mean a note such as `The portable version can only work with one profile. If you want to work with multiple profiles, please use the stantard version.` ?

PR #2497, opened 2019-07-23
Add 319 missing descriptions

hoilc:


bucket/vcredist.json
```diff
@@ -1,5 +1,6 @@
 {
     "homepage": "https://www.visualstudio.com/downloads/",
+    "description": "Microsoft Visual C++ Redistributable for Visual Studio 2008/2010/2012/2013/2019.",
```
rasa: Why not also list 2015 and 2017?

niheaven: Hope not late.

This should be `Microsoft Visual C++ Redistributable for Visual Studio 2008/2010/2012/2013/2015-2019`

2015, 2017 and 2019 are all VC14, and 2019 can replace 2015 and 2017.

PR #2512, opened 2019-07-25
Fix ~195 license tags

linsui:
Accidently force-push before commit ...


linsui:
I didn't change those `Freeware` without a license. It could be changed to `Non-license` easily. What's you suggestion?

rasa:
I removed “non-licensed” from the docs. Let’s just go with “Unknown” if it appears the app doesn’t have a license.

linsui:
OK, I'll change those freeware to unknown and apply new rule.

Luckz:
> Accidently force-push before commit ...

is that what necessitated the `git reset origin/master --hard`?

r15ch13:
@Luckz no, that is related to https://github.com/lukesampson/scoop-extras/pull/2352
The .reg file where added with the wrong line ending. This PR fixed them.

_Sent from my OnePlus5T using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #2544, opened 2019-08-01
Add Ren'Py 7.3.2

MunsterPlop:


bucket/renpy.json
```diff
@@ -0,0 +1,28 @@
+{
+    "homepage": "https://www.renpy.org/",
+    "description": "Popular open source visual novel engine that uses Python for scripting",
+    "license": "MIT",
+    "version": "7.3.2",
+    "url": "https://www.renpy.org/dl/7.3.2/renpy-7.3.2-sdk.7z.exe#dl.7z",
```
Ash258:
```suggestion
    "url": "https://www.renpy.org/dl/7.3.2/renpy-7.3.2-sdk.7z.exe#/dl.7z",
```

bucket/renpy.json
```diff
@@ -0,0 +1,28 @@
...
+    "bin": "renpy.exe",
+    "shortcuts": [
+        [
+            "renpy.exe",
+            "RenPy"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.renpy.org/release_list.html",
+        "regex": "Ren'Py ([\\d\\.]+)"
```
Ash258:
```suggestion
        "regex": "Ren'Py ([\\d.]+)"
```

bucket/renpy.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "autoupdate": {
+        "url": "https://www.renpy.org/dl/$version/renpy-$version-sdk.7z.exe#dl.7z",
```
Ash258:
```suggestion
        "url": "https://www.renpy.org/dl/$version/renpy-$version-sdk.7z.exe#/dl.7z",
```

bucket/renpy.json
```diff
@@ -0,0 +1,28 @@
...
+        "hash": {
+            "url": "https://www.renpy.org/dl/$version/checksums.txt",
```
Ash258:
```suggestion
            "url": "$baseurl/checksums.txt",
```

bucket/renpy.json
```diff
@@ -0,0 +1,28 @@
...
+            "regex": "$sha256.*?$basename"
```
Ash258: I think this could be removed. But needs to be tested.

```suggestion
```

MunsterPlop: I just tested it and it works fine but returns the md5 checksum by default. 
If that's okay with you, I'm commiting this.

Ash258: Ok then go with this. No need to use greedy match all when it is only whitespace.

```suggestion
            "regex": "$sha256\\s+$basename"
```

r15ch13:
Even if it's a tool to create games, I would suggest adding it to the games bucket.
https://github.com/Calinou/scoop-games

MunsterPlop:
> 
> 
> Even if it's a tool to create games, I would suggest adding it to the games bucket.
> https://github.com/Calinou/scoop-games

I got nothing against that but some game engines such as Godot are already in the extras bucket, so I'd ask @Ash258 for some input on this.

Ash258:
I do not mind it here. Since cheat engine is here for example too.

PR #2566, opened 2019-08-04
Add and fix some au.hash

linsui:


Ash258:
Really?? Again this huge PR? 

You know how time consuming is to reveiew this and then each merging from master after update? Completely not worth and mess.

Learn how to user `hub` for posting PRs from command line and post PR for each manfiest.

linsui:
I'll have a try. By the way, could chocolatey be used to get hash?

niheaven:
Of course you can, see https://github.com/lukesampson/scoop/pull/3249 for some attemp. But make sure the files are unique if you use Choco's checksum with something downloaded elsewhere.

linsui:
@niheaven I mean when scoop update a manifest maybe choco haven't update its. Dose it make sense to get hash from choco? Or we should also checkver from choco?

niheaven:
Use checkver from Choco, too. See ScriptCs for example. If the PR is merged, using `choco` in checkver and au are more preferred.

PR #2592, opened 2019-08-08
Archi: Update to 4.5.1 & remove 32bit

hoilc:
https://www.archimatetool.com/downloads/What's%20New%20in%20Archi.pdf

Starting from 4.5, Archi supports only 64-bit versions.

#2308

### DIFF
--- a/bucket/archi.json
+++ b/bucket/archi.json
@@ -21,7 +21,6 @@
         "url": "https://www.archimatetool.com/download/",
         "re": "data-version=\"([\\d.]+)\""
     },
-
     "autoupdate": {
         "architecture": {
             "64bit": {
--- a/bucket/archi.json
+++ b/bucket/archi.json
@@ -9,11 +9,19 @@
             "hash": "md5:63d78b2d8ef298b0ed111691417d1339"
         }
     },
+    "extract_dir": "Archi",
+    "bin": "archi.exe",
+    "shortcuts": [
+        [
+            "archi.exe",
+            "Archi"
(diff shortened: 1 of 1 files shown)


bucket/archi.json
```diff
@@ -2,41 +2,35 @@
...
+            "archi.exe",
+            "Archi"
+        ]
+    ],
     "checkver": {
         "url": "https://www.archimatetool.com/download/",
         "re": "data-version=\"([\\d.]+)\""
     },
-    "extract_dir": "Archi",
+
```
Ash258: Empty line
```suggestion
```

bucket/archi.json
```diff
@@ -2,41 +2,35 @@
     "homepage": "https://www.archimatetool.com/",
     "license": "MIT",
     "description": "Archi® is a free, open source, cross-platform tool and editor to create ArchiMate models.",
```
Ash258:
```suggestion
    "description": "Free, open source, cross-platform tool and editor to create ArchiMate models.",
```

hoilc:
> Move `extract_dir`, `bin` , `shortcuts` properties after `architecture`. (preserve order which I specified)

I was also suggested to put `homepage` and `description` at the top in another PR.

Maybe a tool that adjusts the field order automatically is needed.😂

Ash258:
It would be nice, but it is hard to create such. Automatic deterministic properties sorting needs to cover lots of combinations especially nested sorting of architecture property.

PR #2596, opened 2019-08-08
DBGL: fix checkver

issaclin32:
mentioned in #2308 (Outstanding Excavator issues).

### DIFF
--- a/bucket/dbgl.json
+++ b/bucket/dbgl.json
@@ -20,7 +20,7 @@
             "DOSBox Game Launcher"
         ]
     ],
-    "checkver": "v([\\d.]+)\\s+<sup",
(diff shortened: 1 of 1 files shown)


issaclin32:
> The checkver appears to be working, the file download of the new version is not:
> 

The URL for **v0.90** (`http://members.quicknet.nl/blankendaalr/dbgl/download/dbgl082.zip`) works though.

It looks like that the developer of **DBGL** released v0.90 (dependent on Java 8 to 12) and v0.83 (dependent on Java 7 to 8) at the same time, and put v0.83 in a different download location.


This PR modifies `checkver` from matching the blue box instead of the red box, so that `checkver` recognizes v0.90 instead of v0.83.

checkver

Issue #2384, opened 2019-06-29
[Request] Notepad3 - with the option to replace the default notepad


pedrogarcia:
Notepad3 is a fork of notepad2-mod, but it is still actively developed.
Altough scoop apps are generaly portable, I would like this one to use the default installer that allows me to replace windows default notepad with notepad3. 

Their git repo is here: https://github.com/rizonesoft/Notepad3

PR #2598, opened 2019-08-08
notepad3: Cleanly replace Windows notepad

phgmacedo:
This is how the original installer does it.

The replace-windows-notepad.reg file was not working how I expected.


pedrogarcia:
I have fixed the error that prevented the commit from passing the checks. I believe that this is a better solution than it currently exists, or at least it was what I requested in #2384 and proposed in #2385.
Instead of creating a new context menu option, the PR uses a Windows feature called [Image File Execution Options](https://blogs.msdn.microsoft.com/junfeng/2004/04/28/image-file-execution-options/) to redirect Notepad.exe into notepad3.exe without touching any system files. 
 @linsui @rasa @Ash258

Luckz:
As long as nobody ever force-executes the .reg upon simple installation, this is indeed a better way to handle the .reg aspect.

pedrogarcia:
> As long as nobody ever force-executes the .reg upon simple installation, this is indeed a better way to handle the .reg aspect.

It requires admin, if that's what you mean.

linsui:
I don't know much about it... Don't forget to update hash in notepad3.json.

PR #2602, opened 2019-08-09
bulk-rename-utility: Add version 3.1.0.0

### LINKED ISSUE #2601: Can Bulk Rename Utility be added?
It is free for personal use.

https://www.bulkrenameutility.co.uk/Main_Intro.php

niheaven:
Close #2601

### DIFF
--- a/bucket/bulk-rename-utility.json
+++ b/bucket/bulk-rename-utility.json
@@ -1,0 +1,48 @@
+{
+    "homepage": "https://www.bulkrenameutility.co.uk",
+    "version": "3.1.0.0",
+    "license": {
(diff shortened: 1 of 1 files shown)


bucket/bulk-rename-utility.json
```diff
@@ -0,0 +1,48 @@
+{
+    "homepage": "https://www.bulkrenameutility.co.uk",
+    "version": "3.1.0.0",
+    "license": {
+        "identifier": "Freeware",
```
rasa: It requires a commercial license at times, so let’s use Proprietary, right?

PR #2613, opened 2019-08-10
prusaslicer: Fix checkver

issaclin32:
mentioned in #2308 (Outstanding Excavator issues)

**PrusaSlicer** now put alpha, beta and rc versions in their release tags.
(see: https://github.com/prusa3d/PrusaSlicer/releases)

~This modifies `checkver` to **check all releases** (rather than checking only the latest), so that it will only match the **latest stable release**.~

This modifies `checkver` to allow alpha and RC releases.

### DIFF
--- a/bucket/prusaslicer.json
+++ b/bucket/prusaslicer.json
@@ -26,8 +26,8 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/prusa3d/PrusaSlicer/releases",
-        "regex": "PrusaSlicer-(?<version>[\\d.]+)\\+win64-(?<timestamp>\\d+).zip"
+        "url": "https://github.com/prusa3d/PrusaSlicer/releases/latest",
+        "regex": "PrusaSlicer-(?<version>[\\w-.]+)\\+win64-(?<timestamp>\\d+)\\.zip"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


niheaven:
You'd better change `checkver.regex` to allow `alpha`, etc.. It is also the **LATEST RELEASE**.

bucket/prusaslicer.json
```diff
@@ -26,8 +26,8 @@
         ]
     ],
     "checkver": {
-        "github": "https://github.com/prusa3d/PrusaSlicer",
-        "regex": "PrusaSlicer-(?<version>[\\d.]+)\\+win64-(?<timestamp>\\d+).zip"
+        "url": "https://github.com/prusa3d/PrusaSlicer/releases/latest",
```
Ash258:
```suggestion
        "github": "https://github.com/prusa3d/PrusaSlicer",
```

rasa:
In general, I think alpha/beta/RC versions should go in the versions bucket. Agreed? Either that or an app-beta.json in this bucket (yuk).

niheaven:
For these apps that use `alpha` in `latest release` tag, I think it should be stable enough to stay in `main` bucket rather than `versions`. It is rare, and should be discussed case by case.

PR #2615, opened 2019-08-10
vieas: Add version 5.4.6.0

wenmin92:
[Vieas](http://www.vieas.com/en/software/vieas.html) is an image viewer and converter with combination of two window forms.


bucket/vieas.json
```diff
@@ -0,0 +1,38 @@
+{
+    "homepage": "http://www.vieas.com/en/software/vieas.html",
+    "description": "an image viewer and converter with combination of two window forms.",
```
Ash258:
```suggestion
    "description": "An image viewer and converter with combination of two window forms.",
```

PR #2620, opened 2019-08-12
gogs: Move to 'main'

niheaven:
CLI tool


r15ch13:
Then gitea should move as well. It's a fork of gogs.

PR #2686, opened 2019-08-18
todotxt-net: Persist the right config file

dodorz:


bucket/todotxt-net.json
```diff
@@ -13,9 +13,9 @@
             "todotxt.net"
         ]
     ],
-    "persist": "todotxt.exe.config",
+    "persist": "todotxt.settings",
```
Ash258: This file do not exists in archive

bucket/todotxt-net.json
```diff
@@ -6,6 +6,9 @@
     "url": "https://github.com/benrhughes/todotxt.net/releases/download/v3.3.1/todotxt-portable.zip",
     "hash": "1c0fe0d37d5519af0e701aebcd47ee17a39751386b18f8e216976c8f75bcf636",
     "extract_dir": "ProgramFiles",
+    "pre_install": [
```
linsui:
```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\todotxt.settings\")) { New-Item \"$dir\\todotxt.settings\" -ItemType File | Out-Null }",
```

Ash258: This would be fixed by Excavator

linsui: Wow, thanks!

PR #2697, opened 2019-08-19
testcentric: Change download URL to Choco

niheaven:


rasa:
Github still hosts the .nupkg. Why the switch?

Please note a .zip file is also provided. See https://github.com/TestCentric/testcentric-gui/blob/master/README.md#installation

niheaven:
It is recommanded by author, and switching to Choco will have extracted file hash so as to provide better integrity check.

rasa:
Have we asked the chocolatey admins if we can do this? I wonder if their terms of service allow this use? We are a competitor of sorts, and they offer a paid service.

rasa:
I never heard back from Rob @ Chocolatey.

PR #2706, opened 2019-08-20
TortoiseSVN: Fix icon overlays, add uninstaller script

### LINKED ISSUE #2486: Something wrong with TortoiseSVN icon overlays.
There's no icon sets to choose in settings while icon files aren't missing. And there's also no icon overlay effects in the windows explorer.
image

issaclin32:
Fix #2486

* **TortoiseSVN** now stores its icons under `Program Files\\Common Files`. This creates a symbolic link pointing to the `$dir\Common\TortoiseOverlays`.

* Add **uninstaller script** to remove the symbolic link when uninstalling. Also, users will not have to run `$dir\\tortoisesvn-uninstall.reg` before uninstalling TortoiseSVN.

* `sudo Remove-Item ... -Recurse -Force` has problems dealing with symbolic links. Therefore I use `cmd` instead. If there is a better approach, please let me know.


linsui:
Could you download reg files as vscode manifest?

bucket/tortoisesvn.json
```diff
@@ -58,12 +78,9 @@
         ]
     ],
     "post_install": [
-        "$null, $bucket = find_manifest($app)",
-        "$bucketdir = Find-BucketDirectory($bucket)",
         "$escdir = $dir -replace \"\\\\\",\"\\\\\"",
-        "(gc \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-install.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install.reg\"",
-        "(gc \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-install-tools.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install-tools.reg\"",
-        "cp \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-uninstall.reg\" \"$dir\"",
+        "(gc \"$dir\\tortoisesvn-install.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install.reg\"",
```
Ash258: - Do not use aliases
- Do not use short parameters
- Convert code to proper casing

bucket/tortoisesvn.json
```diff
@@ -78,9 +78,9 @@
         ]
     ],
     "post_install": [
-        "$escdir = $dir -replace \"\\\\\",\"\\\\\"",
-        "(gc \"$dir\\tortoisesvn-install.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install.reg\"",
-        "(gc \"$dir\\tortoisesvn-install-tools.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install-tools.reg\"",
+        "$escdir = $dir -Replace \"\\\\\",\"\\\\\"",
```
Ash258: Operators are not defined with uppercase ..

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-6

Ash258: Double quotes are not needed

bucket/tortoisesvn.json
```diff
@@ -78,9 +78,9 @@
...
+        "(Get-Content \"$dir\\tortoisesvn-install.reg.templ\") | % {$_ -Replace \"%TSVNHOME%\",\"$escdir\"} | Out-File -Encoding OEM \"$dir\\tortoisesvn-install.reg\"",
```
Ash258: - First are positional parameters then named
- DO not use `OEM`, Use `ASCII`
- `%` is alias of `ForEach-Object`
- Single quotes for `%TSVNHOME`

issaclin32: Thanks for the details.

linsui:
You can create junction instead of symlink. Out-Null doesn't work with sudo.

scripts/tortoisesvn/tortoisesvn-install.reg.templ
```diff
@@ -366,15 +366,15 @@ REGEDIT4
...
-"NormalIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\NormalIcon.ico"
-"ModifiedIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\ModifiedIcon.ico"
-"ConflictIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\ConflictIcon.ico"
-"DeletedIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\DeletedIcon.ico"
-"ReadOnlyIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\ReadOnlyIcon.ico"
-"LockedIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\LockedIcon.ico"
-"AddedIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\AddedIcon.ico"
-"IgnoredIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\IgnoredIcon.ico"
-"UnversionedIcon"="%TSVNHOME%\\TortoiseOverlays\\icons\\XPStyle\\UnversionedIcon.ico"
+"NormalIcon"="%TSVNHOME%\\Common\\TortoiseOverlays\\icons\\XPStyle\\NormalIcon.ico"
```
linsui: I install Tortoisesvn and get `C:\Program Files\Common Files\TortoiseOverlays\icons\XPStyle\AddedIcon.ico` here. Is this the path to icon? If then, why we need to create a dir in C:\Program Files\Common Files\ ?

linsui:
I wonder if there is any way to prevent `sudo New-Item` output messages to terminal...

scripts/tortoisesvn/tortoisesvn-install.reg.templ
linsui: I mean if we can change the path of icons, do we need to create a dir in  C:\Program Files\Common Files\ ? I wonder if it works without creating the symlink...

issaclin32: Sorry for the late reply.

If we don't create symlink under `Program Files\Common Files`, it will not work properly.

I think `TortoiseSVN` does not handle all features of the overlay icons through that registry key. For some functions it still loads the icon files through `Program Files\Common Files`.

linsui: Oh, I see...

linsui:
Why not use Junction? It can be removed by Remove-Item.

issaclin32:
> * There are no `%COMMON%` strings in any template
> * `$common` variable is unused
> * Foreach is not needed. Replace operator works on array by default
> * Simplified checkver and autoupdate
> * I am facing redirects to http

Thanks for testing it out.

I found that `sudo` works under Powershell **v5.1** but not **v6.1**.
Would you suggest that we directly stop the install when the command prompt is not running under Admin? (similar to [The KMS manifest](https://github.com/Ash258/scoop-Ash258/blob/master/bucket/KMS.json))

bucket/tortoisesvn.json
```diff
@@ -57,21 +78,28 @@
...
     "notes": [
         "To add TortoiseSVN as a context menu option, run '$dir\\tortoisesvn-install.reg'",
-        "For file associations, run '$dir\\tortoisesvn-install-tools.reg'",
-        "Before uninstalling TortoiseSVN, run '$dir\\tortoisesvn-uninstall.reg' "
+        "For file associations, run '$dir\\tortoisesvn-install-tools.reg'"
     ],
     "checkver": {
         "url": "https://tortoisesvn.net/downloads.html",
-        "re": "tortoisesvn/storage/(?<release>[\\d.]+)/Application/TortoiseSVN-([\\d.]+)-x64-svn-(?<svn>[\\d.]+).msi"
+        "regex": "tortoisesvn/storage/(?<release>[\\d.]+)/Application/TortoiseSVN-([\\d.]+)-x64-svn-(?<svn>[\\d.]+).msi"
```
Ash258: Get fixed from my review.

bucket/tortoisesvn.json
```diff
@@ -57,21 +78,28 @@
...
-        "$null, $bucket = find_manifest($app)",
-        "$bucketdir = bucketdir($bucket)",
-        "$escdir = $dir -replace \"\\\\\",\"\\\\\"",
-        "(gc \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-install.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install.reg\"",
-        "(gc \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-install-tools.reg.templ\") | % {$_ -replace \"%TSVNHOME%\",\"$escdir\"} | out-file -enc oem \"$dir\\tortoisesvn-install-tools.reg\"",
-        "cp \"$bucketdir\\..\\scripts\\tortoisesvn\\tortoisesvn-uninstall.reg\" \"$dir\""
+        "function repl($file) {",
+        "    $esc = $dir.Replace('\\', '\\\\')",
+        "    $it = Get-Item \"$dir\\$file\"",
+        "    ((Get-Content $it.Fullname) -replace '%TSVNHOME%', \"$esc\") | Set-Content \"$dir\\$($it.Basename)\" -Encoding ASCII",
```
Ash258: Now you need COMMON

issaclin32: The icon files can be found under `$dir\common`, but for some reason `TortoiseSVN` still needs to load the icons from `Program Files\Common Files`. I previously modified all the location to `Program Files\Common Files`, but found it unnecessary and reverted it back.
 
(It's messed up. Sorry.)

Ash258: Then just replace current manifest with one from my review. There you have updated checkver, autoupdate and it is working.

For the icons i will try again in few hours because I do not know what is hapenning right now.

bartlibert:
Any update on this? I'm plagued by the issue as well and was wondering if I can help in any way to get this through?

PR #2710, opened 2019-08-20
Everything CLI: Add version 1.1.0.15

niboan:
Added Everything command line interface.

### DIFF
--- a/bucket/everything-cli.json
+++ b/bucket/everything-cli.json
@@ -1,0 +1,17 @@
+{
+    "homepage": "https://www.voidtools.com/",
+    "description": "Command line interface to Everyting.",
+    "license": "MIT",
+    "version": "1.1.0.15",
+    "depends": "everything",
+    "url": "https://www.voidtools.com/ES-1.1.0.15.zip",
+    "hash": "ea7ef4291f94f630dd142d8e0fa7cf90db83750cbb32429888ef405365c79a18",
+    "bin": "es.exe",
+    "checkver": {
+        "url": "https://www.voidtools.com/downloads/",
+        "re": "ES-([\\d.]+).zip"
+    },
+    "autoupdate": {
+        "url": "https://www.voidtools.com/ES-$version.zip"
+    }
+}
--- a/bucket/everything-cli.json
+++ b/bucket/everything-cli.json
@@ -1,5 +1,5 @@
 {
-    "homepage": "https://www.voidtools.com/",
+    "homepage": "https://www.voidtools.com/support/everything/command_line_interface/",
     "description": "Command line interface to Everyting.",
     "license": "MIT",
     "version": "1.1.0.15",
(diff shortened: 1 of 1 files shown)


bucket/everything-cli.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://www.voidtools.com/",
+    "description": "Command line interface to Everyting.",
+    "license": "MIT",
+    "version": "1.1.0.15",
+    "depends": "everything",
```
Ash258:
```suggestion
    "depends": "extras/everything",
```

bucket/everything-cli.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://www.voidtools.com/",
```
Ash258:
```suggestion
    "homepage": "https://www.voidtools.com/support/everything/command_line_interface/",
```

bucket/everything-cli.json
```diff
@@ -0,0 +1,17 @@
...
+    "description": "Command line interface to Everyting.",
+    "license": "MIT",
```
Ash258: Where did you get this information?

bucket/everything-cli.json
```diff
@@ -0,0 +1,17 @@
...
+    "version": "1.1.0.15",
+    "depends": "everything",
+    "url": "https://www.voidtools.com/ES-1.1.0.15.zip",
+    "hash": "ea7ef4291f94f630dd142d8e0fa7cf90db83750cbb32429888ef405365c79a18",
+    "bin": "es.exe",
+    "checkver": {
+        "url": "https://www.voidtools.com/downloads/",
+        "re": "ES-([\\d.]+).zip"
```
Ash258:
```suggestion
        "regex": "ES-([\\d.]+)\\.zip"
```

bucket/everything-cli.json
niboan: From the source code included in the download. It doesn't explicitly say MIT, but it's the MIT license text.

bucket/everything-cli.json
niboan: OK, I didn't like the idea of depending on packages in other buckets, especially from main to extras, but if that's the way it's done...

niboan: Should I just submit a new PR on main?

linsui: Yeah, please close this one and submit a new PR with suggested change in Main.

niboan: A quick search in main showed no other packages depending on packages in extras, only suggest's. I would say keep it in extras...

linsui: I wonder if it is acceptable to add cli into everything.json and move it to main...

niboan: Everything is a GUI tool, and does not fit the [criteria of main](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket). I also don't think the cli fits the criteria, since it depends on a GUI tool.
I don't think the cli can be added to `everything.json`, as it has its own versioning.

PR #2718, opened 2019-08-21
fastcopy: don't create persist dir in pre_install

Retia-Adolf:


bucket/fastcopy.json
```diff
@@ -19,9 +19,6 @@
     "pre_install": [
         "if (!(Test-Path \"$dir/FastCopy2.ini\")) {",
```
Ash258:
```suggestion
        "if (!(Test-Path \"$persist_dir\\FastCopy2.ini\")) {",
```

bucket/fastcopy.json
```diff
@@ -19,9 +19,6 @@
...
         "   Set-Content \"$dir/FastCopy2.ini\" -value '[main]'",
```
Ash258:
```suggestion
        "   Set-Content \"$dir\\FastCopy2.ini\" '[main]' -Encoding ASCII",
````

PR #2723, opened 2019-08-23
jpegview: Add persistence

dodorz:


bucket/jpegview.json
```diff
@@ -24,6 +25,11 @@
         "url": "https://sourceforge.net/projects/jpegview/rss?path=/",
         "re": "/jpegview/([\\d.]+)/JPEGView_"
     },
+    "persist": [
+        "JPEGView.ini",
+        "KeyMap.txt",
+        "ParamDB.db"
```
Ash258: Where did you get information about this file?

When it is created it is always 0bytes and also when you delete it application do not create it.

Ash258: A

Ash258: You also need to copy old folder to persist for backward compatability

Ash258: If you just create it then it cannot be used

A

See ditto manifest how to handle this

issaclin32:
I think **manual persist** (in pre_install/uninstaller) is the right way to do this, because an empty `ParamDB.db` cannot work. Also whether `ParamDB.db` exists depends on user settings.

rashil2000:
The `.db` file can be created like this:

https://github.com/ScoopInstaller/Extras/blob/1cd0ac886e13c86efbded46dedc068be42c8717d/bucket/ditto.json#L18-L26

issaclin32:
> The `.db` file can be created like this:

Thanks for the info, but `ParamDB.db` will only be created when user clicks `right-click -> save parameter to DB`. So this approach cannot work.

The usage of ParamDB can be found at: http://wa2l-wintools.sourceforge.net/man1/jpegview.Readme.1.html

issaclin32:
Superseded by #7632

PR #2729, opened 2019-08-23
winscp: add commandline-friendly binary.

silverkorn:
The `.exe` binary creates another command prompt instance while the `.com` one stays in the same one.

### DIFF
--- a/bucket/winscp.json
+++ b/bucket/winscp.json
@@ -15,7 +15,10 @@
     ],
     "bin": [
         "WinSCP.exe",
-        "WinSCP.com"
+        [
+            "WinSCP.com",
+            "WinSCP-cli"
+        ]
     ],
     "shortcuts": [
         [
--- a/bucket/winscp.json
+++ b/bucket/winscp.json
@@ -13,7 +13,10 @@
         "    ) | Add-Content -Path \"$dir\\winscp.ini\" -Encoding ASCII -Force",
         "}"
     ],
-    "bin": "WinSCP.exe",
+    "bin": [
+        "WinSCP.exe",
+        "WinSCP.com"
+    ],
     "shortcuts": [
         [
             "WinSCP.exe",


bucket/winscp.json
```diff
@@ -13,7 +13,10 @@
         "    ) | Add-Content -Path \"$dir\\winscp.ini\" -Encoding ASCII -Force",
         "}"
     ],
-    "bin": "WinSCP.exe",
+    "bin": [
+        "WinSCP.exe",
+        "WinSCP.com"
```
Ash258: This will not work as `winscp.exe` will be immediatelly overriden with `winspc.com` Add just `.com` to bin.

silverkorn: Hmm... May I suggest this instead?
```json
...
"pre_install": [
   ...
   "Set-Content \"$dir\\winscp-cli.cmd\" (@('@echo off', '\"%~dp0WinSCP.com\" %*') -join \"`r`n\") -Encoding ASCII"
],
"bin": [
    "WinSCP.exe",
    "winscp-cli.cmd"
],
...
```

We could use `winscp-cli` for commandline usage. (Tested and working)

Ash258: Why do you add `winscp.exe` in first?

silverkorn: Backward compatibility?
I'm not sure to understand the question.
The current need is to have a way to also use the commandline version (`WinSCP.com`) along with the common GUI one (`WinSCP.exe`).

Ash258: Nonsense.

So you will prioritize GUI application as command line instead of proper command line application with proper command line argument handler.

silverkorn: Totally agreed but I don't want to upset users that might be using it as it currently is...
This will shim only the `.com` version, no?
Should proceed anyway?

Ash258: This will shim exe file.
Then this exe will be overriden of com shim, which will lead to only com shimmed, which will result for erorrs on updates / installation.

silverkorn: Ok, so `"bin": "WinSCP.com"` only and no backward compatibility, if anyone was using it this way?

Ash258: I am forbidding spawning GUI application from terminal in all cases, as it is completely irrelevant and nonsense, so this decision could not be done by me.

silverkorn: Done.
You can assign the one who can take the final word/action.
Thanks!

PR #2731, opened 2019-08-25
teamviewer: Add post_install to set nosave to 0 for user preferences

robmonte:
This installs the portable Teamviewer meaning that nosave defaults to 1. This means user preferences are not maintained between uses. Setting this to 0 with a regex replace on that line in the .ini file fixes this.

### DIFF
--- a/bucket/teamviewer.json
+++ b/bucket/teamviewer.json
@@ -16,6 +16,11 @@
         ]
     ],
     "persist": "teamviewer.ini",
+    "post_install": [
+        "$iniFile = \"$dir\\teamviewer.ini\"",
+        "$changeRegex = \"(?<=nosave=)[0-9]*\"",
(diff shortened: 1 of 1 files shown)


Ash258:
Where did you get this information?

robmonte:
Had to Google around because it was frustrating me why I kept losing Partner IDs on close. Found this first:
https://community.teamviewer.com/t5/Knowledge-Base/Getting-Started-with-TeamViewer-Portable/ta-p/4799
and then after seeing "nosave=1" on that page,  I searched that up and saw this result:
https://portableapps.com/node/24854

Seeing it twice, I just tested it out locally and it worked. It's worth noting that changing this causes a UAC prompt, because apparently the way Teamviewer stores its preferences is the Registry (sigh). But there isn't a way around this.

pratikpc:
Closing as robmonte:master has been deleted.

PR #2739, opened 2019-08-25
OnionShare: Add versiom 2.1

linsui:


bucket/onionshare.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "2.1",
+    "description": "Securely and anonymously send and receive files.",
+    "homepage": "https://onionshare.org/",
+    "license": "GPL-3.0-only",
+    "url": "https://github.com/micahflee/onionshare/releases/download/v2.1/onionshare-2.1-setup.exe#/dl.7z",
+    "hash": "sha512:395299aa40d8e15066edfd898d7633d283a9cf3a304bf25a06af62c57b79fe767a48b173957ffbddacbffa27cfcbb71013e917134fb003424abc6995f0602943",
+    "bin": [
+        "onionshare-gui.exe",
```
Ash258: Does gui have some sort of command line suport?

bucket/onionshare.json
```diff
@@ -0,0 +1,30 @@
...
+            "onionshare.ico"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/micahflee/onionshare"
+    },
+    "autoupdate": {
+        "url": "https://github.com/micahflee/onionshare/releases/download/v$version/onionshare-$version-setup.exe#/dl.7z",
+        "hash": {
+            "url": "https://chocolatey.org/packages/onionshare/$version",
```
Ash258: I do not think that chocolatey page will exists as soon as new release id sone, which will lead to downloading

bucket/onionshare.json
linsui: No, it's my fault, I forget to add `[]` ...

bucket/onionshare.json
linsui: Yeah, should I checkver from chocolatey? Or just remove au.hash?

Ash258: Just remove it.

Depend on choco should not be used when there are better options.

PR #2740, opened 2019-08-25
Ricochet: add verion 1.1.4

linsui:


bucket/ricochet.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "1.1.4",
+    "description": "Anonymous peer-to-peer instant messaging.",
+    "homepage": "https://github.com/ricochet-im/ricochet",
```
Ash258: https://ricochet.im

bucket/ricochet.json
```diff
@@ -0,0 +1,18 @@
...
+    "innosetup": true,
+    "bin": "ricochet.exe",
+    "shortcuts": [
+        [
+            "ricochet.exe",
+            "Ricochet"
+        ]
+    ],
+    "post_install": "Remove-Item \"$dir\\tor.exe\" -Force"
+}
```
Ash258: No checkver and autoupdate

bucket/ricochet.json
```diff
@@ -0,0 +1,18 @@
...
+    "depends": "tor",
+    "innosetup": true,
+    "bin": "ricochet.exe",
+    "shortcuts": [
+        [
+            "ricochet.exe",
+            "Ricochet"
+        ]
+    ],
+    "post_install": "Remove-Item \"$dir\\tor.exe\" -Force"
```
Ash258: Why do you remove this tor execuable and depending on tor manifest?

bucket/ricochet.json
linsui: It has stoped for more than 2 years... Should I add checkver anf autoupdate?

bucket/ricochet.json
linsui: The bundled tor is outdate and unsafe.

bucket/ricochet.json
linsui: It will redirect to github.

Ash258: I can without problem my VMs physically running in Czechia, Romania, Germany, Switzerland, England.

Problem is on your end.

bucket/ricochet.json
Ash258: There is no reason to not add it. Repository is not archived, no official statement, anything, ...

bucket/ricochet.json
linsui: Thanks, you are right, it's blocked in China.

bucket/ricochet.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "1.1.4",
+    "description": "Anonymous peer-to-peer instant messaging.",
+    "homepage": "https://ricochet.im",
+    "license": "BSD-3-Clause",
+    "url": "https://github.com/ricochet-im/ricochet/releases/download/v1.1.4/ricochet-1.1.4-win-install.exe#/dl.7z",
```
Ash258: Why this rename, when you are using `innosetup`??

linsui: Sorry, my fault...

bucket/ricochet.json
Ash258: 1. Will this work with new versions of tor?
1. Will this work when there is no file `tor.exe` in directory, but in Path

linsui: Yeah, it works.

PR #2741, opened 2019-08-25
Nextcloud: add version 2.5.3

linsui:
Maybe we need something like `-gui` option, so we can always run gui from cmd.

### DIFF
--- a/bucket/nextcloud.json
+++ b/bucket/nextcloud.json
@@ -6,6 +6,7 @@
     "url": "https://download.nextcloud.com/desktop/releases/Windows/Nextcloud-2.5.3-setup.exe#/dl.7z",
     "hash": "c75f667e6a149f5f72db46d15b4ea24a614789a2cf103040c2f9a66fc3482466",
     "bin": [
+        "nextcloudcmd.exe",
         [
             "nextcloudcmd.exe",
             "nextcloud"
--- a/bucket/nextcloud.json
+++ b/bucket/nextcloud.json
@@ -19,7 +19,7 @@
         ]
     ],
     "checkver": {
-        "url": "https://nextcloud.com/install/#install-clients",
+        "url": "https://nextcloud.com/install/",
         "regex": "Latest stable version: ([\\d.]+)"
     },
     "autoupdate": {
--- a/bucket/nextcloud.json
+++ b/bucket/nextcloud.json
@@ -1,0 +1,27 @@
+{
+    "version": "2.5.3",
+    "description": "Desktop sync client for Nextcloud, a self-hosted productivity platform that keeps you in control.",
+    "homepage": "https://nextcloud.com/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://download.nextcloud.com/desktop/releases/Windows/Nextcloud-2.5.3-setup.exe#/dl.7z",
+    "hash": "c75f667e6a149f5f72db46d15b4ea24a614789a2cf103040c2f9a66fc3482466",
+    "bin": [
+        [
+            "nextcloudcmd.exe",
+            "nextcloud"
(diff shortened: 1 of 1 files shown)


bucket/nextcloud.json
```diff
@@ -0,0 +1,27 @@
+{
+    "version": "2.5.3",
+    "description": "Desktop sync client for Nextcloud, a self-hosted productivity platform that keeps you in control.",
+    "homepage": "https://nextcloud.com/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://download.nextcloud.com/desktop/releases/Windows/Nextcloud-2.5.3-setup.exe#/dl.7z",
+    "hash": "c75f667e6a149f5f72db46d15b4ea24a614789a2cf103040c2f9a66fc3482466",
+    "bin": [
```
Ash258: Keep defaults. You cannot be sure if application is not calling it statically.

```suggestion
    "bin": [
        "nextcloudcmd.exe",
````

File: bucket/nextcloud.json [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+        ]
+    ],
+    "shortcuts": [
+        [
+            "nextcloud.exe",
+            "Nextcloud"
+        ]
+    ],
+    "checkver": {
+        "url": "https://nextcloud.com/install/#install-clients",
```
Ash258: Label is only for browsers. Webclients are ignoring them as they cannot be evaluated in most cases

```suggestion
        "url": "https://nextcloud.com/install/",
```

bucket/nextcloud.json
linsui: How about

```
    "bin": [
        "nextcloudcmd.exe",
        [
            "nextcloudcmd.exe",
            "nextcloud"
        ]
    ],
```

or

```
    "bin": [
        "nextcloud.exe",
        [
            "nextcloudcmd.exe",
            "nextcloud"
        ]
    ],
```

Ash258: My suggestion will make it this:

```json
"bin": [
    "cmd",
    [
         "cmd",
         "simple"
    ]
]
```

Both codes you provided are identical and same to what my suggestion is ...

bucket/nextcloud.json
```diff
@@ -0,0 +1,28 @@
+{
+    "version": "2.5.3",
+    "description": "Desktop sync client for Nextcloud, a self-hosted productivity platform that keeps you in control.",
+    "homepage": "https://nextcloud.com/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://download.nextcloud.com/desktop/releases/Windows/Nextcloud-2.5.3-setup.exe#/dl.7z",
+    "hash": "c75f667e6a149f5f72db46d15b4ea24a614789a2cf103040c2f9a66fc3482466",
```
Ash258:
```suggestion
    "hash": "c75f667e6a149f5f72db46d15b4ea24a614789a2cf103040c2f9a66fc3482466",
    "pre_install": "'$R*', '$P*', 'vc_redist*' | ForEach-Object { Remove-Item \"$dir\\$_\" -Force -Recurse }",
```

bucket/nextcloud.json
```diff
@@ -0,0 +1,28 @@
+{
+    "version": "2.5.3",
+    "description": "Desktop sync client for Nextcloud, a self-hosted productivity platform that keeps you in control.",
+    "homepage": "https://nextcloud.com/",
+    "license": "GPL-2.0-or-later",
```
Ash258:
```suggestion
    "license": "GPL-2.0-or-later",
    "suggest": {
        "Microsoft Visual C++ Redistributable 2017": "extras/vcredist2017"
    },
```

PR #2747, opened 2019-08-25
Add PDF-XChange Editor

Velgus:


Ash258:
Is this persist working with simple New-Item? Lots of applications cannot work when custom dat files are not generated by application (or defined without additional headers, format, encoding, ...)

Velgus:
Seems to work fine. The files' data populates as usual after the first launch and close, and setting changes work fine when re-opened.

PR #2753, opened 2019-08-26
Add popeye: Kubernetes cluster resource sanitizer

filippobuletto:
Add popeye from https://github.com/derailed/popeye


Ash258:
Why extras bucket instead of main? I do not see any shortcuts or screenshots of graphical interface.

filippobuletto:
As per "Criteria for including apps in the main bucket" this is not a:

- a reasonably well-known and **widely used** developer tool

If you think this fits better the main bucket I'll close the PR and move to main :)

r15ch13:
@filippobuletto please move it to main.

PR #2756, opened 2019-08-27
QuickViewer: Add version 1.1.6

linsui:
Close https://github.com/lukesampson/scoop-extras/issues/2755

Persist doesn't works, so copy files manully.

### DIFF
--- a/bucket/quickviewer.json
+++ b/bucket/quickviewer.json
@@ -17,8 +17,8 @@
     },
     "installer": {
         "script": [
-            "foreach ($file in @('progress.ini', 'quickviewer.ini')){",
-            "    if(!(Test-Path \"$persist_dir\\$file\")) {",
+            "foreach ($file in @('progress.ini', 'quickviewer.ini')) {",
+            "    if (!(Test-Path \"$persist_dir\\$file\")) {",
             "        New-Item \"$persist_dir\\$file\" -Force | Out-Null",
             "    }",
             "    Copy-Item \"$persist_dir\\$file\" \"$dir\\$file\" -Force",
@@ -27,7 +27,7 @@
     },
     "uninstaller": {
         "script": [
-            "foreach ($file in @('progress.ini', 'quickviewer.ini')){",
+            "foreach ($file in @('progress.ini', 'quickviewer.ini')) {",
             "    Copy-Item \"$dir\\$file\" \"$persist_dir\\$file\" -Force",
             "}"
         ]
--- a/bucket/quickviewer.json
+++ b/bucket/quickviewer.json
@@ -27,8 +27,6 @@
     },
     "uninstaller": {
         "script": "Copy-Item \"$dir\\*\" \"$persist_dir\" -Include 'progress.ini', 'quickviewer.ini' -Force"
-            "Copy-Item \"$dir\\*\" \"$persist_dir\" -Include 'progress.ini', 'quickviewer.ini'"
-        ]
     },
     "bin": "QuickViewer.exe",
     "shortcuts": [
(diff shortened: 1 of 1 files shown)


bucket/quickviewer.json
```diff
@@ -0,0 +1,57 @@
+{
+    "version": "1.1.6",
+    "homepage": "https://kanryu.github.io/quickviewer/",
+    "description": "An image/comic viewer application.",
```
Ash258:
```suggestion
    "description": "An image/comic viewer.",
```

bucket/quickviewer.json
```diff
@@ -0,0 +1,57 @@
...
+            "    if (!(Test-Path \"$persist_dir\\$file\")) {",
+            "        New-Item \"$persist_dir\\$file\" -Force | Out-Null",
+            "    }",
+            "    Copy-Item \"$persist_dir\\$file\" \"$dir\\$file\" -Force",
+            "}"
+        ]
+    },
+    "uninstaller": {
+        "script": [
+            "foreach ($file in @('progress.ini', 'quickviewer.ini')) {",
```
Ash258:
```suggestion
            "Copy-Item \"$dir\\*\" \"$persist_dir\" -Include 'progress.ini', 'quickviewer.ini'",
```

bucket/quickviewer.json
```diff
@@ -0,0 +1,55 @@
...
+            "    if (!(Test-Path \"$persist_dir\\$file\")) {",
+            "        New-Item \"$persist_dir\\$file\" -Force | Out-Null",
+            "    }",
+            "    Copy-Item \"$persist_dir\\$file\" \"$dir\\$file\" -Force",
+            "}"
+        ]
+    },
+    "uninstaller": {
+        "script": [
```
Ash258: I forgot force, sorry..
```suggestion
        "script": "Copy-Item \"$dir\\*\" \"$persist_dir\" -Include 'progress.ini', 'quickviewer.ini' -Force"
```

linsui: How about `"Copy-Item \"$dir\\*.ini\" \"$persist_dir\" -Force"` ? And it works well without force.

Ash258: I do not know what ini files are there. If you want to persist only these two, then just copy these two.

linsui: There is no other ini file. We can simplify it as we do not need to specify which ini file should be persist.

Ash258: There is nothing which could be earned when copying all ini files. Yeah funciton will be shorter, but that is all.

linsui: I wonder will this make the installer more clear?

```
    "installer": {
        "script": [
            "if (!(Test-Path \"$persist_dir\")) { New-Item \"$persist_dir\" -ItemType Directory -Force | Out-Null }",
            "Copy-Item \"$persist_dir\\*.ini\" \"$dir\" -Force"
        ]
    },
    "uninstaller": {
        "script": "Copy-Item \"$dir\\*.ini\" \"$persist_dir\" -Force"
    },
```

Ash258: How will this solve that these configuration files do not exists in archive?

Ash258: Keep it as is now.

linsui: They will be created when first run. So we can persist them only when they exist. We can also do this on those file which can't be created manully. Such as

```
            "    if (Test-Path \"$persist_dir\\$file\") {",
            "        Copy-Item \"$persist_dir\\$file\" \"$dir\\$file\" -Force",
            "    }",
```

Ash258: These special use cases will be covered in lukesampson/scoop#3212

Until persistence is reworked keep it as simple and readable as you can.

linsui: Great! Thanks!

PR #2769, opened 2019-08-28
gcloud: Remove post_install and add notes for init

abarani:
I propose to remove the `post_install` script for Google Cloud SDK mainly for two reasons:
1. Avoid re-initializing Cloud SDK on each update
  Currently the init process must be done on each update otherwise scoop marks the app installation as failed;
2. Allow installing Cloud SDK in an unattended way
  This change allow installing Cloud SDK without user interaction, the user is still able to initialize it manually afterwards with their credentials and settings.

I also added `notes`  field as it is in Git in the main repository.


abarani:
Any comments on this?

PR #2808, opened 2019-09-05
Link to npcap download page during Wireshark installation

gregmac:
image

### DIFF
--- a/bucket/wireshark.json
+++ b/bucket/wireshark.json
@@ -31,6 +31,10 @@
     "suggest": {
         "nmap": "nmap"
     },
+    "notes": [
+        "Wireshark needs a packet sniffing library to capture packets from your network interface.",
+        "Please either install the 'nmap' package or install NPcap from https://nmap.org/npcap/#download."
+    ],
     "post_install": "Write-Host -fore yellow \"*********************`nNote: Wireshark needs a packet sniffing library to capture packets from your network interface, but this must be installed separately. You can download NPcap at https://nmap.org/npcap/#download`n*********************\";",
     "checkver": {
         "url": "https://www.wireshark.org/update/0/Wireshark/0.0.0/Windows/x86/en-US/stable.xml",
--- a/bucket/wireshark.json
+++ b/bucket/wireshark.json
@@ -31,6 +31,7 @@
     "suggest": {
         "nmap": "nmap"
     },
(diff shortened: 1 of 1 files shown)


bucket/wireshark.json
```diff
@@ -31,6 +31,7 @@
     "suggest": {
         "nmap": "nmap"
     },
+    "post_install": "Write-Host -fore yellow \"*********************`nNote: Wireshark needs a packet sniffing library to capture packets from your network interface, but this must be installed separately. You can download NPcap at https://nmap.org/npcap/#download`n*********************\";",
```
Ash258: Or just add npcap manifest and depends on it.
```suggestion
    "notes": [
        "Wireshark needs a packet sniffing library to capture packets from your network interface, but this must be installed separately",
        "You can download NPcap at https://nmap.org/npcap/#download",
    ],
```

linsui:
nmap has been suggested which contains npcap.

gregmac:
I didn't actually realize nmap had npcap (the manifest for nmap doesn't mention it), so in addition to taking @Ash258's suggestion to use `notes` I explicitly call that out. 

Should we update the nmap manifest to mention it includes npcap? This would make it searchable in scoop, at the least. Right now the only way to figure that out is to visit the [nmap download page](https://nmap.org/download.html) and read a few paragraphs down -- not obvious if you're just trying to use wireshark.

bucket/wireshark.json
```diff
@@ -31,6 +31,11 @@
     "suggest": {
         "nmap": "nmap"
     },
+    "notes": [
+        "Wireshark needs a packet sniffing library to capture packets from your network interface.",
+        "Please either install the 'nmap' package or install NPcap from https://nmap.org/npcap/#download."
+    ],
+    "post_install": "Write-Host -fore yellow \"*********************`nNote: Wireshark needs a packet sniffing library to capture packets from your network interface, but this must be installed separately. You can download NPcap at https://nmap.org/npcap/#download`n*********************\";",
```
linsui:
```suggestion
```

bucket/wireshark.json
```diff
@@ -31,6 +31,11 @@
     "suggest": {
         "nmap": "nmap"
```
linsui:
```suggestion
         "pcap": "nmap"
```

Maybe you can also change the suggest as we only need npcap in nmap. This is more clear.

gregmac:
Already fixed in #2930

PR #2828, opened 2019-09-10
Update caprine setup name to new naming scheme

mads256h:
Fixes a 404 download error.


Ash258:
Again? They change it from release to release...

- 2.35 - dots
- 2.36 - dashes with lower case name
- 2.37 - dots
- 2.38 - dashes with upper case

mads256c:
Maybe it would be easier just to grab the first .exe on the releases page.

mads256c:
New changes to this file was merged on the master branch, but it is still broken.

PR #2843, opened 2019-09-14
notepadplusplus: script for replacing default notepad with notepad++

archit9169:
Since the explorer context menu is still not 100% working all the time, this will replace the default notepad with notepad++. Right now this is the best alternative for using context menu.
I do not know much about hashes so i have not included those.


issaclin32:
I am closing this PR because we have `install-context.reg` for **notepad++** now.

PR #2869, opened 2019-09-21
foobar2000: Improve manifest persist handling

Velgus:
Following fixes:
* `user-components` directory is persisted and migration from local AppData to portable mode is handled (handles [custom components](https://www.foobar2000.org/components))
* Playlist migration for arbitrarily old versions of foobar2000 (eg. `playlists-v1.3`, `playlists-v1.4`, potential future version numbers)


bucket/foobar2000.json
```diff
@@ -28,22 +29,28 @@
             "    New-Item \"$persist_dir\\playlists\" -ItemType Directory -Force | Out-Null",
             "    if (Test-Path \"$env:Appdata\\foobar2000\") {",
             "        Write-Host -F yellow \"Copying old '$env:Appdata\\foobar2000' to '$persist_dir'\"",
-            "        Copy-Item \"$env:Appdata\\foobar2000\\*\" \"$dir\" -Recurse -Force",
+            "        Copy-Item \"$env:Appdata\\foobar2000\\*\" \"$dir\" -Force -Recurse",
+            "        if (Test-Path \"$env:Appdata\\foobar2000\\playlists-v*\") {",
+            "            Copy-Item \"$dir\\playlists-v*\" \"$persist_dir\\playlists\" -Force -Recurse",
+            "        }",
+            "        if (Test-Path \"$env:Appdata\\foobar2000\\user-components\") {",
```
Ash258: I do not understand this check. 

If there is this folder in appdata, you are copying, user component folder from dir to persist.

Why is there a user-components directory inside appdata, which is not used then. And you are not copying anything from it...

bucket/foobar2000.json
```diff
@@ -28,22 +29,28 @@
...
-            "    $shortVersion = $version.Split('.')[0, 1] -Join '.'",
-            "    Copy-Item \"$persist_dir\\playlists\" \"$dir\\playlist-v$shortVersion\" -Force -Recurse",
+            "    Copy-Item \"$persist_dir\\playlists\\*\" \"$dir\" -Force -Recurse",
             "}",
             "if (Test-Path \"$persist_dir\\theme.fth\") {",
             "    Copy-Item \"$persist_dir\\theme.fth\" \"$dir\" -Force",
             "}"
         ]
     },
+    "pre_install": "if(!(Test-Path \"$persist_dir\\user-components\")) { New-Item \"$dir\\user-components\" -ItemType Directory -Force | Out-Null}",
```
Ash258: Why??

You are using force, recurse in L37

bucket/foobar2000.json
```diff
@@ -28,22 +29,28 @@
...
     "post_install": "if (is_directory \"$dir\\theme.fth\") { Remove-Item \"$dir\\theme.fth\" -Force }",
     "uninstaller": {
         "script": [
-            "Copy-Item \"$dir\\playlists-v*\\*\" \"$persist_dir\\playlists\" -Force  -Recurse",
+            "Copy-Item \"$dir\\playlists-v*\" \"$persist_dir\\playlists\" -Force  -Recurse",
```
Ash258: This will result in

`$persist_dir\playlists\playlists-vXXX.XX`. Is it correct?

bucket/foobar2000.json
Velgus: It's a migration from when foobar2000 in scoop used to be run in non-portable mode. The line right above there moved the files to `$dir` from AppData already.

bucket/foobar2000.json
Velgus: Yes. Foobar2000 uses `$dir\playlists-vX.X` when it reads playlists. There is no way to support arbitrary version numbers with just scoop's the `persist`.

bucket/foobar2000.json
Velgus: I'm not the one who wrote that migration code originally, but I'll look into a way to get it to only run when needed as well (currently it will always run if you've _ever_ had foobar2000 installed non-portably).

bucket/foobar2000.json
Velgus: Having some weird behaviors with this line - will resolve shortly.

bucket/foobar2000.json
Velgus: Oh, I see what you mean, I'll fix it.

bucket/foobar2000.json
Velgus: Ah, was getting confused. Fixed.

bucket/foobar2000.json
```diff
@@ -28,22 +29,25 @@
...
-            "    $shortVersion = $version.Split('.')[0, 1] -Join '.'",
-            "    Copy-Item \"$persist_dir\\playlists\" \"$dir\\playlist-v$shortVersion\" -Force -Recurse",
+            "    Copy-Item \"$persist_dir\\playlists\\*\" \"$dir\" -Force -Recurse",
             "}",
             "if (Test-Path \"$persist_dir\\theme.fth\") {",
             "    Copy-Item \"$persist_dir\\theme.fth\" \"$dir\" -Force",
             "}"
         ]
     },
+    "pre_install": "if(!(Test-Path \"$persist_dir\\user-components\")) { New-Item \"$dir\\user-components\" -ItemType Directory -Force | Out-Null}",
```
linsui: `persist` will create user-components if it is a directory. Only file need to be create manully.

bucket/foobar2000.json
```diff
@@ -28,22 +29,25 @@
...
     "post_install": "if (is_directory \"$dir\\theme.fth\") { Remove-Item \"$dir\\theme.fth\" -Force }",
```
linsui: `post_install` is useless, I forgot to remove it in https://github.com/lukesampson/scoop-extras/pull/2624 ,sorry...

bucket/foobar2000.json
linsui: So why do you make it nested? Will foobar2000 rename playlist-v1.3 to playlist-v1.4 when update?

Velgus: It doesn't actually need to be nested in the persist directory - I was just keeping the folder from the manifest as I had found it before.

linsui: Well, I still don't understand... And why you copy it twice in `installer` and `uninstaller` ?

Velgus: I'm not sure I understand what you mean copied twice?

The first `if` section in the installer only runs if it's a fresh installation via coop, and possibly runs a migration from local AppData if found there. The `playlists` directory isn't actually a part of the foobar2000 installation, but is used to check if it's a fresh installation via scoop. The else section and uninstaller ensures the `playlists-vX.X` folders are carried over between updates (if `-p` isn't used).

linsui: I mean, why we need to copy the playlist to persist_dir in installer? We will copy it in uninstaller when update.

Velgus: Oh sorry, I thought you were talking about L39. I see what you mean, I'll remove the redundant installer copy.

bucket/foobar2000.json
```diff
@@ -28,22 +29,20 @@
...
             "if (Test-Path \"$persist_dir\\theme.fth\") {",
             "    Copy-Item \"$persist_dir\\theme.fth\" \"$dir\" -Force",
             "}"
         ]
     },
-    "post_install": "if (is_directory \"$dir\\theme.fth\") { Remove-Item \"$dir\\theme.fth\" -Force }",
     "uninstaller": {
         "script": [
-            "Copy-Item \"$dir\\playlists-v*\\*\" \"$persist_dir\\playlists\" -Force  -Recurse",
+            "Copy-Item \"$dir\\playlists-v*\" \"$persist_dir\\playlists\" -Force  -Recurse",
```
linsui:
```suggestion
            "Copy-Item \"$dir\\playlists-v*\" \"$persist_dir\\playlists\" -Force -Recurse",
```

My fault, sorry.
When update from 1.3 to 1.4, the struct is 
`persist_dir > playlist > playlist-v1.3` and `dir > playlist-v1.3` .
Then what will foobar do? Will it rename playlist-v1.3 to playlist-v1.4?
If then, we will get `persist_dir > playlist > playlist-v1.3, playlist-v1.4` and `dir > playlist-v1.3, playlist-v1.4` in the next update.

Velgus: After a bit of testing, how it works is if `playlists-v1.3` is detected in version 1.4, foobar converts them to the new format used in 1.4 automatically. 

So `playlists-v1.3` does need to be moved during upgrades, but isn't actually needed after foobar has converted them (the converted playlists will be needed in `playlists-v1.4`. By default foobar doesn't do anything with the unused earlier playlist folder - after converting, it just leaves it.

linsui: Thanks, I got it. If I just copy playlists-v1.4 to $dir when update from 1.3 as what I did, foobar can't load playlist automatically. But now you also persist the outdate playlist. So I suggest that we only copy the current playlist.

```
    "installer": {
        "script": [
            "Remove-Item \"$dir\\`$*\", \"$dir\\uninstall.exe\" -Force -Recurse",
            "New-Item \"$dir\\portable_mode_enabled\" | Out-Null",
            "if(!(Test-Path \"$persist_dir\")) {",
            "    if (Test-Path \"$env:Appdata\\foobar2000\") {",
            "        Write-Host -F yellow \"Copying old '$env:Appdata\\foobar2000' to '$persist_dir'\"",
            "        Copy-Item \"$env:Appdata\\foobar2000\\*\" \"$dir\" -Recurse",
            "    }",
            "}",
            "else {",
            "    Move-Item \"$persist_dir\\playlists-v*\" \"$dir\"",
            "}",
            "if (Test-Path \"$persist_dir\\theme.fth\") {",
            "    Copy-Item \"$persist_dir\\theme.fth\" \"$dir\"",
            "}"
        ]
    },
    "uninstaller": {
        "script": [
            "$shortVersion = $version.Split('.')[0, 1] -Join '.'",
            "if (Test-Path \"$dir\\playlists-v$shortVersion\") {",
            "    Copy-Item \"$dir\\playlists-v$shortVersion\" \"$persist_dir\" -Recurse",
            "}",
            "if (Test-Path \"$dir\\theme.fth\") {",
            "    Copy-Item \"$dir\\theme.fth\" \"$persist_dir\"",
            "}"
        ]
    },
```

Velgus: Sure - made the change.

linsui:
Note: foobar2000 loads playlist automatically from `playlists` if there is no `playlist-v*` . I'm not sure what we can do to simplify the script.

Velgus:
It depends if we want to support playlist upgrading from 1.3 versions to 1.4 versions. If not, it's fairly simple - we can just copy the `$shortVersion` playlist folder to `playlists`, then delete all `playlists-v*` folders from `$dir`.

If we did want to support it, we could handle it from the uninstaller. Simply have a migration update use the `playlists-v*` versions, so when the user launches it will move the files. Then when running the uninstaller it saves the contents of the latest `$shortVersion` playlist folder to a persisted `playlists` folder.

The question is how does foobar2000 behave if you have out-of-date playlist files in the `playlists` folder? Does it detect them as out-of-date, and update them automatically?

linsui:
The script already supports upgrading. This note may be useful in the future. foobar2000 will load playlist from `playlists` even if they are out of date and generate `playlists_v*` in new format. So we can persist `playlists` but we need to replace files in it when uninstalling. As it works well now, I thought this can be merged. :smile:

Velgus:
> EDIT: if users don't run foobar2000 between two update, they may lose their playlists.

I don't think there's a way we can avoid that short of replicating the process foobar2000 takes to update the playlists - which would be far more effort than it's worth.

bucket/foobar2000.json
```diff
@@ -18,34 +18,35 @@
...
         ]
     },
-    "post_install": "if (is_directory \"$dir\\theme.fth\") { Remove-Item \"$dir\\theme.fth\" -Force }",
     "uninstaller": {
         "script": [
-            "Copy-Item \"$dir\\playlists-v*\\*\" \"$persist_dir\\playlists\" -Force  -Recurse",
+            "$shortVersion = $version.Split('.')[0, 1] -Join '.'",
+            "if (Test-Path \"$dir\\playlists-v$shortVersion\") {",
+            "    Copy-Item \"$dir\\playlists-v$shortVersion\" \"$persist_dir\" -Recurse",
+            "}",
```
linsui:
```suggestion
            "}",
            "else { Copy-Item \"$dir\\playlists*\" \"$persist_dir\" -Recurse }",
```

If there is no new playlist, just persist the old one.

bucket/foobar2000.json
```diff
@@ -18,34 +18,36 @@
...
             "Remove-Item \"$dir\\`$*\", \"$dir\\uninstall.exe\" -Force -Recurse",
-            "New-Item \"$dir\\portable_mode_enabled\" -Force | Out-Null",
-            "if(!(Test-Path \"$persist_dir\\playlists\")) {",
-            "    New-Item \"$persist_dir\\playlists\" -ItemType Directory -Force | Out-Null",
+            "New-Item \"$dir\\portable_mode_enabled\" | Out-Null",
+            "if(!(Test-Path \"$persist_dir\")) {",
             "    if (Test-Path \"$env:Appdata\\foobar2000\") {",
             "        Write-Host -F yellow \"Copying old '$env:Appdata\\foobar2000' to '$persist_dir'\"",
-            "        Copy-Item \"$env:Appdata\\foobar2000\\*\" \"$dir\" -Recurse -Force",
+            "        Copy-Item \"$env:Appdata\\foobar2000\\*\" \"$dir\" -Recurse",
```
ebugusey: I'm against this kind of data migrations in install script because user doesn't have any control over what is happening.
This should be removed and replaced with notice, like in [vscode-portable](https://github.com/lukesampson/scoop-extras/blob/1ef0112d03f4b55fbd4720cf670df4af5098a7f0/bucket/vscode-portable.json#L10-L12).

linsui: This script won't remove any data.

ebugusey: It is still a bad idea to copy data from folder you have no control. You can easily break installation that way.
I'm actually affected by this, because without warning installation script changed in newer version and it copied **everything** from this folder.
There can be other files beside folder this script trying to persist.

ebugusey: Other problem is: this script is akin to **chocolatey** packages, and **scoop** isn't **chocolatey**.

Velgus:
Closing in favor of #2884. I just want the manifest fixed so updates work with plugins personally, don't care the how it's done.

PR #2876, opened 2019-09-23
retroarch: fix hash

hsiktas:
This fixes https://github.com/lukesampson/scoop-extras/issues/2854 and https://github.com/lukesampson/scoop-extras/issues/2858.


hsiktas:
had some tooling issues (too many vscode extensions) and started a new pull request

PR #2884, opened 2019-09-25
foobar2000: partial revert of recent "improvements" (add portable)

ebugusey:
tl; dr; It's not enough to persist `configuration`, `index-data` and `library`
folders. This persistence creates more problems than it solves.

foobar2000 config shouldn't be persisted that way. Entire config folder should
be included in persisted folders, because plugins use that folder to store
their files as well.

But this isn't possible, because "portable mode" for foobar2000 means moving
`$APPDATA/Roaming/foobar2000/*` to location where foobar2000 installed, and this
in return means you can't persist contents of every folder because you
don't know beforehand what to persist.

### DIFF
--- a/bucket/foobar2000.json
+++ b/bucket/foobar2000.json
@@ -23,7 +23,7 @@
     },
     "checkver": {
         "url": "https://www.foobar2000.org/download",
-        "re": "foobar2000_v([\\d\\.]+)\\."
+        "regex": "foobar2000_v([\\d.]+)\\."
     },
     "autoupdate": {
         "url": "https://www.videohelp.com/download/foobar2000_v$version.exe#/dl.7z"
--- a/bucket/foobar2000.json
+++ b/bucket/foobar2000.json
@@ -8,6 +8,12 @@
     },
     "url": "https://www.videohelp.com/download/foobar2000_v1.4.8.exe#/dl.7z",
     "hash": "993315168f76d1e91685805f6c746315c1cc34f73cfd44bf6bddab070897c2bf",
+    "notes": [
+        "WARNING: foobar2000 package is reverting back to non-portable installation.",
+        "If you still want portable version, use foobar2000-portable instead.",
+        "If you want to migrate persisted data to portable version",
+        "please rename $persist_dir to \"foobar2000-portable\"."
+    ],
     "bin": "foobar2000.exe",
     "shortcuts": [
         [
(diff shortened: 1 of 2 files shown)


linsui:
I don't know much about plugins. Maybe a foobar200-portable is a good idea. But please extract file to avoid privilege instead of run installer.

ebugusey:
> But please extract file to avoid privilege instead of run installer.

I have reverted changes. It is now, how it was before. It worked before without privileges.

Installer writes to registry, I can't remove this step. What it writes -- I don't really know, it's not open source and can be easily changed in next version. I don't want to maintain those changes.

ebugusey:
Perhaps it should be moved to **nonportable** bucket.

linsui:
> It worked before without privileges.

Oh, yes, you are right, it doesn't need privilege. 

> Installer writes to registry, I can't remove this step. What it writes -- I don't really know, it's not open source and can easily change in next version. I don't want to maintain those changes.

I don't think you need to worry about that... It's how scoop works.

ebugusey:
@linsui Actually found a way to install without running installer.

bucket/foobar2000.json
```diff
@@ -15,43 +15,15 @@
...
-            "if (Test-Path \"$dir\\theme.fth\") {",
-            "    Copy-Item \"$dir\\theme.fth\" \"$persist_dir\" -Force",
-            "}"
+            "New-Item \"$dir\\user_profiles_enabled\" -Force | Out-Null"
         ]
     },
     "checkver": {
         "url": "https://www.foobar2000.org/download",
-        "regex": "foobar2000_v([\\d.]+)\\."
+        "re": "foobar2000_v([\\d\\.]+)\\."
```
linsui:
```suggestion
        "regex": "foobar2000_v([\\d.]+)\\."
```

ebugusey:
@r15ch13 Can we merge it?

r15ch13:
This change will break updates for people that have foobar installed right now. Please add notes or a script to transition to the new portable version.

bucket/foobar2000.json
```diff
@@ -8,6 +8,12 @@
     },
     "url": "https://www.videohelp.com/download/foobar2000_v1.4.8.exe#/dl.7z",
     "hash": "993315168f76d1e91685805f6c746315c1cc34f73cfd44bf6bddab070897c2bf",
+    "notes": [
+        "WARNING: foobar2000 package is reverting back to non-portable installation.",
+        "If you still want portable version, use foobar2000-portable instead.",
+        "If you want to migrate persisted data to portable version",
+        "please rename $persist_dir to \"foobar2000-portable\"."
+    ],
```
ebugusey: @r15ch13 Is this enough?

PR #2891, opened 2019-09-27
Rename calibre-normal.json to calibre-np.json

OranShuster:
calibre-normal is for the non portable version of calibre
*-normal is not used anywhere else and other non-portable version use the -np suffix. 
you might also argure that calibre-normal(-np) should not be in the "extra" bucket but in the "np" bucket


JanPokorny:
I think the right course would be to add `calibre-np` to the nonportable bucket and possibly mark this one as deprecated.

TheRandomLabs:
I've added `calibre-np` to the nonportable bucket.

PR #2905, opened 2019-09-28
klatexformula:  Add version 4.0.0

linsui:


bucket/klatexformula.json
```diff
@@ -0,0 +1,30 @@
...
+            "klatexformula"
+        ]
+    ],
+    "shortcuts": [
+        [
+            "bin\\klatexformula.exe",
+            "kLatexFormula"
+        ]
+    ],
+    "post_install": "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall*\" -Recurse",
```
Ash258: pre

bucket/klatexformula.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "4.0.0",
+    "homepage": "https://klatexformula.sourceforge.io",
+    "description": "An easy-to-use graphical application for generating images from LaTeX equations.",
+    "license": "GPL-2.0-or-later",
+    "depends": "latex",
```
Ash258: Where is mentioned you need to have miktex installed?

linsui: https://klatexformula.sourceforge.io/downloads#system-requirements

Ash258: > On Windows, you **can** install

Why do you force user to have Miktex installed? Why I cannot use installation of Texlive?

linsui: I'll change it to suggest.

PR #2906, opened 2019-09-28
DeaDBeef: add version 2019-09-08

linsui:


bucket/deadbeef.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "2019-09-08",
+    "description": "A music player",
```
Ash258: There is not mentioned it is unofficial port of this player for unix based OSes

bucket/deadbeef.json
```diff
@@ -0,0 +1,34 @@
...
+    "homepage": "https://deadbeef-for-windows.github.io/",
```
Ash258:
```suggestion
    "homepage": "https://deadbeef-for-windows.github.io",
```

bucket/deadbeef.json
linsui: What's your suggestion? How about `A port of DeaDBeeF music player.` ?

Ash258: > There is not mentioned it is unofficial port

http://deadbeef.sourceforge.net/download.html
> Windows builds
> DeaDBeeF builds for Windows
> Please note, that while it should be possible to build DeaDBeeF for Windows from the official git repository, the builds above are not official and not supported. Please report issues to the DeaDBeeF for Windows Project.

linsui: So `Unofficial port of DeaDBeeF music player.` ?

PR #2907, opened 2019-09-28
Poedit: Add version 2.2.3

linsui:


bucket/poedit.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version": "2.2.3",
+    "homepage": "https://poedit.net/",
```
Ash258:
```suggestion
    "homepage": "https://poedit.net",
```

bucket/poedit.json
```diff
@@ -0,0 +1,26 @@
...
+    "description": "Gettext translations editor.",
+    "license": {
+        "identifier": "Shareware,MIT",
+        "url": "https://poedit.net/download"
```
Ash258: Shareware is not mentioned anywhere on this page

linsui: > Up-to-date source code of the Open Source version of Poedit (sans the Pro features included in the above binaries) is available at GitHub under the terms of the MIT license. 

It has a pro edition. Is it a shareware?

Ash258: Where is said you can distribute pro edition?

linsui: The pro edition can be bought and upgrade in app.

Ash258: Then how could it be shareware? 

> Provided free of charge to users, who are allowed and encouraged to make and share copies of the program

There is not mentioned sharing/distributing of program.

linsui: So freeware?

Ash258: https://github.com/vslavik/poedit


> License
> =========
> Poedit is released under the MIT license and you're free to do whatever you want with it and its source code (well, almost :-) -- see the license text).
> See the COPYING file for details on the program's licensing and the icons/README file for details on the icons.
> Windows and macOS versions contain GNU gettext binaries. They are distributed under the GNU General Public License and their source code is available from http://www.gnu.org/software/gettext. If you have difficulties getting them
from there, email me for a copy of the sources.
>>>

PR #2937, opened 2019-10-05
scrcpy: Add version 1.10

davidxuang:
Adding [scrcpy](https://github.com/Genymobile/scrcpy).


r15ch13:
Already available in the main bucket
https://github.com/ScoopInstaller/Main/blob/master/bucket/scrcpy.json

davidxuang:
so

PR #2971, opened 2019-10-13
Add WinMTR

JanPokorny:


JanPokorny:
@Ash258 Thank you, fixes applied.

PR #2972, opened 2019-10-13
snappy-driver-installer-origin: Add version 1.5.6.705

JanPokorny:
Note that unlike many apps that claim themselves to be a driver installer, SDIO is actually legit.


JanPokorny:
@Ash258 Thanks again, fixes applied.

PR #3001, opened 2019-10-18
Add WireGuard

TheZoker:
Add [WireGuard](https://www.wireguard.com/)

> WireGuard® is an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography. It aims to be faster, simpler, leaner, and more useful than IPsec, while avoiding the massive headache. It intends to be considerably more performant than OpenVPN. WireGuard is designed as a general purpose VPN for running on embedded interfaces and super computers alike, fit for many different circumstances. Initially released for the Linux kernel, it is now cross-platform (Windows, macOS, BSD, iOS, Android) and widely deployable. It is currently under heavy development, but already it might be regarded as the most secure, easiest to use, and simplest VPN solution in the industry.

### DIFF
--- a/bucket/wireguard.json
+++ b/bucket/wireguard.json
@@ -1,0 +1,38 @@
+{
+    "homepage": "https://www.wireguard.com",
+    "description": "Simple and fast VPN client.",
+    "version": "0.1.0",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://download.wireguard.com/windows-client/wireguard-amd64-0.1.0.msi",
+            "hash": "1bba81cc4e95bf09fcd1d1d63afa0ec19ed204407aee770dcce647e3c3b04262"
+        },
+        "32bit": {
+            "url": "https://download.wireguard.com/windows-client/wireguard-x86-0.1.0.msi",
+            "hash": "cc6f49c8f16dcba3d6998dbfdeb0835e5cc1240a59b92959fef82243c4e03df8"
+        }
+    },
+    "extract_dir": "WireGuard",
+    "bin": "wireguard.exe",
+    "shortcuts": [
+        [
+            "wireguard.exe",
+            "WireGuard"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.wireguard.com/install",
+        "regex": "distro-status-uptodate\">([\\d.]+)"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://download.wireguard.com/windows-client/wireguard-amd64-$version.msi"
+            },
+            "32bit": {
+                "url": "https://download.wireguard.com/windows-client/wireguard-x86-$version.msi"
+            }
+        }
+    }
+}


bucket/wireguard.json
```diff
@@ -0,0 +1,34 @@
+{
+    "homepage": "https://www.wireguard.com/",
```
Ash258:
```suggestion
    "homepage": "https://www.wireguard.com",
```

bucket/wireguard.json
```diff
@@ -0,0 +1,34 @@
...
+    "description": "WireGuard® is an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.",
+    "version": "0.0.31",
+    "license": "GPLv2",
```
Ash258:
```suggestion
    "license": "GPL-2.0-or-later",
```

bucket/wireguard.json
```diff
@@ -0,0 +1,34 @@
+{
+    "homepage": "https://www.wireguard.com/",
+    "description": "WireGuard® is an extremely simple yet fast and modern VPN that utilizes state-of-the-art cryptography.",
```
Ash258:
```suggestion
    "description": "Simple and fast VPN client.",
```

bucket/wireguard.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "extract_dir": "WireGuard",
+    "bin": "wireguard.exe",
+    "shortcuts": [
+        [
+            "wireguard.exe",
+            "WireGuard"
+        ]
+    ],
+    "autoupdate": {
```
Ash258:
```suggestion
    "checkver": {
        "url": "https://www.wireguard.com/install",
        "regex": "distro-status-uptodate\">([\\d.]+)"
    },
    "autoupdate": {
```

TheZoker:
@Ash258 Thanks for the feedback! I implemented all the changes. But I have no clue, why the hashes are wrong 😕

linsui:
This doesn't work because wintun is not installed. See https://github.com/ScoopInstaller/Main/issues/397

bucket/wireguard.json
```diff
@@ -0,0 +1,38 @@
+{
+    "homepage": "https://www.wireguard.com",
+    "description": "Simple and fast VPN client.",
+    "version": "0.0.31",
+    "license": "GPL-2.0-or-later",
```
linsui:
```suggestion
    "license": "MIT",
```

> The kernel components are released under the GPLv2, as is the Linux kernel itself. Other projects are licensed under MIT, BSD, Apache 2.0, or GPL, depending on context.

See https://github.com/WireGuard/wireguard-windows

TheZoker:
Hmm that's bad. @Ash258 Do you have any suggestions for this?

TheZoker:
@r15ch13 @Ash258 Can we do something about this?

Error1000:
Could we at least add it to non-portable and install it to Program files and install drivers ( wintun )? I am new here so if I'm missing something please tell me.

linsui:
You can file an issue in the [non-portable bucket](https://github.com/TheRandomLabs/scoop-nonportable).

PR #3011, opened 2019-10-18
Add Icaros

TheRandomLabs:
I'm a little more unsure about this one, and I have the following concerns:

* Administrator rights are required to install the Icaros shell extensions.
* At the moment, if `scoop install icaros` and `scoop uninstall icaros` fail when run without administrator privileges. To make the install/uninstall script automatically show a UAC popup, I'd have to condense it into a single line, which would be a lot messier.
* I've made the installer script enable Icaros for all supported file extensions by default. If this is an issue, I can remove this line.

Would this belong in the extras bucket or the nonportable bucket?


TheRandomLabs:
Sorry about the commits; I kept discovering minor flaws after I thought I'd fixed everything. :p

Ash258:
I am against adding this into official bucket. Lots of registry manipulation and administrative rights.

TheRandomLabs:
I'm inclined to agree. I'll move it to the nonportable bucket.

PR #3021, opened 2019-10-21
pac: Creating matifest for Microsoft PowerApps Command line tools

shytikov:
I have created manifest for PAC (PowerApp CLI). It's used intensively for component development within Microsoft Dynamics applications, namely for custom components.

This is my first manifest, I hope it was done right.


shytikov:
Thank you, @Ash258! Should I implement these changes and push them again?

bucket/microsoft-powerapps-cli.json
```diff
@@ -0,0 +1,20 @@
...
+    "extract_dir": "tools",
+    "bin": "pac.exe",
+    "checkver": {
+        "url": "https://www.nuget.org/packages/Microsoft.PowerApps.CLI",
+        "regex": "([\\d.]+):"
+    },
+    "autoupdate": {
+        "url": "https://www.nuget.org/api/v2/package/Microsoft.PowerApps.CLI/$version#/dl.zip"
+    }
+}
```
linsui:
```suggestion
}

```

bucket/microsoft-powerapps-cli.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "1.0.6",
+    "description": "Simple, single-stop tool for developing customizations and extensions for Microsoft PowerApps.",
+    "homepage": "https://docs.microsoft.com/en-us/powerapps/developer/component-framework/overview",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://www.nuget.org/packages/Microsoft.PowerApps.CLI/1.0.6/License"
+    },
+    "url": "https://www.nuget.org/api/v2/package/Microsoft.PowerApps.CLI/1.0.6#/dl.zip",
```
Ash258: Could you please verify if zip is enough? Maybe rename to 7z is needed as I do not know if native windows ExpandArchive could handle nupkges

Right now I cannot test as I am on mobile.

shytikov: Actually it does not matter what to put after `#` in the url... it will no matter of what will download same `.nupkg` file.

`.nupkg` is just renamed `.zip`, and it seems to be similar approach was taken in other packages, such as `nunit-console`.

Ash258: But there is different handler for zip files and for 7zip files.

Zips are extracted using Windows ExpandArhive function by default
7zips are extracted using 7zip call.

Ash258: Zip is enough

A

PR #3025, opened 2019-10-21
FileZilla-Server: Fix persist

issaclin32:
* Change the target of *Test-Path* from `$dir` to `$persist_dir`.

* **FileZilla Server** cannot handle the empty `FileZilla Server.xml`, which results in an error when saving user settings. (`protocol error: Invalid data, could not import account settings. Could not change account settings`) This fixes the problem by adding required fields to `FileZilla Server.xml`.

* Rename the shim-exe from `filezilla_server` to `filezilla-server`. (more intuitive, because it's the same name as package name `filezilla-server`)


Ash258:
#2425

issaclin32:
Thanks for the info.

It seems that **FileZilla Server** handles the config files in a different way (compared to FileZilla FTP Client). According to [this thread from its official forum](https://forum.filezilla-project.org/viewtopic.php?t=650) , FileZilla Server only stores its settings in `Filezilla Server.xml` under installation directory.

PR #3028, opened 2019-10-22
uget-portable:Add persistence

ef6:
`share\locale` for storing translated documents


Ash258:
What do you mean with `translated documents`? I cannot find any reference to capability of translating downloaded documents or anything like thi

linsui:
> The following downloads are for pre-compiled packages that you can use to install uGet. 7z Archive is the main file that includes the uGet application. Locale 7z is the archive that includes the Translations files, if your preferred language is something other than English, you’ll need to download the Locale file.

See https://ugetdm.com/downloads/windows/

Ash258:
Would not make sense to rather download all locales and extract it?

ef6:
Most people whose native language is not English need to extract translation files from the `LOCALE.7Z` to the `share\locale `

If we add it to the persistence, then when people upgrade the small version(look like 2.2.1->2.2.2), there is no need to add translation files.

ef6:
Of course.I assume the translation files are unchanged when uget upgrade a small version.

linsui:
After https://github.com/lukesampson/scoop/pull/3518 is merged we can update the LOCALE.7Z.

PR #3032, opened 2019-10-22
Pencil: update to 3.1.0

linsui:


bucket/pencil.json
```diff
@@ -1,22 +1,22 @@
 {
-    "version": "3.0.4",
+    "version": "3.1.0",
     "description": "GUI prototyping tool.",
-    "homepage": "https://pencil.evolus.vn/",
+    "homepage": "https://pencil.evolus.vn",
     "license": {
-        "identifier": "GPL-2.0-only|Freeware",
+        "identifier": "GPL-2.0-only",
```
Ash258: It would be nice to mention that commercial license.

I do not know how extractly, but it could be `Custom`.

linsui: I just thought that it makes no sense for end users. Should we add another identifier?

Ash258: Then you could just keep license as string since it is valid SPDX

PR #3045, opened 2019-10-23
Updated wsl-ssh-pageant version to 20190527.1

jeeger:
Updated wsl-ssh-pageant to newer version.

This has the `-force` option, which comes in handy for autostarting the daemon.

### DIFF
--- a/bucket/wsl-ssh-pageant.json
+++ b/bucket/wsl-ssh-pageant.json
@@ -2,7 +2,7 @@
     "homepage": "https://github.com/benpye/wsl-ssh-pageant/",
     "description": "A Pageant -> TCP bridge for use with WSL, allowing for Pageant to be used as an ssh-ageant within the WSL environment.",
     "license": "BSD-2-Clause",
-    "version": "20190513.14",
+    "version": "20190527.1",
     "bin": [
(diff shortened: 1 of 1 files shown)


jeeger:
Ack, feel free to edit the commit message.

jeeger:
Hello! What keeps this from getting merged? I could really use that `--force` option.

linsui:
What's the difference between the two binaries? Would it be better to put the cli into main bucket and live gui here? Then they will get autoupdate.

jeeger:
The GUI can run in the background without keeping a command window open (and supports a tray icon), the command line version can't. Apart from that there's no difference.

Why would splitting them in the main bucket and extra change this?

linsui:
Before https://github.com/lukesampson/scoop/pull/3518 is merged, scoop can only autoupdate the first url. Since they are standalone apps, maybe users want to install only one of them?

Ash258:
Well simple reason why this is not going in is that it is not final release. It is prerelease. If you want it create custom gist with it or push it to versiosn bucket.

jeeger:
Hello!
Ah, thanks for the information.

PR #3049, opened 2019-10-23
musikcube: add version 0.65.0

linsui:


Ash258:
Why there are no release artifacts for 0.65.1 (except macos)?

linsui:
It's a minor fix only for macos.

PR #3058, opened 2019-10-24
papercut: persist rules.json

abejenaru:


bucket/papercut.json
```diff
@@ -5,11 +5,17 @@
     "license": "Apache-2.0",
     "url": "https://github.com/ChangemakerStudios/Papercut/releases/download/5.5.1/Papercut.5.5.1.zip",
     "hash": "ec67c13414057cf9bf4e08002bc39ba148e2e79c7ec1d321a599f05374f7f26d",
+    "pre_install": [
```
Ash258:
```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\rules.json\")) { Set-Content \"$dir\\rules.json\" '[]' -Encoding Ascii }",
```

bucket/papercut.json
```diff
@@ -5,11 +5,17 @@
...
+         "    New-Item -Force -Path \"$dir\" -Name $file -Value '[]' | Out-Null",
+        "}"
+    ],
     "persist": [
         "Incoming",
-        "Logs"
+        "Logs",
+        "rules.json"
     ],
-    "bin": "Papercut.exe",
```
Ash258: Why was bin removed?

abejenaru: It'a s GUI app - the shortcut is enough.

abejenaru:
> Link to documentation where is this configuration file mentioned?

From application log: "Attempting to Load Rules from "C:\Users\XXX\scoop\apps\papercut\current\rules.json"

bucket/papercut.json
Ash258: I do not see a reason to remove binary. There are lots of manifests with applications which are graphical only and have binary.

abejenaru: In this case, launching the shim gives an empty console window that stays on :(
I can't see any benefit from keeping the shim.

PR #3059, opened 2019-10-24
filezilla: fix default settings in config file

abejenaru:
Comment out <Setting name="Config Location">$SOMEDIR/filezilla/</Setting> (doesn't work anyway, $SOMEDIR is not defined) and <Setting name="Cache directory">$USERPROFILE/Documents</Setting>.
Set Disable update check to 1 - don't check for new FileZilla versions as Scoop handles the updates.


Ash258:
#2328 #2425 #3025

PR #3062, opened 2019-10-26
nirlauncher: Simplify manifest

Ash258:
Additional tweaks to 9b1c2f069fcb6fa96836fe688abe0dd01aaf8572

There is no need to call 7zip manually. `Switches` parameter is available

A

@rasa Do not know if your script is still relevant and is using. If yes i will add notes back and tweak script (if needed)


rasa:
I ran https://github.com/rasa/scoops/blob/master/nlp2json.py on it, and there are no new apps. I think the comment is helpful, but not required.

PR #3065, opened 2019-10-27
update keepassxc to 2.5.0

jeremy886:


r15ch13:
Thanks, it seems my update process is not working. This should have been updated.

PR #3089, opened 2019-11-01
mu-editor: Add version 1.0.2

niheaven:


Ash258:
Does this work in all powershell versions? I think ps6 has AsByteStream instead od byte encoding

niheaven:
For PS 6 & 7, it should use `-AsByteStream` here.

Ash258:
Would not be more suitable something like this:

```powershell
if ($PSVersionTable.PSVersion.Major -gt 5) {
	$param = @{
		'AsByteStream' = $true
	}
} else {
	$param = @{
		'Encoding' = 'Byte'
	}
}

get-content $file @param
Set-content $file @param -nonewline
```

bucket/mu-editor.json
```diff
@@ -0,0 +1,55 @@
...
+        },
+        "32bit": {
+            "url": "https://github.com/mu-editor/mu/releases/download/1.0.2/mu-editor_1.0.2_win32.exe#/dl.7z",
+            "hash": "576baea0bd64649583a61bf6ce2836a91b02dce450c4b886d6adc7c35460ae8c"
+        }
+    },
+    "pre_install": [
+        "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst*\" -Recurse -Force",
+        "if ((Get-Command Get-Content).Parameters.ContainsKey('AsByteStream')) {",
+        "   $enc = @{",
```
Ash258: You can keep initialization as one line.

```suggestion
        "   $enc = @{ 'AsByteStream' = $true }",
```

PR #3093, opened 2019-11-01
waterfox: use "classic" branch, add waterfox-current package for "current" branch

issaclin32:

**Waterfox** has [split their development into 2 branches](https://www.waterfox.net/blog/waterfox-2019.10-release-download/).

The **classic** branch is the legacy branch that supports all the old standards (it will not be retired, and will be actively maintained). The **current** branch introduces new features and options that the classic version does not support.

https://www.waterfox.net/ has a **download button** on its homepage, and it leads to the classic branch. 

Therefore, I tend to link `waterfox` package to the **classic branch**, and add another `waterfox-current` package for **current branch**. So that users using the `waterfox` package can keep their software updated.

Please tell me if there is a better idea. Thanks!

### DIFF
--- a/bucket/waterfox-current.json
+++ b/bucket/waterfox-current.json
@@ -18,8 +18,8 @@
         ]
     ],
     "checkver": {
-        "url": "https://www.waterfox.net/update/win64/56.0/en-US/release/update.xml",
-        "xpath": "updates/update/@version"
+        "url": "https://www.waterfox.net/releases/",
+        "regex": "Waterfox%20Current%20(\\d{4}\\.\\d{1,2})%20Setup.exe"
     },
     "autoupdate": {
         "architecture": {
--- a/bucket/waterfox-current.json
+++ b/bucket/waterfox-current.json
@@ -1,0 +1,31 @@
+{
+    "version": "2019.10",
+    "description": "The 100% fresh, free-range, ethical browser.",
+    "homepage": "https://www.waterfox.net",
+    "license": "MPL-2.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://storage-waterfox.netdna-ssl.com/releases/win64/installer/Waterfox%20Current%202019.10%20Setup.exe#/dl.7z",
+            "hash": "d208a40ad818a526da8b8ab8cb43a084658172fea6393a785b3f7a29c00e1f1a"
+        }
+    },
+    "extract_dir": "core",
+    "bin": "waterfox.exe",
+    "shortcuts": [
+        [
+            "waterfox.exe",
+            "Waterfox"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.waterfox.net/update/win64/56.0/en-US/release/update.xml",
+        "xpath": "updates/update/@version"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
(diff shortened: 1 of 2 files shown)


linsui:
https://www.waterfox.net/blog/waterfox-2019.10-release-download/

Classic branch or current version, which is better?

issaclin32:
> https://www.waterfox.net/blog/waterfox-2019.10-release-download/
> 
> Classic branch or current version, which is better?

I tend to add another `waterfox-current` package for `current` version, since https://www.waterfox.net/ puts `classic` version on the download button. 

Maybe they think users that have been using `Waterfox` may be more used to the `classic` version. (not sure about the details, though)

bucket/waterfox-current.json
```diff
@@ -0,0 +1,31 @@
...
+    "extract_dir": "core",
+    "bin": "waterfox.exe",
+    "shortcuts": [
+        [
+            "waterfox.exe",
+            "Waterfox"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.waterfox.net/update/win64/56.0/en-US/release/update.xml",
```
linsui: This maybe doesn't work.

```xml
<patch type="complete" URL="https://storage-waterfox.netdna-ssl.com/releases/win64/update/waterfox-classic-2019.10.en-US.win64.complete.xz.mar" hashFunction="SHA512" hashValue="137a696c61d46783759a6661b7d406d14c5175de7bcd0c8fc759361ab76ad5a25fc15bd9ae3b5050360bfa667cbd8bf59940cce0ed6104dd6e2532108a6f2f26" size="86254889"/>
```

issaclin32: Do you mean the SHA512 hashValue cannot be applied to `Waterfox Current 2019.10 Setup.exe`?  

or the update channel `https://www.waterfox.net/update/win64/56.0/en-US/release/update.xml` may not be maintained in the future?

I think `https://storage-waterfox.netdna-ssl.com/releases/win64/update/waterfox-classic-2019.10.en-US.win64.complete.xz.mar` (and the hash value) is intended to be used by the update program of Waterfox browser.

linsui: ```waterfox-classic-2019.10.en-US.win64.complete.xz.mar```

This url ckeck update for version 56.0 and return classic-2019.10, I wonder if it works for current branch and classic branch. Maybe we need to update this url?

issaclin32: https://www.waterfox.net/blog/waterfox-2019.10-release-download/

Looks like the classic and current branch will be released at the same time.
However, I found there is no content at `https://www.waterfox.net/update/win64/2019.10/en-US/release/update.xml`, so it is possible that the XML channel will no longer be maintained.

Therefore, it makes more sense to check version updates from the homepage.

linsui: > However, I found there is no content at https://www.waterfox.net/update/win64/2019.10/en-US/release/update.xml, so it is possible that the XML channel will no longer be maintained.

It's correct. Only when there is a new version there is content. I agree to change checkver because it seems that its update url is not as stable as firefox.

rasa:
The files descriptions should end with `(classic)` and `(current)`, and it would be helpful to document in either the description, notes, or comments, what the differences between the two versions are, IMO.

PR #3099, opened 2019-11-02
flux: extract instead install

linsui:
fix https://github.com/lukesampson/scoop-extras/issues/1928 ? If we don't run the uninstaller, we won't lose anything.


h404bi:
It should. I didn't test.

PR #3102, opened 2019-11-02
Gammy: Add version 0.9.52

linsui:


Ash258:
There is `0.9.52a` release. https://github.com/Fushko/gammy/releases/tag/v0.9.52a

Why there are no binaries for it? Or 0.9.52 is just replaced with new `a` releases?

linsui:
There are some releases without binaries. I don't know why but as we checkver from its website it's ok.

Issue #2924, opened 2019-10-02
update flutter 1.9


DestinyHunter:

Findulas:
I changed the version regexp locally as stated in #2439. Is there a plan to change it in this bucket?

ratatoeey:
will need some testing as some popular manifests may need testing. and sometimes software updates break things.
- regards @ratatooey

Issue #2439, opened 2019-07-12
how to update flutter bucket


chatterzhao:
how to update flutter bucket

issaclin32:
You mean the `flutter` **package** under "extras" bucket?

Try `scoop update flutter`
(or `scoop update flutter -f` if you want to force re-install.)

chatterzhao:
> You mean the `flutter` **package** under "extras" bucket?
> 
> Try `scoop update flutter`
> (or `scoop update flutter -f` if you want to force re-install.)

Sorry, i want update manifest(我想学习怎么升级flutter.json)

rikuben:
Change the checkver regex from `"windows_v([\\w-.]+)-stable"` to `"windows_v([\\w+-.]+)-stable"`.
Then, run `.\bin\checkver.ps1 flutter -u`

chatterzhao:
@rikuben 
thanks


 :octocat: [From gitme Android](http://flutterchina.club/app/gm.html)

PR #3113, opened 2019-11-04
Flutter: 脚本变更

fanlide:
#2924 #2439

### DIFF
--- a/bucket/flutter.json
+++ b/bucket/flutter.json
@@ -1,21 +1,17 @@
 {
     "homepage": "https://flutter.dev/",
-    "version": "1.5.4-hotfix.2",
+    "version": "v1.9.1+hotfix.6",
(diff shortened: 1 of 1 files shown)


fanlide:
Version number is not satisfied `^[\w\.\-_]+$`, resulting in verification does not pass, please fix under. 

image

linsui:
See https://github.com/lukesampson/scoop-extras/pull/2770 https://github.com/lukesampson/scoop/pull/3613

fanlide:
@linsui I gave up regex, and used jsonpath. I believe my way is better.

linsui:
I agree, see https://github.com/ScoopInstaller/Main/pull/366 😂

fanlide:
I don't think adopt8 is necessary. Other jdk can still. I installed zulu8,. I don't want to install adopt8.

PR #3114, opened 2019-11-05
mozbackup: Add version 1.5.1

owendaprile:
Add [MozBackup](http://mozbackup.jasnapaka.com/).
It backs up profiles from Mozilla programs.

### DIFF
--- a/bucket/mozbackup.json
+++ b/bucket/mozbackup.json
@@ -1,0 +1,18 @@
+{
+    "version": "1.5.1",
+    "description": "MozBackup is a simple utility for creating backups of Mozilla Firefox, Mozilla Thunderbird, Mozilla Sunbird, Flock, SeaMonkey, Mozilla Suite, Spicebird, Songbird and Netscape profiles.",
+    "homepage": "http://mozbackup.jasnapaka.com/",
+    "license": {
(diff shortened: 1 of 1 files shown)


linsui:
It's open source, see http://mozbackup.jasnapaka.com/development.php . Would it be better to download from sourceforge for https?

owendaprile:
Like @linsui said, it's licensed under the three licenses I originally had. 
I'm not really sure what the extra files it has are, as it doesn't have any settings exposed in the GUI and does it's behavior doesn't change on subsequent runs.

Ash258:
Source code is licensed under all of that, but artifacts are freeware as stated on homepage.

> This program is freeware (even for commercial use) and works on Windows

See vscode as other example. Code is licensed under [MIT](https://github.com/microsoft/vscode/blob/master/LICENSE.txt), but final software is under [freeware proprietary license](https://code.visualstudio.com/License/)

bucket/mozbackup.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "1.5.1",
+    "description": "Create backups of profiles from Mozilla software.",
+    "homepage": "http://mozbackup.jasnapaka.com/",
```
Ash258:
```suggestion
    "homepage": "http://mozbackup.jasnapaka.com",
```

Ash258 (approved):

PR #3125, opened 2019-11-06
ydl-ui: Update to version 2.6.2 (manually)

issaclin32:

* [ydl-ui](https://github.com/Maxstupo/ydl-ui/releases) changed their file name back to the original pattern. This fixes the download URL.

* This also adds `download-list.json` and `ydl-ui.json` to *persist*.

* The *installer.script* will modify `ydl-ui.json` to disable update. (if previous persist data does not exist)

### DIFF
--- a/bucket/ydl-ui.json
+++ b/bucket/ydl-ui.json
@@ -8,10 +8,10 @@
     "extract_dir": "YDL_UI_v2.6.2_Portable",
     "installer": {
         "script": [
-            "if(!(Test-Path \"$persist_dir\\ydl-ui.json\")){",
+            "if ((!(Test-Path \"$persist_dir\\ydl-ui.json\")) -and (Test-Path \"$dir\\ydl-ui.json\")) {",
             "    $cont = Get-Content \"$dir\\ydl-ui.json\" | ConvertFrom-Json",
             "    $cont.CheckForUpdates = $false",
-            "    $cont | ConvertToPrettyJson | Set-Content \"$dir\\ydl-ui.json\"",
+            "    $cont | ConvertToPrettyJson | Set-Content \"$dir\\ydl-ui.json\" -Encoding Ascii -Force",
             "}"
         ]
     },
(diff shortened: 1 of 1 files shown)


issaclin32:
> This manifest is not really portable
> 
> Persist `download-list.json`, `ydl-ui.json` and disable updates

Thanks for the info. 

I implemented "disable update" using string replace. It is better than copying the whole `ydl-ui.json` from `(bucket_dir)\scripts` because there may be more items added into `ydl-ui.json` in the future.

Not sure if it's the best approach or not.

Ash258:
You could always use something like `$cont = (Get-Content ...); $cont.CheckForUpdates = $false; $cont | ConvertToPrettyJson | Set-Content l..`

But do not know if it is needed. Maybe for readability it could be better, cannot tell. Choose one and go with it.

issaclin32:
> You could always use something like `$cont = (Get-Content ...)`

Thanks for the reply.
I will choose your approach because it is easier to expand. (when you want to add more options into the config file)

bucket/ydl-ui.json
```diff
@@ -2,20 +2,33 @@
...
+    "version": "2.6.2",
+    "url": "https://github.com/Maxstupo/ydl-ui/releases/download/v2.6.2/YDL_UI_v2.6.2_Portable.zip",
+    "hash": "774a4bf579eaa253f4712d561f7a6b0b545a959fec8e6f17078963a453ef8542",
+    "extract_dir": "YDL_UI_v2.6.2_Portable",
+    "installer": {
+        "script": [
+            "if(!(Test-Path \"$persist_dir\\ydl-ui.json\")){",
+            "    $cont = Get-Content \"$dir\\ydl-ui.json\" | ConvertFrom-Json",
+            "    $cont.CheckForUpdates = $false",
+            "    $cont | ConvertToPrettyJson | Set-Content \"$dir\\ydl-ui.json\"",
```
Ash258:
```suggestion
            "    $cont | ConvertToPrettyJson | Set-Content \"$dir\\ydl-ui.json\" -Encoding Ascii -Force",
```

bucket/ydl-ui.json
```diff
@@ -2,20 +2,33 @@
...
+        ]
+    },
     "bin": "YDL-UI.exe",
     "shortcuts": [
         [
             "YDL-UI.exe",
             "YDL-UI"
         ]
     ],
+    "persist":[
```
Ash258:
```suggestion
    "persist": [
```

PR #3135, opened 2019-11-07
as-ssd: Add version 2.0.6821.41776

digrouz:


digrouz:
how did found the hash  and the checkver ? :)

Ash258:
Just add hash property with some random string adn execute checkver with force parameter.

For update url

1. Use Fiddler
1. Enable listening (if none)
1. Click `Help -> Updates`
1. See relevant requests in fiddler
1. Copy request URL

digrouz:
nice, thanks for the infos!

PR #3156, opened 2019-11-11
lazydocker: Add version 0.7.5

reicheltp:
`Lazydocker` is "The lazier way to manage everything docker".

It is widely used with >36k downloads and >12k stars on GitHub.

As of version 0.7.5 it supports windows as platform.
See: https://github.com/jesseduffield/lazydocker/releases/tag/v0.7.5


reicheltp:
@Ash258 Done! See: https://github.com/ScoopInstaller/Main/pull/558

PR #3160, opened 2019-11-13
Update autoupdate paths for notepadplusplus.json

idan-weizman:
Notepad++ download repository path has changed.
This commit updates it to the new path.

P.S. There does not seem to be an https path for those downloads. A real shame.
This should be changed back to https when it (hopefully) becomes available again.
For now, the hashes could at least be extracted from github instead, using https.


linsui:
duplicate of https://github.com/lukesampson/scoop-extras/pull/2962 https://github.com/lukesampson/scoop-extras/pull/3123

PR #3172, opened 2019-11-14
hddguardian: Add version 0.7.1

digrouz:


Ash258:
Using snapfiles and codeplex is not really a good idea for wider user base

image

digrouz:
The fact is that the official website is there now :(

digrouz:
are you suggesting to keep this in a private bucket ?

Ash258:
I am only telling that this cannot go into official bucket unless there is safe provider for homepage and storage. What you will do with it is up to you.

digrouz:
understood :)

PR #3173, opened 2019-11-14
ldap-browser: Add version 4.5.19808.0

abejenaru:


abejenaru:
> * Simple extraction is not working. Application is not started without installation
>   
>   * @abejenaru Look for solution

I thought it was portable, but I see the installed version was giving that impression. I cannot find a working solution, I think I should drop the PR.

PR #3196, opened 2019-11-16
openvpn: Add requirement checks

Ash258:
I hope I602 is not going to change

### DIFF
--- a/bucket/openvpn.json
+++ b/bucket/openvpn.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.4.8-I602",
+    "version": "2.4.8",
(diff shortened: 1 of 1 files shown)


r15ch13:
Hm, I guess it will probably change. For 2.4.7 it is i607 and there are a couple of commits that got those. https://github.com/lukesampson/scoop-extras/commits/806d09252b2aca0d2897882b3144da75da131206/bucket/openvpn.json
Didn't notice that when I changed it in https://github.com/lukesampson/scoop-extras/commit/c0813a20089673390536ae718fd3e09066534d44

Ash258:
There is no problem to change it back as part of version if there is going to be problem.

PR #3226, opened 2019-11-20
calibre: fix url

jawys:
Some time ago (calibre v4.1.0) I noticed failure during downloading from fosshub. I assume they changed their URL(s) for downloading?
First I checked if it is possible to correct the URL on fosshub itself but I realized that there is always website overhead since fosshub's downloads work via query parameter.
So finally I changed the url to `download.calibre-ebook.com` which seems to work pretty well.

Is there some testing pipeline to check my new change because I mainly tested downloading with curl?

regards


Ash258:
What about `calibre-normal`??

PR #3235, opened 2019-11-22
gatling: Add version 3.3.1

### LINKED ISSUE #3217: [Request] Gatling - Load Testing tool
https://gatling.io/open-source/start-testing/

Gatling is a developer tool to load test web applications

LoZeno:
Adds gatling version 3.3.1

Closes #3217

Gatling DEPENDS on Java, but I added it as a "suggest" because it doesn't look like "depends" can allow the user to choose whether to use the oracle version or openjdk version.

Also, I've put it here in extras instead of main because gatling recorder is not a commandline tool and pops a window application open (unlike k6)

### DIFF
--- a/bucket/gatling.json
+++ b/bucket/gatling.json
@@ -13,7 +13,7 @@
     "persist": [
         "conf",
         "results",
-        "user-files"        
+        "user-files"
     ],
     "env_set": {
         "GATLING_HOME": "$dir"
--- a/bucket/gatling.json
+++ b/bucket/gatling.json
@@ -10,6 +10,10 @@
         "gatling.bat",
         "recorder.bat"
     ],
+    "persist": [
+        "user-files",
+        "results"
+    ],
     "shortcuts": [
         [
             "recorder.bat",
--- a/bucket/gatling.json
+++ b/bucket/gatling.json
@@ -1,14 +1,14 @@
 {
     "homepage": "https://gatling.io",
-    "description": "Async Scala-Akka-Netty based Load Test Tool",
+    "description": "Async load test tool for web applications",
     "version": "3.3.1",
     "license": "Apache-2.0",
     "url": "https://repo1.maven.org/maven2/io/gatling/highcharts/gatling-charts-highcharts-bundle/3.3.1/gatling-charts-highcharts-bundle-3.3.1-bundle.zip",
     "hash": "sha1:6d9630ef1c3af22b217bb7cef3c6179924886469",
     "extract_dir": "gatling-charts-highcharts-bundle-3.3.1",
     "bin": [
-        "gatling.bat",
-        "recorder.bat"
+        "bin\\gatling.bat",
+        "bin\\recorder.bat"
     ],
     "persist": [
         "user-files",
--- a/bucket/gatling.json
+++ b/bucket/gatling.json
@@ -11,8 +11,9 @@
         "bin\\recorder.bat"
     ],
     "persist": [
-        "user-files",
-        "results"
+        "conf",
+        "results",
+        "user-files"        
     ],
     "env_set": {
         "GATLING_HOME": "$dir"
--- a/bucket/gatling.json
+++ b/bucket/gatling.json
@@ -1,0 +1,36 @@
+{
+    "homepage": "https://gatling.io",
+    "description": "Async Scala-Akka-Netty based Load Test Tool",
+    "version": "3.3.1",
+    "license": "Apache-2.0",
+    "url": "https://repo1.maven.org/maven2/io/gatling/highcharts/gatling-charts-highcharts-bundle/3.3.1/gatling-charts-highcharts-bundle-3.3.1-bundle.zip",
+    "hash": "sha1:6d9630ef1c3af22b217bb7cef3c6179924886469",
+    "extract_dir": "gatling-charts-highcharts-bundle-3.3.1",
+    "bin": [
+        "gatling.bat",
+        "recorder.bat"
+    ],
+    "shortcuts": [
+        [
+            "recorder.bat",
+            "Gatling Recorder"
+        ]
+    ],
+    "suggest": {
+        "JDK": [
+            "java/oraclejdk",
+            "java/openjdk"
+        ]
+    },
+    "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/gatling.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "checkver": {
+        "url": "https://search.maven.org/solrsearch/select/?q=g:io.gatling.highcharts+AND+a:gatling-charts-highcharts-bundle",
+        "jp": "$.response.docs[0].latestVersion"
+    },
+    "autoupdate": {
+        "url": "https://repo1.maven.org/maven2/io/gatling/highcharts/gatling-charts-highcharts-bundle/$version/gatling-charts-highcharts-bundle-$version-bundle.zip",
+        "extract_dir": "gatling-charts-highcharts-bundle-$version",
+        "hash": {
+            "url": "$baseurl/gatling-charts-highcharts-bundle-$version-bundle.zip.sha1"
```
Ash258:
```suggestion
            "url": "$url.sha1"
```

bucket/gatling.json
```diff
@@ -0,0 +1,36 @@
...
+    "hash": "sha1:6d9630ef1c3af22b217bb7cef3c6179924886469",
+    "extract_dir": "gatling-charts-highcharts-bundle-3.3.1",
+    "bin": [
+        "gatling.bat",
+        "recorder.bat"
+    ],
+    "shortcuts": [
+        [
+            "recorder.bat",
+            "Gatling Recorder"
```
Ash258: Is there icon inside archive, which could be used?

Ash258:
You are not persisting `user-files` and `results`. What are these folders for?

bucket/gatling.json
LoZeno: As far as I can see, no icons or images are provided in the archive unfortunately

LoZeno:
I missed the `user-files` and `results` folders from the zip... they are used as the entry point where to include the files with the code for the requests to execute (`user-files`) and as the output folder with the stats from the load test results (`results`). I'll fix the manifest to persist them.

bucket/gatling.json
```diff
@@ -0,0 +1,43 @@
+{
+    "homepage": "https://gatling.io",
+    "description": "Async Scala-Akka-Netty based Load Test Tool",
+    "version": "3.3.1",
+    "license": "Apache-2.0",
+    "url": "https://repo1.maven.org/maven2/io/gatling/highcharts/gatling-charts-highcharts-bundle/3.3.1/gatling-charts-highcharts-bundle-3.3.1-bundle.zip",
+    "hash": "sha1:6d9630ef1c3af22b217bb7cef3c6179924886469",
+    "extract_dir": "gatling-charts-highcharts-bundle-3.3.1",
+    "bin": [
+        "gatling.bat",
```
Ash258:
```suggestion
        "bin\\gatling.bat",
```

bucket/gatling.json
```diff
@@ -0,0 +1,43 @@
...
+        "recorder.bat"
```
Ash258:
```suggestion
        "bin\\recorder.bat"
```

bucket/gatling.json
```diff
@@ -0,0 +1,43 @@
...
+    ],
+    "persist": [
```
Ash258: ??

A

bucket/gatling.json
```diff
@@ -0,0 +1,43 @@
+{
+    "homepage": "https://gatling.io",
+    "description": "Async Scala-Akka-Netty based Load Test Tool",
```
Ash258:
```suggestion
    "description": "Async load test tool for web applications",
```

bucket/gatling.json
```diff
@@ -0,0 +1,43 @@
...
+    "bin": [
+        "gatling.bat",
+        "recorder.bat"
+    ],
+    "persist": [
+        "user-files",
+        "results"
+    ],
+    "env_set": {
+        "GATLING_HOME": "$dir"
```
Ash258: Where is explicitly said you need this variable? Since scoop register binary, it should not be needed.

LoZeno:
> 
> 
> > Since 3.0, Gatling requires Scala 2.12. Gatling is not compatible with Scala 2.11 nor Scala 2.13.
> 
> > Gatling supports 64bits OpenJDK 8 and OpenJDK 11 with HotSpot. Other JVMs such as JDK 12+, client JVMs, 32bits systems or OpenJ9 are not supported.
> 
> You are suggesting oraclejdk, which is clearly not supported with this description.

That's the requirement for the Java plugin (Maven/Gradle), not for the Gatling Bundle (which is what this manifest is intended for).
The requirement for the bundle is the following:

> In order to run Gatling, you need to have a JDK installed. Gatling requires at least JDK8. Then, we recommend that you use an up-to-date version.

bucket/gatling.json
LoZeno: When you run it: if you run recorder.bat without setting that environment variable, this happens:
image

bucket/gatling.json
LoZeno: I'm a bit worried about persisting THIS conf folder: across the versions of the bundle, the .conf files have changed (adding both files AND entries in each file); what happens if a new version requires a new entry in the gatling.conf file but the previous version's is persisted? It might cause errors.

Ash258: So you are worried, but you still persist it for some reason.

LoZeno: Because after my comment I asked a few people who have been using gatling longer than I do and they told me they'd rather run the risk of having to deal with changes in the config structure than having to re-set the personalised configuration after a new version.

bucket/gatling.json
```diff
@@ -0,0 +1,44 @@
...
+    "hash": "sha1:6d9630ef1c3af22b217bb7cef3c6179924886469",
+    "extract_dir": "gatling-charts-highcharts-bundle-3.3.1",
+    "bin": [
+        "bin\\gatling.bat",
+        "bin\\recorder.bat"
+    ],
+    "persist": [
+        "conf",
+        "results",
+        "user-files"
```
Ash258:
```suggestion
        "user-files"
```

PR #3249, opened 2019-11-24
ungoogled-chromium: Update to 78.0.3904.108-1 (manually)

rasa:


Ash258:
This cannot work.

- Curly bracket substitution is not supported
- Version in URl is different than manifest version
	- `78.0.3904.108-1` => `78.0.3904.108-1.1`
- Filename is `ungoogled-chromium_78.0.3904.108-1.1_windows-x64.zip`

A

rasa:
> This cannot work.

@Ash258 Good point. I will try to rework it.

rasa:
@Ash258 Installed fine locally. I should *always* test locally from now on.

PR #3261, opened 2019-11-26
wincompose: Create settings.ini for persist

### LINKED ISSUE #1512: Wincompose creates folder instead of file
When Wincompose is installed, a persist folder called "settings.ini" is created. For Wincompose to work, it should be a file, and not a folder, called settings.ini.

bb010g:
Fixes #1512 like #1852, by preempting folder creation with a file.

---
This change is [Reviewable](https://reviewable.io/reviews/lukesampson/scoop-extras/3261)

### DIFF
--- a/bucket/wincompose.json
+++ b/bucket/wincompose.json
@@ -8,6 +8,7 @@
     "extract_dir": "WinCompose-NoInstall-0.9.3",
     "bin": "wincompose.exe",
     "persist": "settings.ini",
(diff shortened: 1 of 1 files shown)


bucket/wincompose.json
```diff
@@ -8,6 +8,9 @@
     "extract_dir": "WinCompose-NoInstall-0.9.3",
     "bin": "wincompose.exe",
     "persist": "settings.ini",
+    "pre_install": [
```
Ash258:
```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\settings.ini\")) { New-Item \"$dir\\settings.ini\" | Out-Null }",
```

PR #3276, opened 2019-11-27
ziglang-edge: Add ziglang edge

Nebulavenus:
Adds ziglang nightly


niheaven:
Move to https://github.com/ScoopInstaller/Versions, and run `checkver -f` first.

The version contains '+', so https://github.com/lukesampson/scoop/pull/3613 is needed.

Nebulavenus:
Moved to ScoopInstaller/Versions#185

PR #3287, opened 2019-12-01
arduino-portable: fix persistence

### LINKED ISSUE #3282: Arduino Portable installation not working
d0bcf77d0add5f06d9075503ba3d6d45acfe2795 changed _arduino-portable.json_ to persist 3 .txt files under the _hardware_ folder but these files do not exist in a standard installation.  Perhaps some function within the IDE creates them at a later point in time, but they are not in the .zip file that scoop expands so when the shortcuts are created they default to creating folder shortcuts.  When the IDE tries to start it finds these "files" and tries to load them but they are folder shortcuts instead of text files so the IDE fails to load.

Deleting the three .txt folder shortcuts fixes the problem and the IDE loads.
I am seeing this problem on Windows 10.

BlackPowerade:
Closes #3282

### DIFF
--- a/bucket/arduino-portable.json
+++ b/bucket/arduino-portable.json
@@ -7,6 +7,10 @@
     "hash": "sha512:cec645783225e2edf88acd6e759180434f7b1f6ffa5d7532d16ba37a976d3a0581cde64ea9c6e3331d04a32afc4265db0929f7e1b0786fd2748b0244143b2b47",
     "extract_dir": "arduino-1.8.10",
     "bin": "arduino.exe",
+    "pre_install": [
+        "'arduino\\avr\\platform.local.txt', 'arduino\\avr\\boards.local.txt', 'platform.local.txt' | ForEach-Object {",
+        "    if (!(Test-Path \"$persist_dir\\hardware\\$_\")) { New-Item \"$dir\\hardware\\$_\" | Out-Null }}"
+    ],
     "shortcuts": [
         [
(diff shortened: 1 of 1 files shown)


bucket/arduino-portable.json
```diff
@@ -7,6 +7,10 @@
     "hash": "sha512:cec645783225e2edf88acd6e759180434f7b1f6ffa5d7532d16ba37a976d3a0581cde64ea9c6e3331d04a32afc4265db0929f7e1b0786fd2748b0244143b2b47",
     "extract_dir": "arduino-1.8.10",
     "bin": "arduino.exe",
+    "pre_install": [
+        "'arduino\\avr\\platform.local.txt', 'arduino\\avr\\boards.local.txt', 'platform.local.txt' | ForEach-Object {",
+        "    if (!(Test-Path \"$persist_dir\\hardware\\$_\")) { New-Item \"$dir\\hardware\\$_\" | Out-Null }}"
```
Ash258:
```suggestion
        "    if (!(Test-Path \"$persist_dir\\hardware\\$_\")) { New-Item \"$dir\\hardware\\$_\" | Out-Null }",
```

bucket/arduino-portable.json
```diff
@@ -7,6 +7,10 @@
...
+    ],
```
Ash258:
```suggestion
        "}"
    ],
```

PR #3294, opened 2019-12-02
archi: Update to version 4.6.0

The-Loeki:
As the download path know seems to contain some unknowable hash, I've removed the `autoupdate` stanza


Ash258:
As hotfix why not. If someone found way how to properly autoupdate post PR.

r15ch13:
Easy: https://www.archimatetool.com/js/x.txt

niheaven:
Yes, I found `x.txt` before, but we need special url treatment, but it's not so useful for this single app...

PR #3300, opened 2019-12-03
xtremsplit: Add version 1.2.0.54

digrouz:
Any idea how the "12" can be managed in the autoupdate url ?


Ash258:
Applications without English homepage are not accepted into official buckets, as it is not suitable for majority of users.

digrouz:
Too bad, the application itself have an English parameter. I guess this is tipically french to create french only websites :)

PR #3307, opened 2019-12-04
fontbase: Add version 2.10.3

Ash258:


r15ch13:
The trusty [Fiddler](https://github.com/lukesampson/scoop-extras/blob/master/bucket/fiddler.json) finds another one: https://releases.fontba.se/win/latest.yml

PR #3308, opened 2019-12-04
spotify: Add nightly

Ash258:
Instllation, uninstallation and application itself is working well.
Currently being tested. I will see in few days how updates are behaving in custom folder

### DIFF
--- a/bucket/spotify.json
+++ b/bucket/spotify.json
@@ -1,0 +1,29 @@
+{
+    "version": "nightly",
+    "description": "Client for spotify music streaming service",
+    "homepage": "https://www.spotify.com",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://www.spotify.com/au/legal/end-user-agreement"
+    },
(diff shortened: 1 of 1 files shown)


ddavness:
image

Worked for like a couple times until I hit this.
Htting `Retry` will hang the installer forever. So will an attempt to abort it. (Only way is to bring it down via Task Manager).

Since then, the installation is now bricked. I have a suspicion that this might be happening to Premium users and might be due to the offline mode feature (download stuff).

## Repro Steps
- Log in from a premium account;
- Download your list of liked songs;
- Reboot (I guess);
- Congratulations, your installation is bricked.

ddavness:
As a side note, though, it'd be awesome if we could keep data in `persist` (login info, downloaded songs, etc.) and only delete it if we uninstalled with `-p` - even if that meant that we had to mess with the AppData directory (basically replace the `AppData/Roaming/Spotify` directory with a junction to `scoop/persist/spotify`)

PR #3319, opened 2019-12-06
miniget: Add version 1.1.10.1830

wenmin92:
[MiniGet](http://www.miniget001.com/): a Lightweight and Fast Download Manager & BitTorrent Client

### DIFF
--- a/bucket/miniget.json
+++ b/bucket/miniget.json
@@ -1,10 +1,10 @@
 {
     "homepage": "http://www.miniget001.com",
     "description": "A lightweight and fast download manager & BitTorrent client.",
-    "version": "1.1.10.1666",
+    "version": "1.1.10.1830",
     "license": "Unknown",
     "url": "http://www.miniget001.com/download/MiniGet.zip",
-    "hash": "6a77c803df8e0b714e91cac800b1a605a7ca1e65ea814d25827fe2536178ab94",
+    "hash": "8ad26e1853681c4e4ee678fa4eac03040967eb50c41e917dcbd9d016b305edc5",
     "extract_dir": "MiniGet",
     "persist": [
         "Config",
--- a/bucket/miniget.json
+++ b/bucket/miniget.json
@@ -1,0 +1,19 @@
+{
+    "homepage": "http://www.miniget001.com/",
+    "description": "a Lightweight and Fast Download Manager & BitTorrent Client.",
+    "version": "1.1.10.1666",
+    "license": "Unknown",
+    "url": "http://www.miniget001.com/download/MiniGet.zip",
+    "hash": "6a77c803df8e0b714e91cac800b1a605a7ca1e65ea814d25827fe2536178ab94",
+    "extract_dir": "MiniGet",
+    "shortcuts": [
+        [
+            "MiniGet.exe",
+            "MiniGet"
+        ]
+    ],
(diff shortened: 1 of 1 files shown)


bucket/miniget.json
```diff
@@ -0,0 +1,19 @@
+{
+    "homepage": "http://www.miniget001.com/",
+    "description": "a Lightweight and Fast Download Manager & BitTorrent Client.",
```
Ash258:
```suggestion
    "description": "A lightweight and fast download manager & BitTorrent client.",
```

r15ch13:
Website is offline.

Ash258:
No. It is available 

A

r15ch13:
Strange, I got 404 when tested back then. :sweat_smile:

wenmin92:
> No persist for json file and config, Torrent folders?

Thanks, I will check later.

Ash258:
Is there any specific dependency? On my workstation at work I am experiencing this:


image

wenmin92:
> Is there any specific dependency? On my workstation at work I am experiencing this:

I have asked the author about this problem, but not replied yet. I think this may be a package issue, the exe installer is Ok. We can wait some days or use the exe version(can be extract by 7z).

issaclin32:
> Is there any specific dependency? On my workstation at work I am experiencing this:

I got the same problem. When I search in the menu and set the language to *English*, the problem was solved.

I guess this problem will happen to all the PCs with a **default language that is not supported by Miniget**. `miniget` did not handle these situation properly (which is to set the language to *English*),  therefore all the dialogs appears to be empty. (I have not test it on a PC with English interface, so I can not be completely sure.)

It is possible to fix to problem by manually tweaking `MiniGetConfig.cfg`, but that will be error-prone and hard to maintain.
hex

wenmin92:
Hi @issaclin32 , Thanks for your test. My computer Language is English while Region is China. Still not work.

wenmin92:
Sorry for lost update. I'm guessing whether the author might update, so I went to the website today and found that there was indeed a small update on December 28. Tested, this new updated version can be used normally.

Ash258:
I will test again after I am back in the office on all workstations.

r15ch13:
Off-Topic: Regarding my network problem with the domain.
It gets blocked by the [maleware list](https://hosts-file.net/default.asp?s=www.miniget001.com) from [hpHosts](https://hosts-file.net/). It was added in 02-04-2018.

Ash258:
Default language after clean install is Espanol.

image

@wenmin92 Any response from author about https://github.com/lukesampson/scoop-extras/pull/3319#issuecomment-575765996 ?

wenmin92:
There is no other response. 
My clean install language is Chinese. I think the language is related to the system language setting.

rashil2000:
@issaclin32 @wenmin92 any updates on this?

issaclin32:
The official website (http://www.miniget001.com/) has been discontinued.
Also I cannot find a new one.

Normally I will try to find a file that matches the same hash, and add it to [Binary](https://github.com/ScoopInstaller/Binary).

But for the case of **MiniGet**, I would not do that because the software involves outbound connections, and it is risky to use an outdated one.

rashil2000:
Oh, thanks for the quick update. Closing this then.

PR #3321, opened 2019-12-06
Update todolist.json homepage link

vatosyorgos:
The Homepage-link is not set to the correct address, every other link in todolist.json seems to be right tho.


linsui:
duplicate of https://github.com/lukesampson/scoop-extras/pull/3318

PR #3325, opened 2019-12-07
emacs: Change to main download

### LINKED ISSUE #3302: Nearest GNU mirror down for me, can't install emacs
Hi, 

I attempted to install emacs but the emacs.json in scoop-extras attempts to get the file from https://ftpmirror.gnu.org. The problem is that this server seems to fetch from the nearest mirror and in my case, this mirror has been down for a while. I could use a VPN, but the solucion should be a bit more elegant, either upstream from https://ftpmirror.gnu.org or from scoop itself.

BlackPowerade:
Closes #3302 
Might be a bit slower for some users, but should be more reliable.

### DIFF
--- a/bucket/emacs.json
+++ b/bucket/emacs.json
@@ -5,11 +5,11 @@
     "version": "26.3",
     "architecture": {
         "64bit": {
-            "url": "https://ftpmirror.gnu.org/gnu/emacs/windows/emacs-26/emacs-26.3-x86_64.zip",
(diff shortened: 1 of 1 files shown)


Ash258:
For me this is approximately 24 times slower on 3 out of 4 regions, which i have for testing (Czech, Switzerland, Canada)

ddavness:
I could argue this is more of an issue in their side - if one of their mirrors is down then they should attempt the second nearest mirror instead (or just remove it from their list if this downtime is extended...)

PR #3350, opened 2019-12-15
reaper: Use portable configuration

jarosz:
Hi.

During startup, reaper.exe checks for the presence of reaper.ini file in the current directory (means portable install) and if it's not present, default data is created in %APPDATA%/REAPER directory. Data installed in scoop/apps/reaper (e.g. themes) is ignored.

Default clean install:
2019-12-15 17_55_37-REAPER v6 01_x64 - EVALUATION LICENSE

Expected result:
2019-12-15 17_56_51-REAPER v6 01_x64 - EVALUATION LICENSE

Scoop doesn't provide default reaper.ini file for portable installation, so I've made this PR to fix it and provide persistent config on update.

There's also reaper-vstplugins64.ini file being created, but I didn't include it in "persist" array, because I don't know how to support different filename for 32-bit architecture. The file contains only some cached data for vst plugins, which is going to be generated again if the file is missing.

Unfortunately, this change is going to break current REAPER scoop installs, because files are being kept in %APPDATA%/REAPER. Maybe it's the desired way for people using the software, so nobody has reported this behaviour yet. Maybe separate manifest reaper-portable.json should be created. If so, reaper.json should install data to ~~%APPDATA%/REAPER~~ $dir/InstallData.


bucket/reaper.json
```diff
@@ -33,9 +33,29 @@
     "installer": {
         "script": [
             "Move-Item \"$dir\\`$INSTDIR`$_8_\\*\" -Destination \"$dir\"",
-            "Remove-Item \"$dir\\`$*\", \"$dir\\Uninst*\" -Recurse -Force"
+            "Remove-Item \"$dir\\`$*\", \"$dir\\Uninst*\" -Recurse -Force",
+            "New-Item -ItemType \"file\" -Path \"$dir\" -Name \"reaper.ini\" -Value \"[REAPER]\nwnd_state=0\n\" | Out-Null",
```
Ash258: What does `wnd_state=0` mean?

jarosz: this is the default reaper.ini content if you tick the checkbox "Portable install" while installing without scoop. If this file is empty, reaper.exe behaves like it's not present and creates default files in %APPDATA%/REAPER.

jarosz:
> APPDATA contains lots of folders inside non portable installation. How these should be handled?

If REAPER is installed without scoop in a non-portable way it unpacks data to the InstallData folder in Program Files\REAPER and then copies it into %APPDATA%/REAPER on the first run. I it's installed with "Portable" option checked, all data is unpacked into main installation directory.

I think the scoop script should behave like it was a portable install, since REAPER supports it.

Ash258:
So there are no configs inside any of that folder?

jarosz:
Hi!

I've had some time to return to this issue.

> So there are no configs inside any of that folder?

Well there might be some. It seems that reaper creates them on the fly. Some are very important, like vst presets, fx chains, project and track templates. Scoop could make them persistent, but I don't know the full list.

That's why I've abandoned the idea of handling persistent config by scoop, and returned to %APPDATA%.

I've changed -Destination parameter to $dir/InstallData, so reaper.exe behaves like it's an ordinary install, creates config in %APPDATA%\REAPER and copies necessary content. New-Item is required, because if the destination directory doesn't exist, Move-Item will take first subdirectory ("ColorThemes") and call it InstallData, therefore preventing color themes from being copied to %APPDATA% by reaper.exe.

bucket/reaper.json
```diff
@@ -32,7 +32,8 @@
     ],
     "installer": {
         "script": [
-            "Move-Item \"$dir\\`$INSTDIR`$_8_\\*\" -Destination \"$dir\"",
+            "New-Item -Path \"$dir\" -ItemType \"directory\" -Name \"InstallData\"",
+            "Move-Item \"$dir\\`$INSTDIR`$_8_\\*\" -Destination \"$dir\\InstallData\"",
```
Ash258, line 36: I really do not understand why this change is needed, when the application works without this change.

```suggestion
            "Rename-Item \"$dir\\`$INSTDIR`$_8_\" 'InstallData'",
```

jarosz: It does if you're upgrading, since necessary content (themes, effects, etc) is already in %APPDATA%/REAPER. With clean install I get the result shown on the first screenshot.

Steps to reproduce:
1. Uninstall reaper
2. Backup and remove %APPDATA%/REAPER directory
3. scoop install reaper
4. Run program

Why?
Reaper setup program has two modes, "ordinary" install and portable.
1. In portable mode, executable and program data is unpacked to a single directory, on the first run reaper.exe checks for reaper.ini (which is created by setup program with "wnd_state=0" content inside), if it's not present, it behaves like an ordinary install - that's the case with reaper.json in master branch.
My first idea was to make configs persistent, but there is much more data which should persist and maintaining it would be hard and risky (imagine loosing hours of work put in presets being lost with a reaper update).
2. In ordinary mode, on the first run, reaper.exe copies content of InstallData subfolder (created by setup program) to %APPDATA%/REAPER. Without InstallData folder (current reaper.json in master) there is no data to copy.

jarosz: Now I think i misundersrtood the question.
"Rename-Item \"$dir\\`$INSTDIR`$_8_\" 'InstallData'"
is fine, but I didn't know this PowerShell command.

tech189 (commented):
I can confirm that with these changes, the expected theme is installed and the program looks as expected. It looks like it is unavoidable to keep the program's themes/settings out of AppData. I just make one suggestion to remove New-Item output.

rashil2000:
@tech189 could you raise a new superceding PR for this change? This one's quite stale.

PR #3351, opened 2019-12-15
Update libreoffice-fresh.json

omerakgoz34:
Fix "https" error


Ash258:
This is not good way how to fix. As there is automatic load balancing and mirror finding, which will lead into http => https => http mirror and you have same problem

Download Manually or with aria2 it should work.

rasa:
I agree with @Ash258. If the error is an expired cert, that’s an upstream error. A better solution would be to add a `—no-check-certificate` option. What, exactly, is the error.

PR #3353, opened 2019-12-16
hexchat: Add hexchat-text shim

BlackPowerade:
Adds text only hexchat option.

### DIFF
--- a/bucket/hexchat.json
+++ b/bucket/hexchat.json
@@ -14,7 +14,16 @@
         }
     },
     "innosetup": true,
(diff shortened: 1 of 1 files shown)


bucket/hexchat.json
```diff
@@ -14,7 +14,16 @@
         }
     },
     "innosetup": true,
-    "bin": "hexchat.exe",
+    "bin": [
```
Ash258:
```suggestion
    "bin": [
        "hexchat.exe",
        "hexchat-text.exe",
    ],
```

PR #3356, opened 2019-12-16
claws-mail: Change download URL

SirFireNewt:
The update commit for claws-mail was missing the new url and file hash for the installer


SirFireNewt:
I appreciate it Ash!   
It almost feels wrong to push your changes, I would have just made an issue but I thought I could just change the url & hash and save someone else the effort :p

PR #3365, opened 2019-12-18
qbittorrent-portable: remove --portable option

linsui:
Close https://github.com/lukesampson/scoop-extras/issues/3363 
I don't know why not download from fosshub.


r15ch13:
https://github.com/lukesampson/scoop/issues/3802

PR #3368, opened 2019-12-19
chromium: Update to version 79.0.3945.88

niheaven:


justinkb:
chromium 79 keeps crashing after a few seconds, even in incognito with all extensions disabled

PeeZu:
> chromium 79 keeps crashing after a few seconds, even in incognito with all extensions disabled

same issue here...  it's surely related to the ungoogled chromium version used as there is no more other available artifacts on https://github.com/macchrome/winchrome/releases/

gitolicious:
Same! If #3415 is taking much longer, can we at least roll back to a stable version?

PR #3372, opened 2019-12-20
infoqube: Update to version 0.9.114q (manually)

issaclin32:
**infoqube** has changed the layout of their website, causing `checkver` to fail. This fixes the problem.

Also modified `description` to make it briefer.

### DIFF
--- a/bucket/infoqube.json
+++ b/bucket/infoqube.json
@@ -1,13 +1,13 @@
 {
     "homepage": "http://www.infoqube.biz/",
-    "description": "An Information Management System that provides all the tools you need to simplify your life.",
-    "version": "0.9.113f",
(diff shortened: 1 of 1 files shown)


bucket/infoqube.json
```diff
@@ -17,7 +17,7 @@
     "persist": "Users",
     "checkver": {
         "url": "http://www.infoqube.biz/download",
-        "re": "InfoQube ([\\d.]+\\w?)"
+        "regex": "InfoQube([\\d.]+\\w?)Portable.zip"
```
Ash258:
```suggestion
        "regex": "InfoQube([\\d.]+\\w?)Portable\\.zip"
```

PR #3374, opened 2019-12-20
treesize-free: Fix url and checkver

issaclin32:

* They changed the layout of their website. Therefore `checkver` needs to be modified.

* The **download url** on `downloads.jam-software.com` is not working. Their website provides multiple download locations, but both *North America* and *Europe* leads to `downloads.jam-software.de`.

* Not sure why **hash** value changed. Probably due to some minor updates.

### DIFF
--- a/bucket/treesize-free.json
+++ b/bucket/treesize-free.json
@@ -1,13 +1,13 @@
 {
     "homepage": "https://www.jam-software.com/treesize_free",
-    "description": "Quickly capture directory sizes and memory guzzlers.",
+    "description": "Capture directory sizes and memory guzzlers.",
     "version": "4.4.1",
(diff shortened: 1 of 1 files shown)


bucket/treesize-free.json
```diff
@@ -17,9 +17,9 @@
     ],
     "checkver": {
         "url": "https://www.jam-software.com/treesize_free/changes.shtml",
-        "regex": "Version\\s+([\\d.]+):"
+        "regex": "Version ([\\d.]+)\\</h3\\>"
```
Ash258: Why do you escape anchor?

```suggestion
        "regex": "Version\\s+([\\d.]+)</h3>"
```

PR #3376, opened 2019-12-20
yatqa: Update to version 3.9.8.3b (manually)

issaclin32:

**yatqa** now include letters (such as 3.9.8.3**b**) in the version number. Therefore `checkver` needs to be modified.

### DIFF
--- a/bucket/yatqa.json
+++ b/bucket/yatqa.json
@@ -1,13 +1,13 @@
 {
     "homepage": "https://yat.qa/",
     "description": "Manage TeamSpeak 3 servers and instances using a query interface",
-    "version": "3.9.8.3",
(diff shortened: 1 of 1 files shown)


bucket/yatqa.json
```diff
@@ -22,7 +22,7 @@
     "post_install": "Remove-Item \"$dir\\`$PLUGINSDIR\" -Recurse",
     "checkver": {
         "url": "https://yat.qa/.update",
-        "regex": "v([\\d.]+)\\s+\\("
+        "regex": "v([\\d.]+\\w?)\\s+\\("
```
Ash258:
```suggestion
        "regex": "v([\\w.]+)\\s+\\("
```

PR #3377, opened 2019-12-20
vidi: Update to version 0.2.5.5 (manually)

issaclin32:

* **ViDi** has changed their file name patterns. ~Therefore `autoupdate` needs to be modified.~ Removing `autoupdate` since their latest version is marked as **experimental**.

* ~ViDi now depends on VCRedist 2015. (was VCRedist 2017 previously) (see: https://github.com/hasherezade/ViDi/releases)~
    -> Visual C++ 2015, 2017 and 2019 all share the same redistributable files, therefore use `extras/vcredist2019` instead.

### DIFF
--- a/bucket/vidi.json
+++ b/bucket/vidi.json
@@ -1,16 +1,16 @@
 {
     "homepage": "https://hshrzd.wordpress.com/vidi-visual-disassembler/",
     "description": "A tool for static analysis of PE files, based on bearparser & capstone",
-    "version": "0.2.5.6",
+    "version": "0.2.5.5",
     "license": "BSD-2-Clause",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


niheaven:
> Visual C++ 2015, 2017 and 2019 all share the same redistributable files.

Ref: https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads

So use `extras/vcredist2019` should be better.

niheaven:
> Visual C++ 2015, 2017 and 2019 all share the same redistributable files.

Ref: https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads

So use `extras/vcredist2019` should be better.

Ash258:
This version is alpha.

image

issaclin32:
Since version 0.2.5.6 is **experimental**, I will remove `autoupdate`. This package has to be updated manually to get only **stable versions**.

PR #3380, opened 2019-12-21
qalculate: Add version 3.6.0

linsui:


Ash258:
Do not add two manifests. Portable and normal manifests, which are inside buckets are result of portable function addition after manifest was added. It is nonse to add both of them.

1. `git rm bucket/qalculate.json`
1. `git mv bucket/qalculate-portable.json bucket/qalculate.json`

r15ch13:
Neat! 👍

PR #3382, opened 2019-12-21
parsecd: Add Parsec Daemon

ftk:


bucket/parsecd.json
```diff
@@ -0,0 +1,44 @@
+{
+    "version": "2019.6.19",
+    "description": "Simple, Low-Latency Game Streaming.",
```
Ash258:
```suggestion
    "description": "Simple, Low-Latency Game Streaming",
```

bucket/parsecd.json
```diff
@@ -0,0 +1,44 @@
...
+    "homepage": "https://parsecgaming.com",
+    "license": "Freeware|Proprietary",
```
Ash258: Where did you get this information?

bucket/parsecd.json
```diff
@@ -0,0 +1,44 @@
...
+            "url": "https://builds.parsecgaming.com/package/parsec-flat-windows.zip",
+            "hash": "c5e69da443eb0c834e2960928999063326263a638e260076c15346316f40e0ba"
+        },
+        "32bit": {
+            "url": "https://builds.parsecgaming.com/package/parsec-flat-windows32.zip",
+            "hash": "5b7aec1183ce5993419677625854a1cd67fd0cc8aef30ed7a3a88638315baeb1"
+        }
+    },
+    "bin": "parsecd.exe",
+    "persist": [
```
Ash258: - Any of this file do not exists inside archive. Precreate them inside pre_install block
	- Post URL to documentation about each of them
- Move this property after shortcuts

bucket/parsecd.json
```diff
@@ -0,0 +1,44 @@
+{
+    "version": "2019.6.19",
```
Ash258: What version do parsec use? This one or is there some other versioning?

ftk:
It seems like a problematic app to be run portable.
From my observations:
1. parsecd first checks if current working directory is the same where parsecd.exe resides.
2. then it downloads dll from somewhere and creates/uses config files in this directory
3. if it's not the same dir, then it creates Appdata/Roaming/Parsec and does step 2 in this directory

PR #3387, opened 2019-12-21
Materialize: Add version 1.78

ddavness:
A program that converts images into materials that you can use in video game development.


bucket/materialize.json
```diff
@@ -0,0 +1,24 @@
...
+    "homepage": "http://www.boundingboxsoftware.com/materialize/",
+    "license": "GPL-3.0-or-later",
+    "architecture": {
+        "64bit": {
+            "url": "http://www.boundingboxsoftware.com/materialize/download/Materialize_1.78.zip",
+            "hash": "b0f6e3a905781e85dcd7db2383646bce51b8984ab93d1e3d13be42d5bea8904a"
+        }
+    },
+    "bin": "Materialize.exe",
+    "shortcuts": [["Materialize.exe", "Materialize"]],
```
Ash258: ...

```suggestion
    "shortcuts": [
        [
            "Materialize.exe",
            "Materialize"
        ]
    ],
```

bucket/materialize.json
```diff
@@ -0,0 +1,24 @@
...
+    "suggest": {
+        "vcredist": "extras/vcredist2019"
```
Ash258: Link inside FAQ and download page lead into 2013 version.

bucket/materialize.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "checkver": {
+        "url": "http://www.boundingboxsoftware.com/materialize/getkey.php",
+        "regex": "download/Materialize_(\\d+\\.\\d+).zip"
```
Ash258:
```suggestion
        "regex": "download/Materialize_([\\d.]+)\\.zip"
```

bucket/materialize.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "1.78",
+    "description": "Image-to-material converter for use in video games",
+    "homepage": "http://www.boundingboxsoftware.com/materialize/",
+    "license": "GPL-3.0-or-later",
```
Ash258: > GNU General Public License v3. Use it for whatever you want.

There is nothing about future versions

```suggestion
    "license": "GPL-3.0-only",
```

bucket/materialize.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "1.78",
+    "description": "Image-to-material converter for use in video games",
```
Ash258:
```suggestion
    "description": "Image-to-material converter for video game development",
```

bucket/materialize.json
ddavness: I assumed that the "or later" statement was implicit. My bad

bucket/materialize.json
ddavness: I'm going to test this before changing the VC version...

ddavness:
Apparently the `Materialize_Data` folder isn't more than the program's libraries (Unity, Mono, etc.) and some obscure dev-side configs you probably shouldn't be touching anyway.

User settings are saved in a file called `settings.txt`, which is not there by default, and not in the `Materialize_Data` folder (that's the only thing we have got to persist...)

bucket/materialize.json
ddavness: Update: It does indeed require 2013. 2015/17/19 won't do.

ddavness:
There you go, persisting should work now.

bucket/materialize.json
```diff
@@ -0,0 +1,35 @@
...
+        }
+    },
+    "bin": "Materialize.exe",
+    "shortcuts": [
+        [
+            "Materialize.exe",
+            "Materialize"
+        ]
+    ],
+    "pre_install": [
```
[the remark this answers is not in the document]
ddavness: Not a fan of one-liners, but ¯\_(ツ)_/¯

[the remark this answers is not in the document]
Ash258: This is not for oneliners. Main reason is that awefull write-output with redirection, when you have native command

PR #3393, opened 2019-12-23
calibre: Fix bin

luisfernandobarrera:
calibre-portable not working, should be calibre.exe


Ash258:
I do not think you want to cal `calibre.exe` You should rather be called `calibre-portable.exe` from root.

Ash258:
Since the file was moved the `calibre` shortcut will most likely be broken too.

Ash258:
image

PR #3405, opened 2019-12-24
Add ripgrep-all

ammgws:


Ash258:
https://scoop-docs.now.sh/apps/?query=rga
https://github.com/ScoopInstaller/Main/blob/master/bucket/rga.json

ammgws:
Argh, was searching for `ripgrep-all` so that's why I didn't find it.

PR #3407, opened 2019-12-24
foobar2000-encoders: Update to version 2019-12-23 (manually)

issaclin32:


niheaven:
The random hash is per session, so use original download source please.

issaclin32:
**videohelp.com** has updated their files. Therefore this PR is not needed.

PR #3414, opened 2019-12-26
authy: Add nightly

ddavness:
Authy is a multi-device 2FA (aka Two Factor Authentication) client, being the most popular 2FA app in mobile devices.

# Considerations about this PR

Like some custom installer-based apps (like Spotify), installing Authy via Scoop is presenting us some challenges (apart from very poor developer-sided documentation):

## Nightly-versioned manifest
I am aware that Authy versions are numbered (at the moment of writing this, we're at version `1.7.2`). However, from a technical standpoint, it doesn't seem feasible to do so:

- Authy comes with an auto-updater, so it's possible that the application breaks free from the scoop directory. (Still need to investigate this)
- Older versions are not made (publicly) available for us to download. I'm unable to look for an older version and older endpoints seem to be no longer existing.

## Double extraction
To get what we want out of the installer, we need to perform two extractions:

- The first one, to get the NuGet package out (e.g. `authy-electron-1.7.2-full.nupkg`
- And then extract that package.

At the moment it's impossible to perform this double extraction without extremely hacky methods (for example, the manifest downloading an installer script and then running that script in `post_install` - and even that doesn't allow us to set up shortcuts without their own very hacky methods, etc.) - so clearly scoop itself might need to be adjusted to this case.

### Possible solutions
At the moment the solution that involves the least work would be either:
- Copying the download file into the directory before running `pre_install`, allowing us to run 7z directly over it without copying it manually from cache;
- - Alternatively, if there are manifests that rely on an empty directory for `pre_install`, add a second step between making the download available to use and before installation (shimming, etc.)

**OR**

- Allowing us to specify more than one extraction step

## Persisting

The application's persisting data is stored on `AppData/Roaming/Authy Desktop` (apparently an hardcoded path).

# Checklist

- [x] Download the files
- [x] Extract the contents (manually) and avoiding running the installer.
- [ ] Extract the contents (automatically)
- [ ] Monitor and make sure autoupdate doesn't ruin the party.
- [ ] *Anything else that might be missing?*

### DIFF
--- a/bucket/authy.json
+++ b/bucket/authy.json
@@ -1,6 +1,6 @@
 {
     "homepage": "https://authy.com",
-    "description": "The .",
+    "description": "The 2FA client for all devices.",
     "license": {
         "identifier": "Freeware",
         "url": "https://www.twilio.com/legal/tos"
--- a/bucket/authy.json
+++ b/bucket/authy.json
@@ -1,0 +1,30 @@
+{
+    "homepage": "https://authy.com",
+    "description": "The .",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://www.twilio.com/legal/tos"
+    },
+    "version": "nightly",
+    "architecture": {
+        "32bit": {
+            "url": "https://electron.authy.com/download?channel=stable&arch=x32&platform=win32&version=latest&product=authy#dl.7z"
+        },
+        "64bit": {
+            "url": "https://electron.authy.com/download?channel=stable&arch=x64&platform=win32&version=latest&product=authy#dl.7z"
+        }
+    },
+    "pre_install": [
+        "# Manual extraction",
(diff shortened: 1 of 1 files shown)


Ash258:
I have no problem with https://github.com/Ash258/Scoop-Ash258/blob/master/TODO/AAAuthy.json

ddavness:
Well then, to avoid work duplication, please disregard this.

Ash258:
I am not gonna work on it.

ddavness:
Mind if I ask why? I can take the time to start from what you've left (or, if the manifest is complete, maybe moving it off from the TODO?)

Ash258:
Waste of time for me. I am not user of this application, that manifest exists for really long time when I considered this app, then I switched to different one.

ddavness:
For further information, see #3421

PR #3421, opened 2019-12-28
authy: Add version 1.7.2

ddavness:
From https://github.com/Ash258/Scoop-Ash258/blob/master/TODO/AAAuthy.json

**Checklist:**
 - [x] Link Authy's AppData to `persist` so that `scoop uninstall -p authy` clears all data.
- [x] Make sure we're in control of the update process (this is, Authy autoupdate will not interfere).

~~Should we import (as in preserve) Authy's existing `AppData`, or just perform a clean installation?~~

### DIFF
--- a/bucket/authy.json
+++ b/bucket/authy.json
@@ -28,7 +28,7 @@
     ],
     "checkver": {
         "url": "https://s3.amazonaws.com/authy-electron-repository-production/_squirrel/authy/stable/x64/RELEASES",
-        "regex": "authy-electron-([\\d\\.]+)-full\\.nupkg",
+        "regex": "authy-electron-([\\d.]+)-full\\.nupkg",
         "reverse": true
     },
     "autoupdate": {
--- a/bucket/authy.json
+++ b/bucket/authy.json
@@ -24,7 +24,6 @@
         "}",
         "New-Item -ItemType Junction -Path $appdata -Target $persist_dir | Out-Null"
     ],
-    "persist": "void",
     "bin": [
         [
             "Authy Desktop.exe",
--- a/bucket/authy.json
+++ b/bucket/authy.json
@@ -15,7 +15,7 @@
     },
     "extract_dir": "lib\\net45",
     "pre_install": [
-        "$appdata = \"$env:APPDATA\\Authy Desktop\"",
+        "$appdata = \"$env:APPDATA\\Authy\u0020Desktop\"",
         "if (!(Test-Path $persist_dir)) {",
         "   New-Item -ItemType Directory -Path $persist_dir | Out-Null",
         "}",
@@ -37,7 +37,7 @@
         ]
     ],
     "uninstaller": {
-        "script": "Remove-Item -Confirm:$false -Force -Recurse -Path \"$env:APPDATA\\Authy Desktop\""
+        "script": "Remove-Item -Confirm:$false -Force -Recurse -Path \"$env:APPDATA\\Authy\u0020Desktop\""
     },
     "checkver": {
         "url": "https://s3.amazonaws.com/authy-electron-repository-production/_squirrel/authy/stable/x64/RELEASES",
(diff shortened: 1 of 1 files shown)


ddavness:
This will work if you've never installed Authy before via their installer. However, if the user is coming from Authy installed via their installer (they don't clean their AppData folder), they're going to lose all data. Should we transfer said data or is it ok to just go ahead and clean it?

ddavness:
As useful as it is, it probably is not a good idea.

ddavness:
Looks good to me, ready for review.

bucket/authy.json
```diff
@@ -0,0 +1,63 @@
...
+        ]
+    ],
+    "shortcuts": [
+        [
+            "Authy Desktop.exe",
+            "Authy Desktop"
+        ]
+    ],
+    "uninstaller": {
+        "script": "Remove-Item -Confirm:$false -Force -Recurse -Path \"$env:APPDATA\\Authy Desktop\""
```
Ash258: > -Confirm:$false -Force

???

bucket/authy.json
```diff
@@ -0,0 +1,63 @@
...
+            "authy"
+        ]
+    ],
+    "shortcuts": [
+        [
+            "Authy Desktop.exe",
+            "Authy Desktop"
+        ]
+    ],
+    "uninstaller": {
```
Ash258: Why would you do this? So much not needed operations on each update. Removing and then linking...

bucket/authy.json
```diff
@@ -0,0 +1,63 @@
...
+            "url": "https://s3.amazonaws.com/authy-electron-repository-production/authy/stable/1.7.2/win32/x64/authy-electron-1.7.2-full.nupkg",
+            "hash": "sha1:53646438bd9c05efa4790b655ab1d056302a6a7f"
+        },
+        "32bit": {
+            "url": "https://s3.amazonaws.com/authy-electron-repository-production/authy/stable/1.7.2/win32/x32/authy-electron-1.7.2-full.nupkg",
+            "hash": "sha1:adb40cd6a325585b3b2fc32c4ccf29d8d1bb08da"
+        }
+    },
+    "extract_dir": "lib\\net45",
+    "pre_install": [
```
Ash258: Why? 

Any of manifest is doing this. There is no reason to start with this.
There is no research about how applications will handle link. What they will do to this folder and how they accesss them. 

Remove it. Same for uninstaller.

ddavness:
You're right, the approach sucks. (Leaving the discussion open), but here's why I did it:

**It boils down to Authy storing the sensitive data in** `AppData`.

Well, sure, they're not the only one to do so. For example, Discord also does it for their data (including the login data). This means that the following sequence of commands would preserve your session:

`scoop uninstall discord`
`scoop install discord`

But we're talking about access tokens. And the uninstaller authy provides - `Update.exe --uninstall` doesn't clean the folder (in my opinion, it should).

As an user of this application, though: I can assure you the app didn't complain and worked just as intended. (This is, the app was "fooled" into writing it's data in persist just fine...)

bucket/authy.json
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    "shortcuts": [
+        [
+            "Authy Desktop.exe",
+            "Authy Desktop"
+        ]
+    ],
+    "checkver": {
+        "url": "https://s3.amazonaws.com/authy-electron-repository-production/_squirrel/authy/stable/x64/RELEASES",
+        "regex": "authy-electron-([\\d\\.]+)-full\\.nupkg",
```
Ash258:
```suggestion
        "regex": "authy-electron-([\\d.]+)-full\\.nupkg",
```

bucket/authy.json
```diff
@@ -0,0 +1,50 @@
+{
+    "version": "1.7.2",
+    "description": "Free desktop app for two factor authentication.",
```
Ash258: Free is convered by license

```suggestion
    "description": "Two factor authentication client",
```

PR #3422, opened 2019-12-29
age: Add version 1.0.0-beta2

avoidwork:
age ([homepage](https://age-encryption.org)) is a simple, modern and secure encryption tool with small explicit keys, no config options, and UNIX-style composability.

- Only a 64-bit version exists

### DIFF
--- a/bucket/age.json
+++ b/bucket/age.json
@@ -1,0 +1,15 @@
+{
+    "homepage": "https://age-encryption.org/",
+    "version": "1.0.0-beta2",
+    "description": "A simple, modern and secure encryption tool with small explicit keys, no config options, and UNIX-style composability.",
+    "license": {
+        "identifier": "BSD 3-Clause",
+        "url": "https://github.com/FiloSottile/age/blob/master/LICENSE"
+    },
+    "url": "https://github.com/FiloSottile/age/releases/download/v1.0.0-beta2/age-v1.0.0-beta2-windows-amd64.zip",
+    "extract_dir": "age",
+    "bin": [
+        "age.exe",
+        "age-keygen.exe"
+    ]
+}
--- a/bucket/age.json
+++ b/bucket/age.json
@@ -2,16 +2,13 @@
     "homepage": "https://age-encryption.org/",
     "version": "1.0.0-beta2",
     "description": "A simple, modern and secure encryption tool with small explicit keys, no config options, and UNIX-style composability.",
-    "license": {
-        "identifier": "BSD 3-Clause",
-        "url": "https://github.com/FiloSottile/age/blob/master/LICENSE"
+    "license": "BSD-3-Clause",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/FiloSottile/age/releases/download/v1.0.0-beta2/age-v1.0.0-beta2-windows-amd64.zip",
+            "hash": "76c616ee56c2a2e182b50504d7a3b8ca390ee8bc2083a1dfd3cb7da50dcf70cc"
+        }
     },
-     "architecture": {
-         "64bit": {
-             "url": "https://github.com/FiloSottile/age/releases/download/v1.0.0-beta2/age-v1.0.0-beta2-windows-amd64.zip",
-             "hash": "76c616ee56c2a2e182b50504d7a3b8ca390ee8bc2083a1dfd3cb7da50dcf70cc"
-         }
-     },
     "extract_dir": "age",
     "bin": [
         "age.exe",
--- a/bucket/age.json
+++ b/bucket/age.json
@@ -19,6 +19,10 @@
         "regex": "tag/v([\\\\w.]+)"
     },
     "autoupdate": {
-        "url": "https://github.com/FiloSottile/age/releases/download/v$version/age-v$version-windows-amd64.zip"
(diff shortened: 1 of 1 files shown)


bucket/age.json
```diff
@@ -0,0 +1,15 @@
+{
+    "homepage": "https://age-encryption.org/",
+    "version": "1.0.0-beta2",
+    "description": "A simple, modern and secure encryption tool with small explicit keys, no config options, and UNIX-style composability.",
+    "license": {
+        "identifier": "BSD 3-Clause",
+        "url": "https://github.com/FiloSottile/age/blob/master/LICENSE"
+    },
+    "url": "https://github.com/FiloSottile/age/releases/download/v1.0.0-beta2/age-v1.0.0-beta2-windows-amd64.zip",
```
ddavness: Since that we only have a 64-bit version of the app, we need to specify the `architecture` difference - otherwise our folks who use 32-bit Windows will be able to install the app, but not run it! (and that's weird)

We also need to specify the zip's checksum. You can get the checksum of a file with a tool like `hashcheck`, also available via scoop (`scoop bucket add extras` and `scoop install hashcheck`)

```suggestion
     "architecture": {
         "64bit": {
             "url": "https://github.com/FiloSottile/age/releases/download/v1.0.0-beta2/age-v1.0.0-beta2-windows-amd64.zip",
             "hash": "76c616ee56c2a2e182b50504d7a3b8ca390ee8bc2083a1dfd3cb7da50dcf70cc"
         }
     }
```

bucket/age.json
```diff
@@ -0,0 +1,15 @@
+{
+    "homepage": "https://age-encryption.org/",
+    "version": "1.0.0-beta2",
+    "description": "A simple, modern and secure encryption tool with small explicit keys, no config options, and UNIX-style composability.",
+    "license": {
+        "identifier": "BSD 3-Clause",
+        "url": "https://github.com/FiloSottile/age/blob/master/LICENSE"
+    },
```
ddavness: Since this is a standard SPDX license, we don't need to point to their `LICENSE` text.

```suggestion
    "license": "BSD-3-Clause",
```
(Note that you'll then need to fix the manifest because GitHub suggestions cannot overwrite more than one line)

bucket/age.json
linsui: And autoupdate also need to be fixed.

ddavness: Just checked that now, and it indeed doesn't work (because at the moment these are all prereleases)

Ash258: It works. But you need to specify regex

ddavness: In particular, we need to predict the wording they use. For the sake of simplicity, I'm thinking of `\/releases\/tag\/v([\\d.]+(-beta\\d+)?)` (The first part is needed, otherwise it'll catch garbage). Any way to shorten this even further?

Ash258: ... `tag\v([\\w.]+)`

bucket/age.json
avoidwork: Updated with 6dd2ac5447ed4e6df2d309851883540f70b78411

bucket/age.json
ddavness: Doesn't work because it doesn't match `-beta2` - so maybe `tag/v([\\w.]+(-[\\w]+)?)`?

avoidwork: Updated with de8a66f98ae4eb0252bf547257167221eadd7105

Ash258: Forgot dash, `\w` contains only underscore

bucket/age.json
```diff
@@ -0,0 +1,24 @@
...
+        }
+    },
+    "extract_dir": "age",
+    "bin": [
+        "age.exe",
+        "age-keygen.exe"
+    ],
+    "checkver": {
+        "github": "https://github.com/FiloSottile/age",
+        "regex": "tag/v([\\\\w.]+)"
```
Ash258:
```suggestion
        "regex": "tag/v([\\w-.]+)"
```

bucket/age.json
avoidwork: @linsui to correct `autoupdate` do i need to dupe the bin key inside of it? I'm not understanding the mistake based on the wiki.

bucket/age.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "autoupdate": {
+        "url": "https://github.com/FiloSottile/age/releases/download/v$version/age-v$version-windows-amd64.zip"
```
Ash258:
```suggestion
        "architecture": {
            "64bit": {
                "url": "https://github.com/FiloSottile/age/releases/download/v$version/age-v$version-windows-amd64.zip"
            }
        }
```

bucket/age.json
avoidwork: Fixed with 3407b306267ee08cd19d76e7723de0a7e648e6a9

bucket/age.json
linsui: I mean you need to use `architecture` in autoupdate. :smile:

avoidwork:
@Ash258 are there still changes to make? I can't seem to access to view where you made that selection, unless i "unresolve" the conversation.

PR #3423, opened 2019-12-29
quicktextpaste: Add version 5.41

worldadventurer:
Another really useful freeware from SoftwareOK (also the creator of dont-sleep, which is already in this bucket). I used this json file as my starting point: https://github.com/lukesampson/scoop-extras/blob/master/bucket/dont-sleep.json


worldadventurer:
Thanks for the feedback, @Ash258 ! I've updated my pull request. Please review. :)

PR #3435, opened 2020-01-01
CUDA8: Add version 8.0.61

issaclin32:
**CUDA** ([homepage](https://developer.nvidia.com/cuda-toolkit)) is NVIDIA's runtime library and development toolkit for GPU-accelerated applications.

* This package follows the pattern of [VCRedist packages](https://github.com/lukesampson/scoop-extras/blob/8a1d5cd2d16c5796a30e9f3e07eb755171c0463a/bucket/vcredist2013.json).

* This package only contains CUDA 8.0. I will send another PR for other CUDA versions (`9.0`, `9.1`, `9.2`, `10.0`, `10.1`) if this PR is accepted.

* **Naming** of other CUDA packages: Is it acceptable to use `cuda10.1.json` as the package name? (or should I name it as `cuda10-1` instead?)

* This package will install CUDA Toolkit **without installing the display driver**. It can be installed on a PC without a compatible NVIDIA graphics card. Also, different versions of CUDA can be installed on the same PC.

* Installer **arguments** can be found in the **Installation Guide Windows -> Install the CUDA Software** section in CUDA Toolkit Documentation ([CUDA 8.0](https://docs.nvidia.com/cuda/archive/8.0/cuda-installation-guide-microsoft-windows/index.html)) ([CUDA 9.0](https://docs.nvidia.com/cuda/archive/9.0/cuda-installation-guide-microsoft-windows/index.html)) ([CUDA 10.0](https://docs.nvidia.com/cuda/archive/10.0/cuda-installation-guide-microsoft-windows/index.html))

* **hash**: The MD5 hash value comes from https://developer.nvidia.com/compute/cuda/8.0/Prod2/docs/sidebar/md5sum-txt.

* **license**: I marked the `license` as **Proprietary** according to the [Wikipedia article](https://en.wikipedia.org/wiki/CUDA). Also there are some restrictions described in the EULA.
> Subject to the terms of this Agreement, NVIDIA hereby grants you a non-exclusive, non-transferable license, without the right to sublicense (except as expressly provided in this Agreement) to:
>1. Install and use the SDK,
>2. Modify and create derivative works of sample source code delivered in the SDK, and
>3. Distribute those portions of the SDK that are identified in this Agreement as distributable, as incorporated in object code format into a software application that meets the distribution requirements indicated in this Agreement.
(https://docs.nvidia.com/cuda/eula/index.html)

### DIFF
--- a/bucket/cuda8.json
+++ b/bucket/cuda8.json
@@ -1,0 +1,20 @@
+{
+    "version": "8.0.61",
+    "description": "Runtime library and development toolkit for GPU-accelerated applications",
+    "homepage": "https://developer.nvidia.com/cuda-toolkit",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://docs.nvidia.com/cuda/eula/index.html"
+    },
+    "url": "https://developer.nvidia.com/compute/cuda/8.0/Prod2/local_installers/cuda_8.0.61_win10-exe#/setup.exe",
+    "hash": "md5:f857c29b8121099090524b90290cfbf6",
+    "post_install": [
+        "if (-not (is_admin)) { error 'This package requires admin privileges to install'; break }",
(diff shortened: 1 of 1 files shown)


linsui:
CUDA without drivers has been in Main bucket https://github.com/ScoopInstaller/Main/blob/master/bucket/cuda.json . But this manifest should be useful. Just wonder where it should live.

issaclin32:
> CUDA without drivers has been in Main bucket https://github.com/ScoopInstaller/Main/blob/master/bucket/cuda.json . But this manifest should be useful. Just wonder where it should live.

Oops. Didn't notice that. Thanks for the info.

I think the method used in https://github.com/ScoopInstaller/Main/blob/master/bucket/cuda.json should be better (rather than running the installer). I just don't know if **not running the installer** will make any difference (for example, some of the components may not work as expected).

I will close the PR for now, and add the manifest to the *nonportable* bucket if needed.

issaclin32:
By the way. I think it will be better to have other versions of CUDA in the `versions` bucket.

linsui:
Yes, other versions are useful. The difference is that without running installer some drivers won't be installed.

issaclin32:
@linsui I found that **env_set** for the CUDA package is not working as expected, but it can be fixed by replacing `"CUDA_PATH": "."` with `"CUDA_PATH": "$dir"`. Not sure if this is due to some recent changes in *Scoop* or what...
env

linsui:
Sorry, that's my fault... evn_set is not only for path so scoop won't add a prefix for it.

PR #3456, opened 2020-01-05
q-dir: Add version 7.99

worldadventurer:
Another freeware utility from softwareok - this one is a quad window file manager

### DIFF
--- a/bucket/q-dir.json
+++ b/bucket/q-dir.json
@@ -43,7 +43,7 @@
     "extract_dir": "Q-Dir",
     "persist": [
         "Q-Dir.ini",
-	    "Favoriten"
+        "Favoriten"
     ],
     "checkver": {
         "url": "https://www.softwareok.com/?seite=Microsoft/Q-Dir/History",
--- a/bucket/q-dir.json
+++ b/bucket/q-dir.json
@@ -40,7 +40,7 @@
             ]
         }
     },
-    "pre_install": "mv $original_dir\\Q-Dir\\Q-Dir*.* $original_dir\\",
+    "extract_dir": "Q-Dir",
     "persist": "Q-Dir.ini",
     "checkver": {
         "url": "https://www.softwareok.com/?seite=Microsoft/Q-Dir/History",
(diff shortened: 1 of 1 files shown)


bucket/q-dir.json
```diff
@@ -0,0 +1,63 @@
...
+            ],
+            "shortcuts": [
+                [
+                    "Q-Dir.exe",
+                    "Q-Dir"
+                ]
+            ]
+        }
+    },
+    "pre_install": "mv $original_dir\\Q-Dir\\Q-Dir*.* $original_dir\\",
```
linsui:
```suggestion
    "extract_dir": "Q-Dir",
```

worldadventurer:
> What's the `Favoriten` directory? Is it needed to be persist?
As far as I can tell (have been using this software for years), the Favoriten directory is a suggested directory for storing Favorites from inside the software. But, it's not even the default directory for when you go to save a Favorite item for the first time. We could delete the Favoriten directory if that's a best practice for scoop?

linsui:
> But, it's not even the default directory for when you go to save a Favorite item for the first time.

Do you mean there is another default directory? 😕  I test and a new item is saved in the `Favoriten` directory. If this is the defauld location, it should be persist.

worldadventurer:
I see your point now. :) Ok, I just added the Favoriten folder to persist. Thanks for catching that!

bucket/q-dir.json
```diff
@@ -0,0 +1,66 @@
...
+                [
+                    "Q-Dir.exe",
+                    "Q-Dir"
+                ]
+            ]
+        }
+    },
+    "extract_dir": "Q-Dir",
+    "persist": [
+        "Q-Dir.ini",
```
Ash258: Move Favoriten folder as first item

bucket/q-dir.json
```diff
@@ -0,0 +1,66 @@
...
+                [
+                    "Q-Dir_x64.exe",
+                    "Q-Dir"
+                ]
+            ]
+        },
+        "32bit": {
+            "url": "https://www.softwareok.com/Download/Q-Dir_Portable.zip",
+            "hash": "850e2e17dcb88ac808ab7d25789fa5f8f552341d75ecf04af936188f144223a6",
+            "bin": [
```
Ash258: "bin": "Q-Dir.exe",

bucket/q-dir.json
worldadventurer: ok, done

bucket/q-dir.json
worldadventurer: ok, done

worldadventurer:
Hi @Ash258, would you please re-review after I made the changes you suggested? :)

PR #3465, opened 2020-01-07
imageglass: Update to version 7.5.1.1

linsui:
https://github.com/lukesampson/scoop-extras/issues/2308
When updating from old version it does not work unless `igconfig.xml` is removed. This should be an upstream issue.

### DIFF
--- a/bucket/imageglass.json
+++ b/bucket/imageglass.json
@@ -1,33 +1,62 @@
 {
-    "version": "7.0.7.26",
-    "homepage": "https://imageglass.org/",
-    "description": "A lightweight, versatile image viewer.",
+    "version": "7.5.1.1",
+    "homepage": "https://imageglass.org",
+    "description": "A lightweight, versatile image viewer",
     "license": "GPL-3.0-only",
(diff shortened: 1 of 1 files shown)


Ash258:
> When updating from old version it does not work unless `igconfig.xml` is removed. This should be an upstream issue.

This does not make sense. What do not work? What update? Application autoupdate or what?

linsui:
The app won't work. Nothing happens when I try to start it.  Some settings are not compatible I guess. See https://imageglass.org/docs/app-configs. But I see no message box.

Ash258:
Then it is pretty nonsense to update without taking additional steps or informing user how to deal with it.

linsui:
How about a note `If this app doesn't work maybe you need to clean '$dir\\igconfig.xml' .` ?

rasa:
Can this be approved? I think a note would be helpful, but it shouldn’t block approval.

Ash258:
I am not gonna approve/merge this change, which will not work for most of users.

Feel free to merge on your own @rasa

PR #3482, opened 2020-01-12
everything: Persist Bookmarks.csv

HUMORCE:
File **Bookmarks.csv** saves a list of bookmarks (Also known as: search filter).

### DIFF
--- a/bucket/everything.json
+++ b/bucket/everything.json
@@ -14,10 +14,17 @@
         }
     },
     "pre_install": [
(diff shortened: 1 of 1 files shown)


bucket/everything.json
```diff
@@ -20,6 +20,7 @@
     "bin": "Everything.exe",
     "shortcuts": [
         [
+            "Bookmarks.csv",
```
Ash258: Shortcuts?

```suggestion
```

PR #3446, opened 2020-01-03
Update flutter --> 1.12.13-hotfix.5

vidarkongsli:

PR #2770, opened 2019-08-28
Upgrade Flutter to windows_v1.7.8+hotfix.4-stable

hanabi1224:

PR #2526, opened 2019-07-29
flutter new stable: 1.7.8+hotfix.4

ench0:

PR #3489, opened 2020-01-14
flutter: Update to version 1.12.13-hotfix.7

### LINKED ISSUE #2439: how to update flutter bucket

### LINKED ISSUE #2924: update flutter 1.9

### LINKED ISSUE #3490: The exact "+" sign would be decoded as a space
Hi all.
I got into this issue while submitting #3489 which involves the `+` sign. Idk why but the `flutter team` did use an exact `+` sign in the releases. And when I changed the `checkver` in order to support the `+` sign and tried to `autoupdate` it, then I got a stubborn error. 

After some debugging... It turns out that the `[System.Web.HttpUtility]::UrlDecode` method in `lib/autoupdate.ps1` line `175` unexpectedly ( yet reasonable ) decoded the `+` of the `$basename` variable to an empty space. Hence the jsonpath `$.releases[?(@.archive =~ /.*$basename/)].sha256` failed

Now that the cause of this issue is clear, may I ask if we could just replace all the `+` in the `$basename` part of the URL with a `%2B` before decoding it or provide a better approach to cope with similar situations?

* *Are there some more special characters that may lead people into this trap*?
* *Will the replacing `+` with `%2B` harm the other logics*?
* *Is there a possibility that some apps do use `+` to represent a space*?

Any ideas?

sinloss:
- Closes #2439
- Closes #2526
- Closes #2770 
- Closes #2924
- Closes #3113
- Closes #3360
- Closes #3446 
- Closes #3490

As the flutter officially used `+` instead of `-` to concatenate `hotfix` in their version string since *2019/7/9* or before, the scoop version has stuck at `1.5.4-hotfix.2` yet the real-world version has already been advanced to `1.12.13+hotfix.5`.

So here is this PR.

### DIFF
--- a/bucket/flutter.json
+++ b/bucket/flutter.json
@@ -35,7 +35,7 @@
         "url": "https://storage.googleapis.com/flutter_infra/releases/stable/windows/flutter_windows_v$matchVer+$matchBuild-stable.zip",
         "hash": {
             "url": "https://storage.googleapis.com/flutter_infra/releases/releases_windows.json",
-            "jsonpath": "$.releases[?(@.archive =~ /.*flutter_windows_v$matchVer[+]$matchBuild-stable.zip/)].sha256"
+            "jsonpath": "$.releases[?(@.archive =~ /.*flutter_windows_v$matchVer[-+]$matchBuild-stable.zip/)].sha256"
         }
     },
     "suggest": {
--- a/bucket/flutter.json
+++ b/bucket/flutter.json
@@ -1,14 +1,14 @@
 {
     "homepage": "https://flutter.dev/",
-    "version": "1.5.4-hotfix.2",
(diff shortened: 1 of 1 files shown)


linsui:
dulplicate of https://github.com/lukesampson/scoop-extras/pull/3113 https://github.com/lukesampson/scoop-extras/pull/3113

sinloss:
Close this because of duplication

sinloss:
Reopen this PR as the other approaches did not really fix the whole issue. So I introduced this approach I used locally, and it works fine including the hash extraction. This PR would also change the `hash.jsonPath` of `autoupdate` in order to avoid the controversial character `+` mentioned in lukesampson/scoop#3613

PR #3494, opened 2020-01-15
keybase: Add version 5.5.0

ddavness:
https://keybase.io

## Blocked by keybase/client#22039

Apparently they assume that all the binaries that they need to execute reside in `%localappdata%` - within a scoop installation, this is false, and the programs will not run.

Alongside with that, the `env_set` field (currently) serves no use as they don't actually look for it for anything (yet)


ddavness:
No updates on this at the moment. This PR might remain stale until the end of time... ¯\\\_(ツ)_/¯

pratikpc:
@ddavness if it becomes possible again, do create a PR for the same

PR #3501, opened 2020-01-16
oripa: Add version 0.36

linsui:


bucket/oripa.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "0.36",
+    "homepage": "http://mitani.cs.tsukuba.ac.jp/oripa/",
+    "description": "Drawing software dedicated to designing the crease patterns of origami",
+    "license": "GPL-3.0-or-later",
+    "url": "https://github.com/oripa/oripa/releases/download/v0.36/oripa-0.36.jar#/oripa.jar",
+    "hash": "5e72f6aea5722e80a11fa8098cf3cd1f9c21be67014da39360abf640be194ebc",
+    "pre_install": [
+        "Expand-7zipArchive \"$dir\\oripa.jar\" -Switches 'icon\\oripa.gif'",
+        "Set-Content \"$dir\\oripa.bat\" 'start javaw -jar %~dp0\\oripa.jar %*' -Encoding ASCII",
```
Ash258: Do not hardcore string, which are already variable, to just save few character of escaping.

```suggestion
        "Set-Content \"$dir\\oripa.bat\" \"start javaw -jar %~dp0\\$fname %*\" -Encoding ASCII",
```

bucket/oripa.json
```diff
@@ -0,0 +1,36 @@
...
+        "$stream = [System.IO.File]::OpenWrite(\"$dir\\icon.ico\")",
+        "$bitmap = [System.Drawing.Image]::FromFile(\"$dir\\icon\\oripa.gif\")",
+        "([System.Drawing.Icon]::FromHandle($bitmap.GetHicon())).Save($stream)"
```
Ash258: I bet the icon will be blurry.

linsui: This script works well for gif pictures. Only png will be broken.

Ash258:
I am having weird issue. Can anyone test with native W10 powershell 5?

A

bucket/oripa.json
```diff
@@ -0,0 +1,36 @@
...
+    "version": "0.36",
+    "homepage": "http://mitani.cs.tsukuba.ac.jp/oripa/",
+    "description": "Drawing software dedicated to designing the crease patterns of origami",
+    "license": "GPL-3.0-or-later",
+    "url": "https://github.com/oripa/oripa/releases/download/v0.36/oripa-0.36.jar#/oripa.jar",
+    "hash": "5e72f6aea5722e80a11fa8098cf3cd1f9c21be67014da39360abf640be194ebc",
+    "pre_install": [
+        "Expand-7zipArchive \"$dir\\oripa.jar\" -Switches 'icon\\oripa.gif'",
+        "Set-Content \"$dir\\oripa.bat\" \"start javaw -jar %~dp0\\$fname %*\" -Encoding ASCII",
+        "$stream = [System.IO.File]::OpenWrite(\"$dir\\icon.ico\")",
```
ddavness: **This is for Windows Powershell 5.1 (Native)**
Apparently you need to grab the type before using `[System.Drawing]`

```suggestion
        "Add-Type -AssemblyName System.Drawing",
        "$stream = [System.IO.File]::OpenWrite(\"$dir\\icon.ico\")",
```

Neverthless, even after this fix, for some reason the icon doesn't show up - does it relate to the fact that the source is actually not a square? (Dimensions are 30x31 pixels)

image

linsui:
When I tested yesterday it worked but now it doesn't work any more...
The code comes from https://stackoverflow.com/a/13764392 . It seems it doesn't works for image not square. I don't know why it works in pscore as this is totally a C# script.

ddavness:
Not sure how hard is it (might be trivial, might be impossible) but one could try adding a column of empty pixels (alpha 0) so that we hit 31x31...

ddavness:
A better long term answer to this problem is actually filing an issue in their repository and ask them to "fix" the icon. (And actually provide an `.ico` file while they're at that, because converting it is still ugly)

linsui:
> Not sure how hard is it (might be trivial, might be impossible) but one could try artificially adding a column of empty pixels (alpha 0) so that we hit 31x31...

It works. But I don't know how to do that in C#.

ddavness:
Well, I managed to achieve resizing the icon to 31x31, but that didn't work... Neither did resizing it to 32x32 (which is a standard size...), even when clearing icon cache and all that jazz... 😞:

Here's the full `pre_install` snippet if you want (also I added it as new suggestion option to the review above):

```suggestion
    "pre_install": [
        "Expand-7zipArchive \"$dir\\oripa.jar\" -Switches 'icon\\oripa.gif'",
        "Set-Content \"$dir\\oripa.bat\" \"start javaw -jar %~dp0\\$fname %*\" -Encoding ASCII",
        "Add-Type -AssemblyName System.Drawing",
        "$stream = [System.IO.File]::OpenWrite(\"$dir\\icon.ico\")",
        "$bitmap = [System.Drawing.Image]::FromFile(\"$dir\\icon\\oripa.gif\")",
        "$icon = New-Object System.Drawing.Bitmap 32,32",
        "$iconStream = [System.Drawing.Graphics]::FromImage($icon)",
        "$iconStream.Clear([System.Drawing.Color]::Black)",
        "$iconStream.DrawImage($bitmap, 0, 0, 30, 31)",
        "([System.Drawing.Icon]::FromHandle($icon.GetHicon())).Save($stream)",
        "$icon.Dispose()",
        "$stream.Dispose()"
    ],
```

@Ash258 @linsui See if that works on your machines. It doesn't seem to work on mine, but hopefully it's specific to it...

linsui:
It works for me! Thanks! But I wonder will @Ash258 approve this...

ddavness:
I mean, with the `pre_install` script taking around 40% of the manifest, everything because of an icon... 🤔 - here's the alternatives I can think of:

- Fix the icon by hand and host it somewhere. The manifest will then grab the icon;
- File an issue/PR for an actual icon to be provided.

Anyway, we should agree that more time was spent on this issue than we'd care to admit.

PR #3502, opened 2020-01-16
vdesk: Add version 1.2.0

Teoxoy:


Teoxoy:
I got the hash using powershell `Get-FileHash`.

> Move it to Main bucket with fixed manifest below

Ok, thanks! But why is [`cmdow`](https://github.com/lukesampson/scoop-extras/blob/master/bucket/cmdow.json) in the `extras` bucket instead of the `main` one?

PR #3507, opened 2020-01-19
awake: Add shortcuts

wenmin92:
Awake is not a command line program. Should remove `bin`, and add `shortcuts`.


ddavness:
I'd say keep `bin`, since that someone could use this in a script.
Even a lot of GUI-only apps (e.g. Blender) still have a `bin` field in their manifest, so it's best to keep it.

wenmin92:
Thanks @ddavness , I'll add bins and remember this in the future.

PR #3508, opened 2020-01-19
iconsextract: Add version 1.47

wenmin92:
[IconsExtract](https://www.nirsoft.net/utils/iconsext.html) - Extract icon/cursor stored in EXE, DLL, OCX, CPL files.

### DIFF
--- a/bucket/iconsextract.json
+++ b/bucket/iconsextract.json
@@ -15,8 +15,6 @@
     ],
     "persist": "iconsext.cfg",
     "checkver": "IconsExtract v([\\d.]+)",
-        "regex": "IconsExtract v([\\d.]+)"
-    },
     "autoupdate": {
         "url": "https://www.nirsoft.net/utils/iconsext.zip"
     }
--- a/bucket/iconsextract.json
+++ b/bucket/iconsextract.json
@@ -14,7 +14,7 @@
         ]
     ],
     "persist": "iconsext.cfg",
-    "checkver": {
+    "checkver": "IconsExtract v([\\d.]+)",
         "regex": "IconsExtract v([\\d.]+)"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/iconsextract.json
```diff
@@ -0,0 +1,23 @@
...
+    "pre_install": "if(!(Test-Path(\"$persist_dir\\iconsext.cfg\"))) { New-Item \"$dir\\iconsext.cfg\" | Out-Null }",
+    "bin": "iconsext.exe",
+    "shortcuts": [
+        [
+            "iconsext.exe",
+            "IconsExtract"
+        ]
+    ],
+    "persist": "iconsext.cfg",
+    "checkver": {
```
Ash258:
```suggestion
    "checkver": "IconsExtract v([\\d.]+)",
```

Ash258:
I do not think duplicating manifests is good idea. Nirlauncher is in extras and separate bucket already exists with all tools. Same as JetBrains and SysInternals.

Better to improve quality of nirsoft bucket.

wenmin92:
Although this app is created by _nirsoft_, the `nirsoft` bucket does not contains it.
We cannot search the contained apps by scoop. I searched "iconsectract", but not find, so I PR this app.
I think a whole Nirlauncher is not a good idea, users may not need them all, people will only search some single apps, if them can't search, they didn't know the nirlauncher contains it.

Ash258:
> nirsoft bucket does not contains it

If you are talking about old nirsoft bucket, then lukesampson/scoop#2808

image

> I searched "iconsectract"

Name of app is not `iconsextract`, but `iconsext` (homepage). lukesampson/scoop#3446

wenmin92:
Thanks for these infomation. 
Since this app was already in the https://github.com/MCOfficer/scoop-nirsoft bucket, I will close this PR.

PR #3510, opened 2020-01-19
added manifest for daily rstudio build

edavidaja:
Rstudio publishes a preview version of its editor, which this manifest installs.


cderv:
@edavidaja do you plan on moving this PR to the versions bucket ?

PR #3512, opened 2020-01-20
coopeliasim: Add version 4.0.0

wngfra:
Added install manifest of CoppeliaSim robot simulator


wngfra:
Hi, the application doesn’t store options internally. All user options are stored with the user created scene.
Thanks for the fix.

PR #3515, opened 2020-01-22
nswagstudio: Add version 13.2.2

foriequal0:
~~`NSwagStudio.*.nupkg` is built with this [NSwagStudio.nuspec](https://github.com/RicoSuter/NSwag/blob/master/src/NSwagStudio.Chocolatey/NSwagStudio.nuspec). It contains `NSwagStudio.msi` inside `tools\` and some chocolately ps1 files, and bunch of other files outside of the directory. That's why I've used `"extract_to"`.~~

### DIFF
--- a/bucket/nswagstudio.json
+++ b/bucket/nswagstudio.json
@@ -1,0 +1,34 @@
+{
+    "version": "13.2.2",
+    "description": "The OpenAPI/Swagger API toolchain for .NET and TypeScript",
+    "homepage": "https://github.com/RicoSuter/NSwag",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://ci.appveyor.com/api/buildjobs/pl7r86mp8k60ccx2/artifacts/src/NSwagStudio.Installer/bin/Release/NSwagStudio.msi",
+            "hash": "f77401db0c3c41a32af760cef9eb1cc8d86e2537f59b4e3a862a5aee6d96b573"
+        }
+    },
+    "extract_dir": "Rico Suter\\NSwagStudio",
+    "bin": [
+        "NSwagStudio.exe",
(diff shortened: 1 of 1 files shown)


foriequal0:
It nags to update to 13.2.2, but it hasn't released to Chocolately.
13.2.2 is only released to Appveyor, but it is SPA and it seems to require at least two REST API calls to locate the build artifact. (the first one is to get the build job id, and the second one is to get the artifact of the build job)
I think Chocolately is our best way.

foriequal0:
I found where they put the update URL from here:
https://github.com/RicoSuter/NSwag/blob/c7cdefa8d1b1dd429d7561f990eb03c8988b30fe/src/NSwagStudio/Views/MainWindow.xaml.cs#L58
I can fix the lagging version with it but the check isn't made to be disabled. A user might see the update notification for a while.

foriequal0:
I've changed it to directly download `msi`. The update page provides the version and the download URL in XML, but I couldn't capture the version and the job id both with XPath. I resorted to the regex.

foriequal0:
Oh sorry. I found lately what you've written behind toggle buttons. I'll apply that.

bucket/nswagstudio.json
```diff
@@ -22,12 +22,12 @@
     ],
     "checkver": {
         "url": "https://rsuter.com/Projects/NSwagStudio/updates.php",
-        "regex": "(?sm)<version>([\\d.]+).*?https://(?<url>.*)</download"
+        "regex": "(?sm)<version>([\\d.]+).*?https://ci.appveyor.com/(?<url>.*)</download"
```
foriequal0: 👍

foriequal0:
I'll squash the commit and reset the author. I want to change the author email with my personal email.

PR #3516, opened 2020-01-22
multicommander: Update to version 9.6.1.2582

pcrama:
Multicommander's website displays an older version than the latest version offered by MultiUpdate (companion tool of multicommander): by using the same version info download as MultiUpdate, we can track the latest version more closely.

I took the opportunity to update multicommander to the latest version at the same time.

### DIFF
--- a/bucket/multicommander.json
+++ b/bucket/multicommander.json
@@ -30,8 +30,8 @@
         ]
     ],
     "checkver": {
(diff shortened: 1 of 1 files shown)


pcrama:
Thank you for the suggestion.  I modified the first commit to include it and force-pushed the modifications.

PR #3521, opened 2020-01-22
aimp: Update to version 4.60.2170

RawBytes28:


Ash258:
Latest version listed on page is still old. 

I added harckaround using their update service. I have no idea how long it will survive.

PR #3526, opened 2020-01-24
 fadetop: Add version 3.1.0.171

pcrama:
I added two apps to remind myself to take regular breaks.

Both have their quirks:
- FadeTop: is a portable app but I was unable to make the persist mechanism work because the Settings.xml is a file next to the executable, not in a directory next to the executable
- stretchly: is not really portable, it stores its settings in AppData\Roaming\stretchly


pcrama:
Thank you for the suggestions and the fix for the persistence.

PR #3553, opened 2020-02-03
vidi: Update to 0.2.5.6 (add au.url)

rasa:


Ash258:
Autoupdate was removed intentionally to not get alpha versions

image

image

PR #3558, opened 2020-02-03
qbittorrent: Update checkver to use sf.net

rasa:
Addresses: The operation has timed out. URL https://www.qbittorrent.org/download.php is not valid. See #2308.


Ash258:
It looks like it was one time issue.

PR #3559, opened 2020-02-03
logexpert: Change to Choco download

niheaven:


rasa:
Why is this required? Have the app maintainers suggested we do this?

niheaven:
You can extract hash in Choco page.

rasa:
> You can extract hash in Choco page.

IMO, that’s not a compelling enough reason. We should only be using chocolatey’s servers, if there is no other option for the binaries.

PR #3605, opened 2020-02-16
wavebox10-pro: Add version 10.0.69.2

r15ch13:
see https://github.com/lukesampson/scoop-extras/pull/3604


Ash258:
what is difference between normal?

r15ch13:
The newer version that's only available if one paid for it. I didn't test if this also works for free accounts. (I have a subscription)
The normal version will show a prompt to update when used with a paid account.

PR #3609, opened 2020-02-17
xmind: Add version 11.1.2

wenmin92:
[XMind: ZEN](https://www.xmind.net/zen/) is the most professional and popular mind mapping tool.


bucket/xmind-zen.json
```diff
@@ -0,0 +1,38 @@
+{
+    "version": "10.0.1",
```
Ash258:
```suggestion
    "version": "10.0.2",
```

bucket/xmind-zen.json
```diff
@@ -0,0 +1,38 @@
...
+    "shortcuts": [
+        [
+            "XMind ZEN.exe",
+            "XMind ZEN"
+        ]
+    ],
+    "checkver": {
+        "url": "https://dl2.xmind.cn/latest-win64.yml",
+        "regex": "XMind-ZEN-for-Windows-64bit-([\\d.]+)-(?<time>\\d+)\\.exe"
+    },
```
Ash258:
```suggestion
    "checkver": {
        "url": "https://www.xmind.net/download/",
        "regex": "(?sm)Windows-32bit-([\\d.]+)-(?<time32>\\d+)\\.exe.*?Windows-64bit-[\\d.]+-(?<time64>\\d+)\\.exe"
    },
```

bucket/xmind-zen.json
```diff
@@ -0,0 +1,38 @@
...
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-64bit-$version-$matchTime.exe#/dl.7z"
+            }
```
Ash258:
```suggestion
                "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-64bit-$version-$matchTime64.exe#/dl.7z"
            },
            "32bit": {
                "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-32bit-$version-$matchTime32.exe#/dl.7z"
            }
```

bucket/xmind-zen.json
```diff
@@ -0,0 +1,38 @@
...
+            "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-64bit-10.0.1-202001022330.exe#/dl.7z",
+            "hash": "298d2e0eddc3fbb897ac134d6a2d70274a821d490dc7e4b84cd560a63de429ce",
+            "installer": {
+                "script": [
+                    "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\"",
+                    "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall*\" -Force -Recurse"
+                ]
+            }
+        }
+    },
```
Ash258: Recalculate hashes and try with this

```suggestion
        "64bit": {
            "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-64bit-10.0.2-202002120025.exe#/dl.7z",
            "hash": "asdfasdfasd",
            "installer": {
                "script": "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\""
            }
        },
        "32bit": {
            "url": "https://dl2.xmind.cn/XMind-ZEN-for-Windows-32bit-10.0.2-202002120038.exe#/dl.7z",
            "hash": "faskdjfsdla",
            "installer": {
                "script": "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-32.7z\" \"$dir\""
            }
        }
    },
    "post_install": "Remove-Item \"$dir\\`$*\", \"$dir\\Uninstall*\" -Recurse",
```

Ash258:
I cannot really test as downloading is painfully slow for all mirrors they provide.

I would prefer for scoop `http://dl2.xmind.net/xmind-downloads/XMind-....zip`

wenmin92:
> I cannot really test as downloading is painfully slow for all mirrors they provide.
> 
> I would prefer for scoop `http://dl2.xmind.net/xmind-downloads/XMind-....zip`

Thanks for your hard work. The download url has changed.

Ash258:
> prefer for scoop ;**http**:/...;

image

wenmin92:
> > prefer for scoop ;**http**:/...;
> 
> image

Ahh, sorry, I lost it. Will change now.

rashil2000:
@wenmin92 can you update the existing xmind8.json (rename to just xmind.json) instead of creating a new manifest? Thanks.

wenmin92:
> @wenmin92 can you update the existing xmind8.json (rename to just xmind.json) instead of creating a new manifest? Thanks.

xmind8 is still available, do we really want to deprecate it?
https://www.xmind.net/xmind8-pro/

niheaven:
Move `xmind8` to `versions` and add a new xmind.

rashil2000:
Xmind 8 is around 2 years old now, and latest is version 10. So we should update it.

Moving to `Versions` bucket is not deprecation. This is standard procedure for manifests. For example, PostgreSQL 13 still receives updates, but the latest is 14. So 14 is in `Main` bucket, and 13 is in `Versions` bucket.

wenmin92:
Okay, that's fine to do.

But xmind's `checkver` is having trouble with the 32-bit and 64-bit download links containing a timestamp variable in addition to the version and architecture. In fact the actual download links are fetched separately by requesting different URLs.

https://github.com/ScoopInstaller/Scoop/issues/4611

rashil2000:
That's fine. In the meanwhile, you can use a script in checkver, like this: https://github.com/ScoopInstaller/Extras/pull/7586/files

wenmin92:
> That's fine. In the meanwhile, you can use a script in checkver to follow links, like this: https://github.com/ScoopInstaller/Extras/pull/7586/files

Thanks, I have considered this method before, but without an example, I did not try it. It is easiest if the architecture block is directly supported.

Now that I have an example, I will try to update it.

niheaven:
image

wenmin92:
@niheaven Please add details, I can't understand what you're trying to say with a few simple words or a screenshot.

niheaven:
XMind 8 is still here, so you should move xmind8 to `versions` bucket as I said.

> Move `xmind8` to `versions` and add a new xmind.

niheaven:
I've said very clearly and don't know why you thumb down.

wenmin92:
> I've said very clearly and don't know why you thumb down.

I wish you could be more friendly and add enough information instead of standing there and simply giving orders.

niheaven:
> xmind8 is still available, do we really want to deprecate it? [xmind.net/xmind8-pro](https://www.xmind.net/xmind8-pro/)

You knew scoop as a long-time contributor and should deeply understand our bucket's structure.

It's an advice to your comment, and I'm sorry that you thought this as an order. Keep simple, you mean I must add 'please' to every comment?

wenmin92:
You don't need to add "please", but please give people a sense of goodwill.

rashil2000:
> It is easiest if the architecture block is directly supported.

Yes, but it's non-trivial, so it could take time.

The CI failed because autoupdate URL is not in URL format. See: 

image


You can fix it by doing this - https://github.com/ScoopInstaller/Extras/pull/7586/commits/8217c8863f9a499f1942fe001cc848fe73f03c03

bucket/xmind.json
```diff
@@ -8,14 +8,14 @@
     },
     "architecture": {
         "64bit": {
-            "url": "https://dl3.xmind.net/XMind-for-Windows-64bit-11.1.2-202111071931.exe#/dl.7z",
+            "url": "https://dl3.xmind.net//XMind-for-Windows-64bit-11.1.2-202111071931.exe#/dl.7z",
```
rashil2000, line 11: Why use 2 slashes? This URL now results in 403 error. Same for autoupdate. You can remove the beginning slash in checker script by `.TrimStart('/')`

rashil2000 (changes requested):

bucket/xmind.json
wenmin92: Apologize for my carelessness

PR #3625, opened 2020-02-25
xampp: Add version 8.0.12-0

### LINKED ISSUE #3570: [Request] Add XAMPP
It's like LAMP, but for Windows (available for other OSes too) and has nice GUI. It contains: Apache web server, MariaDB, PHP interpreted language, and Perl. Aside these, there are Apache Tomcat, MercuryMail server, FileZilla FTP, and an out-to-dated SQLite3 binary (from MercuryMail). It has shell script that bind `%PATH%` and other environment variables in `xampp_shell.bat`.

Site: https://www.apachefriends.org

Here's the recommendation list of persistent directories:
```
apache\conf
htdocs
mysql\data
php\php.ini
tmp
```

Personally, I don't mind these directories and files are not persistent (except `htdocs/`) because these can be customized.

issaclin32:
close #3570

**XAMPP** ([homepage](https://www.apachefriends.org/)) is an Apache distribution containing MariaDB, PHP, and Perl.

Notes:

* If there are more files that needs to be in *bin* / *shortcut* / *persist* , please tell me.

* The Portable Version **does not** contain *FileZilla FTP* and *Mercury Mail Server*. The service installations are also disabled. I tried to extract files from the non-portable installer, but the installer could not be properly extracted by *7-zip*, *innounp* or *Uniextract*.

* `vcredist2017` is required according to the readme file.
> PHP in this package needs the Microsoft Visual C++ 2017 Redistributable package from
http://www.microsoft.com/en-us/download/. Please ensure that the VC++ 2017 runtime
libraries are installed on your system.

Since Visual C++ 2015, 2017 and 2019 all share the same redistributable files, I use **vcredist 2019** in `suggest`.
(https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads)

* *Sendmail* can work independently without *Mercury Mail Server*. (see `$dir\sendmail\ReadMe.html` for details)

* The `$dir\tmp` folder is for saving PHP sessions. I think it does not need to be in *persist*. (see `$dir\tmp\why.tmp`)
> Why this tmp-Folder?
PHP need it for saving
the Sessions. 
So please do not delete it!

### DIFF
--- a/bucket/xampp.json
+++ b/bucket/xampp.json
@@ -4,7 +4,7 @@
     "license": "GPL-2.0-only",
     "description": "Apache distribution containing MariaDB, PHP, and Perl",
     "notes": "Follow the instructions on '$dir\\readme_en.txt' to set up XAMPP.",
-    "suggests": {
+    "suggest": {
         "Visual C++ 2017 Redistributable": "extras/vcredist2017"
     },
     "architecture": {
--- a/bucket/xampp.json
+++ b/bucket/xampp.json
@@ -1,0 +1,55 @@
+{
+    "homepage": "https://www.apachefriends.org/index.html",
+    "version": "7.4.2-0",
+    "license": "GPL-2.0-only",
+    "description": "Apache distribution containing MariaDB, PHP, and Perl",
+    "notes": "Follow the instructions on '$dir\\readme_en.txt' to set up XAMPP.",
+    "suggests": {
+        "Visual C++ 2017 Redistributable": "extras/vcredist2017"
+    },
+    "architecture": {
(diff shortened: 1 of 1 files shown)


Ash258 (dismissed):
Otherwise manifest looks good, but I cannot properly review full functionality with all persists, etc, ...

rashil2000:
I think they have a newer release (8)

rashil2000:
SSL error again in the PullRequestHandler, but manifest looks good to me 👍

rashil2000 (approved):

PR #3636, opened 2020-03-01
nushell: add manifest with version 0.10.0

littleli:
Nu draws inspiration from projects like PowerShell, functional programming languages, and modern CLI tools. Rather than thinking of files and services as raw streams of text, Nu looks at each input as something with structure.


Ash258:
https://github.com/ScoopInstaller/Main/blob/master/bucket/nu.json

PR #3653, opened 2020-03-05
wizmouse: Add version 1.7.0.3

Dabombber:
There's already a couple of antibody-software.com programs in this bucket, but it's missing this quite useful utility.

> WizMouse is a mouse enhancement utility that makes your mouse wheel work on the window currently under the mouse pointer, instead of the currently focused window. This means you no longer have to click on a window before being able to scroll it with the mouse wheel. This is a far more comfortable and practical way to make use of the mouse wheel.
>
> WizMouse can also optionally enable the mouse wheel in applications that don't support it. It does this by translating mouse wheel commands into scroll bar commands that all applications can understand and process.

### DIFF
--- a/bucket/wizmouse.json
+++ b/bucket/wizmouse.json
@@ -8,7 +8,7 @@
     "pre_install": "if (-not (Test-Path \"$persist_dir\\wizmouse.ini\")) { New-Item \"$dir\\wizmouse.ini\" -ItemType File | Out-Null }",
     "bin": [
         "WizMouse.exe",
-        "WizMouseConfig.exe",
+        "WizMouseConfig.exe"
     ],
     "shortcuts": [
         [
--- a/bucket/wizmouse.json
+++ b/bucket/wizmouse.json
@@ -18,7 +18,7 @@
         [
             "WizMouseConfig.exe",
             "WizMouseConfig"
-        ],
+        ]
     ],
     "persist": "wizmouse.ini",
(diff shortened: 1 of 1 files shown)


bucket/wizmouse.json
```diff
@@ -0,0 +1,28 @@
+{
+    "version": "1.7.0.3",
+    "description": "Allows mouse wheel scrolling for the window under the pointer instead of the currently focused window.",
```
Ash258:
```suggestion
    "description": "Mouse enhancement utility that makes mouse wheel work on the window currently under the mouse pointer.",
```

bucket/wizmouse.json
```diff
@@ -0,0 +1,28 @@
...
+            "WizMouse.exe",
+            "WizMouse"
+        ],
+        [
+            "WizMouseConfig.exe",
+            "WizMouseConfig"
+        ]
+    ],
+    "persist": "wizmouse.ini",
+    "checkver": "v([\\d.]+)",
```
Ash258:
```suggestion
    "checkver": "([\\d.]+)\\?",
```

Ash258: .

Dabombber: Wouldn't `ver ([\\d.]+)` be better since it'll only match the top number

Ash258: No

PR #3667, opened 2020-03-08
falconx: Change shortcut

janymal:


janymal:
> Why did you change version?

To force update of the app

Ash258:
What? It will stay only for 1 hour max. Then it will be fixed by bot.

Remove the version change

janymal:
Okay, if you say so. Done

PR #3668, opened 2020-03-08
Transmission: add env variable TRANSMISSION_HOME to manifest

nixxo:
TRANSMISSION_HOME env variable allows settings to be in the persist folder and not in the APPDATA folder.


Ash258:
As soon as someone update the manifest, current configuration will be lost.

nixxo:
ok.

Issue #3654, opened 2020-03-05
[Request] Add LAVFilters


Meng5in:
LAVFilters is contained in lav-filters-megamix-np, but I can't install it separately.

[https://github.com/Nevcairiel/LAVFilters](url)

SorYoshino:
Hello, @Voya9e .

lavfilters has been added to the `Nonportable` bucket. You can install it using `scoop install nonportable/lavfilters-np`.

Please close this issue.

PR #3675, opened 2020-03-10
lavfilters: Add version 0.74.1

Dabombber:
As requested in #3654.

Registered directly instead of using the provided batch files so it'd be easier to change if lukesampson/scoop#3412 ever gets done.

I don't think it's currently possible to make shortcuts using variables (eg run `$env:SystemDirectory\rundll32.exe`, or use `$dir` in an argument) so I generated them on install.

I'm also not quite sure about the license. [free-manga-downloader ](https://github.com/riderkick/FMD/blob/master/license.txt) is listed as `GPL-2.0-only`, while [xournalpp](https://github.com/xournalpp/xournalpp/blob/master/LICENSE) is `GPL-2.0-or-later` even though they're exactly the same.


Ash258:
I really do not want to see such system manipulation in manifests in official buckets.

Unsubscribing. Maybe someone else will have other opinion

PR #3729, opened 2020-03-18
potplayer@200317: Hash fix

wenmin92:
The checkver only checks main version, not build version, we should change the checkver more reliable.


Ash258:
/verify

Do not fix hashes manually. Open issue

PR #3740, opened 2020-03-19
ibmcloud-cli@0.22.1: Fix hash

se35710:
Current manifest for 0.22.1 contains wrong hashes.


Ash258:
Just open issue instead of manually fixing.

PR #3753, opened 2020-03-22
Adding gradle latest release candidate

d3vel0per:
Adding gradle latest release candidate as the current GA ie. 6.2.2 fails with latest JDK GA ie. JDK14. More details here: https://github.com/gradle/gradle/issues/10248


Ash258:
Why extras bucket??

Open this PR into versions bucket

d3vel0per:
Because the documentation here: https://github.com/lukesampson/scoop/wiki/Buckets mentioned that "There is also an optional **versions bucket containing older versions** of some well-known packages". Also I did find quite a few googlechrome*** version manifests in extras.
 
Anyways raised a new PR in versions bucket now.

PR #3775, opened 2020-03-27
GIMP: twain,2.exe appears to have been renamed to twain,3.exe

kimbirkelund:
It appears that a file was renamed from one version to the next. The install script tried to delete twain,2.exe; it now deletes twain,3.exe.

I have no idea if it is the right thing to do or how to test it, but the install works again.

### DIFF
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -12,7 +12,7 @@
             "Get-ChildItem -Filter '*.debug' -Recurse | Remove-Item -Recurse",
             "if ($architecture -eq '64bit') {",
             "   Rename-Item 'lib\\gimp\\2.0\\plug-ins\\twain\\twain,1.exe' 'twain.exe'",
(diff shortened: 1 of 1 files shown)


pigsflew:
it doesn't give an error on install, but now every time i launch I get a bunch of errors that say "The application was unable to start correctly (0xc000007b)"

First
Second
Third

After which the GIMP tells me there's a new version available (2.10.18) even though this is 2.10.18-2. Attempting to install gimp@2.10.18 fails

scoop install gimp@2.10.18

It looks like here the url is not being built correctly and the MD5 is not being detected. Same thing happens for `scoop install gimp@2.10.18-1` as well.

scoop install gimp@2.10.18-1

pigsflew:
I think I've figured it out; I suggest #3814 instead.

kimbirkelund:
I use gimp minimally so I'm in no way a qualified to say how best to install it. I'm fine with closing this PR in favor of the other if it makes more sense.

PR #3814, opened 2020-04-05
gimp: Fix installation

pigsflew:
Don't persist the install lib/plug-ins directory; there's binaries in it that may not work between installs. User plugins should be installed to e.g. %APPDATA%\GIMP\2.10\plug-ins.

Also uses matchHead and matchTail in the autoupdate url which I believe should correct the install gimp@version feature. (unsure how to test this; `scoop install .\gimp.json@2.10.18` didnt work the way i expected it to).

Twain now allows for twain,3 and twain,2 which was causing a breakage. This should allow new and old versions to install correctly.


Ash258:
@rasa What are you talking about??

Completely nonsense what you said. All of the substitutions are automatic/native.

Not talking about that autoupdate was disabled on purpose since, binaries and shortcut properties use hardcoded version in name...

The installation is broken since it was added by @niheaven . 32bit version cannot be installed by clearly looking at the code.

rasa:
I installed this PR locally, and it insatlled/run fine. Afterwards, the autoupdate was removed in https://github.com/lukesampson/scoop-extras/commit/1e5b39cb41ad8d1a815951850b4b9c4bb38966fa#diff-7b1d642f5782ee85e8feb8dab179f116
It sounds like it shouldn’t be merged in its present form as the 32bit arch is broken? Should we remove the 32bit arch?

niheaven:
https://github.com/lukesampson/scoop/pull/3502 is the further solution for such complex inno type installer, but scoop-core doesn't move on for a long time...

BTW, the 32bit installer works as I added the manifest, but since I didn't use 32bit, it may be broken soon after.

Ash258:
How can it work, when you are moving directory, which exists only in 64bit architecture?

image

niheaven:
For GIMP 2.10.18, using `innounp -x -d"gimp18" -c'{app}' -y .\gimp-2.10.18-setup.exe`

gimp-2.10.18-setup

Since 32bit and 64bit are in one installer, there are some duplicate files that won't be extracted without `-a` switch. These files are in '32' subdir of 64bit / in base dir of 32bit.

But 2.10.20 doesn't have '32' subdir with `-x` flag, I'll check it again. https://github.com/lukesampson/scoop/pull/3502 should work fine, 'cos it extracts files according to `install_script.iss`.

Issue #3694, opened 2020-03-15
[Request] Add Folding@Home


NatoBoram:
https://foldingathome.org

> **The Folding@home software runs while you do other things.**
> 
> While you keep going with your everyday activities, your computer will be working to help us find cures for diseases like cancer, ALS, Parkinson’s, Huntington’s, Influenza and many others.

PR #3818, opened 2020-04-07
foldingathome: Add version 7.5.1

pigsflew:
for issue #3694 . Must be installed global (installer requires sudo).

### DIFF
--- a/bucket/foldingathome.json
+++ b/bucket/foldingathome.json
@@ -1,0 +1,33 @@
+{
+    "version": "7.5.1",
+    "homepage": "https://foldingathome.org/",
+    "description": "Folding@home is a distributed computing project focused on disease research.",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://foldingathome.org/support/faq/opensource/"
+    },
+    "url": "https://download.foldingathome.org/releases/public/release/fah-installer/windows-10-32bit/v7.5/fah-installer_7.5.1_x86.exe",
+    "hash": "md5:5b02471b6919244ccfbdc0d6d72aaf98",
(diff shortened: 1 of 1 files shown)


pigsflew:
It downloads the installer and runs it. It's because installer is set to an empty object, which by default, has `file` property set to the last URL downloaded.

However, I could see wanting to be explicit here, as this might not be a reliable setting.

I'm actually not sure i like how this works to begin with. It requires the UAC, it doesn't uninstall cleanly... I'm going to look at this again but I don't really know what i'm doing and this one's more complex than the others.

Ash258:
Yes I know. But you are not passing any arguments to it. Which most likely will result into installing it into program files or similar junk, which is unacceptable in official bucket

pigsflew:
That's more than fair.

I have no idea what to do about this one; extracting the exe fails (though I can make 7zip do it "sorta" but i don't know what i need to do to make it cleanly set up that way).

They have an [FAHClient only tar.bz2](https://download.foldingathome.org/releases/public/release/fahclient/windows-10-32bit/v7.5/) that i was able to find, but it's also got an error on unpacking, *and* the FAHClient references the other executables that aren't present in this pack.

I can continue to take stabs at this on my own time; in the meantime I'm closing this mr, thanks!

Ash258:
Try to open PR in non-portable bucket with this version of manifest.

pigsflew:
Actually, I want to try to rewrite it to use the build from [PortableApps](https://portableapps.com/apps/utilities/folding-at-home-portable) - added last week, but it might do the trick! I'll update this and resubmit tonight, thanks for your patience!

PR #3847, opened 2020-04-12
bandizip: Update download URL

zhangwang97:


Ash258:
I do not understand.

Why? Link `https://www.bandisoft.com/bandizip/dl.php?portable` is permanent redirect to `https://dl.bandisoft.com/bandizip.std/BANDIZIP-PORTABLE.ZIP`, which contains both 32bit and 64versions.

Both URLs defined in manifest could be downloaded without any issue (https://dl.bandisoft.com/bandizip/BANDIZIP-PORTABLE-GL-64BIT.ZIP, https://dl.bandisoft.com/bandizip/BANDIZIP-PORTABLE-GL-32BIT.ZIP)

zhangwang97:
> I do not understand.
> 
> Why? Link `https://www.bandisoft.com/bandizip/dl.php?portable` is permanent redirect to `https://dl.bandisoft.com/bandizip.std/BANDIZIP-PORTABLE.ZIP`, which contains both 32bit and 64versions.
> 
> Both URLs defined in manifest could be downloaded without any issue (https://dl.bandisoft.com/bandizip/BANDIZIP-PORTABLE-GL-64BIT.ZIP, https://dl.bandisoft.com/bandizip/BANDIZIP-PORTABLE-GL-32BIT.ZIP)

The file I download from URL `https://dl.bandisoft.com/bandizip/BANDIZIP-PORTABLE-GL-64BIT.ZIP` is version 6.26, instead of version 7.03. I notice that 6.26 is the last version of version 6, so I guess the official changed the download address since version 7. 
I found URL `https://www.bandisoft.com/bandizip/dl.php?portable` and it works fine.

jku1995:
I encountered the same problem.​:sweat_smile:​

wenmin92:
Maybe we should use `https://dl.bandisoft.com/bandizip.std/BANDIZIP-PORTABLE.ZIP` instead.
Note: we also have to change the `shortcuts` and `bin`.

Ash258:
@wenmin92 The manifest has to be reworked. I will do it as soon as I have time.

PR #3864, opened 2020-04-15
kitty: Persist SshHostKeys, Jumplist

### LINKED ISSUE #3734: Kitty app should persist more folders
Besides of Sessions, SshHostKeys and Jumplist folder should also be persisted, or else upgrading will cause all accepted host keys lost, and taskbar jumplist cleared.

mark05e:
- Closes #3734

Add more persist folders - SshHostKeys, Jumplist

### DIFF
--- a/bucket/kitty.json
+++ b/bucket/kitty.json
@@ -39,9 +39,9 @@
         ]
     ],
     "persist": [
+        "Jumplist",
         "Sessions",
         "SshHostKeys",
-        "Jumplist",
         "kitty.ini"
     ],
     "checkver": {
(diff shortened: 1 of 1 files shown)


mark05e:
Is there anything more I can add do for this?

kraml:
During my daily use I only have the SshHostKeys, Jumplist and Sessions folders created by Kitty.

But according to [Kitty's doc](http://www.9bis.net/kitty/#!pages/Portability.md) there are more:
* Commands
* Folders
* Launcher
* Sessions
* Sessions_Commands
* SshHostKeys

mark05e:
seems like the docs missed out on `Jumplist` folder,
@Ash258 - what do you propose to do here?

PR #3867, opened 2020-04-16
nathanchance_wsl2-linux-kernel: Add version wsl2-cbl-kernel-next-20200416-v8

koumaza:
WSL2 Kernel


Ash258:
There is no reason why some random fork of something, which is not application should be in any of official buckets.

koumaza:
OK.
Thank you very much.

PR #3900, opened 2020-04-20
Add pasteboard (pbcopy/pbpaste)

uzxmx:
This package provides two executables: `pbcopy.exe` and `pbpaste.exe` whose purpuses are the same as OSX's pbcopy/pbpaste.

Though `clip.exe` shipped with Windows works well, it seems Windows missed an utility to get contents from clipboard using CLI. One workaround is to use `powershell.exe -command Get-Clipboard`, but that's too slow on WSL. Another drawback is it cannot supporting converting Windows line break `\r\n` to UNIX line break `\n`.

The `pbpaste.exe` provided by this package runs faster than `powershell.exe -command Get-Clipboard`, and supports line break conversion.

So this package can provide some values to others.

### DIFF
--- a/bucket/pasteboard.json
+++ b/bucket/pasteboard.json
@@ -1,0 +1,19 @@
+{
+    "version": "1.2.0",
+    "homepage": "https://github.com/uzxmx/pasteboard",
+    "description": "Clipboard utilities for Windows (like OSX pbcopy/pbpaste)",
+    "license": "MIT",
+    "url": "https://github.com/uzxmx/pasteboard/releases/download/v1.2.0/pasteboard-v1.2.0-windows.zip",
+    "hash": "2290667c7740a666c2d0341291538500285d737e98aa6c7c66b1217a99ebca9e",
+    "bin": [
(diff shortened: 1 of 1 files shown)


uzxmx:
@Ash258  I've made a PR to the original author. There are two purposes for that PR:

* Add --lf option to pbpaste.exe, so we can use it correctly on WSL.
* The package was uploaded to chocolatey. For me, it's easier to use scoop than chocolatey. I only need to build a release and define an app.json, and scoop can help me do the rest of stuff. So I also added Travis CI to build a release whenever a tag is made.

Since that original repo has been inactive for several years, so I suspect the author may not respond in time.

If you think this package is welcome to this bucket, and when the original author accepts that PR, I'll suggest using scoop to do automatic release to that author. But no matter whether that author accepts the suggestion, I like the scoop way, and can maintain that scoop support.

rasa:
@uzxmx Great find, and sorry for the delay!

uzxmx:
@rasa  Thanks for the approval.

PR #3901, opened 2020-04-20
Updated startskript

StarPicard:
Solves a problem when a user has a userfolder with a space in the name (like "John Doe")


pratikpc:
It looks like the languagetool pre-install has been updated and there is a conflict.

https://github.com/ScoopInstaller/Extras/blob/master/bucket/languagetool.json#L15

If the error still remains, do open a new PR.

PR #3912, opened 2020-04-22
wezterm: Add version 20200505-090057-31c6155f

nightscape:
A GPU-accelerated cross-platform terminal emulator and multiplexer written by @wez and implemented in Rust

### DIFF
--- a/bucket/wezterm.json
+++ b/bucket/wezterm.json
@@ -1,15 +1,15 @@
 {
-    "version": "20200406-151651-5b700e4",
+    "version": "20200505-090057-31c6155f",
     "description": "GPU-accelerated terminal emulator",
     "homepage": "https://github.com/wez/wezterm",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/wez/wezterm/releases/download/20200406-151651-5b700e4/WezTerm-windows-20200406-151651-5b700e4.zip",
-            "hash": "5f1b0898c54b0dd69a56c9ee2c2f3d2c4c39f29e2b80acb3320515b833a9f687"
+            "url": "https://github.com/wez/wezterm/releases/download/20200505-090057-31c6155f/WezTerm-windows-20200505-090057-31c6155f.zip",
+            "hash": "accb1b66b07395c03cd5d2ba331e0fc8c97f8df372828d5646cb764615e4cebf"
         }
     },
-    "extract_dir": "WezTerm-windows-20200406-151651-5b700e4",
+    "extract_dir": "WezTerm-windows-20200505-090057-31c6155f",
     "bin": "wezterm.exe",
     "shortcuts": [
         [
--- a/bucket/wezterm.json
+++ b/bucket/wezterm.json
@@ -18,12 +18,15 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/wez/wezterm/releases",
-        "regex": "WezTerm-windows-([\\d]+)-([\\d.]+)-([\\d\\w]+)",
-        "replace": "${1}-${2}-${3}"
+        "github": "https://github.com/wez/wezterm",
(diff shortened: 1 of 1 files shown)


bucket/wezterm.json
```diff
@@ -0,0 +1,25 @@
...
+    "shortcuts": [
+        [
+            "wezterm.exe",
+            "Wezterm"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/wez/wezterm/releases",
+        "regex": "WezTerm-windows-([\\d]+)-([\\d.]+)-([\\d\\w]+)",
+        "replace": "${1}-${2}-${3}"
```
Ash258, line 19:
```suggestion
        "github": "https://github.com/wez/wezterm",
        "regex": "WezTerm-windows-([\\w-]+)"
```

bucket/wezterm.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "autoupdate": {
+        "url": "https://github.com/wez/wezterm/releases/download/$version/WezTerm-windows-$version.zip",
```
Ash258, line 22:
```suggestion
        "architecture": {
            "64bit": {
                "url": "https://github.com/wez/wezterm/releases/download/$version/WezTerm-windows-$version.zip"
            }
        },
```

bucket/wezterm.json
```diff
@@ -0,0 +1,25 @@
+{
+    "version": "20200406-151651-5b700e4",
+    "description": "GPU-accelerated terminal emulator",
+    "homepage": "https://github.com/wez/wezterm",
+    "license": "Apache-2.0",
+    "url": "https://github.com/wez/wezterm/releases/download/20200406-151651-5b700e4/WezTerm-windows-20200406-151651-5b700e4.zip",
+    "hash": "5f1b0898c54b0dd69a56c9ee2c2f3d2c4c39f29e2b80acb3320515b833a9f687",
```
Ash258, line 7:
```suggestion
    "architecture": {
        "64bit": {
            "url": "https://github.com/wez/wezterm/releases/download/20200406-151651-5b700e4/WezTerm-windows-20200406-151651-5b700e4.zip",
            "hash": "5f1b0898c54b0dd69a56c9ee2c2f3d2c4c39f29e2b80acb3320515b833a9f687"
        }
    },
```

nightscape:
@Ash258 applied your suggestions, thanks!
It might be best to squash-merge now.

bucket/wezterm.json
```diff
@@ -0,0 +1,32 @@
...
+    "bin": "wezterm.exe",
+    "shortcuts": [
+        [
+            "wezterm.exe",
+            "Wezterm"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/wez/wezterm",
+        "regex": "WezTerm-windows-([\\w-]+)"
```
wez, line 22: I think this was better before; the version format is described here:
https://wezfurlong.org/wezterm/config/lua.html#weztermversion
There are only three components to the version string.

It is important because I have plans to add eg: `.msi` or `.exe` windows downloads in the future and this regex will be broken by that.

nightscape: @wez the man himself, cool 😄 
Would the correspondence between download file name and version still be the same?
The new regex is more lenient, so it should be more flexible in terms of changes to the version format,
as long as the URL to the download can still be string-interpolated in the same way.

I don't mind either way though :wink:

wez: I'm not familiar with this configuration in scoop, but my assumption is that the capture defined in this regex is used to populate the `$version` token in the `autoupdate` section below.

If that is the case then it feels best to constrain the regex to match the version triple to avoid problems if/when I start to add files named something like (for a slightly but not overly contrived example): `WezTerm-windows-DATE-TIME-HASH-debuginfo.zip` from falsely matching and being used as a candidate update!

PR #3922, opened 2020-04-24
Update scripts for Notepad3.

xx256xx:
1. The previous scripts are designed to add "open with notepad" into the context menu, not to replace Windows Notepad.
2. I rewrite the scripts for replacing/restoring Windows Notepad based on [this post](https://www.winhelponline.com/blog/replace-notepad-text-editor-notepad-plus-association/). These two scripts require the administrator privilege.
3. I rename the original scripts to "add-into-context-menu.reg" and "remove-from-context-menu.reg".


xx256xx:
Update the scripts to add/remove Notepad3 into/from the context menu without requiring the administrator privilege. Add an icon for this context menu entry.

PR #3970, opened 2020-05-03
ccleaner: Persist branding.dll, CCleaner.dat

vuwjzl17985:
Close https://github.com/lukesampson/scoop-extras/issues/3951.
Persist branding.dll and CCleaner.dat where the registration information is stored.


Ash258:
You did not persist anything. The files are incorrectly created since the branding.dll file has to be valid, not just empty.

Always start with clean branch without junk commits not related to this change.

PR #3971, opened 2020-05-03
chromium: Add chromimum shims

Neiromaster:
The regular version of Chromium already has the correct name

### DIFF
--- a/bucket/chromium-dev-nosync.json
+++ b/bucket/chromium-dev-nosync.json
@@ -8,7 +8,12 @@
         "jsonpath": "$.chromium.windows.download",
         "regex": "v([\\d.]+-r\\d+)-Win64"
     },
-    "bin": "chrome.exe",
+    "bin": [
+        [
+            "chrome.exe",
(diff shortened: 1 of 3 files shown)


Neiromaster:
And chrome.exe will not grind each other when installing multiple versions of the application?

r15ch13:
`chrome.exe` will get overwritten by installing/updating another version.
But it should still be added because a user expects to have it after installing just one version of chromium.

Neiromaster:
But if I want to install both **Google chrome**  and **Chromium** , then I will not be able to run them together from the terminal

Ash258:
Why not??

1. Install chromium -> chrome.exe or chromimum.exe
1. Install chrome -> chrome.exe will replace chrome.exe from chromium

Reset in case of advanced usages

Neiromaster:
Good.  If I put Chromium, and then Chrome, then everything will be fine.  But if I run "scoop update *", I won’t know what will start with chrome.exe.  A kind of Chrome Schrödinger

PR #3975, opened 2020-05-04
handbrake: Use portable configuration

### LINKED ISSUE #3919: Handbrake: Use portable configuration
Properly migrate from non portable configuration and configure portable mode:

- `portable.ini.template` -> `portable.ini`
- persist:
	- `storage`
	- `tmp`

### LINKED ISSUE #3973: handbrake@1.3.2: download via aria2 failed
```
ERROR Download failed! (Error 3) Resource was not found
ERROR https://download.handbrake.fr/handbrake/releases/1.3.2/HandBrake-1.3.2-x86_64-Win_GUI.zip
    referer=https://download.handbrake.fr/handbrake/releases/1.3.2/
    dir=D:\Users\tim\Scoop\cache
    out=handbrake#1.3.2#https_download.handbrake.fr_handbrake_releases_1.3.2_HandBrake-1.3.2-x86_64-Win_GUI.zip

ERROR & 'D:\Users\tim\Scoop\apps\aria2\current\aria2c.exe' --input-file='D:\Users\tim\Scoop\cache\handbrake.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/5.1 (Windows NT 10.0; Win64; x64; Desktop)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn --enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=15272 --continue
```

Dabombber:
This closes #3973, closes #3919

Prior to 1.1.1 the GUI version only came in exe.

Sometimes the data in the zip is in a Handbrake folder other times it's Handbrake $version, and with this release it's at the top level. It's a bit of a mess, maybe there's a more elegant way of handling it.

portable.ini might contain user settings hardware.enabled for example, and it looks like there's more planned, so that's persisted.

Not sure if I've gotten data migration right. ATM it moves data, but that might break separate non-portable installs so perhaps it should be copied instead. Temp data should only stick around if handbrake crashes, so it shouldn't need migration.

### DIFF
--- a/bucket/handbrake.json
+++ b/bucket/handbrake.json
@@ -9,14 +9,9 @@
             "hash": "f2adb660ff24aa62677aa5a67421e76e0c588f9a5567f74819384a351574d2e7"
         }
     },
+    "extract_dir": "HandBrake",
     "installer": {
         "script": [
-            "@(\"HandBrake\", \"HandBrake $version\") | %{",
-            "    if (Test-Path \"$dir\\$_\") {",
-            "        Get-ChildItem -Path \"$dir\\$_\" -Force -Recurse | Move-Item -Destination \"$dir\" -Force",
-            "        Remove-Item \"$dir\\$_\" -Force",
-            "    }",
-            "}",
             "if (!(Test-Path \"$persist_dir\\portable.ini\")) {",
             "    if (!(Test-Path \"$persist_dir\\storage\") -And (Test-Path \"$env:APPDATA\\HandBrake\")) {",
             "        New-Item \"$persist_dir\" -ItemType Directory -ErrorAction SilentlyContinue | Out-Null",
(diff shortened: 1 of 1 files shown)


Ash258:
> Prior to 1.1.1 the GUI version only came in exe.

Where did you get this information? How should I understand this?

The application works without any issue from zip.

Dabombber:
> Where did you get this information? How should I understand this?
> 
> The application works without any issue from zip.

The [github releases](https://github.com/HandBrake/HandBrake/releases), or the [previous versions](https://handbrake.fr/old.php) page. It means that `scoop install handbrake@1.1.0` and earlier wont work, which isn't really a problem.

PR #3984, opened 2020-05-04
handbrake: use extract_dir

TheZoker:


r15ch13:
`extract_dir` is a better option for that. :smile:

PR #4000, opened 2020-05-06
Update foxit-reader.json

y1zhou:
Original `url` returned a 404 error as there's something wrong with the versioning in it.


Ash258:
#3998

PR #4011, opened 2020-05-08
kotatogram: Add version 1.3.4

batkiz:


Ash258:
Forks of already appliations are not subject of addition to official bucket.

batkiz:
In fact, it's **a 3-rd party client** of telegram, which contains much of [enhanced features](https://github.com/kotatogram/kotatogram-desktop#features). Just it's introduced as "telegram fork" by the author. The relationship between Kotatogram and telegram is just like chrome and chromium, we could add it to this bucket.

PR #4027, opened 2020-05-10
vscode-insiders: Add version info

Neiromaster:


Ash258:
Not all insiders versions come with both architectures and unless you have commit ID in version property you will not get updates.

Have nightly version with force updates is much more convenient.

PR #4042, opened 2020-05-11
New 0.73.2.14 KiTTY version

cyd01:
- new version 0.73.2.14
- new default kitty.ini configuration file (with more documentation)


pratikpc:
It's already updated to v0.74

And there are conflicts.

It would be better if you could create a new PR instead

pratikpc:
@cyd01 BTW is Kitty's homepage correct?

Is it kitty.9bis.net or www.9bis.com/kitty?

cyd01:
It's https://www.9bis.net/kitty

PR #4089, opened 2020-05-18
attempt to fix axiom.json

camjac251:
I haven't tested this and it could possibly not work.

The current manifest seems broken since it needed manual updates several times so hopefully this can fix it.
I borrowed from https://github.com/lukesampson/scoop-extras/blob/master/bucket/tremulous.json for the checkver variable. It might have additional items that aren't needed in the regex.

If tags are used:
https://github.com/GrangerHub/tremulous/tags
https://github.com/MattMcManis/Axiom/tags

### DIFF
--- a/bucket/axiom.json
+++ b/bucket/axiom.json
@@ -18,7 +18,10 @@
             "Axiom"
         ]
     ],
-    "checkver": "/releases/download/v([\\w.-]+)/Axiom\\.zip",
+    "checkver": {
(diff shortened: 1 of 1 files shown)


Ash258:
Nothing needed manual updates. I only rushed my normal pipeline, which suppress Excavator to provide closing directive.

Everything works fine without changes....

image

camjac251:
Ah ok. I'll be honest and say I don't know what you mean. Will that mean that updates could go out sooner in the future?

PR #4141, opened 2020-05-26
add vscode-github integration support

vuwjzl17985:
1. Move existing vscode registry files into a subfolder
2. When one installs the vscode, copy reg files from **bucket** rather than download from web
3. Add registry for vscode-github integration support


vuwjzl17985:
Thank you for your feedback. I think it would be better if the changes are split into two pull requests: one for making reg files copied, the other for adding github integration.

PR #4144, opened 2020-05-27
DeepL: Fix pre_install error

StarsbySea:
A small fix to fix the installation error.

### DIFF
--- a/bucket/deepl.json
+++ b/bucket/deepl.json
@@ -19,7 +19,7 @@
     "pre_install": [
         "$null = New-Item \"$dir\\DeepL\" -Type Directory -Force",
         "$null = New-Item \"$dir\\DeepL\\packages\" -Type Directory -Force",
-        "Move-Item \"$dir\\net45\" \"$dir\\DeepL\\app-$version\"",
+        "Move-Item \"$dir\\net45\\*\" \"$dir\\DeepL\\app-$version\"",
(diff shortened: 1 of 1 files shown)


Ash258:
There is no error

image

StarsbySea:
> 
> 
> There is no error
> 
> image

Thank you for your feedback. But it seems that the original script gets installed incorrectly on my computer, and after modifications it works properly.
My environment is: Microsoft Windows [version 10.0.18363.836]
image
@Ash258

StarsbySea:
I believe the following code explains the problem I'm having.
https://github.com/lukesampson/scoop-extras/blob/d91dcc2d3bdc301f83552ace3be3e85dae78560d/bucket/filezilla.json#L35-L39
The strange thing is that if I run this command in a separate PowerShell terminal it doesn't produce an error. But I think it's better to modify it for compatibility? Is the modified script not working on your computer or is there some other problem?
Thank you for your review.

StarsbySea:
Request for review @Ash258

StarsbySea:
When I quit Kaspersky Internet Security 2020, deepl installed properly. When I turned the Kaspersky back on, it wouldn't install properly again.
Can the developer test this?
Should I give feedback to the Kaspersky team on this?
But why can't they be merged when a simple modification to the script can make it more compatible?

Ash258:
There is no reason to merge this when there is no issue

StarsbySea:
Errors don't happen anymore.

PR #4165, opened 2020-05-29
totalcommander@9.51: Update persist

stasjok:
Total Commander: update persistent files.

* **plugins**: default installation directory for plugins (e.g. if you open plugin archive, totalcmd by default installs it in this directory)
* **VERTICAL.BAR**, **vertical.br2**: vertical (between panels) button bar
* **contplug.ini**, **fsplugin.ini**, **lsplugin.ini**, **pkplugin.ini**: default config files for plugins settings (https://www.ghisler.ch/wiki/index.php/Ini-settings)

Unfortunately it's not the all possible files. Various plugins can use their own config files. For example SFTP plugin from Total Commander developer is using sftpplug.ini, his WebDAV plugin uses tcwebdav.ini and so on. It's the same problem as in foobar2000: #2884. So it could be better to split it to normal and portable version. But this PR should cover most cases.

### DIFF
--- a/bucket/totalcommander.json
+++ b/bucket/totalcommander.json
@@ -33,15 +33,22 @@
         "Move-Item \"$dir\\e\\*\" \"$dir\"",
         "Remove-Item -Recurse \"$dir\\d\", \"$dir\\e\", \"$dir\\_tmp\", \"$dir\\$fname\"",
         "\"[Configuration]`r`nUseIniInProgramDir=7\" | Out-File -Encoding ASCII \"$dir\\wincmd.ini\"",
-        "ForEach ($emptyFile in $manifest.persist) {New-Item \"$dir\\$emptyFile\" -ItemType file -ErrorAction SilentlyContinue | Out-Null}"
+        "ForEach ($emptyFile in $manifest.persist | Where-Object { $_ -ne 'plugins' }) {New-Item \"$dir\\$emptyFile\" -ItemType file -ErrorAction SilentlyContinue | Out-Null}"
     ],
     "persist": [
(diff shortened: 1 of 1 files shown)


Ash258:
Toolbars will not work with precreation. They will be created empty and then total commander will make all toolbars empty instead of defaults.

stasjok:
Are you sure? It works in current version of manifest for DEFAULT.BAR. And it works for VERTICAL.BAR also based on my testing.

Ash258:
In case of clean installation without any persisted data?

I remember one of the file managers is having these issues. But it could be unreal or doublecmd or similar

stasjok:
Yes. I've tested a clean installation. Then I changed some settings, buttons, installed some plugins. Then uninstall, install. It worked for me.

stasjok:
It works because `DEFAULT.BAR` and `VERTICAL.BAR` are already in an installer archive. An install script doesn't create those files (New-Item silently fails). So they are either defaults (from installer) or are copied from persist directory by scoop. `default.br2` and `vertical.br2` are created empty, but total commander overrides them with something after first startup.

If I empty those files manually after installation, then indeed a toolbar would be empty. But it shouldn't happen (files are already there; if not, then toolbar would be empty anyway).

stasjok:
Probably `default.br2` and `vertical.br2` aren't needed at all. As I understand it's just an icon cache. It should be regenerated if not exist. [1](https://ghisler.ch/board/viewtopic.php?p=153291#p153291), [2](https://www.ghisler.ch/wiki/index.php/Files#Additional_standard_files)

PR #4177, opened 2020-05-30
deltachat: Add version 1.3.4

linsui:


bucket/deltachat.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "1.3.4",
+    "description": "Email-based secure instant messaging",
+    "homepage": "https://delta.chat",
```
Ash258, line 4:
```suggestion
    "homepage": "https://delta.chat/en/",
```

image

bucket/deltachat.json
```diff
@@ -0,0 +1,36 @@
...
+            "hash": "9f6f1db789a8ca2a59e2c6c7423d6d128892b4ec038537a4a3cc22f1110d8b20",
+            "installer": {
+                "script": [
+                    "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\"",
+                    "Remove-Item \"$dir\\`$PLUGINSDIR\" -Recurse"
+                ]
+            }
+        }
+    },
+    "bin": "DeltaChat.exe",
```
Ash258, line 18: What are available parameters??

linsui: None. Just for consistence with other manifest.

bucket/deltachat.json
linsui: I guess it is redirected to different url based on the location. Users can see the page in their languages.

bucket/deltachat.json
Ash258: Then remove it.

bucket/deltachat.json
Ash258: Use english

kunago:
I used a slightly different approach. Instead of unpacking the DeltaChat.exe file I left it the way it was, using a portable version. Upon start a folder called "DeltaChatData" is created which stores all data. Then it is easy to make the folder persistent and keep the data separately.

Feel free to see https://bitbucket.org/kunago/scoop-bucket/src/master/bucket/deltachat.json

linsui:
@kunago Thanks! I didn't notice that it's portable.

Ash258:
Its not. After launch it is extracted into appdata/local temp and then executed from it. As all electron "portable" binaries.

Ash258:
https://github.com/lukesampson/scoop-extras/pull/2087#issuecomment-488344372

image

linsui:
@Ash258 Yeah, the drawback is it's slow...

kunago:
So far I found it the best solution but if someone will come up with something that can persist the database, then that would be nicer.

Issue #3370, opened 2019-12-20
archi@4.6.0: hash check failed


ktarou:
```
Scoop was updated successfully!
Installing 'archi' (4.6.0) [64bit]
Downloading https://www.archimatetool.com/downloads/51f64a6b21595fe0b/Archi-Win64-4.6.0.zip (-1 B)...

Checking hash of Archi-Win64-4.6.0.zip ... ERROR Hash check failed!
App:         extras/archi
URL:         https://www.archimatetool.com/downloads/51f64a6b21595fe0b/Archi-Win64-4.6.0.zip
First bytes: 3C 21 64 6F 63 74 79 70
Expected:    78f2f45b32386154bbf28bef512c9e78
Actual:      04a09d4ce884fc4846489d8c8995cba2

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/lukesampson/scoop-extras/issues/new?title=archi%404.6.0%3a+hash+check+failed
```

xycui:
![image](https://user-images.githubusercontent.com/2989428/71363500-56561d00-25d4-11ea-9274-efb22ca1910e.png)

Ash258:
Download URL change each day (look like). The hash is correct but for the page you downloaded. If you open the downloaded file you will see HTML instead of archive.

Best possible solution is to remove archi. There is no way how to deal with it. As soon as developers get back correct release process it could be added.

xycui:
![image](https://user-images.githubusercontent.com/2989428/71363697-ee540680-25d4-11ea-9de0-8c1d497aa6be.png)
@Ash258 I don't think the current url is the correct one. We should not use the url including any changing hash.

Ash258:
As I said. The URL (hahs) is changing (each day). Not on update. There is nothing which could be done.

xycui:
@Ash258 Do you think it is possible for us to use github release url from github repo 
![image](https://user-images.githubusercontent.com/2989428/71363813-4985f900-25d5-11ea-91b1-3b91ad877da7.png)
Url like this https://github.com/archimatetool/archi/archive/release_4.6.0.zip

PR #4190, opened 2020-05-31
archi@4.6.0: download url fix

mark05e:
issue ref: https://github.com/lukesampson/scoop-extras/issues/4056


Ash258:
As I said already. The ID change on regulate basis #3370

mark05e:
Understood. URL might change tomorrow.

PR #4199, opened 2020-06-02
flutter: Update to version 1.17.2

sinloss:
- Closes #4023

Also, change the `autoupdate` strategy to adapt the new versioning manner while keeping the compatibility with former versioning manners.

This is implemented by just make the matching of `delim` and `build` optional and incorporate the character `v` in the `version`.

### DIFF
--- a/bucket/flutter.json
+++ b/bucket/flutter.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.12.13-hotfix.9",
+    "version": "1.17.2",
     "description": "Google’s mobile app SDK for crafting high-quality native interfaces on iOS and Android",
     "homepage": "https://flutter.dev",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


ThatNerdyPikachu:
Is there a reason this isn't merged yet?
Can I install this manually until this is merged -- will it conflict when this is eventually merged?

sinloss:
> Is there a reason this isn't merged yet?
> Can I install this manually until this is merged -- will it conflict when this is eventually merged?

You might want to create a **local** git bare repo as a custom bucket, then you can copy this manifest into it.

say we have `C:\foobar` folder, we can:
* `cd C:\foobar`
* `git init --bare`
* `scoop bucket add foobar C:\foobar`
* copy the manifest file directly to the `<your scoop home>\buckets\foobar`
* whether you `commit` and `push` or not, you can now install it from your own bucket.

PR #4208, opened 2020-06-03
libreoffice-fresh: Fixed Shortcuts

medoni:


Ash258 (changes requested):
They changed it already like 3 or 5 times. If it happen in feature I am for deprecating this.

Revert everything you have done and add `    "extract_dir": "LibreOffice",`

PR #4217, opened 2020-06-04
Update blender.json to v2.83.0

Sebkasanzew:


Sebkasanzew:
Does anyone know, what I missed?

shiena:
Other pull requests are also failing in `continuous-integration/appveyor/pr`, so it's probably not a problem.

Celeborn2BeAlive:
Oups, did not see this pull request (did the same in #4218 started at the same time approximately).
However I've also added the change in the update URL.

PR #4219, opened 2020-06-04
Added AusweisApp2

JamesCullum:
Added AusweisApp2, which is software used to identify with a german ID card.

Source code is available at https://github.com/Governikus/AusweisApp2

Autoupdater & semver embedded, installation tested and it runs fine


JamesCullum:
Tested it myself - do we need a 3rd party to verify or can we go without?

PR #4224, opened 2020-06-05
mobaxterm: Add non-portable version & persist Custom.mxtpro

vuwjzl17985:
1. Add a installer version mobaxterm
2. Rename the original mobaxterm as mobaxterm-portable
3. Rename shortcut (remove "Personal")
4. Persist "Custom.mxtpro"


Ash258:
Portable mobaxterm HAS to be without prefix.

I really do not think nonportable applications should be added when there is portable one already

Add it to versions or somewhere else.

Remove all changes you did except the persisting

PR #4237, opened 2020-06-09
android-clt: Add version 6609375

kidonng:
Android Command Line Tools (`android-clt`, the name can be discussed) is the successor of [`android-sdk`](https://github.com/lukesampson/scoop-extras/blob/master/bucket/android-sdk.json).

If this gets merged we probably want to modify [what `android-studio` suggests](https://github.com/lukesampson/scoop-extras/blob/02d4c9f9a0780e6e7fba618ac929bcd5af0b9466/bucket/android-studio.json#L33) too.

### DIFF
--- a/bucket/android-clt.json
+++ b/bucket/android-clt.json
@@ -1,0 +1,39 @@
+{
+  "version": "6514223",
+  "description": "Android command line tools",
+  "homepage": "https://developer.android.com/studio/",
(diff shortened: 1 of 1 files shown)


Ash258:
Move this PR into Main bucket and request review from someone from ScoopInstaller/maintainers

PR #4244, opened 2020-06-10
lazarus: add "env_set" and "bin" sections to the manifest

AntonOks:
making sure the tools are also found on the CLI and in scripts

### DIFF
--- a/bucket/lazarus.json
+++ b/bucket/lazarus.json
@@ -32,6 +32,15 @@
             "--debug"
         ]
     ],
+    "env_set": {
+                    "LazarusDir": "$dir",
+                    "LazDir": "$dir"
+               },
+    "bin": [
+                "lazarus.exe",
+                "lazbuild.exe",
+                "startlazarus.exe"
+            ],
     "checkver": "lazarus-([\\d.]+)-fpc-(?<fpc>[\\d.]+)-win64.exe",
     "autoupdate": {
         "architecture": {


rashil2000:
I don't think this should be done for IDEs. If it's a compiler or library package, then it makes sense.

rdipardo:
> If it's a compiler or library package, then it makes sense.

Just to clarify, the complete Free Pascal Compiler toolchain is bundled with the installer:

```sh-session
$ Get-ChildItem -Name -Recurse $env:userprofile/scoop/apps/lazarus/current/fpc | Out-Host -Paging

3.2.0\bin\i386-win32\ar.exe
3.2.0\bin\i386-win32\as.exe
3.2.0\bin\i386-win32\bin2obj.exe
3.2.0\bin\i386-win32\chmcmd.exe
3.2.0\bin\i386-win32\chmls.exe
3.2.0\bin\i386-win32\cldrparser.exe
3.2.0\bin\i386-win32\cmp.exe
3.2.0\bin\i386-win32\compileserver.exe
3.2.0\bin\i386-win32\cp.exe
3.2.0\bin\i386-win32\cpp.exe
3.2.0\bin\i386-win32\cvsco.tdf
3.2.0\bin\i386-win32\cvsdiff.tdf
3.2.0\bin\i386-win32\cvsup.tdf
3.2.0\bin\i386-win32\data2inc.exe
3.2.0\bin\i386-win32\delp.exe
3.2.0\bin\i386-win32\diff.exe
3.2.0\bin\i386-win32\dlltool.exe
3.2.0\bin\i386-win32\fp.ans
3.2.0\bin\i386-win32\fp.exe
3.2.0\bin\i386-win32\fp.rsj
3.2.0\bin\i386-win32\fp32.ico
3.2.0\bin\i386-win32\fpc.cfg
3.2.0\bin\i386-win32\fpc.exe
3.2.0\bin\i386-win32\fpcjres.exe
[ . . . ]
```

On one hand, this PR only provides CLI access to the IDE launcher and `lazbuild.exe`; on the other hand, `lazbuild` is directly equivalent to `MsBuild` in functionality. Just like Visual Studio comes with a Developer Command Prompt, a user may want/need to build a project at the command line, for example in CI/CD workflows.

rashil2000:
Okay, a bin entry for `lazbuild.exe` makes sense.

What is the difference between `LazDir` and `LazarusDir` env vars? Why are they set to same directory?

rdipardo:
> What is the difference between `LazDir` and `LazarusDir` env vars? Why are they set to same directory?

`$LazarusDir` is the absolute path to the compiler installation; the IDE is configured based on it:

lazarusDir

Granted, the IDE currently works *without* defining a system-wide variable (I think the installer writes it into a config),
but someone with multiple compilers could switch between them by pointing `$LazarusDir` at a different path, same as you configure `$JAVA_HOME` or `$ANDROID_SDK_ROOT`.

The IDE uses `$LazDir` to find user settings and cached data; by default it resolves to `%LocalAppData%\lazarus`:

lazDir-resolved

There's less of a good reason to expose this globally, unless you're switching user profiles. But there are examples in the Free Pascal community of using `$LazDir` and `$LazarusDir` interchangeably: <https://forum.lazarus.freepascal.org/index.php/topic,46436.msg330934.html#msg330934>

PR #4259, opened 2020-06-10
sdl2: Update to version 2.0.12

Naomi010Sentzke:
I dont know whats the point of downloading the source code instead of the actual runtime libraries and why theres no auto update in this manifest. Who made this ?
Ideally, this manifest should use regsvr32 to register the dlls, but that probably wont be allowed in the extras bucket.


tresf:
> I dont know whats the point of downloading the source code instead of the actual runtime libraries

Because they contained the pkg-config files.

> Who made this ?

Originally, I did.

> Ideally, this manifest should use regsvr32 to register the dlls

Registering it system-wide can break stuff.   Please don't do this. <#

PR #4262, opened 2020-06-11
picpick@5.1.0: hash fix

mark05e:
- Closes https://github.com/lukesampson/scoop-extras/issues/4260


mark05e:
realized that there might be some automated action for this that will make this PR irrelevant.

PR #4265, opened 2020-06-11
windbg: Add version 10.0.19041.0

NotWearingPants:
WinDbg is the official Windows debugger, made by Microsoft.
It is used to debug native Windows programs in both user mode and kernel mode.

To install it, you have to use the Windows SDK installer and only check "Debugging Tools for Windows". The SDK installer is a WiX installer which contains links to the actual installers (and hashes), so I extract the manifests inside and download the WinDbg installer.

The only problem is that because checkver doesn't allow running scripts, we can only get the version of the Windows SDK, which seems to be almost the version of WinDbg itself (SDK is 10.0.19041.0, WinDbg is 10.0.19041.1). The real WinDbg version is written inside the WiX installer, which I print when installing.

### DIFF
--- a/bucket/windbg.json
+++ b/bucket/windbg.json
@@ -1,56 +1,62 @@
 {
     "homepage": "https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/",
     "description": "The official Windows debugger",
-	"version": "10.0.19041.0",
-	"url": "https://go.microsoft.com/fwlink/p/?linkid=2120843#/temp/sdk-setup.exe",
-	"hash": "d53f651370f87484b78622e30dfb1a41920b501e4041035771c0d785561f47d5",
-	"installer": {
-		"script": [
-			"Expand-DarkArchive \"$dir\\sdk-setup.exe\" \"$dir\\temp\"",
-			"$setupSettings = (Select-Xml -Path \"$dir\\temp\\UX\\UserExperienceManifest.xml\" -XPath '/').Node.UserExperienceManifest.Settings",
-			"$version = $setupSettings.ProductName.Split(' ')[-1]",
-			"Write-Output \"Actual version is $version\"",
-			"$downloadRootPointer = $setupSettings.SourceResolution.DownloadRoot",
-			"$downloadRoot = (Invoke-WebRequest -Method Head -Uri $downloadRootPointer -MaximumRedirection 0 -ErrorAction Ignore).BaseResponse.Headers['Location']",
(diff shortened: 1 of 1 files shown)


bucket/windbg.json
```diff
@@ -0,0 +1,62 @@
...
+            "$setupSettings = (Select-Xml -Path \"$dir\\temp\\UX\\UserExperienceManifest.xml\" -XPath '/').Node.UserExperienceManifest.Settings",
+            "$version = $setupSettings.ProductName.Split(' ')[-1]",
+            "Write-Output \"Actual version is $version\"",
+            "$downloadRootPointer = $setupSettings.SourceResolution.DownloadRoot",
+            "$downloadRoot = (Invoke-WebRequest -Method Head -Uri $downloadRootPointer -MaximumRedirection 0 -ErrorAction Ignore).BaseResponse.Headers['Location']",
+            "$manifestPayloads = (Select-Xml -Path \"$dir\\temp\\UX\\manifest.xml\" -XPath '/').Node.BurnManifest.Payload",
+            "$windbgPayload = $manifestPayloads | Where-Object { $_.FilePath.Contains(\"X$($architecture.Replace('bit', '')) Debuggers And Tools\") }",
+            "$windbgDownloadPath = $windbgPayload.SourcePath",
+            "$windbgHash = $windbgPayload.Hash",
+            "dl_with_cache $app $version ($downloadRoot + $windbgDownloadPath) \"$dir\\temp\\windbg-setup.msi\"",
```
Ash258, line 19: Never do this in official buckets. Downloading some random files without hash in bunch of lines of code is dangerous.

If this will not be fixed it will be closed.

NotWearingPants: How normally?
There is no way to get the download url with autoupdate, because the actual url is inside another WiX installer as I've explained.
How is that different than using checkver+autoupdate to get the url from a webpage?

What's your suggestion? I even used `dl_with_cache` to make it have an indicative progress bar.

Ash258: Using `url` property.

Then use it in own bucket.

Allowing manifests with such practice will lead into junk and dangerous software to be added into official buckets.

NotWearingPants: @Ash258 

This is an *official* Microsoft product, which has a complex installation process.

This is the exact opposite of junk.

The problem here is that scoop can't handle this installer indirection.

If git had a complex installer which downloads another installer, would you also not include it in the official buckets?

Please reconsider. This is a very useful app for many people, created *by Microsoft* especially *for Windows*.
scoop is supposed to be a Windows app installer, it's like not having .NET in scoop.

PR #4268, opened 2020-06-12
vuplayer: add version 4.9

linsui:
Portable mode can only read settings, but won't save settings to file. So I didn't make it portable.

### DIFF
--- a/bucket/vuplayer.json
+++ b/bucket/vuplayer.json
@@ -1,0 +1,36 @@
+{
+    "version": "4.9",
+    "description": "Audio player",
+    "homepage": "http://www.vuplayer.com",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


Ash258:
> but won't save settings to file.

What does this mean??

- https://github.com/jfchapman/VUPlayer/blob/66789122b0f760c8e69c722e940e834b0fa7b76b/VUPlayer.cpp#L50
- https://github.com/jfchapman/VUPlayer/blob/66789122b0f760c8e69c722e940e834b0fa7b76b/VUPlayer.cpp#L1774-L1792
- https://github.com/jfchapman/VUPlayer/blob/66789122b0f760c8e69c722e940e834b0fa7b76b/Settings.cpp#L39-L45

linsui:
When I use the portable option, it will read settings from an ini file. But after closing it, new settings won't be write back to the file. It seems that users need to export settings manully. In nonportable mode it works well.

PR #4181, opened 2020-05-30
powertoys: add note about context menu entries

JanPokorny:
Closes #4937

PR #3257, opened 2019-11-25
Powertoys: added .reg files for PowerRename context menu item

ianharmon:
the .reg files' URLs will need to be updated before this PR gets merged

PR #4287, opened 2020-06-15
add possibility to add powerrename context

janicmikes:
PowerToys adds a tool that can be used from the context menu of the explorer.

Similar to vscode manifest this PR will create an additional install-context.reg and uninstall-context.reg to allow the installation of such context entries.

Depending on --global setting the reg file will write to local_machine or local_user values to match the install mode.


janicmikes:
This would replace  #4181 with a solution like it was proposed in #3257 

This would still allow the manifest to be fully portable with the ability to add some stuff to the registry if (needed/wanted) by the installer. But without the need to actually run the msi installer.

janicmikes:
Let me add the entries for the markdown and svg preview handlers. Then this manifest will be complete.

PR #4288, opened 2020-06-15
tegrarcmsmash: Add version 1.2.1-3

TheLastZombie:
Adds TegraRcmSmash to the bucket. It's not a GUI application, but it's not really a ["well-known and widely used developer tool"](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) either.

Description taken from the [official website](https://switchtools.sshnuke.net/), license via [GitHub](https://github.com/rajkosto/TegraRcmSmash/blob/master/LICENSE).

### DIFF
--- a/bucket/tegrarcmsmash.json
+++ b/bucket/tegrarcmsmash.json
@@ -1,0 +1,22 @@
+{
+    "version": "1.2.1-3",
+    "description": "A reimplementation of fusee-launcher in C++ for Windows platforms.",
+    "homepage": "https://switchtools.sshnuke.net/",
+    "license": "GPL-3.0-or-later",
+    "url": "https://files.sshnuke.net/TegraRcmSmash1213.zip",
+    "hash": "3a850b8f7b75bdda00f52e51edd0f43c234fbe532b9c328efbb1e17a2111828f",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


Ash258:
I do not think exploits should be in official buckets. Like why would you want it here? What will users get?

TheLastZombie:
To be fair, I should have specified that while this is based on and uses an exploit, it is rather an end-user application with the purpose of sending payloads to Nintendo Switch devices. This could be compared to, say, booting your Android device to a custom ROM or plugging a bootable USB drive into your PC.

That being said:

> why would you want it here?

For the same reason all other programs are listed here, quick installation/uninstallation without having to worry about unzipping, going to the webpage to find the latest download link or having to add its directory to the %PATH%.

> What will users get?

Just that, a command-line program to send payloads to their Switch devices. Presumably what they want when they type `scoop install tegrarcmsmash`.

> Opening gates for such tools will lead into associating scoop with malwarous software.

I do get where you're coming from here though. I can't say that I manage a project with such a large userbase, so you probably know better than me what's best.

Ash258:
So to sum it up. 

Easier, faster and in general better download of virus thanks to Scoop 👍. 

Definitely not gonna happen in official bucket.

PR #4305, opened 2020-06-18
hokus: Add version 0.6.1

niheaven:
https://www.hokuscms.com

https://github.com/julianoappelklein/hokus

### DIFF
--- a/bucket/hokus.json
+++ b/bucket/hokus.json
@@ -22,8 +22,7 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/julianoappelklein/hokus/releases",
-        "regex": "tree/([\\d.]+)"
+        "github": "https://github.com/julianoappelklein/hokus"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/hokus.json
```diff
@@ -0,0 +1,35 @@
...
+    },
+    "shortcuts": [
+        [
+            "Hokus CMS.exe",
+            "Hokus CMS"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/julianoappelklein/hokus/releases",
+        "regex": "tree/([\\d.]+)"
```
Ash258, line 26: You can go with classic github

```suggestion
        "github": "https://github.com/julianoappelklein/hokus"
```

niheaven: There isn't stable (or Latest) release now, so `github` doesn't work.

Ash258: It will be redirected.

PR #4309, opened 2020-06-18
Add Toggl, Iridium to extras bucket

dtger:
Verified working using custom git repo, feel these apps will be more at home in extras bucket though.


Ash258:
Open 1 PR per manifest/application

dtger:
ah apologies, incoming shortly

PR #4315, opened 2020-06-19
vc14-19 update 14.26.28720.3 to 14.27.28914.0

tymes:
vc14-19 update 14.26.28720.3 to 14.27.28914.0


tymes:
I've just realized the SHA256 hash for this Microsoft download is in the URL... is there anyway we can take advantage of this?

PR #4319, opened 2020-06-20
ibmcloud-cli: Move to main bucket

### LINKED ISSUE #4318: Move ibmcloud-cli to main bucket

chawyehsu:
close #4318


Ash258:
Why do you move manifests when there is no support for it??

This will only open gates for users to report the manifest was removed...

PR #4335, opened 2020-06-24
defold: Add version 1.2.170

ftk:


ftk:
> 
> 
> https://github.com/defold/defold/tree/dev/editor#requirements
> 
> > Java 11 (installation instructions - Note that we do not yet officially support development with a newer version of Java than 11.
> > Leiningen (install instructions)
> 
> https://github.com/defold/defold/tree/dev/editor#windows

This seems to be instructions for building from source code. The zip archive already contains jdk11.

Changed checkver to https://d.defold.com/stable/ (however zip always has the same filename - issues with scoop cache?)

Also, if there is a way to fake LOCALAPPDATA env var, it could be used for persist

bucket/defold.json
```diff
@@ -21,12 +21,13 @@
...
     "checkver": {
-        "github": "https://github.com/defold/defold"
+        "url": "https://d.defold.com/stable/",
+        "regex": "{\"url\": \"(?<url>[^\"]+)\", *\"name\": \"Windows\"}.*?\"tag\": \"([^\"]+)\""
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/defold/defold/releases/download/$version/beta.editor2.Defold-x86_64-win32.zip"
+                "url": "$matchUrl"
```
Ash258, line 30: Not valid. Use previous version

bucket/defold.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "checkver": {
+        "github": "https://github.com/defold/defold"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/defold/defold/releases/download/$version/editor-alpha.editor2.Defold-x86_64-win32.zip"
+            }
+        }
```
Ash258, line 31: As I said. Just remove it.

```suggestion
```

anderlli0053:
What is up with this now? Should it be added or what? I don't understand..

PR #4345, opened 2020-06-27
add YNote app

xuzikuan12:


Ash258:
Application without English homepage are not accepted in official buckets.

PR #4386, opened 2020-07-06
xenu: Add version 1.3.8

jfastnacht:


bucket/xenu.json
```diff
@@ -0,0 +1,20 @@
...
+            "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Setup.exe\", \"$dir\\`uninst.exe\" -Force -Recurse"
+        ]
+    },
+    "shortcuts": [
+        [
+            "Xenu.exe",
+            "Xenu"
+        ]
+    ]
+}
```
pratikpc, line 20: Would need autoupdate or checkver

issaclin32: I don't think *checkver* is needed in this case.
`(current version: 1.3.8 from September 4th, 2010)`
http://home.snafu.de/tilman/xenulink.html

bucket/xenu.json
```diff
@@ -0,0 +1,20 @@
+{
+    "homepage":  "http://home.snafu.de/tilman/xenulink.html",
+    "description": "Xenu's Link Sleuth checks Web sites for broken links",
+    "version":  "1.3.8",
+    "license":  "Unknown",
+    "url": "http://home.snafu.de/tilman/XENU.ZIP#/dl.7z",
```
rashil2000, line 6: Is the dl.7z part required? It's already zip

bucket/xenu.json
```diff
@@ -0,0 +1,20 @@
+{
+    "homepage":  "http://home.snafu.de/tilman/xenulink.html",
+    "description": "Xenu's Link Sleuth checks Web sites for broken links",
+    "version":  "1.3.8",
+    "license":  "Unknown",
+    "url": "http://home.snafu.de/tilman/XENU.ZIP",
+    "hash": "sha256:319108b72709e32c250bb49bd02739f26d37b8a9d5a9bb3bbfa488d7354e197f",
```
rashil2000, line 7: `sha256:` can also be omitted

PR #4392, opened 2020-07-08
ilspy: Update to version 6.0.0.5836

omjadas:


bucket/ilspy.json
```diff
@@ -14,9 +14,9 @@
     ],
     "checkver": {
         "github": "https://github.com/icsharpcode/ILSpy",
-        "regex": "v[\\d.]+/ILSpy_binaries_([\\d.]+)\\.zip"
+        "regex": "v(?<tag>[\\d.]+)/ILSpy_binaries_(?<version>[\\d.]+)\\.zip"
     },
     "autoupdate": {
-        "url": "https://github.com/icsharpcode/ILSpy/releases/download/v$matchHead/ILSpy_binaries_$version.zip"
+        "url": "https://github.com/icsharpcode/ILSpy/releases/download/v$matchTag/ILSpy_binaries_$version.zip"
```
Ash258, line 20:
```suggestion
        "url": "https://github.com/icsharpcode/ILSpy/releases/download/v$majorVersion.$minorVersion/ILSpy_binaries_$version.zip"
```

bucket/ilspy.json
```diff
@@ -14,9 +14,9 @@
     ],
     "checkver": {
         "github": "https://github.com/icsharpcode/ILSpy",
-        "regex": "v[\\d.]+/ILSpy_binaries_([\\d.]+)\\.zip"
+        "regex": "v(?<tag>[\\d.]+)/ILSpy_binaries_(?<version>[\\d.]+)\\.zip"
```
Ash258, line 17:
```suggestion
        "regex": "v[\\d.]+/ILSpy_binaries_([\\d.]+)\\.zip"
```

bucket/ilspy.json
omjadas: image


The tag is only major.minor when there is no patch version, when there is a patch version is it major.minor.patch.

Ash258: Then it will be changed for next version

PR #4394, opened 2020-07-08
keepassxc@2.6.0: Persist config folder

### LINKED ISSUE #4395: KeePassXC config did not persist upon update to 2.6.0

erogy:
- Closes #4395 

Since version 2.6.0 KeePassXC stores its portable configuration in the 'config' folder. Which makes it very convenient to persist.


Ash258:
You cannot just persist it.

Users which do have old configuration will lost it and start with fresh. You need to do migration of the configuration.

erogy:
> You need to do migration of the configuration.

Done.

PR #4397, opened 2020-07-09
vswhere: Move to main bucket

chawyehsu:
block https://github.com/ScoopInstaller/Main/pull/1147


Ash258:
I already asked you why do you do this when there is no support. And you still go and open other PR.

Implement support for manifest migration and then this could be considered and possibly merged. I am not gonna merge this unless there is support.

https://github.com/lukesampson/scoop-extras/pull/4319#issuecomment-647002117

chawyehsu:
I'm NOT ASK you to merge this **now**, I just create the pull-request for a reference. If you're not happy with this then please close this. 

And P.S. you removed some manifests (without any notices and reviews) even the deprecated feature is not implemented, which **already** created confusions:

```
Desktop ~
$ scoop status
Scoop is up to date.
Updates are available for:
    vscode: 1.46.0 -> 1.46.1
These app manifests have been removed:
    dnspy
Desktop ~
$
```

Ash258:
Every commit, which remove such manifests have description why there were removed.

chawyehsu:
So you had those good reasons, which only left in the commit logs, to **delete** manifests without PR/reviews because you have the power to commit directly? And I became the bad boy after creating some manifest migration PRs for issue references.

And it's funny that they are the same thing - deleting manifests.

Ash258:
Yes, they are the same. But the difference is the reason.

This PR is not needed to be opened or even considered. The manifest is already available and could be installed.

These manifests, which I deleted, are broken. They cannot be installed. They will not be updated. So the user will get the message with `manifest removal`, but do not need to be concerned and do not have to do anything to get updates (in contrast to this PR; he needs to investigate which bucket was the target of migration and then change it manually in `scoop-install.json` file). Installation of them leads to opening issues and forcing me to explain the reason to each "user" and them asking why it is still available when it is broken. I am not gonna waste my and users time with this.

Yes it would be better to have `deprecated` property, which blocks the installation completely (or suggest alternative), but it is not there and I am not gonna wait the next 6 months for it, and wasting my time explaining it to "users".

rasa:
Forgive me if I merged this in error, but it seems we can’t have the app in both buckets. If I misunderstood, my apologies.

PR #4411, opened 2020-07-13
mumble: Fixed hash for version 1.3.2

tynril:
The hash present in the manifest does not match the correct hash of the `mumble-1.3.2.msi` file as verified by GPG.

This PR updates the hash to its correct value for that version.


Ash258:
Do not post hash fix PRs. Open issue instead

PR #4416, opened 2020-07-14
trafficmonitor: Adopt portable configuration

shelmor:
- The license is [no longer MIT](https://github.com/zhongyang219/TrafficMonitor/blob/master/LICENSE)
- Persist configuration between installations (~~and migrate from the old manifest if needed~~)
- Ensure the app runs in portable mode
- Disable update check upon the first launch

### DIFF
--- a/bucket/trafficmonitor.json
+++ b/bucket/trafficmonitor.json
@@ -17,13 +17,11 @@
         }
     },
     "extract_dir": "TrafficMonitor",
-    "installer": {
-        "script": [
-            "if (!(Test-Path \"$persist_dir\\config.ini\")) { Set-Content \"$dir\\config.ini\" @('[general]', 'check_update_when_start = false') -Encoding Ascii }",
-            "if (!(Test-Path \"$persist_dir\\history_traffic.dat\")) { New-Item \"$dir\\history_traffic.dat\" -ItemType File | Out-Null }",
-            "if (!(Test-Path \"$persist_dir\\global_cfg.ini\")) { Set-Content \"$dir\\global_cfg.ini\" @('[config]', 'portable_mode = true') -Encoding Ascii }"
-        ]
-    },
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\config.ini\")) { Set-Content \"$dir\\config.ini\" @('[general]', 'check_update_when_start = false') -Encoding Ascii }",
+        "if (!(Test-Path \"$persist_dir\\history_traffic.dat\")) { New-Item \"$dir\\history_traffic.dat\" | Out-Null }",
+        "if (!(Test-Path \"$persist_dir\\global_cfg.ini\")) { Set-Content \"$dir\\global_cfg.ini\" @('[config]', 'portable_mode = true') -Encoding Ascii }"
+    ],
     "shortcuts": [
         [
             "TrafficMonitor.exe",
--- a/bucket/trafficmonitor.json
+++ b/bucket/trafficmonitor.json
@@ -1,8 +1,9 @@
 {
     "version": "1.78",
-    "description": "TrafficMonitor is a network monitoring suspension window software in Windows platform. It can display the current network speed, CPU and memory usage. It also support the functions of display in the taskbar, change skin and historical traffic statistics.",
+    "description": "Network monitoring suspension window software",
     "homepage": "https://github.com/zhongyang219/TrafficMonitor",
     "license": {
+        "identifier": "Anti 996-License-1.0",
         "url": "https://github.com/zhongyang219/TrafficMonitor/blob/master/LICENSE"
     },
     "architecture": {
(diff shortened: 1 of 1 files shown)


bucket/trafficmonitor.json
```diff
@@ -1,8 +1,10 @@
 {
-    "homepage": "https://github.com/zhongyang219/TrafficMonitor",
-    "description": "TrafficMonitor is a network monitoring suspension window software in Windows platform. It can display the current network speed, CPU and memory usage. It also support the functions of display in the taskbar, change skin and historical traffic statistics.",
-    "license": "MIT",
     "version": "1.78",
+    "description": "TrafficMonitor is a network monitoring suspension window software in Windows platform. It can display the current network speed, CPU and memory usage. It also support the functions of display in the taskbar, change skin and historical traffic statistics.",
+    "homepage": "https://github.com/zhongyang219/TrafficMonitor",
+    "license": {
+        "url": "https://github.com/zhongyang219/TrafficMonitor/blob/master/LICENSE"
```
Ash258, line 6:
```suggestion
        "identifier": "Anti 996-License-1.0",
        "url": "https://github.com/zhongyang219/TrafficMonitor/blob/master/LICENSE"
```

bucket/trafficmonitor.json
```diff
@@ -1,8 +1,10 @@
 {
-    "homepage": "https://github.com/zhongyang219/TrafficMonitor",
-    "description": "TrafficMonitor is a network monitoring suspension window software in Windows platform. It can display the current network speed, CPU and memory usage. It also support the functions of display in the taskbar, change skin and historical traffic statistics.",
-    "license": "MIT",
     "version": "1.78",
+    "description": "TrafficMonitor is a network monitoring suspension window software in Windows platform. It can display the current network speed, CPU and memory usage. It also support the functions of display in the taskbar, change skin and historical traffic statistics.",
```
Ash258, line 3:
```suggestion
    "description": "Network monitoring suspension window software",
```

bucket/trafficmonitor.json
```diff
@@ -14,7 +16,27 @@
...
+        ]
+    },
+    "uninstaller": {
+        "script": [
+            "if (!(Test-Path \"$persist_dir\")) {",
+            "   New-Item \"$persist_dir\" -ItemType Directory | Out-Null",
+            "   Copy-Item \"$dir\\config.ini\" \"$persist_dir\" -ErrorAction SilentlyContinue -Force",
+            "}"
+        ]
+    },
```
Ash258, line 32: Why?

bucket/trafficmonitor.json
```diff
@@ -14,7 +16,27 @@
         }
     },
     "extract_dir": "TrafficMonitor",
-    "bin": "TrafficMonitor.exe",
+    "installer": {
+        "script": [
+            "if (!(Test-Path \"$persist_dir\\config.ini\")) { Set-Content \"$dir\\config.ini\" @('[general]', 'check_update_when_start = false') }",
+            "Set-Content \"$dir\\global_cfg.ini\" @('[config]', 'portable_mode = true')"
```
Ash258, line 22: Always use `-Encoding Ascii` for *-Content cmdlets

Ash258: Why global is not persisted?

Why user cannot change it? If you always configure it statically users changes will be always overriden.

bucket/trafficmonitor.json
```diff
@@ -14,7 +16,27 @@
...
+            "}"
+        ]
+    },
+    "shortcuts": [
+        [
+            "TrafficMonitor.exe",
+            "TrafficMonitor"
+        ]
+    ],
+    "persist": "config.ini",
```
Ash258, line 39: - https://github.com/zhongyang219/TrafficMonitor/blob/167e9940df72f8d9dca8a9eb187c3020ad8a4686/TrafficMonitor/TrafficMonitor.cpp#L555
- https://github.com/zhongyang219/TrafficMonitor/blob/167e9940df72f8d9dca8a9eb187c3020ad8a4686/TrafficMonitor/TrafficMonitor.cpp#L551

bucket/trafficmonitor.json
Ash258: This will never work.

Ash258: Flow of potential update/installation

1. User will execute update (force update)
1. Old manifest will be loaded
	1. No uninstaller.script property defined in old manifest
1. User will loose configuration
	1. You will not copy anything
1. New version installed
1. User will update
1. This will never be executed.

erogy: It will execute once when people who has already installed this app update their manifest since the old one does not persist anything.

bucket/trafficmonitor.json
erogy: The global_cfg.ini file has only one setting and it's what's being set in the manifest. If one doesn't want the portable version why even bother installing it with scoop?

Ash258: Not your decision to make.

erogy:
> There is no migration from appdata

Not sure what would be the appropriate way to do that. The app's default setting is to run in portable mode, so the only time the configuration is even created under appdata is when a user manually changes the setting. And if that's the case I don't know if it needs to be migrated at all.

pratikpc:
Closing as shelmor:trafficmonitor has been deleted.

PR #4417, opened 2020-07-14
mini-sftp-server: Add version 2.25

mark05e:
- Closes https://github.com/lukesampson/scoop-extras/issues/4375

### DIFF
--- a/bucket/mini-sftp-server.json
+++ b/bucket/mini-sftp-server.json
@@ -1,0 +1,42 @@
+{
+    "version": "2.25",
+    "description": "A simple mini sftp server with GUI controls",
+    "homepage": "http://www.coreftp.com/server/",
+    "license": "Freeware",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/mini-sftp-server.json
```diff
@@ -0,0 +1,42 @@
+{
+    "version": "2.25",
+    "description": "A simple mini sftp server with GUI controls",
+    "homepage": "http://www.coreftp.com/server/",
+    "license": "Freeware",
```
Ash258, line 5: Source?

mark05e: > Source?

Are you asking about the source for the license details? If yes - it is mentioned on the same page.

image

If you are asking about the source code - there is none available.

mark05e:
> Where are configuration stored?

As per this [discussion on portablefreeware.com](https://www.portablefreeware.com/forums/viewtopic.php?p=29438#p29438), it seems to be writing in the registry. 😓

PR #4426, opened 2020-07-17
android-sdk: Persist system-images

amantoor:
AVD images persists while upgrading.


Ash258:
Source/Documentation?

Where did you get the information about this?

amantoor:
I am user of this package and when I create a new AVD it download images to system-images folder which should persist over updates. There are one more issues like it create .android folder in user home and have AVD placed in that folder which should also stay in persistent folders.

PR #4430, opened 2020-07-19
Update openvpn client to latest 3.2.0 version

OranShuster:


Ash258:
There is no new version available

image

OranShuster:
There is the community version and the official client. they are both free but the community version is open sourced.

Ash258:
THen it cannot be added here when they are completely different clients maintained by different teams.

PR #4444, opened 2020-07-24
gimp: Add gimp-console bin

tresf:
* Fixes install error `twain,2.exe` renamed to `twain,3.exe`, fixes: (**Edit:** fixed prior via #3814)
    ```diff
    Extracting gimp-2.10.20-setup-1.exe ... done.
    Running installer script..
    - Remove-Item : Cannot find path 'C:\Users\owner\scoop\apps\gimp\2.10.20-1\lib\gimp\2.0\plug-ins\twain\twain,2.exe' because it does not exist.
    - At line:5 char:4
    -    Remove-Item 'lib\gimp\2.0\plug-ins\twain\twain,2.exe'
    -    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    - CategoryInfo          : ObjectNotFound: (C:\Users\owner\...ain\twain,2.exe:String) [Remove-Item], ItemNotFoundException
    - FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.RemoveItemCommand
    ```

* Add command-line tool to match Windows:`gimp-console` (was `gimp`).
* ~~Adds `gimp` as command to start GUI~~<br>Removed, see https://github.com/lukesampson/scoop-extras/pull/4444#pullrequestreview-454620032

### DIFF
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -42,7 +42,7 @@
         [
             "bin\\gimp-console-2.10.exe",
             "gimp-console",
-            "gimp",
+            "gimp"
         ],
         "bin\\gimptool-2.0.exe",
         [
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -41,17 +41,13 @@
         "bin\\gimp-console-2.10.exe",
         [
             "bin\\gimp-console-2.10.exe",
-            "gimp-console"
+            "gimp-console",
+            "gimp",
         ],
         "bin\\gimptool-2.0.exe",
         [
             "bin\\gimptool-2.0.exe",
             "gimptool"
-        ],
-        "bin\\gimp-2.10.exe",
-        [
-            "bin\\gimp-2.10.exe",
-            "gimp"
         ]
     ],
     "shortcuts": [
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -12,7 +12,6 @@
             "Get-ChildItem -Filter '*.debug' -Recurse | Remove-Item -Recurse",
             "if ($architecture -eq '64bit') {",
             "   Rename-Item 'lib\\gimp\\2.0\\plug-ins\\twain\\twain,1.exe' 'twain.exe'",
-            "   Remove-Item 'lib\\gimp\\2.0\\plug-ins\\twain\\twain,2.exe'",
             "   Get-ChildItem -Filter '*,1*' -Recurse | Remove-Item",
             "   Get-ChildItem -Filter '*,*' -Recurse | Rename-Item -NewName { $_.name -Replace ',\\d','' }",
             "   New-Item '32\\etc', '32\\share', 'share\\gimp\\2.0\\fonts' -ItemType 'Directory' | Out-Null",
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -47,12 +47,17 @@
         [
             "bin\\gimptool-2.0.exe",
             "gimptool"
+        ],
+        "bin\\gimp-2.10.exe",
+        [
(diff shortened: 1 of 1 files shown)


bucket/gimp.json
```diff
@@ -41,12 +41,17 @@
         "bin\\gimp-console-2.10.exe",
         [
             "bin\\gimp-console-2.10.exe",
-            "gimp"
+            "gimp-console"
         ],
```
Ash258, line 45:
```suggestion
        [
            "bin\\gimp-console-2.10.exe",
            "gimp"
        ],
        [
            "bin\\gimp-console-2.10.exe",
            "gimp-console"
        ],
```

bucket/gimp.json
```diff
@@ -41,12 +41,17 @@
...
         "bin\\gimptool-2.0.exe",
         [
             "bin\\gimptool-2.0.exe",
             "gimptool"
+        ],
+        "bin\\gimp-2.10.exe",
+        [
+            "bin\\gimp-2.10.exe",
+            "gimp"
```
Ash258, line 54:
```suggestion
```

tresf:
> Not a single reason to have GUI applications shim, when there is command line executable

Well the reason was that the `gimp-console` does not launch a GUI, rendering the GUI unusable.  According to the repo rules, `scoop-extras` is the place for GUI apps (assuming what's NOT allowed in min goes here, per [Criteria-for-including-apps-in-the-main-bucket](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket))

Ash258:
There is no reason to have the GUI shim, because there is no reason to spawn the GUI from command line.

Does the gui executable have some parameter handler?

tresf:
> there is no reason to spawn the GUI from command line.

Is that a strict policy?  I don't disagree from a programming perspective, but I'm not sure how one would know that.

> Does the gui executable have some parameter handler?

Unfortunately, no.  It's not a dual-handler like `blender.exe`, it just fires the GUI.  To your point, the command-line should be able to as well.  There's splash handers, new instance handlers, etc.  I'm just not sure why it deadlocks.  I'd happily revert it.  It was added for my own testing purposes and the CLI `gimp` is slightly more consistent with POSIX scripts (albeit windows scripts would naturally have `-console` in the name, diverging a bit.

I know tools like Chocolately permit UI stubs.  This is my first experience with a UI stub with scoop.

bucket/gimp.json
```diff
@@ -41,6 +41,7 @@
         "bin\\gimp-console-2.10.exe",
         [
             "bin\\gimp-console-2.10.exe",
+            "gimp-console",
             "gimp"
         ],
```
Ash258, line 46: This was not requested

```suggestion
            "gimp"
        ],
        [
            "bin\\gimp-console-2.10.exe",
            "gimp-console"
        ],
```

bucket/gimp.json
```diff
@@ -43,6 +43,11 @@
             "bin\\gimp-console-2.10.exe",
             "gimp"
         ],
+        "bin\\gimp-console-2.10.exe",
```
Ash258, line 46: Already there... L41

```suggestion
```

bucket/gimp.json
tresf: I'm still curious about scoop's policy on invocation of the GUI stub.  Personally, I launch UIs from command line fairly often.  e.g. `notepad myfile.txt`.  I would still strongly recommend that `gimp` launches the UI and `gimp-console` launches the console, which is how the standard install behaves.  Marking as resolved, because that's a policy that scoop needs to make and I have no investment in the decision.

PR #4451, opened 2020-07-25
regester: Add version 2.0.1

ssfjhh:
Regester is a regular expression test tool.


ssfjhh:
> Application without English homepage are not accepted in official buckets

The language of this software can be set to English. Is it not accepted still? it's a so simple software that I believe this sentence "Regester is a regular expression test tool." is enough to describe its function.

PR #4470, opened 2020-07-28
element@1.7.2: Fix bin and shortcuts

erogy:
https://github.com/vector-im/riot-desktop/commit/94086f247a8c89630ddf3c134e4af50c18d1af3b

### DIFF
--- a/bucket/element.json
+++ b/bucket/element.json
@@ -12,13 +12,13 @@
     "extract_dir": "lib\\net45",
     "bin": [
         [
-            "element (riot).exe",
+            "element.exe",
             "element-desktop"
         ]
     ],
     "shortcuts": [
         [
-            "element (riot).exe",
+            "element.exe",
             "Element"
         ]
     ],
--- a/bucket/element.json
+++ b/bucket/element.json
@@ -11,14 +11,15 @@
     },
     "extract_dir": "lib\\net45",
     "bin": [
+        "Element.exe",
         [
-            "element.exe",
+            "Element.exe",
             "element-desktop"
         ]
     ],
     "shortcuts": [
         [
-            "element.exe",
+            "Element.exe",
             "Element"
         ]
     ],


bucket/element.json
```diff
@@ -12,13 +12,13 @@
...
         [
-            "element (riot).exe",
+            "element.exe",
             "element-desktop"
         ]
     ],
     "shortcuts": [
         [
-            "element (riot).exe",
+            "element.exe",
```
Ash258, line 21:
```suggestion
            "Element.exe",
```

bucket/element.json
```diff
@@ -12,13 +12,13 @@
     "extract_dir": "lib\\net45",
     "bin": [
         [
-            "element (riot).exe",
+            "element.exe",
```
Ash258, line 15:
```suggestion
            "Element.exe",
```

bucket/element.json
```diff
@@ -12,13 +12,13 @@
     "extract_dir": "lib\\net45",
     "bin": [
         [
```
Ash258, line 14:
```suggestion
        "Element.exe",
        [
```

erogy: Why do we need two different shims for the same executable?

Ash258: Becuase in documentaiton they use `element-desktop` ...
image

erogy: Yeah I know, but why can't we just have `element-desktop` then?

Ash258: Because it is its name? Original executable.

Before it was not created before its stupid `(riot) in name. Now they removed and it could be added.

Ash258:
Next time you will not apply changes using Github suggestions PR will be immediately closed

image

erogy:
> Next time you will not apply changes using Github suggestions, PR will be immediately closed

May I ask you why? Sure, it's not hard and I'll do it from now on, but what's wrong with the approach I use? Just curious.

Ash258:
- Automatic conversation resolve
- Avoid stupid mistakes by manual changes of PR author
- Co-author
- Exact purpose of suggestions

PR #4560, opened 2020-08-09
nwjs(-sdk): Fix conflict with native scoop files

### LINKED ISSUE #3932: nw.js name collision with file manifest.json
When you download NWJS through scoop, you get two files called manifest.json and install.json that are not in the normal distribution. These files are used by scoop to cache information about the bucket, and various information about the application.

Chromium uses manifest.json and this causes a name collision which is painful. It causes a malfunction because the format scoop uses does not agree with chromium behind the scenes. You will be given a popup window explaining this when it is run in a directory without a file package.json, then nw.js will enter a forever-running state (not a crash). If you run it in a directory with a package.json file (which is, how you kick off nwjs apps normally), I believe it gives a quiet failure and enters a forever running state, but have to double check.

To get it out of this state you need to run powershell (or something equivalent) and use tasklist to get the pid of nw, and then kill it with taskkill -pid -F [PID of nw.exe] - otherwise you can't uninstall nw.js, or start a new instance of nw.js. Using task manager, you can't see the running instance of nw.exe in my experience, hence the need to use tasklist-taskkill in powershell. The -F (force) flag is necessary, the task will not exit without it.

This is trivial to fix, because you can delete or rename 'manifest.json'. Finding that out and dealing with it was annoying, so I'm leaving this here. Also, I'm not sure what that does to the installation or scoops functioning (if it is a quick fix that works, and doesn't mess with updates, then that's my main concern as an end user).

omerakgoz34:
Scoop's installation manifest was conflicting with NWjs' manifest system.

- Closes #3932

### DIFF
--- a/bucket/godot-mono.json
+++ b/bucket/godot-mono.json
@@ -16,7 +16,8 @@
     "pre_install": [
         "(Get-ChildItem \"$dir\\Godot_*\\Godot_*.exe\" | Rename-Item -NewName \"godot.exe\");",
         "(Get-ChildItem \"$dir\\Godot_*\\*\" | Move-Item -Destination \"$dir\");",
-        "Remove-Item \"$dir\\Godot_*\""
+        "Remove-Item \"$dir\\Godot_*\"",
+        "New-Item \"$dir\\_sc_\" -ItemType \"file\" | Out-Null"
     ],
     "bin": "godot.exe",
     "shortcuts": [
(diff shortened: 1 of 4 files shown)


omerakgoz34:
@Ash258 Why "thumbs down" emoji?

NW.js was throwing me a manifest error every time before my fix :D

Screenshot_8
Screenshot_9

Ash258:
1. Use `extract_to` property and set it to static `nwjs`
1. Request changes to core to move from generic `install.json`, `manifest.json`
	- Ref: Ash258/Scoop-Core#18, Ash258/Scoop-Core@26c301768bd83161a1910131d93c8517ad1da5b3, Ash258/Scoop-Core@d805aa195f26d827d913b558bc4d3ed44bc34244

Ash258:
Do you really think it is good idea to put changes related to godot into PR related to NWJS??

PR #4566, opened 2020-08-10
godot(-mono): Add Persistent Data

omerakgoz34:


omerakgoz34:
https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html
Screenshot_11
:D

omerakgoz34:
Install and Update works with just the "persist" setting :P

Ash258:
So user who already used this software and have the configuration in `%APPDATA%\Godot` will have to start again after update.

bucket/godot.json
```diff
@@ -13,14 +13,21 @@
...
     "shortcuts": [
         [
             "godot.exe",
             "Godot"
         ]
     ],
+    "persist": "editor_data",
+    "post_install": [
+        "if (!(Test-Path \"$persist_dir\\editor_data\\editor_settings-3.tres\") -and (Test-Path \"$Env:AppData\\Godot\\editor_settings-3.tres\")) {Copy-Item -Path \"$Env:AppData\\Godot\\*\" -Destination \"$persist_dir\\editor_data\" -Recurse -Force}"
+    ],
```
Ash258, line 30: - Move it directly after `pre_install`
- Why do you check for `editor_settings-3.tres`?
- `$Env` is defined as `$env`
- `$Env:AppData` == `$env:APPDATA`

Ash258: - Move `post_install` property **directly** after `pre_install` property
- Why did you check for editor_settings-3.tres?
- Why do you use `Measure-Object`?

omerakgoz34: Q: Why did you check for editor_settings-3.tres?
A: No, I didn't use it this time :D

Q: Why do you use Measure-Object?
A: Testing for the folder is empty or not

Ash258: > Why did you check for editor_settings-3.tres?

Past tense. Why did you do it in first time??

> A: Testing for the folder is empty or not

Does not make sense... Absolutely wasted `measure-object` call when you can use direct array property.

image

Ash258: `if (((Get-ChildItem \"$persist_dir\\editor_data\" -Force | Select-Object -First 1 | Measure-Object).Count -eq 0) -and ((Get-ChildItem \"$env:APPDATA\\Godot\" -Force | Select-Object -First 1 | Measure-Object).Count -ne 0)) {Copy-Item -Path \"$env:APPDATA\\Godot\\*\" -Destination \"$persist_dir\\editor_data\" -Recurse -Force}`

Why did you ever think this is good approach??

`gci` calls on directory, always return array. Classic nature of native powershell cmdlets

So this code will do the following:

1. Gci call
	1. In case the directory do not exists, it will fail and throw exception
	1. -Force parameter does not make sense.
1. Limit the output of gci into 1 Item for no reason
1. Nonsensual `measure-object` call just for slowing down the execution
1. `Copy-Item` call
	1. It will produce junk in case the `editor_data` folder do not exists (first ever case on update)

Your approach of copying all files:

image


Correct:

image

omerakgoz34: I don't know PowerShell. I just want to contribute to scoop repo.
I wasn't know how to test a folder is empty or not. Then, today I found a way in a forum.

> It will produce junk in case the editor_data folder do not exists

but scoop creates persist folder on update

without persist and post_install:
Screenshot_16

with persist and post_install (fake update):
Screenshot_17

Ash258: Then create issue instead of posting non functional broken PR, when you do not know powershell...

Ash258: Forget it. I will do it on my own

omerakgoz34: >:///

PR #4598, opened 2020-08-17
miniconda3: Update to version 4.9.2

### LINKED ISSUE #4481: miniconda3 doesn't get autoupdate
Currently the latest version of miniconda3 is 4.8.3 but in manifest file, version is still 4.7.12.1.
looks like the installer's filename changed to `Miniconda3-py37_4.8.3-Windows-x86_64.exe` (or `Miniconda3-py38_4.8.3-Windows-x86_64.exe` which based on Python 3.8) so it doesn't match checkver's regexp pattern (`Miniconda3-([\\d.]+)-Windows`).

https://repo.anaconda.com/miniconda/

oerpli:
Closes #4481


Ash258:
Is it good idea to limit it into python 3.8?

How will this behave if I have python 3.7 installed instead? Or this will install full python

oerpli:
I am not sure whether the whole versioning hell of python can/should be solved by scoop. 

If we care about this, having multiple manifests for specific versioning needs is something that might be a good approach, e.g:

- miniconda3: Installs some default version (3.8.x seems to be the correct choice for me, as it's also shipped with e.g. the current Ubuntu LTS release)
- miniconda37: Python3.7
- miniconda38: Python3.8
- ...

Though how the shims would then be created in a consistent or useful manner eludes me currently.

oerpli:
To better address your questions:
> Is it good idea to limit it into python 3.8?
I think yes
> How will this behave if I have python 3.7 installed instead? Or this will install full python
If you have installed the previously available version of miniconda3 via scoop (python 3.7) it will update to 3.8.

oerpli:
To add an additional point:
I just found that there is a scoop-bucket with versioned versions of several dev-tools (including python):

https://github.com/ScoopInstaller/Versions

I think that:

1. for users that need a specific python-version, this would be the better choice than conda.
2. if many users need an specific older miniconda version, a bucket like extras-versions would be the better approach then polluting this bucket with out-of-date packages.

magicgoose:
Any plans to merge? The patch looks good IMO (although I'm not aware of any possible gotchas about how it will actually work)

PR #4605, opened 2020-08-18
windowgrid: Add version 1.3.1.1

### LINKED ISSUE #3746: [Request] Add WindowGrid
WindowGrid is a very handy tool for anyone in need to work with multiple windows on a single or multiple screens. It allows to nicely resize a window to match some part of a screen and basically organize multiple windows to be able to work flawlessly on various tasks.

This is a software link: http://windowgrid.net

kunago:
- Closes #3746

### DIFF
--- a/bucket/windowgrid.json
+++ b/bucket/windowgrid.json
@@ -1,0 +1,18 @@
+{
+    "version": "1.3.1.1",
+    "description": "A modern window management program for Windows.",
+    "license": "Unknown",
+    "homepage": "http://windowgrid.net",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
The executable inside archive is setup, not application

image

kunago:
Once you pass the setup and choose to make the app portable, it creates a default "WindowGrid.settings" file in the installation dir. It's a JSON file which also contains this:
```
{
    "App":
    {
        "Portable": true,
        ...
    },
    ...
}
```

This makes the app portable. You cannot just save this part which I tried, it requires whole JSON to be generated. Is there any standard way to create or download from the default settings file if it does not exist yet?

Ash258:
Do not download or create full hardcoded json. Such huge static files will not survive updates and will be fragile.

kunago:
Unfortunately the program does not seem to be developed as of now but still belongs to those working the best.
So if this does not fit scoop policy, feel free to close the request.

PR #4634, opened 2020-08-24
ytmdesktop: Add version 1.12.1

jcwillox:
Adds the well known 3rd party desktop client for Youtube Music
* https://ytmdesktop.app/
* https://github.com/ytmdesktop/ytmdesktop

### DIFF
--- a/bucket/ytmdesktop.json
+++ b/bucket/ytmdesktop.json
@@ -1,0 +1,36 @@
+{
+    "version": "1.12.1",
+    "description": "A Desktop App for YouTube Music",
+    "homepage": "https://ytmdesktop.app",
+    "license": "CC0-1.0",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/ytmdesktop.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "1.12.1",
+    "description": "A Desktop App for YouTube Music",
```
Ash258, line 3: I am missing clarification that this is not official product from Google.

Add something like `unofficial`, `third-party`, ...

Ash258 (changes requested):

bucket/ytmdesktop.json
jcwillox: Agreed, I just grabbed that description from their GitHub page.

Ash258 (approved):

PR #4651, opened 2020-08-27
Added superslicer manifest

Duckle29:
I added a manifest for SuperSlicer. A deriviative of PrusaSlicer.


Ash258 (changes requested):
> it's a personal fork for my use. 

I do not think this should be added in official bucekt

Duckle29:
That's fair. It's seen a bit of a wider use than just that, but I suppose with no commitment past personal use from the repo owner than, it may be shaky. Any known buckets it'd fit in, or would I be better off making my own?

PR #4656, opened 2020-08-27
kitty: Use portable executable

tb4jc:
Create shortcut to portable version else all settings and sessions are stored to registry instead of created directories.


Ash258 (changes requested):
Revert what you have done and add following lines into `pre_install`:

`        "Move-Item \"$dir\\kitty_portable.exe\" \"$dir\\kitty.exe\" -Force"`

tb4jc:
Hm, I think it's nicer to keep the non-portable version 'kitty.exe' instead of overwriting it. Else, if there's anybody who likes to use it, can't use it anymore.

Ash258:
No

tb4jc:
Well, of course, we can have different opinions. But instead of just a 'no', would be more helpful to understand decisions.
I've changed it anyhow.

Ash258:
There is no reason to explain clear intention of all scoop manifests and whole ecosystem...

https://github.com/lukesampson/scoop-extras/pull/4656/files#diff-856a56e68c9f9bf337a418d147d3fcbcR42-R47

It should be there since the manifest was added. That is why there is default config creation in pre_install script. Since the author of the application changed the packages like 15 times, it got broken after few updates.

Now finally it works as intended until author change it again.

As soon as I will do the migration it will work as desired.

Ash258:
> Create shortcut to portable version else all settings and sessions are stored to registry instead of created directories.

This is not true

PR #4690, opened 2020-09-02
nvcleanstall: Add version 1.7.0

ghost-420:


Ash258 (changes requested):
- Indent using 4 spaces
- Checkver cannot work
- Move hash property after url

Ash258:
/verify it will most likely fail as base repo do not exists

Ash258 (approved):

PR #4692, opened 2020-09-02
noise-supression-for-voice: Add version 0.9

ghost-420:


Ash258 (commented):
There are no executables/binaries/shortcuts.

Scoop is meant to be for applications.

PR #4707, opened 2020-09-04
Update darktable.json with latest NSIS exe file

rpeyron:
Script to cope with new NSIS exe, some hints :  
- NSIS executable command-line arguments: /S for silent install and /D= for destination folder
- RunAsInvoker to avoid the elevation asked by the NSIS setup
- pipe to Out-Null to force to wait for the end of the installation

Also re-added checkver and autoupdate

### DIFF
--- a/bucket/darktable.json
+++ b/bucket/darktable.json
@@ -1,20 +1,36 @@
 {
-    "version": "2.4.4",
-    "description": "A virtual lighttable and darkroom for photographers",
+    "version": "3.2.1",
     "homepage": "https://www.darktable.org/",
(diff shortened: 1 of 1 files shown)


rasa (approved):

Ash258:
@rasa Really you pushed it with this name? I though it is clear that format `<manifest>(@<version>)?: Change` is clear.

rasa:
> @rasa Really you pushed it with this name? I though it is clear that format `<manifest>(@<version>)?: Change` is clear.

Yeah, my bad, sorry. I did it again yesterday. I will strive to be more careful.

PR #4726, opened 2020-09-09
Update retroarch hashes

SirFireNewt:
Retroarch build server was hacked and had everything deleted [see link](https://www.libretro.com/index.php/buildbot-server-up-and-running-again-the-status-and-future-plans/) so these are the hashes for the new build

### DIFF
--- a/bucket/retroarch.json
+++ b/bucket/retroarch.json
@@ -6,11 +6,11 @@
     "architecture": {
         "64bit": {
             "url": "https://buildbot.libretro.com/stable/1.9.0/windows/x86_64/RetroArch.7z",
-            "hash": "70ca7d174a9687fb7029806340e30cc297cf4d65f8cd886764778bdd2a042cd4"
+            "hash": "dfb1f230d672e74cc69b1fcc10c2a1232de02c9ff357a9fe793de86c1795d4bb"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


equinox:
I'm not entirely sure what's holding the fix up, but there's already a fix over here https://github.com/lukesampson/scoop-extras/pull/4570

Ash258:
> What holds the update?

You have it inside this PR description: THe server was attacked. Merging it in meantime could infect scoop users (that is the reason why hash fixes are not merged automatically and they are investigated, or at least in extras repo it is this way)

Ther files changed about 4 times, since the first change.

How can I know the new files are the correct and they were not infected? When developers do not provide checksum there is no way how to validate the hash changes. Some files are signed, but some are not.

Until developers say it was fully fixed it will be merged.

Duplicate of #4570

SirFireNewt:
Aah sorry my bad, saw the update failing and figured I'd just go ahead and update the hash, didn't think to check for a pr since it had been a while

As far as I am aware the devs have cleaned up basically everything including build server and github org but I appreciate the caution

link is their response saying "Buildbot and Github mostly restored" which was posted aug 27th tomorrow the 10th will be two weeks from that post so perhaps it would be safe to merge sometime soon

PR #4738, opened 2020-09-11
android-studio: Suggest android-clt

kidonng:
Now that ScoopInstaller/Main#1362 has been merged, we can suggest the new SDK instead.

### DIFF
--- a/bucket/android-studio.json
+++ b/bucket/android-studio.json
@@ -30,7 +30,10 @@
     },
     "extract_dir": "android-studio",
     "suggest": {
-        "SDK": "android-clt"
+        "SDK": [
+            "android-clt",
+            "android-sdk"
+        ]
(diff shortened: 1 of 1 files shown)


bucket/android-studio.json
```diff
@@ -30,7 +30,7 @@
     },
     "extract_dir": "android-studio",
     "suggest": {
-        "SDK": "android-sdk"
+        "SDK": "android-clt"
```
Ash258, line 33: Add both

Ash258 (changes requested):

pratikpc:
@kidonng should Flutter also depend on CLT instead of the SDK?

kidonng:
@pratikpc I think so, you are welcome to try submit a PR

PR #4740, opened 2020-09-11
Added KiCAD nightly

Duckle29:
Currently KiCAD doesn't provide hashes for the binaries. I'm asking if they could provide that, as it'd avoid a 1gb download just for calculating hashes when running checkver.

### DIFF
--- a/bucket/kicad-nightly.json
+++ b/bucket/kicad-nightly.json
@@ -1,6 +1,6 @@
 {
     "version": "17403.fc1665ff2",
-    "description": "Electronics Design Automation Suite",
+    "description": "Nightly build of KiCAD: An Electronics Design Automation Suite",
     "homepage": "https://kicad-pcb.org",
     "license": "GPL-3.0-only",
     "architecture": {
--- a/bucket/kicad-nightly.json
+++ b/bucket/kicad-nightly.json
@@ -1,0 +1,44 @@
+{
+    "version": "17403.fc1665ff2",
+    "description": "Electronics Design Automation Suite",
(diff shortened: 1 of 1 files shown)


Ash258 (commented):
How often the updates are done?

Duckle29:
There's a build every day, the version is: `commits_since_beginning_of_version_control.9_hex_digits_of_commit` so a build might not actually be different from day to day (but will have a different hash) however that's rare. 

If you mean updates to the manifest. Not sure. I talked to the dev team a bit, and they were open to having me compute hashes of the build artifacts. Alternatively I think we found Jenkins gives an MD5 fingerprinter.
Extra alternatively, the exes are signed, so while it's not an option (that I'm aware of) in the manifest, they suggested checking the exe with [Sign Tool](https://docs.microsoft.com/en-us/dotnet/framework/tools/signtool-exe)

Ash258:
Then move this PR into versions or nightly bucket. There is no reason to have spam commit with this.

Duckle29:
will do. I wasn't aware there was a nightly bucket. I found the firefox-nightly in here and figured this was the place :)

igitur:
@Duckle29 Did you find the nightly bucket? Do you perhaps have the URL? Thanks

Duckle29:
> @Duckle29 Did you find the nightly bucket? Do you perhaps have the URL? Thanks

While there is a nightly it seems to only have one manifest and isn't following the normal bucket layout, so I decided to put it in Versions: https://github.com/ScoopInstaller/Versions

(nightlies: https://github.com/ScoopInstaller/Nightlies)

PR #4747, opened 2020-09-12
uget: Update to version 2.2.3-1

linsui:


Ash258 (commented):
There is no changelog for this version, neither homepage do not mention it.

Where is developer statement regarding this?

linsui:
I don't know how to access their developer and their website has not been updated since 2.2.1. I don't know why. Is that a problem?

Ash258:
Yes it is. There is no official statement about this file and not even changelog, which indicates usesrs, should not update to this.

linsui:
Their Ubuntu package has been updated to [2.2.3](https://ubuntu.pkgs.org/20.10/ubuntu-universe-amd64/uget_2.2.3-2_amd64.deb.html). Looks like their website is outdated. [Here](https://sourceforge.net/projects/urlget/files/uget%20%28stable%29/2.2.3/) is the changelog. Since it the uget (stable) floder, it should be stable enough to use.

Ash258:
But the changelog do not mention the `-1`. What does it mean? Why the version format break?

linsui:
I don't know... Hot fix I guess.

PR #2371, opened 2019-06-28
musicbee: Add autoupdate

rasa:
Installed locally fine.

PR #4752, opened 2020-09-13
musicbee: Fix license

BlackPowerade:


bucket/musicbee.json
```diff
@@ -2,8 +2,7 @@
     "version": "3.3.7491",
     "description": "The ultimate music player to easily manage, find, and play music, podcasts, web radio stations, etc",
     "homepage": "https://getmusicbee.com",
-    "license": "GPL-2.0-or-later",
-    "notes": "Currently it's not possible to make Plugins and Skins persistent",
```
Ash258, line 6: Why is it possible now?

BlackPowerade: I think so. 
Scoop used to have a bug when persisting things that already existed (https://github.com/lukesampson/scoop/pull/2732) but this change was merged after the original manifest was created. (Nov 16, 2018)
Newer versions of musicbee come with the AppData and Library folder already, and scoop does not have a problem persisting them.

Ash258: How is the manifest addition date relevant to this line?? The persisting message was added in #2371, because scoop does not support merging the persisted folders, hence it cannot handle when developers add new native plugin or skins.

BlackPowerade: I see. I thought the note was added shortly after the original manifest was created

Ash258: Revert everything related to persisting and keep only license change

Ash258 (changes requested):
Use link for license

https://musicbee.fandom.com/wiki/FAQ#Are_there_any_limitations_on_using_MusicBee.3F

BlackPowerade:
That is not the license. 
Musicbee has no stated license. It is freeware.

Ash258:
The FAQ explains it is freeware. Use it as URL so users can verify it...

PR #4753, opened 2020-09-13
wexond: Fix manifest

sh4869221b:
Corrections to the wexond repository name and download files have been made


Ash258 (changes requested):
Downloading and extracting few kilobytes web setup is completely nonsense.

PR #4788, opened 2020-09-19
brave: Update to version 1.14.81

### LINKED ISSUE #4606: brave-browser update
brave-browser is updated with the inclusion of sync-v2.
please update the manifest so that the update is available.

operator-name:
~~Newer releases of brave offer zip files for win32, simplifying the installation process.~~

Closes #4606.

### DIFF
--- a/bucket/brave.json
+++ b/bucket/brave.json
@@ -8,12 +8,12 @@
     },
     "architecture": {
         "64bit": {
-            "url": "https://github.com/brave/brave-browser/releases/download/v1.14.81/brave-v1.14.81-win32-x64.zip",
-            "hash": "703ca8d7b9cc7aac7ec7cc610167701f18d68d1a26176d3dc12eeb5241a7aa34"
+            "url": "https://github.com/brave/brave-browser/releases/download/v1.14.81/BraveBrowserStandaloneSilentSetup.exe",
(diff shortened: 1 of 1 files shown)


operator-name:
Actually don't merge this yet, I have misunderstood the files. I will have to investigate where `BraveBrowserSetup.exe` finds it's files.

operator-name:
It looks like [`chocolatey` simply runs the setup.exe.](https://chocolatey.org/packages/brave/1.14.83)

operator-name:
The options seem to be [using the default installer](https://support.brave.com/hc/en-us/articles/360042816611-Why-isn-t-Brave-updating-automatically-on-Windows-), using [portable apps' build](https://github.com/portapps/brave-portable/releases) or another third party. Please advise.

Ash258 (changes requested):
Application which are installed into other locations than SCOOP are not accepted in official buckets.

Move this PR into non-portable bucket

operator-name:
What about using `https://github.com/portapps/brave-portable/releases/download/1.13.82-63/brave_installer-win64.exe`, or should that be under `brave-portable.json`?

Ash258:
If you want to use random third party "portable" installer use your own bucket.

operator-name:
Noted.

PR #4790, opened 2020-09-20
cyberduck: Add version 7.6.1.33485

integzz:


integzz:
Sorry, I am not sure what you're pointing out here? Do you mean that I have to add the `persist` path instead of storing the profiles in `%APPDATA%\\Cyberduck\\Profiles` , which is mentioned in `notes`?

Ash258 (changes requested):

Ash258 (approved):

PR #4812, opened 2020-09-24
qv2ray: pre_install tweaks

kidonng:
- Disable automatic update check
- Fix potential config error


Ash258 (commented):
Source/Documetantation for this changes?

kidonng:
- Disable automatic update check: https://github.com/Qv2ray/Qv2ray/blob/380049ecfad4fd65ca845b6f0af26bec7a2a706d/src/components/update/UpdateChecker.cpp#L32-L33
- Fix potential config error: if `logLevel` is not predefined, Qv2ray will generate a wrong value and cause the program to crash when connecting. I have confirmed the bug with the author, but unfortunately I can't publicize the details.

PR #4844, opened 2020-10-01
bitwarden: Use portable version to prevent installations

### LINKED ISSUE #4843: Bitwarden Should Use Portable Version (Current Version Leads to Non-Portable Installation)
Bitwarden currently extracts from the NSIS file instead of using the portable version that is readily available [on the GitHub page](https://github.com/bitwarden/desktop/releases). Currently it is "Bitwarden-Portable-1.22.2.exe".

For configuration files, it should be as simple as creating an empty folder called "bitwarden-appdata" (if it doesn't exist) and persisting it (it is usually generated the first time you run the portable version).

This should be fixed, as the NSIS version currently being used forces an actual **install** (that appears in "Add and remove programs") whenever the auto-updater detects a new update. This is apparently disabled by default in the  portable version.

Unfortunately the portable version does not have a hash that can be checked against anywhere, but this trade-off is worth it in my opinion for not having an app repeatedly install itself on my system whenever a new update comes out.

Velgus:
- Closes #4843

### DIFF
--- a/bucket/bitwarden.json
+++ b/bucket/bitwarden.json
@@ -14,6 +14,7 @@
         ]
     ],
     "persist": "bitwarden-appdata",
+    "post_install": "if(Test-Path \"$env:Appdata\\Bitwarden\") { Copy-Item \"$env:Appdata\\Bitwarden\\*\" \"$persist_dir\\bitwarden-appdata\" -Recurse -Force }",
     "checkver": {
         "github": "https://github.com/bitwarden/desktop"
     },
--- a/bucket/bitwarden.json
+++ b/bucket/bitwarden.json
@@ -14,7 +14,12 @@
         ]
     ],
     "persist": "bitwarden-appdata",
-    "post_install": "if(Test-Path \"$env:Appdata\\Bitwarden\") { Copy-Item \"$env:Appdata\\Bitwarden\\*\" \"$persist_dir\\bitwarden-appdata\" -Recurse -Force }",
+    "post_install": [
+        "if(Test-Path \"$env:Appdata\\Bitwarden\") {",
+        "    Copy-Item \"$env:Appdata\\Bitwarden\\*\" \"$persist_dir\\bitwarden-appdata\" -Recurse -Force",
+        "    Remove-Item \"$env:Appdata\\Bitwarden\" -Recurse",
+        "}"
+    ],
     "checkver": {
         "github": "https://github.com/bitwarden/desktop"
     },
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
Data migration from non-portable installation is required, otherwise you will completely ignore users configuration and force them to start clean

Velgus:
In my experience, it already loses user configuration between each update anyways with the current setup, I'm not sure why, but every time I update, my theme and settings return to default. This, if anything, should fix that going forward.

Ash258 (changes requested):
It uses classic electron "portale" behavior, which is against scoop principle.

https://github.com/lukesampson/scoop-extras/pull/2087#issuecomment-488344372

Configure portable configuration using 7z files or close the PR

A

Velgus:
It's already using the NSIS version currently, the problem is, like I said, it "installs" (gets a full "Add or remove programs" entry) whenever there's a new Bitwarden update. There is no way to prevent this behavior on Bitwarden's end.

I personally think that is a larger break of "scoop principle" than using a temporary AppData folder that self-deletes when closed.

Ash258:
Add or remove entry is not against scoop principle at all.
Portable, silent installers also creates it, which is not something scoop should and would prevent.

https://github.com/bitwarden/desktop/blob/38ecc3b74b39d675f37eae0b86b06d47b084bc08/src/main.ts#L40-L41

You do not need the portable executable at all to support portable configuration.

For the autoupdate, reach to developers, and request the ability to disable it.

Also noone force you to download and install the update https://github.com/bitwarden/jslib/blob/e1d42f95d9c662bd4ffc8221bc4e5b9b96a095a6/src/electron/updater.main.ts#L65

Velgus:
That code is bugged in some way - it installs even if you reject it. This has happened to me numerous times since I started installing Bitwarden with Scoop - I close the update window and reject installing an update, and it still plants a new installation in the add or remove 

>  Add or remove entry is not against scoop principle at all.

It is absolutely against [scoop principle](https://github.com/lukesampson/scoop/wiki/So-What).
- Lets you stay on the command line, where you can work fast
- avoids GUIs whenever possible, keeps you on the command line

Currently, it requires actions outside of the command line in order to manage install.

- installs applications independently and in a self-contained way (which means less conflicts, easy to undo installs)

The separate installation entry means it is no longer self-contained. You need to perform actions outside of scoop to fix it.

Ash258:
Not it not. You install it from command line and it works. No issue.

When there is bug in application, scoop cannot do anything about it.

Report problem to bitwarden developers and then come back and adopt portable configuration.

PR #4863, opened 2020-10-05
Update ProcessHacker and add nightly

SpiralP:
Updated links in `processhacker.json` so they don't have to redirect, and added a `processhacker-nightly.json` that tracks their nightly builds.

The way ProcessHacker checks for updates is found in their code [here](
https://github.com/processhacker/processhacker/blob/872bb111afe49a1d4b209924f7d37f9edf93f949/plugins/Updater/updater.c#L328) which is called from [here](https://github.com/processhacker/processhacker/blob/872bb111afe49a1d4b209924f7d37f9edf93f949/plugins/Updater/updater.c#L472), which shows it checks `wj32.org` first then `processhacker.sourceforge.io` (so I decided to just check wj32.org).  I got checkver and autoupdate to work so far with this config.

Let me know if I need to change anything, this is my first pr to this repo!  Thanks!

### DIFF
--- a/bucket/processhacker-nightly.json
+++ b/bucket/processhacker-nightly.json
@@ -1,0 +1,54 @@
+{
+    "version": "3.0.3456",
+    "description": "A powerful, multi-purpose tool that helps you monitor system resources, debug software and detect malware.",
+    "license": "GPL-3.0-only",
+    "architecture": {
+        "32bit": {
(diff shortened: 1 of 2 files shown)


Ash258 (changes requested):
Remove nightly manifest from this PR and do not post nighly versions in Extras

SpiralP:
My bad, I'll just use my own repo thanks

PR #4867, opened 2020-10-05
tableplus: Update to version 3.10.4.142

OranShuster:


OranShuster:
@Ash258 does `package-fix-needed`  mean there is an issue with my change? since you didn't comment anything i thought it was ok

Ash258:
I will comment when I have time for fixing the clear problem with autoupdate and version checking.

OranShuster:
`checkver` is working fine
change `autoupdate` to use `$matchHead` and now that works also

PR #4879, opened 2020-10-08
gimp: Update to version 2.10.22

Nirose:
update to 2.10.22


Ash258 (changes requested):
32bit installation does not work

image

Nirose:
help needed, I'm not good with powershell

rasa:
> 32bit installation does not work

But that’s unrelated to the version bump, yes? Should we accept this, and open another ticket re the install fail?

PR #4900, opened 2020-10-11
powertoys: Update to version 0.23.2

imasaru:


Ash258 (changes requested):
Installation is not handled... This will spawn installer, which will install into default location (Program files)

imasaru:
@Ash258 Oops, just realized that. Thank you!

PR #4920, opened 2020-10-19
speedfan: Use installer

### LINKED ISSUE #4905: speedfan 4.52 fails (requires elevation)
Installing speedfan results in:

```
speedfan452.exe (2.9 MB) [====================================================================================] 100%
Checking hash of speedfan452.exe ... ok.
Running installer... error.
ERROR Exception calling "Start" with "0" argument(s): "The requested operation requires elevation"
Installation aborted. You might need to run 'scoop uninstall speedfan' before trying again.
Installing 'speedfan' (4.52) [64bit]
Running pre-install script...
```

There is no UAC prompt, and scoop.exe exits with error code `1`. 

The scoop app dir contains only the installer, and three 0-byte files created for persist:

```
    Directory: ~\scoop\apps\speedfan\4.52


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        2020-10-13   9:35 AM        3086696 speedfan452.exe
-a----        2020-10-13  10:00 AM              0 speedfanevents.cfg
-a----        2020-10-13  10:00 AM              0 speedfanparams.cfg
-a----        2020-10-13  10:00 AM              0 speedfansens.cfg
```

It appears speedfan installs some kind of system driver, so it's likely it _does_ require escalation -- it just shouldn't be failing in scoop like this, with no UAC prompt. This is the same version in the original PR (#1074) and 4.52 appears to be released in 2015, so I'm assuming this isn't a new issue, or something specific to my system.

gregmac:
Fixes #4905


Ash258 (changes requested):
Why innounp could not be used?

If it requires installation and admin rights, the manifest will be removed as it is against scoop’s [principles](https://github.com/lukesampson/scoop/wiki/Chocolatey-Comparison).

gregmac:
Very possible it should be removed, then.

It's an NSIS installer, with contents:

```
   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
1996-04-03 15:33:26 .....                      1524  $SYSDIR\giveio.sys
2012-12-29 16:59:38 .....                     13425  $SYSDIR\speedfan.sys
2012-12-29 16:59:38 .....                     12896  $SYSDIR\speedfan.sys
2015-02-10 13:56:14 .....                     49303  $TEMP\sfextra.dll
2015-02-10 13:56:14 .....                     49303  sfextra.dll
2016-06-29 14:01:22 .....                   2383015  speedfan.exe
2006-08-02 06:11:14 .....                    470930  speedfan.chm
2016-06-29 14:22:48 .....                     21862  speedfan.txt
2006-10-01 13:39:58 .....                     55861  pciidsdata.csv
2012-03-26 13:06:16 .....                     28619  configs.zip
                    .....                       275  uninstall.exe.nsis
------------------- ----- ------------ ------------  ------------------------
2016-06-29 14:22:48                         3087013  11 files
```

Notably, `speedfan.sys` or `giveio.sys` are installed as system drivers. I'm also not sure why there's two `speedfan.sys` files.

So while the contents could be extracted, there's no way around the system driver part. I haven't tried running without the driver to see if it still functions, but I don't think that's worth investigating deeply.

I'll let those in the project make the call on this. This was originally added in https://github.com/lukesampson/scoop-extras/pull/1074 and approved by @r15ch13.

rasa:
> If it requires installation and admin rights, the manifest will be removed as it is against scoop’s [principles](https://github.com/lukesampson/scoop/wiki/Chocolatey-Comparison).

We already have two sudo apps in our [main](https://github.com/ScoopInstaller/Main/search?q=sudo) bucket, and three in [extras](https://github.com/lukesampson/scoop-extras/search?q=sudo&type=code). While I understand wanting to limit new apps that depend on sudo, it would serve our users to provide them apps that require admin rights.

What about allowing admin-only apps to depend on sudo, but then if a non-admin user tries to install a admin-only app, an error message of “You need administrator rights to install this app.” is shown? If the user adds `—elevate`, scoop tries to elevate before installing the app. I know this might be moving some of the sudo logic into scoop, but it just seems cleaner from a user’s point of view.

The same thing could happen when a non-admin user tries to install with `—global`. They’ll get an error message unless they use `—elevate`.

Also, we might be conflating avoiding installer popups (which are avoidable) with avoiding UAC-popups (which aren’t, right?). I don’t think installers that pop up should be allowed in main, so maybe we move the two that are there into extras? Or maybe we move all five into a new bucket named `drafts`. Chocolatey avoids installer popups using precompiled Auto-Hotkey scripts. We could do the same.

Another option would be to add a flag `“interactive: true”` to popup installs, and if the user uses `scoop —batch` the installation would simply fail with an error message.

gregmac:
On further consideration, I disagree with @Ash258's change request.

I think it's desirable to avoid requiring UAC/sudo as much as possible, but that the practical matter is that sometimes it is unavoidable.  I'm not sure why Scoop should be anti-UAC/admin on principle -- "admin only if absolutely necessary" seems very reasonable to me. 

There's also basically three different types of admin access an app can need:

* Admin required at install time
* Admin required at runtime
* Admin optional at runtime

Limiting to applications that install without admin eliminates a handful of useful apps, and limiting to anything that don't require admin at runtime eliminates several more.

As a (somewhat) hypothetical question: If it's just "admin at install" that isn't allowed, then would it be acceptable to change something like speedfan so it does UAC and installs the needed system drivers on first runtime (instead of at install time), and why or why not? It's certainly more work to maintain that, and I'm not sure what difference it makes from a user perspective.

---

This PR was created because right now, if you install speedfan without doing `sudo` first the install fails, and this is just annoying and unlike every other scoop app. This PR makes Scoop request UAC, and if denied or not available it just won't install.

If the desired outcome is to remove all admin-requiring apps, a new issue/PR for that should be opened, the debate on whether that's a good idea or not can happen there, and this PR can be closed as obsolete (or more likely: merged into the forked bucket that will almost certainly result from that decision taking place).

If we want to do more per @rasa's comment, I think that's beyond the scope of this PR (and would be in in the main project repository anyway). A new issue/PR can be created and then this can be closed as obsolete.

issaclin32:
**Speedfan's** installer can be extracted by 7-zip. But direct download does not work for http://www.almico.com/instspeedfan452.exe. Will work on that later.

rashil2000:
Since the app requires admin access _and_ installs system drivers, I really think it should be removed from here. While I can understand the need for sudo, the second part is really not what scoop is intended for.

Non-portable bucket would be the best fit for this application.

issaclin32:
The non-portable package is inactive for a while. I'll wait until *TheRandomLabs* respond, and send this to there.

rashil2000:
> **Speedfan's** installer can be extracted by 7-zip. But direct download does not work for http://www.almico.com/instspeedfan452.exe. Will work on that later.

Scoop is somehow able to download it. Does it work after extraction with 7zip?

issaclin32:
> Scoop is somehow able to download it. Does it work after extraction with 7zip?

It can be extracted by 7-zip, but **SpeedFan** will not start properly because the "speedfan service" is not installed.

bucket/speedfan.json
```diff
@@ -5,14 +5,15 @@
...
     "hash": "e2ccb3c0d23f0d04ee8057f5ce3861eea952fb20694c1656c9805b1d4cd922ff",
-    "depends": "sudo",
-    "pre_install": "foreach ($emptyFile in $manifest.persist) { New-Item \"$dir\\$emptyFile\" | Out-Null }",
+    "pre_install": [
+        "foreach ($emptyFile in $manifest.persist) { New-Item \"$dir\\$emptyFile\" | Out-Null }",
+        "Invoke-ExternalCommand \"$dir\\$fname\" -ArgumentList('/S', \"/D=$dir\") -RunAs | Out-Null"
+    ],
     "installer": {
         "script": "sudo \"$dir\\$fname\" /S /D=$dir"
     },
```
rashil2000, line 14:
```suggestion
```

This is no longer required

rashil2000 (changes requested):

PR #4943, opened 2020-10-20
wslgit : Shim "wslgit.exe" as "git"

Witchilich:


Ash258:
I am strictly against this change as for most of the users this is not suitable.

If you want to achieve such use case, configure the shim on your own manually.

rasa:
This is a very opinionated change. I would suggest hosting this in your own repo, as most people won’t want this.

PR #4944, opened 2020-10-20
Update snipaste 1.16.2 -> 2.5.1-Beta

scvoet:


Ash258:
Beta versions are not accepted in extras bucket, when there is stable version avaialble

scvoet:
But 

> Beta versions are not accepted in extras bucket, when there is stable version avaialble

However, the current version of snipaste is so old that many of the new features are not available. 1.x was released in 2018, while 2.5 is a 2020 product. Also Beta appears in all 2.x version naming.

PR #5004, opened 2020-10-29
Add youtube-dlc app manifest

JFRN:
Not adding to https://github.com/ScoopInstaller/Main because I don't consider it to be "a reasonably well-known and widely used developer tool". 

Youtube-dlc is at least as of now more actively developed than youtube-dl (over 300 commits ahead) and seems to have fixes for extractors that are buggy in the original project.


Ash258:
https://github.com/lukesampson/scoop-extras/pull/4756#issuecomment-691927490

PR #5005, opened 2020-10-29
MagicaVoxel@0.99.6: Fix URL

### LINKED ISSUE #4964: magicavoxel@0.99.6: Download failed
Installing 'magicavoxel' (0.99.6) [64bit]
URL https://github.com/ephtracy/ephtracy.github.io/releases/download/0.99.6/MagicaVoxel-0.99.6-win64.zip is not valid

The correct file name is "MagicaVoxel-0.99.6<strong>.2</strong>-win64.zip".

### LINKED ISSUE #5290: MagicaVoxel version doesn't match
Hello,

When I tried to install *MagicaVoxel*, I got a 404 not found exception.

I went on and looked at both *MagicaVoxel* and [scoop json definition](https://github.com/lukesampson/scoop-extras/blob/master/bucket/MagicaVoxel.json).

Version is listed as **0.99.6** but the binaries are on version **0.99.6.2**. That **.2** revision causes a mismatch between versions.

I don't know if there's any easy way to fix this, seeing that the version in the path and the version in the binaries aren't the same anymore in that release.

`https://github.com/ephtracy/ephtracy.github.io/releases/download/$version/MagicaVoxel-$version-win64.zip` won't work right now, as `$version` should be once 0.99.6 and once 0.99.6.2

I could do a pull request to update properly set it at the moment but I fear that it would break again with next version. 

Anyone has already seen that case and knows how to properly handle this?

Uinelj:
- Closes #4964
- Closes #5290

new version of 0.99.6 changed download paths.
the commit contains updated download path and corresponding hashes.

### DIFF
--- a/bucket/MagicaVoxel.json
+++ b/bucket/MagicaVoxel.json
@@ -5,14 +5,14 @@
     "license": "Unknown",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/ephtracy/ephtracy.github.io/releases/download/0.99.6/MagicaVoxel-0.99.6-win64.zip",
-            "hash": "2de37d619ecc53f411cfcf914329685d57903288d2b993f9e5c3ec906521a9c2",
-            "extract_dir": "MagicaVoxel-0.99.6-win64"
+            "url": "https://github.com/ephtracy/ephtracy.github.io/releases/download/0.99.6/MagicaVoxel-0.99.6.2-win64.zip",
+            "hash": "169dcddbea7dad30c0713d9ca211d445525501aceb9e12c8a6b51e7afad11420",
+            "extract_dir": "MagicaVoxel-0.99.6.2-win64"
         },
         "32bit": {
-            "url": "https://github.com/ephtracy/ephtracy.github.io/releases/download/0.99.6/MagicaVoxel-0.99.6-win32.zip",
-            "hash": "4d6d8b277c2c2c5911b3f3b7be680e6368c7e7ce7754aba8efd847a14a826e54",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
Did you try to install it using these files?

I cannot execute it using this build

Uinelj:
Hello,

I've tried to install it and successfully did so, using `scoop install MagicaVoxel.json` (verified today too)
I however tried on a amd64 architecture only.

Ash258:
It do not work with any version on my end. It got terminated immediately and log following to event viewer:

```log
Faulting application name: MagicaVoxel.exe, version: 0.0.0.0, time stamp: 0x5f6ff557
Faulting module name: unknown, version: 0.0.0.0, time stamp: 0x00000000
Exception code: 0xc0000005
Fault offset: 0x00000000
Faulting process id: 0xa6c
Faulting application start time: 0x01d6b218f74a65a9
Faulting application path: C:\Users\vagrant\scoop\apps\A\current\MagicaVoxel.exe
Faulting module path: unknown
Report Id: 8bd07a14-d738-4a74-a7f0-77a3ea6c6622
Faulting package full name: 
```

A

MunsterPlop:
I see this hasn't had any feedback in a while.
The problem is that magicavoxel's repo has tags with no build version whereas the downloadable archives do.
Unless they change it, the recipe should probably use a different way to get the proper archive name.

pratikpc:
@Uinelj I think a fix for the same was added and this file has conflicts

If the issue is not yet resolved, can you create a new PR for the same

PR #5014, opened 2020-10-30
lingoes: Add version 2.9.2

snowman:


Ash258 (changes requested):
- Remove checkver and autoupdate
- Why do you prefer installer over zip?

image

snowman:
There is only `one` version of the portable zip, but the installer has both two `x32` and `x64` versions.

Ash258:
/verify Download do not work, causing weird redirect

A

B

snowman:
That's not I could help, so should close this pull request?

PR #5019, opened 2020-10-30
add tcping

Rehtt:


Ash258 (changes requested):
There is no reason to add some random go implementation, created yesterday without any userbase or test, when there are much better alternatives

https://elifulkerson.com/projects/tcping.php

Ash258:
OPen 1 PR per manfiest

PR #5030, opened 2020-10-31
terminus: Adopt portable configuration

rosegayming:
See https://github.com/Eugeny/terminus#portable

### DIFF
--- a/bucket/terminus.json
+++ b/bucket/terminus.json
@@ -14,6 +14,9 @@
         }
     },
     "bin": "Terminus.exe",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
You need to handle the data migration from non portable path

dudeamax99:
Should the old folder in `%APPDATA%` be removed in a case like this or left?

rashil2000:
Added in 58acfc51afb9199353721b2bcc505fd8fba6b907

PR #5034, opened 2020-11-01
t-clock: force directory creation in $persist_dir

### LINKED ISSUE #5033: t-clock: persist folder not created when it doesn't exist
[New-Item](https://github.com/starise/scoop-extras/blob/5bd794d4cffab121425fb816d78fa99682c0f497/bucket/t-clock.json#L11) needs to *Force* the creation of the container folder in `$persist_dir` to work correctly.

Edit: got it. changed uninstaller script to check if persist dir exists.

starise:
Fixes #5033

### DIFF
--- a/bucket/t-clock.json
+++ b/bucket/t-clock.json
@@ -8,7 +8,7 @@
     "installer": {
         "script": [
             "if (-not (Test-Path \"$persist_dir\\T-Clock.ini\")) {",
-            "    New-Item \"$dir\\T-Clock.ini\" -ItemType File | Out-Null",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
This will not fix the issue... The line you changed is creatting a FILE in $dir.

You need to change the uninstaller script.

```json
        "script": [
            "ensure \"$persist_dir\" | Out-Null",
            "if (Test-Path \"$dir\\T-Clock.ini\") { Copy-Item \"$dir\\T-Clock.ini\" \"$persist_dir\"
        ] }"
```

PR #5036, opened 2020-11-01
paperwork: Add version 2.0

akuntsch:
`checkver` is a bit ugly because the jsonpath implementation does not allow selecting multiple fields which would make the regex simpler. Currently it heavily relies on the format of the Gitlab API response.

### DIFF
--- a/bucket/paperwork.json
+++ b/bucket/paperwork.json
@@ -2,7 +2,7 @@
     "version": "2.0",
     "description": "Personal document manager",
     "homepage": "https://openpaper.work",
-    "license": "GPL-3.0-only",
+    "license": "GPL-3.0-or-later",
     "url": "https://download.openpaper.work/windows/x86/paperwork-master-24ba2639.zip",
     "hash": "85414f4789e000b60ede495df8e314c7328d47ae16a163c1110e0d1a0b9cbdcd",
     "shortcuts": [
--- a/bucket/paperwork.json
+++ b/bucket/paperwork.json
@@ -1,0 +1,24 @@
+{
+    "version": "2.0",
+    "description": "Personal document manager",
+    "homepage": "https://openpaper.work",
+    "license": "GPL-3.0-only",
+    "url": "https://download.openpaper.work/windows/x86/paperwork-master-24ba2639.zip",
+    "hash": "85414f4789e000b60ede495df8e314c7328d47ae16a163c1110e0d1a0b9cbdcd",
+    "shortcuts": [
+        [
+            "paperwork.exe",
+            "Paperwork",
+            "",
+            "data/paperwork_64.ico"
+        ]
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/paperwork.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "2.0",
+    "description": "Personal document manager",
+    "homepage": "https://openpaper.work",
+    "license": "GPL-3.0-only",
```
Ash258, line 5: A

```suggestion
    "license": "GPL-3.0-or-later",
```

bucket/paperwork.json
```diff
@@ -0,0 +1,24 @@
...
+    "url": "https://download.openpaper.work/windows/x86/paperwork-master-24ba2639.zip",
+    "hash": "85414f4789e000b60ede495df8e314c7328d47ae16a163c1110e0d1a0b9cbdcd",
+    "shortcuts": [
+        [
+            "paperwork.exe",
+            "Paperwork",
+            "",
+            "data/paperwork_64.ico"
```
Ash258, line 13:
```suggestion
            "data\\paperwork_64.ico"
```

bucket/paperwork.json
```diff
@@ -0,0 +1,24 @@
...
+        ]
+    ],
+    "checkver": {
+        "url": "https://gitlab.gnome.org/api/v4/projects/World%2FOpenPaperwork%2Fpaperwork/repository/tags?order_by=name",
```
Ash258, line 17: A

Go through the source code/reverse engineer and provide native version check.

Ash258 (changes requested):
I am against adding application without releases to the official buckets. The checkver will not be executed until there is global release of the application. Leading to the potential of purging old installers/build artifacts and download failures.

bucket/paperwork.json
akuntsch: https://openpaper.work/beacon/latest is used for the update check. This gives me the version but for the download URL I need the commit hash.

akuntsch:
> I am against adding application without releases to the official buckets. The checkver will not be executed until there is global release of the application. Leading to the potential of purging old installers/build artifacts and download failures.

What do you mean by releases? Releases which would show up at https://gitlab.gnome.org/World/OpenPaperwork/paperwork/-/releases or do release git tags count?

For the version I just went with what Arch Linux uses (they build the package from source obviously): https://github.com/archlinux/svntogit-community/blob/packages/paperwork/trunk/PKGBUILD#L14

Even with a release in their GitLab repo the problem of obtaining the download URL would probably persist because past releases did not contain compiled assets and seemingly the only file under https://download.openpaper.work/windows without commit hash in its name seems to be https://download.openpaper.work/windows/amd64/paperwork-master-1.2.4.zip.

Not sure if the same kind of file will be available for 2.0 in the future.

issaclin32 (approved):

PR #5048, opened 2020-11-04
archwsl: Don’t erase file system on upgrades

### LINKED ISSUE #3800: archwsl update deletes existing installation including the entire filesystem
rip

rasa:
Fixes #3800 (i haven’t tested it, but it should work)

### DIFF
--- a/bucket/archwsl.json
+++ b/bucket/archwsl.json
@@ -36,7 +36,7 @@
     ],
     "uninstaller": {
         "script": [
-            "if ($cmd -eq 'uninstall') { return }",
+            "if ($cmd -ne 'uninstall') { return }",
             "$res = Invoke-ExternalCommand \"$dir\\Arch.exe\" 'isregd'",
             "if(!$res) { error 'ArchWsl is not registered!'; return }",
             "Invoke-ExternalCommand \"$dir\\Arch.exe\" 'clean', '-y' | Out-Null"
--- a/bucket/archwsl.json
+++ b/bucket/archwsl.json
@@ -36,6 +36,7 @@
     ],
     "uninstaller": {
         "script": [
+            "if ($cmd -eq 'uninstall') { return }",
             "$res = Invoke-ExternalCommand \"$dir\\Arch.exe\" 'isregd'",
             "if(!$res) { error 'ArchWsl is not registered!'; return }",
             "Invoke-ExternalCommand \"$dir\\Arch.exe\" 'clean', '-y' | Out-Null"


bucket/archwsl.json
```diff
@@ -36,6 +36,7 @@
     ],
     "uninstaller": {
         "script": [
+            "if ($cmd -eq 'uninstall') { return }",
```
akuntsch, line 39:
```suggestion
            "if ($cmd -eq 'update') { return }",
```

rasa: You’r correct. Thanks for catching that.

isavegas:
Looking forward to this! I was extremely lucky that I backed up my heavily modified Arch instance before it got wiped in an update.

rasa:
Can someone test this fix, and confirm it works?

escape0707:
> Can someone test this fix, and confirm it works?

If do a force update will be sufficient to test, I can backup my Linux and  try it now.

escape0707:
Well it still don't work as expected, because we haven't changed the `post_install` script where in line 32 `arch install $dir\rootfs.tar.gz\` is still invoked unconditionally. Look this up in `arch --help` or [archwsl offcial wiki](https://github.com/yuk7/ArchWSL/wiki/How-to-Use(for-Installed-Instance), we can see it will restore from a `tar.gz` backup. In this case it the initial clean `rootfs` and thus cover our data.

I suggest that in the `post_install` script (which will run after install and update) we also bypass the execution of `arch install` on `$cmd -eq update`.

escape0707:
Well, it still don't work.

Here is what I tried:

```powershell
scoop uninstall archwsl
cd ~\scoop\buckets\extras\bucket\
git restore -s db31afbe25803b9ae7ca88176d57b6ba86162156~ archwsl.json
scoop install archwsl
arch
```

```shell
touch a
useradd -m myaccount
exit
```

```powershell
arch config --default-user myaccount
arch
```

```shell
touch b
```

```powershell
git reset --hard
scoop update archwsl
arch
```

```shell
ls -l
ls /home
```

At this moment, what happened is the default user is back to `root` and files `a` and `b` and user `myaccount` don't exist.

I'm still trying to find a possible solution without touching the code of `Arch.exe` right now...

escape0707:
@rasa It works, but other users should be aware that you have to manually insert this change into your old, current version of archwsl's `manifest.json`, which should be `~\scoop\apps\archwsl\20.4.3.0\manifest.json` before doing the update. Or the old `uninstaller` script will still wipe your ArchLinux data.

Other than that, I still hold two little confusions. The first is that why not use `if ($cmd -eq 'update') { return }`? I personally think that's more descriptive about its concern and precise. Maybe more future-proof, too.

The other is that because that we didn't add a symmetric check-and-bypass in the `post_install` script, so that script will result in an (not severe but still shouldn't be ignored) error like this:

```powershell
scoop update archwsl
```

```
archwsl: 20.4.3.0 -> 20.10.31.0
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
Updating one outdated app:
Updating 'archwsl' (20.4.3.0 -> 20.10.31.0)
Downloading new version
Loading Arch.zip from cache.
Checking hash of Arch.zip ... ok.
Uninstalling 'archwsl' (20.4.3.0)
Running uninstaller script...
Removing shim for 'Arch'.
Unlinking ~\scoop\apps\archwsl\current
Installing 'archwsl' (20.10.31.0) [64bit]
Loading Arch.zip from cache.
Extracting Arch.zip ... done.
Linking ~\scoop\apps\archwsl\current => ~\scoop\apps\archwsl\20.10.31.0
Creating shim for 'Arch'.
Creating shortcut for Arch Linux (Arch.exe)
Persisting data
Running post-install script...
ERROR: This instance is already installed.
If you want to reinstall it, you can remove it with the "clean" command.
ERROR Exit code was -2147467260!
ERROR ArchWsl installation failed!
'archwsl' (20.10.31.0) was installed successfully!
Notes
-----
Even when you are logging in as 'root', some operations (like service command) require Windows administrator privileges
```

I suggest we also add the check for `post_install` so it prevents us from potential problems caused by changes to `arch install`, and avoid the unnecessary copy in the `post_install` script, etc.

PR #5063, opened 2020-11-06
registry-workshop: Add version 5.0.1

wenmin92:
[Registry Workshop](http://www.torchsoft.com/en/rw_information.html) is an advanced registry editor. It is a perfect replacement for RegEdit and RegEdt32 which shipped with Windows.

PRed before, but closed because it is a paid application. https://github.com/lukesampson/scoop-extras/pull/2630
Although Registry Workshop is a paid application, it is the best registry editor and we should include it.

### DIFF
--- a/bucket/registry-workshop.json
+++ b/bucket/registry-workshop.json
@@ -1,0 +1,49 @@
+{
+    "description": "An advanced registry editor.",
+    "homepage": "http://www.torchsoft.com/en/rw_information.html",
+    "license": "Shareware",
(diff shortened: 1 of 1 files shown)


issaclin32:
I am closing this because the discussion is settled in #7170.

IMO, the biggest reason "not" to add these paid apps is about **maintainance**. If there is a problem about installation, our maintainers may not be able to test and fix it.

issaclin32:
By the way, you can [start your own bucket](https://github.com/ScoopInstaller/BucketTemplate) to include these apps.

PR #5065, opened 2020-11-06
xyplorer: Add version 21.20.0200

wenmin92:
[Old xyplorer](https://www.xyplorer.com/free.php) is a free edition which has been discontinued as of 10-Jan-2017.


issaclin32:
see #5066

PR #5066, opened 2020-11-06
xyplorerfree: rename xyplorer to xyplorerfree

wenmin92:
[Old xyplorer](https://www.xyplorer.com/free.php) is a free edition which has been discontinued as of 10-Jan-2017.


issaclin32:
Thanks for the info.
We can just keep package name unchanged, see [syncback](https://github.com/ScoopInstaller/Extras/blob/master/bucket/syncback.json) for example.

I will remove `checkver` and `autoupdate` in react to this.

PR #5068, opened 2020-11-06
Add Postman Canary entry

Bibz87-CG:
I noticed Postman's canary channel wasn't available in the bucket and I wanted to use it. That's pretty much all there is to it.


Ash258:
Move this PR to https://github.com/ScoopInstaller/Versions

PR #5079, opened 2020-11-08
fontforge: Add binary

BasixKOR:
This enables CLI usages for FontForge.


Ash258 (commented):
What are available parameters?

BasixKOR:
Notably `-script`, which can be used to run Python scripts written for FontForge.

PR #5090, opened 2020-11-11
netch: Fix default configuration

chawyehsu:
Missing initial field settings causes app exception.

### DIFF
--- a/bucket/netch.json
+++ b/bucket/netch.json
@@ -12,7 +12,7 @@
     "extract_dir": "Netch",
     "pre_install": [
         "ensure \"$dir\\data\" | Out-Null",
-        "if (!(Test-Path \"$persist_dir\\data\\settings.json\")) { Set-Content \"$dir\\data\\settings.json\" '{ \"CheckUpdateWhenOpened\": false }' -Encoding Ascii -Force }"
+        "if (!(Test-Path \"$persist_dir\\data\\settings.json\")) { Set-Content \"$dir\\data\\settings.json\" '{\"CheckUpdateWhenOpened\":false,\"Server\":[]}' -Encoding Ascii -Force }"
     ],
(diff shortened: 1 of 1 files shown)


bucket/netch.json
```diff
@@ -12,7 +12,7 @@
     "extract_dir": "Netch",
     "pre_install": [
         "ensure \"$dir\\data\" | Out-Null",
-        "if (!(Test-Path \"$persist_dir\\data\\settings.json\")) { Set-Content \"$dir\\data\\settings.json\" '{ \"CheckUpdateWhenOpened\": false }' -Encoding Ascii -Force }"
+        "if (!(Test-Path \"$persist_dir\\data\\settings.json\")) { Set-Content \"$dir\\data\\settings.json\" '{\"CheckUpdateWhenOpened\":false,\"Server\":[]}' -Encoding Ascii -Force }"
```
Ash258, line 15:
```suggestion
        "if (!(Test-Path \"$persist_dir\\data\\settings.json\")) { Set-Content \"$dir\\data\\settings.json\" '{ \"CheckUpdateWhenOpened\": false, \"Server\": [] }' -Encoding Ascii -Force }"
```

Ash258 (changes requested):

bucket/netch.json
chawyehsu: spaces make no sense, contents of `settings.json` will be overwritten when the app start.

Ash258: There is no reason to provide ugly minified json

chawyehsu: There is no reason to add meaningless spaces to a one-line JSON string which will be overwritten

Ash258: It is not about the final file, but about the maintainability of the manifest.

PR #5093, opened 2020-11-11
playnite: Persist additional folders

Skyline404:
Fixed ?

### DIFF
--- a/bucket/playnite.json
+++ b/bucket/playnite.json
@@ -22,6 +22,9 @@
         ]
     ],
     "persist": [
+    	"Backup",
+    	"Extensions",
+    	"Themes",
         "browsercache",
         "cache",
         "ExtensionsData",
--- a/bucket/playnite.json
+++ b/bucket/playnite.json
@@ -22,9 +22,9 @@
         ]
     ],
     "persist": [
-    	"Backup",
-    	"Extensions",
-    	"Themes",
+        "Backup",
+        "Extensions",
+        "Themes",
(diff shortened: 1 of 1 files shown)


Ash258:
Do you really think it is necessary to open new PR without closing the first one??

Skyline404:
> 
> 
> Do you really think it is necessary to open new PR without closing the first one??

Sry didnt know how to close pr :)

Ash258 (commented):
How do you update Extension from UI?

Skyline404:
> 
> 
> How do you update Extension from UI?

Double click on pext extension

Ash258 (changes requested):
A

There is no update feature of the extensions inside Playnite UI, which effectively prevents it from being persisted as there is no merging support for the persisted folders. So as soon as developers pack new version of the native extensions it will not be updated as it will use old versions. Which will lead to the misbehavior of application and user have to man manually struggle with tracing down the changed extensions and fixing it. Which is unacceptable.

Skyline404:
> 
> 
> There is no update feature of the extensions inside Playnite UI, which effectively prevents it from being persisted as there is no merging support for the persisted folders. So as soon as developers pack new version of the native extensions it will not be updated as it will use old versions. Which will lead to the misbehavior of application and user have to man manually struggle with tracing down the changed extensions and fixing it. Which is unacceptable.
> 
> A

I see, default installer persists custom extensions on update somehow

PR #5095, opened 2020-11-12
x64dbg package is currently broken. Updating to snapshot_2020-11-12_05-12

alecnunn:
Current x64dbg package is not usable as the previous URL/zip file is no longer available. This is an intermediary fix. I will do some more reading on Github's API and see if there is a way to fetch the current existing build/release's hash (I'll be honest, this is my first time contributing to a package management solution, so I am not sure what all to expect, but hope to learn).


Ash258:
0649004b59753d9d3250ad693422cd2d63a1834a

PR #5142, opened 2020-11-18
qimgv-video: Add version 0.9.1

bradenhilton:


bucket/qimgv-video.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": "0.9.1",
+    "description": "Qt5 image viewer",
```
Ash258, line 3:
```suggestion
    "description": "Fast, configurable and easy to use image viewer with video support",
```

bucket/qimgv-video.json
```diff
@@ -0,0 +1,39 @@
...
+    "extract_dir": "qimgv",
+    "bin": "qimgv.exe",
+    "shortcuts": [
+        [
+            "qimgv.exe",
+            "qimgv"
+        ]
+    ],
+    "persist": [
+        "conf",
```
Ash258, line 25: I would persist `cache` also

Ash258 (changes requested):

bucket/qimgv-video.json
bradenhilton: That's fine with me, I will implement the changes now, but the non-video manifest does not do this.

Ash258 (approved):

PR #5151, opened 2020-11-20
qv2ray: Make v2ray dependency optional

kidonng:
The software can run without using `v2ray` at all, so making it a optional dependency.


Ash258 (changes requested):
Remove one line is not enough

image

bucket/qv2ray.json
```diff
@@ -22,15 +22,18 @@
...
+        "        @{",
+        "            'config_version' = 14",
+        "            'logLevel' = 2",
+        "            'kernelConfig' = @{",
+        "                'v2CorePath_win'   = \"$current\\v2ray.exe\"",
+        "                'v2AssetsPath_win' = \"$current\"",
+        "            }",
+        "        } | ConvertTo-Json -Depth 20 | Out-File \"$dir\\config\\Qv2ray.conf\"",
+        "    }",
         "}"
```
Ash258, line 37: Completely ignoring default configuration in case of v2ray is not installed via scoop is not really good idea.

```suggestion
        "if (!(Test-Path \"$persist_dir\\config\\Qv2ray.conf\")) {",
        "    ensure \"$dir\\config\" | Out-Null",
        "    New-Item \"$dir\\config\\QV2RAY_FEATURE_DISABLE_AUTO_UPDATE\" | Out-Null",
        "    $config = @{",
        "        'config_version' = 14",
        "        'logLevel' = 2",
        "    }",
        "    $v2ray_dir = appdir 'v2ray' $global | Join-Path -ChildPath 'current'",
        "    if (Test-Path $v2ray_dir) {",
        "         $config.Add('kernelConfig', @{",
        "             'v2CorePath_win'   = \"$v2ray_dir\\v2ray.exe\"",
        "             'v2AssetsPath_win' = \"$v2ray_dir\"",
        "         })",
        "    }",
        "    ConvertTo-Json $config -Depth 20 | Out-File \"$dir\\config\\Qv2ray.conf\" -Encoding Ascii",
        "}"
```

Ash258 (changes requested):
> The software can run without using v2ray at all, so making it a optional dependency.

Can you provide any screenshot, which support this claim?

bucket/qv2ray.json
kidonng: Qv2ray will generate a default config file if one is not found. This is the default behavior and won't cause issues.

kidonng:
> Can you provide any screenshot, which support this claim?

It's in the docs: https://qv2ray.net/en/plugins/usage.html#what-is-a-plugin

image

PR #5159, opened 2020-11-22
cherrytree: Update to version 0.99.21

### LINKED ISSUE #5145: Cherrytree autoupdate broken
Cherytree for Windows is now 64-bit only, rather than 32-bit only, so checkver can't find the link or hash. Both are still listed on their [homepage](https://www.giuspen.com/cherrytree/#downl).

Right now, Cherrytree is at version 0.39.4 while the latest is 0.99.20.

Vixb1122:
- Closes #5145

### DIFF
--- a/bucket/cherrytree.json
+++ b/bucket/cherrytree.json
@@ -1,27 +1,35 @@
 {
-    "version": "0.39.4",
+    "version": "0.99.21",
     "description": "Hierarchical note taking application, featuring rich text and syntax highlighting.",
     "homepage": "https://www.giuspen.com/cherrytree/",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
So lets completely trash old configuration and force users to start all over. Sounds like good idea.

PR will be closed as invalid if you do not handle the persist migration

Vixb1122:
64-bit programs cannot read 32-bit configurations, and if a 32-bit configuration is found when a configuration is saved, it is deleted (causing the symbolic link to be deleted) and the 64-bit configuration is recreated.
@Ash258

PR #5184, opened 2020-11-25
panwriter: Add version 0.6.5

schance995:
Note, technically pandoc is not required but [homepage](https://panwriter.com/) says
> You also have to install pandoc to export to most formats.

That's why I put pandoc in "depends', without it the app is not so useful.

### DIFF
--- a/bucket/panwriter.json
+++ b/bucket/panwriter.json
@@ -1,0 +1,26 @@
+{
+    "version": "0.6.5",
+    "license": "GPL-3.0-only",
+    "description": "Markdown editor with pandoc integration and paginated preview",
+    "url": "https://github.com/mb21/panwriter/releases/download/v0.6.5/PanWriter-Setup-0.6.5.exe#dl.7z",
+    "homepage": "https://panwriter.com/",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):

bucket/panwriter.json
```diff
@@ -0,0 +1,34 @@
...
+        "64bit": {
+            "url": "https://github.com/mb21/panwriter/releases/download/v0.6.5/PanWriter-Setup-0.6.5.exe#/dl.7z",
+            "hash": "66275b2ccf8981beb4d42ca0e2655c5f667db1b2df6217fefcfbd2e9dc726c4b",
+            "pre_install": [
+                "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\"",
+                "Remove-Item \"$dir\\`$*\", \"$dir\\Uninstall*\" -Recurse"
+            ]
+        }
+    },
+    "bin": "PanWriter.exe",
```
Ash258, line 17: Even though, there is command line support for passing file as argument, the binary is completely unusable from terminal

```suggestion
```

A

Ash258 (changes requested):

Ash258 (approved):

PR #5214, opened 2020-11-29
ungoogled-chromium-portable: fix autoupdate

schance995:
Fixed hardcoded autoupdate url, and updated regex to match for 32 bit specifically since sometimes the release page is 64 bit only.


pratikpc:
Closed as original repo no longer exists

PR #5229, opened 2020-12-02
Set default grafana-server homepath

emanb29:


Ash258 (commented):
Why? Documentation? Anything?

emanb29:
So the program runs

emanb29:
Sorry about the delay @Ash258 -- here's a bit more context:

From `grafana-server.exe --help`:
```$ grafana-server.exe --help
Usage of <path\to>\grafana-server.exe:
  -config string
        path to config file
  -convey-json
        When true, emits results in JSON blocks. Default: 'false'
  -convey-silent
        When true, all output from GoConvey is suppressed.
  -convey-story
        When true, emits story output, otherwise emits dot output. When not provided, this flag mirrors the value of the '-test.v' flag
  -homepath string
        path to grafana install/home path, defaults to working directory
  -packaging string
        describes the way Grafana was installed (default "unknown")
  -pidfile string
        path to pid file
  -profile
        Turn on pprof profiling
  -profile-port uint
        Define custom port for profiling (default 6060)
  -tracing
        Turn on tracing
  -tracing-file string
        Define tracing output file (default "trace.out")
  -v    prints current version and exits
```

When run as-is with no parameters and the current installation:

```
Grafana-server Init Failed: Could not find config defaults, make sure homepath command line parameter is set or working directory is homepath
```

By adding 'homepath' to the appropriate scoop-managed directory (`scoop/apps/grafana/<version>`), the grafana server runs as expected, eg using config files from `scoop/apps/grafana/<version>/conf`

rashil2000:
Hey @emanb29, is this PR still valid? 

I can help you get it merged.

emanb29:
To be fully honest, I'm not sure. I moved away from grafana and I'm tied down with other commitments for the near future (thru January or so) 

There's also the broader question of whether this approach to config/arguments is the one scoop follows in general, which I'm not certain of. For example, I believe the kafka target doesn't work "out of box" either, so maybe I should infer that the general philosophy is "make the binary accessible but not necessarily runnable"?

rashil2000:
Can you explain what you mean by _doesn't work "out of box"_?

If it doesn't work, no use of adding a manifest, right? 

There are plenty of manifests that specify predetermined binary arguments. If they are required in order to function, we will add them.

emanb29:
By "out of box", I meant (lacking precise commands because I'm responding from my phone):
- install a manifest (eg, grafana) with scoop
- invoke the main application with no additional arguments (eg, grafana-server)

When doing this, at least as of when I opened this PR, the application fails to run (for example, printing the usage helpdoc copied above).

If I remember correctly, the kafka manifest (with the zookeeper-server-start and kafka-server-start shims it installs) has similar behavior, despite a default config being present in the artefacts installed by scoop.

There may be other examples, but grafana and kafka are the two I have most recently worked around.

rashil2000:
Okay, then please reopen the PR, we can merge it.

rashil2000:
@emanb29 can you resolve the conflicts?

PR #5254, opened 2020-12-05
chipmunk: add executable chipmonk.exe to allow run from command line

Slach:


Ash258:
There is no command line support

image

PR #5287, opened 2020-12-11
Update SweetHome3d to 6.4.2

jovandeginste:


jovandeginste:
Any clue as to why the check fails?

As an FYI: the previous manifest would not succeed; for some reason the download always stalls after a few (50?) MB; even downloading the 7z-file through Firefox didn't succeed. This is why I looked into the details. Downloading/installing v6.4.2 worked without problem.

pratikpc:
@jovandeginste SweetHome3D is already at 6.6

If there are new updates to be made, please create a new PR

jovandeginste:
Okay

PR #5302, opened 2020-12-14
liquidsoap: Add version 1.4.3

earnestma:


bucket/liquidsoap.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "1.4.3",
+    "description": "Audio and video streaming language",
+    "homepage": "https://www.liquidsoap.info",
+    "license": "GPL-2.0",
```
Ash258, line 5: Invalid SPX identifier

bucket/liquidsoap.json
```diff
@@ -0,0 +1,22 @@
...
+            "extract_dir": "liquidsoap-v1.4.3-win64"
+        }
+    },
+    "bin": "liquidsoap.exe",
+    "checkver": {
+        "github": "https://github.com/savonet/liquidsoap"
+    },
+    "autoupdate": {
+        "url": "https://github.com/savonet/liquidsoap/releases/download/v$version/liquidsoap-v$version-win64.zip",
+        "extract_dir": "liquidsoap-v$version-win64"
```
Ash258, line 19: 64 bit

bucket/liquidsoap.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "1.4.3",
+    "description": "Audio and video streaming language",
```
Ash258, line 3: Not clear description

Ash258 (changes requested):

bucket/liquidsoap.json
earnestma: Their license didn't specify whether it was GPL-2.0-only or GPL-2.0-or-later, so it's now GPL-2.0-only unless they change/ specify it.

earnestma: Nope. On second thought it makes more sense to include later versions.

> This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 2 of the License, or
    (at your option) any later version.

Ash258 (approved):

PR #5305, opened 2020-12-14
cppcryptfs: Add version 1.4.3.5

akuntsch:
Not sure if a nonportable dependency is allowed in the extras bucket.

No `autoupdate` because it does not work for multiple urls at the moment (https://github.com/lukesampson/scoop/pull/3518).

### DIFF
--- a/bucket/cppcryptfs.json
+++ b/bucket/cppcryptfs.json
@@ -1,0 +1,37 @@
+{
+    "version": "1.4.3.5",
+    "description": "Implementation of the gocryptfs encrypted overlay filesystem in C++ for Windows.",
+    "homepage": "https://github.com/bailey27/cppcryptfs",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": [
+                "https://github.com/bailey27/cppcryptfs/releases/download/1.4.3.5/cppcryptfs.exe",
+                "https://github.com/bailey27/cppcryptfs/releases/download/1.4.3.5/cppcryptfsctl.exe"
+            ],
+            "hash": [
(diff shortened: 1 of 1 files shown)


bucket/cppcryptfs.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "bin": [
+        "cppcryptfsctl.exe",
+        "cppcryptfs.exe"
+    ],
+    "shortcuts": [
+        ["cppcryptfs.exe", "cppcryptfs"]
+    ],
+    "checkver": "github",
+    "depends": "dokany-np"
```
Ash258, line 36: Completely agains scoop's philosophy to depends on non portable application from third party bucket.

Close PR if it does not work without.

Ash258 (changes requested):

rasa:
IMO, this package would best be added to the bucket it has a dependency on, in this case the [non-portable](https://github.com/TheRandomLabs/scoop-nonportable) bucket.

bucket/cppcryptfs.json
akuntsch: It does for sure not work without dokany. Could also use `suggest` instead of `depends` and let the user handle the installation of dokany himself, either via dokany-np or some other manual method.

akuntsch:
> IMO, this package would best be added to the bucket it has a dependency on, in this case the [non-portable](https://github.com/TheRandomLabs/scoop-nonportable) bucket.

That's fine with me as well.

akuntsch:
I will create a PR for the non-portable bucket.

PR #5319, opened 2020-12-16
taskbarx: Update to version 1.6.3.0

Lunchb0ne:
This also accounts for the split in the architecture of the new versions of the app. The auto-update URL has also been added to account for the different versions.

### DIFF
--- a/bucket/taskbarx.json
+++ b/bucket/taskbarx.json
@@ -1,10 +1,18 @@
 {
-    "version": "1.6.2.0",
+    "version": "1.6.3.0",
     "description": "Center Windows taskbar icons with a variety of animations and options.",
     "homepage": "https://chrisandriessen.nl/web/taskbarx/",
     "license": "MIT",
-    "url": "https://github.com/ChrisAnd1998/TaskbarX/releases/download/1.6.2.0/TaskbarX_1.6.2.0.zip",
-    "hash": "eacbd14a13f3abbbddf48aeb7d834411a843d7aa700d27d5718c495b182db520",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/ChrisAnd1998/TaskbarX/releases/download/1.6.3.0/TaskbarX_1.6.3.0_x64.zip",
+            "hash": "959cb3f56878fd41e79e2f7c62c053d1b490f5fd46d356d59f9f6168db622b12"
+        },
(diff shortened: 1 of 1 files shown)


Lunchb0ne:
Also, on a side note, is the regex under the checkver needed? I saw the code for the script that does the version checking and seems like it substitutes the regex with another one for getting the latest GitHub release version, if github is specified as the target

bucket/taskbarx.json
```diff
@@ -1,10 +1,18 @@
 {
-    "version": "1.6.2.0",
+    "version": "1.6.3.0",
     "description": "Center Windows taskbar icons with a variety of animations and options.",
     "homepage": "https://chrisandriessen.nl/web/taskbarx/",
```
Ash258, line 4:
```suggestion
    "homepage": "https://chrisandriessen.nl/taskbarx",
```

bucket/taskbarx.json
```diff
@@ -17,6 +25,13 @@
         "regex": "TaskbarX_([\\d.]+)\\.zip"
```
Ash258, line 25:
```suggestion
```

Ash258 (changes requested):

bucket/taskbarx.json
```diff
@@ -1,10 +1,18 @@
...
+        "64bit": {
+            "url": "https://github.com/ChrisAnd1998/TaskbarX/releases/download/1.6.3.0/TaskbarX_1.6.3.0_x64.zip",
+            "hash": "959cb3f56878fd41e79e2f7c62c053d1b490f5fd46d356d59f9f6168db622b12"
+        },
+        "32bit": {
+            "url": "https://github.com/ChrisAnd1998/TaskbarX/releases/download/1.6.3.0/TaskbarX_1.6.3.0_x86.zip",
+            "hash": "2c2e7c121e278de3b9bc8dce6c9701394b87a1b3279f5b8552d947d1733d3f20"
+        }
+    },
     "bin": "TaskbarX.exe",
```
Ash258, line 16: I do not think this is good idea.

All configurations are saved into the scheduled task parameter field, so launching it without defined parameters is not desired. Or rather, launching it from any other place than task or configurator does not make sense

```suggestion
```

Ash258 (changes requested):

PR #5321, opened 2020-12-17
kdeconnect: Add version 1.4

akuntsch:
Closes https://github.com/lukesampson/scoop-extras/issues/3908


akuntsch:
This might not really work because the artifacts in their Jenkins are not available forever. It seems that they only keep the last 10 builds.

Ash258:
There is no reason to have this in extras if there is not stable release

> An early release of KDE Connect on Windows is

Issue #5084, opened 2020-11-10
PowerToys have not been updated for a long time.


qcts33:
PowerToys have released version 0.25.0, but only the 0.21.0 is available in the bucket.

https://github.com/microsoft/PowerToys/releases
https://github.com/lukesampson/scoop-extras/commits/master/bucket/powertoys.json

gitolicious:
That's probably because the installer was changed from an MSI installer to EXE with version 0.23. Looks like the new one is no longer "portable", but that would need to be checked.
`PowerToysSetup-0.25.0-x64.exe --silent --skip_dotnet_install --no_start_pt --log_level error`

rashil2000:
FYI, [v0.29](https://github.com/microsoft/PowerToys/releases/tag/v0.29.0), released today, allows custom location flags.

gitolicious:
Thanks for the info, I will check if this makes it "scoopable" again.

gitolicious:
@Sn0wCrack was quicker than me, happy to see this working again!

PR #5391, opened 2020-12-30
powertoys: update to version 0.29.0

Sn0wCrack:
This uses the new --install_dir argument to install the app into the correct scoop directory.

A "proper" uninstall requires invocation of .uninstall() on the WMI object. however I have found no issues with uninstalling and re-installing without the use of this, however I believe it should be cleaner using this method.

This should hopefully solve #5084

### DIFF
--- a/bucket/powertoys.json
+++ b/bucket/powertoys.json
@@ -1,26 +1,28 @@
 {
-    "version": "0.21.1",
+    "version": "0.29.0",
     "description": "A set of utilities for power users to tune and streamline their Windows experience for greater productivity.",
     "homepage": "https://github.com/microsoft/PowerToys",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/microsoft/PowerToys/releases/download/v0.21.1/PowerToysSetup-0.21.1-x64.msi",
-            "hash": "ef63a70fc1c4b718410d3129d1691ab3814b4f7d1cea402e6507b0133ca2f09c"
+            "url": "https://github.com/microsoft/PowerToys/releases/download/v0.29.0/PowerToysSetup-0.29.0-x64.exe",
+            "hash": "bc635394b092a11958b2c6af66dbfe0f8d1c2dd52ce758d5963125061241ac64"
         }
     },
-    "extract_dir": "PowerToys",
-    "shortcuts": [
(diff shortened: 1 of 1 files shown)


Sn0wCrack:
I've noticed that version 0.29.3 got pushed for Powertoys, I'll update this PR tonight to be that version.

Sn0wCrack:
This now targets Powertoys 0.29.3

Jogai:
What about the preview? Or should that go into https://github.com/ScoopInstaller/Versions?

gitolicious:
Works like a charm, thanks!

Jogai:
@gitolicious I dont see the 0.29.x version showing up?

gitolicious:
As the pull request is still open you will need to switch to the source branch or manually copy the file to your local repo for now.

Jogai:
Its ok, I'm using the preview anyway. Why are PR's open for a long time?

gitolicious:
Who knows. Maybe @Ash258 is just too busy.

Ash258:
I am not going to merge installers usage, which flood users systems with own shortcuts and other junks. Look for someone else who will merge it.

gitolicious:
Well, at least we know the reason now... Does it mean we need to move it to a bucket with non-portable apps?

foisonocean:
@Ash258 Hi Ash258, is there any documentations about which applications are acceptable in this bucket?

Jogai:
https://github.com/TheRandomLabs/scoop-nonportable has the latest powertoys

ahmadalli:
I think thus is related and would be great to fix

crutkas:
Can we shift this to auto-update based off the github?  https://scoop-docs.now.sh/docs/concepts/App-Manifest-Autoupdate.html#adding-checkver-to-a-manifest

gitolicious:
@crutkas, what do you mean by that? `checkver` and `autoupdate` both point to Github.

gitolicious:
I think the alternative approach by @7crabs in https://github.com/lukesampson/scoop-extras/pull/5669 is worth a look.

MunsterPlop:
> I think the alternative approach by @7crabs in #5669 is worth a look.

Indeed, this looks like it would solve most of what keeps this from being merged.

PR #5396, opened 2020-12-31
jellyfin: Add version 10.6.4

mertd:
Hi,

Jellyfin is a open source fork of Emby, a media server similar to Plex.
This is my first manifest, so I am happy about any feedback I can get.

Also, please squash and merge these changes (if you don't do this anyway) when ready, as my commits are small and many.

### DIFF
--- a/bucket/jellyfin.json
+++ b/bucket/jellyfin.json
@@ -1,9 +1,9 @@
 {
-    "version": "10.7.0-rc1",
-    "url": "https://repo.jellyfin.org/releases/server/windows/stable/combined/jellyfin_10.7.0~rc1.zip",
+    "version": "10.6.3",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
Switch to portable https://github.com/Ash258/Scoop-Ash258/blob/master/bucket/Jellyfin.json

mertd:
I am not sure I understand what you are requesting exactly. The manifest in this PR is using the same URL (and release of Jellyfin) as the one in your personal bucket. Considering that your manifest is more detailed, I do not mind closing this PR in favour of the manifest in your bucket.

If you are referring to the "more portable" release of Jellyfin than this one, the one that is actually named "portable", I am afraid that it does not contain a runnable jellyfin.exe file and would introduce an additional dependency on the .net core runtime. Would that still be preferrable to this release, considering that the app itself (although not the data) is portable?

Ash258 (approved):

PR #5400, opened 2020-12-31
add windows-terminal-quake

Deliganli:
windows-terminal-quake is a companion to windows-terminal. Enabling features such as opacity, quake-style dropdown and such.

disclaimer: I am not the maintainer of the program, I did this file and it works for me.


Ash258:
I am strongly against adding such applications in any of official buckets.

Set of few random functions, which extend application, which is under active development is completely nonsense to maintain here. It will introduce bad precedence for adding next junk forks of applications, which will not be used by anyone.

All of these provided by this are already requested in windows terminal.

PR #5409, opened 2021-01-02
imhex: Add version 1.5.0

TheRandomLabs:


Ash258 (approved):

bucket/imhex.json
```diff
@@ -0,0 +1,27 @@
...
+    "description": "Hex editor",
+    "homepage": "https://github.com/WerWolv/ImHex",
+    "license": "GPL-2.0-only",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/WerWolv/ImHex/releases/download/v1.5.0/ImHex_1.5.0_Windows.zip",
+            "hash": "79e5dfcf0b5a997571180574a3d66fe469923139b70e933ab990874f84fbe391"
+        }
+    },
+    "bin": "ImHex.exe",
```
Ash258, line 12: This is not good idea.

When you start it from command line it will create ini file in current directory. Completely nonsensual flooding of random ini files everywhere is not something, which would be acceptable.

The executable also block the whole terminal, and the window cannot be closed when executed from command line so it is completely unusable.

image

```suggestion
```

Ash258 (changes requested):
Persist

image

Ash258 (approved):

PR #5431, opened 2021-01-04
streamlink: Update to version 2.0.0

Zagitta:
I don't entirely understand what was going on previously with concatenating python and stuff together but the 2.0.0 zip file includes the needed .exe files so I have removed the concatenation.

### DIFF
--- a/bucket/streamlink.json
+++ b/bucket/streamlink.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.7.0",
+    "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
What is `streamlinkw.exe` for?

Zagitta:
Seems to be an alternative version built with `pythonw.exe` rather than `python.exe`, as far as I can tell the only difference is whether or not a terminal window pops up when you use it: https://github.com/streamlink/streamlink/pull/2326

PR #5453, opened 2021-01-06
nextdns-gui: Add version 2.0.1

benjaminwolkchen:


Ash258:
Incompatible with scoop philosophy. It needs tap driver installation and windows service running. All of this is done by installer, and when you extract it user needs to handle it on their own, which is unacceptable. Running this installer is unacceptable too. 

Try non-portable bucket

PR #5455, opened 2021-01-06
Update fritzing.json

lucasiro:


Ash258:
This is directly against vendors interest

image

PR #5470, opened 2021-01-08
winauth: portable configuration

Hxppdv:
put configuration file in the persist folder, make it portable


Ash258 (changes requested):
Notify user about this change and do the migration from non portable

Ash258 (changes requested):
Notify user about this change and do the migration from non portable

Ash258 (changes requested):
Notify user about this change and do the migration from non portable

bucket/winauth.json
```diff
@@ -6,6 +6,19 @@
     "url": "https://github.com/winauth/winauth/releases/download/3.6.2/WinAuth-3.6.2.zip",
     "hash": "sha512:8288a970946c149382c1d8919905fda5381d04f24a4ceed1ec4f3fc7fda0a7c8f5c1fc062b7b312dca55531eecf04327ca0743afbb82ff08acedf4c7c9c4ed1f",
     "bin": "WinAuth.exe",
+    "pre_install": [
+        "if(!(Test-Path(\"$persist_dir\\winauth.xml\"))) { New-Item \"$dir\\winauth.xml\" -type file | Out-Null ",
+        "    $env:LOCALAPPDATA, $env:APPDATA | ForEach-Object {",
```
Ash258, line 11: Why both appdatas?

> If your configuration file (winauth.xml, normally stored in your Windows roaming profile

https://winauth.github.io/winauth/

Ash258 (changes requested):

bucket/winauth.json
Hxppdv: Oh, I didn't figure out where the directory was, but now it looks like you should just need APPDATA

bucket/winauth.json
```diff
@@ -5,11 +5,23 @@
     "license": "GPL-3.0-only",
     "url": "https://github.com/winauth/winauth/releases/download/3.6.2/WinAuth-3.6.2.zip",
     "hash": "sha512:8288a970946c149382c1d8919905fda5381d04f24a4ceed1ec4f3fc7fda0a7c8f5c1fc062b7b312dca55531eecf04327ca0743afbb82ff08acedf4c7c9c4ed1f",
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\winauth.xml\")) {",
+        "    New-Item \"$dir\\winauth.xml\" | Out-Null ",
+        "    $env:APPDATA | ForEach-Object {",
```
Ash258, line 11: Why do you need foreach for one item??

Ash258 (changes requested):

Ash258:
For real??

Ash258:
A

Ash258:
Link do not work

A

Ash258:
No reason to persist. Reverting change

rasa:
Thanks, @Ash258, for catching this. Sorry for the extra work.

Hxppdv:
@rasa @Ash258 
> As a precaution, WinAuth also makes an automatic encrypted backup of each newly created authenticator into your registry, or .xml file if running in portable mode.
[site](https://winauth.github.io/winauth/lost-forgotten-password.html)

I apologize for the extra work caused by the lack of adequate testing.
But according to the winauth website, persist is still necessary because the authenticator is stored inside winauth.xml

> Link do not work
> 
> A

As for the problem of link not working, I think it's caused by something or some feature of winauth that I don't know.
I observed the link situation, after changing some settings in winauth's settings, winauth seems to create a new winauth.xml file, and the old file is named winauth.xml.bak (is this a feature of scoop?). 

Apparently persist/winauth.xml is linked to winauth.xml.bak instead of winauth.xml, which is what causes persist not to work, I don't know if it's a problem with winauth or what, I hope to find a workaround to solve this problem.

PR #5477, opened 2021-01-09
openshot: Use native executable as shortcut target

FTOH:
Fix: https://github.com/lukesampson/scoop-extras/issues/5476

`launch-win.bat` uses for debug, see https://github.com/OpenShot/openshot-qt/blob/8dffb2cd507e93576d24b671927e247614264c02/freeze.py#L185-L186

### DIFF
--- a/bucket/openshot.json
+++ b/bucket/openshot.json
@@ -22,10 +22,8 @@
     "bin": "openshot.ps1",
     "shortcuts": [
         [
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
Remove lines 21-22

FTOH:
@Ash258 Done

PR #5482, opened 2021-01-11
Replaces ACLs with default inherited ACLs for rolloutfile.tv13

igitur:
When running teamviewer, it changes the ACLs of `rolloutfile.tv13` so that a normal uninstall isn't possible with Scoop/Shovel. 

More info at https://community.teamviewer.com/English/discussion/49721/teamviewerportable-rolloutfile-tv13-only-adminrights

This PR adds a `pre_uninstall` property (supported only by Shovel), so that it replaces ACLs with default inherited ACLs for rolloutfile.tv13, and then the file isn't locked when trying to uninstall.

cc @Ash258


Ash258:
Nonsense to open PR with feature, which is not implemented in master.

igitur:
Such a charming guy.

PR #5494, opened 2021-01-13
Update brave.json

jeremy886:
This Brave browser is very outdated so I updated it to the newest available release version. Unfortunately, I am not sure how to update the "autoupdate". I figured out that if it says "Release vXX.YY.ZZ" then it is the right one here: https://github.com/brave/brave-browser/releases?after=v1.20.71
However, I can't code this pattern in the json file.

### DIFF
--- a/bucket/brave.json
+++ b/bucket/brave.json
@@ -8,11 +8,11 @@
     },
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


Ash258:
Not working. These are debug files, not real application.

https://github.com/lukesampson/scoop-extras/commit/e5d78a5270b7f699563da4ede217d31db52b159b#commitcomment-41583111

https://github.com/brave/brave-browser/issues/1444#issuecomment-427574506

PR #5500, opened 2021-01-14
add UPYUN Storage Command Tool

huangnauh:
https://github.com/upyun/upx


Ash258:
Applications without english homepage are not accepted in official buckets.

PR #5523, opened 2021-01-17
marktext-portable: Add version 0.16.3

ouyangwanxi:


Ash258 (changes requested):
Why do you add new manifest instead of simply migrating the data??

ouyangwanxi:
> 为什么要添加新清单而不是简单地迁移数据？

mistake

PR #5538, opened 2021-01-21
networkmanager: Add version 2021.1.0

diggit:
Addition of [NETworkManager](https://borntoberoot.net/NETworkManager/) - graphical utility for network troubleshooting and discovery.

Depends on dotnet desktop runtime 5.0. AFAIK the only package providing this runtime in dotnet-sdk which is overkill, but works...


bucket/networkmanager.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "2021.1.0",
+    "description": "Tool for managing networks and network troubleshooting",
+    "homepage": "https://borntoberoot.net/NETworkManager/",
+    "license": "GPL-3.0",
+    "suggest": {
+        "vcredist": "dotnet-sdk"
+    },
+    "##": "only desktop runtime from dotnet is required",
```
Ash258, line 9:
```suggestion
```

bucket/networkmanager.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "2021.1.0",
+    "description": "Tool for managing networks and network troubleshooting",
+    "homepage": "https://borntoberoot.net/NETworkManager/",
+    "license": "GPL-3.0",
```
Ash258, line 5: Invalid spdx identifier

bucket/networkmanager.json
```diff
@@ -0,0 +1,24 @@
...
+    "hash": "425517A996CE52BB2ADFF05D02537817C954329402AAE9F05773B36C61997E71",
+    "bin": "NETworkManager.exe",
+    "extract_dir": "NETworkManager",
+    "shortcuts": [
+        [
+            "NETworkManager.exe",
+            "NETworkManager"
+        ]
+    ],
+    "checkver": "github",
```
Ash258, line 20: This cannot work since hompeage is not set to github repository

bucket/networkmanager.json
```diff
@@ -0,0 +1,24 @@
...
+    "autoupdate": {
+        "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/$version/NETworkManager_$version_Archiv.zip"
```
Ash258, line 22: Add hash extraction

bucket/networkmanager.json
Ash258: This is not true.

image

bucket/networkmanager.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": "2021.1.0",
+    "description": "Tool for managing networks and network troubleshooting",
+    "homepage": "https://borntoberoot.net/NETworkManager/",
+    "license": "GPL-3.0",
+    "suggest": {
+        "vcredist": "dotnet-sdk"
+    },
+    "##": "only desktop runtime from dotnet is required",
+    "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/2021.1.0/NETworkManager_2021.1.0_Archiv.zip",
```
Ash258, line 10: 64bit only
https://github.com/BornToBeRoot/NETworkManager/releases/tag/2021.1.0

> Migrated to x64 (The x86 version has to be uninstalled manually)!

Ash258 (changes requested):

bucket/networkmanager.json
diggit: This comment was meant for dotnet-sdk only. Honestly I am not even sure whit brought WebView2 to ma PC.

bucket/networkmanager.json
```diff
@@ -1,13 +1,17 @@
 {
     "version": "2021.1.0",
     "description": "Tool for managing networks and network troubleshooting",
-    "homepage": "https://borntoberoot.net/NETworkManager/",
-    "license": "GPL-3.0",
+    "homepage": "https://github.com/BornToBeRoot/NETworkManager",
```
Ash258, line 4: Use previous homepage

Ash258 (changes requested):

bucket/networkmanager.json
diggit: Ok fixed, I wrote version extraction regex to extract it from github release page, because github checkver method requires github homepage as you noted earlier. Though, I am not sure how flexible should such regexes be.

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
...
+    "extract_dir": "NETworkManager",
+    "shortcuts": [
+        [
+            "NETworkManager.exe",
+            "NETworkManager"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/BornToBeRoot/NETworkManager/releases/latest",
+        "regex": "<a href=\"/BornToBeRoot/NETworkManager/releases/tag/([0-9\\.]+)\">\\1</a>"
```
Ash258, line 25:
```suggestion
        "github": "https://github.com/BornToBeRoot/NETworkManager"
```

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
...
+    "suggest": {
+        ".NET": "dotnet-sdk"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/2021.1.0/NETworkManager_2021.1.0_Archiv.zip",
+            "hash": "425517a996ce52bb2adff05d02537817c954329402aae9f05773b36c61997e71"
+        }
+    },
+    "bin": "NETworkManager.exe",
```
Ash258, line 15: What are available parameters?

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "2021.1.0",
+    "description": "Tool for managing networks and network troubleshooting",
+    "homepage": "https://borntoberoot.net/NETworkManager/",
+    "license": "GPL-3.0-or-later",
+    "suggest": {
+        ".NET": "dotnet-sdk"
```
Ash258, line 7: why SDK when you need runtime?

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
...
+            "NETworkManager.exe",
+            "NETworkManager"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/BornToBeRoot/NETworkManager/releases/latest",
+        "regex": "<a href=\"/BornToBeRoot/NETworkManager/releases/tag/([0-9\\.]+)\">\\1</a>"
+    },
+    "autoupdate": {
+        "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/$version/NETworkManager_$version_Archiv.zip",
```
Ash258, line 28: 64bit

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
...
+        "hash": {
+            "url": "https://github.com/BornToBeRoot/NETworkManager/releases/tag/$version",
+            "find": "<td><a href=\"$url\">Archiv</a></td>(?:\\n)?<td><code>([a-fA-F0-9]{64})</code></td>(?:\\n)?</tr>"
```
Ash258, line 31: `(?sm)$basename.*?>$sha256</` should be enough

Test before changing

bucket/networkmanager.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "2021.1.0",
+    "description": "Tool for managing networks and network troubleshooting",
+    "homepage": "https://borntoberoot.net/NETworkManager/",
+    "license": "GPL-3.0-or-later",
```
Ash258, line 5: Where is explicitly stated it is or later?

Ash258 (changes requested):

bucket/networkmanager.json
diggit: Not sure what do you mean. NETworkManager is graphical app AFAIK does not accept any command line arguments.

bucket/networkmanager.json
diggit: Because [`dotnet`](https://github.com/ScoopInstaller/Main/blob/master/bucket/dotnet.json#L6) manifest suggests to use `dotnet-sdk` instead, which also packs runtime. My removed comment was just about that...

bucket/networkmanager.json
Ash258: Then why do you add binary??

bucket/networkmanager.json
Ash258: Why not suggest just runtime??

bucket/networkmanager.json
diggit: :facepalm: I looked up correct string and coped wrong one. Of course it is v3 only.

bucket/networkmanager.json
diggit: Well because user (at least I) would search for `dotnet` in scoop official buckets and would install outdated version, which is not sufficient.

bucket/networkmanager.json
diggit: Because I blindly coped from other manifest not fully understanding contents. Will fix that.

bucket/networkmanager.json
diggit: Not sure what do you mean here.
Tried:
```patch
-        "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/$version/NETworkManager_$version_Archiv.zip",
+        "64bit": {
+            "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/$version/NETworkManager_$version_Archiv.zip"
+        },
```

, but it causes scoop to throw error:

`ERROR Exception calling ".ctor" with "1" argument(s): "Invalid URI: The URI is empty."`

Probably because `32bit` part is missing...

bucket/networkmanager.json
Ash258: So rather fill end-user's system with trash 200MB something, which they will never use instead of updating the runtime manifest?

bucket/networkmanager.json
```diff
@@ -0,0 +1,32 @@
...
+        "64bit": {
+            "url": "https://github.com/BornToBeRoot/NETworkManager/releases/download/2021.1.0/NETworkManager_2021.1.0_Archiv.zip",
+            "hash": "425517a996ce52bb2adff05d02537817c954329402aae9f05773b36c61997e71"
+        }
+    },
+    "bin": "NETworkManager.exe",
+    "extract_dir": "NETworkManager",
+    "shortcuts": [
+        [
+            "NETworkManager"
```
Ash258, line 19: What??

Ash258 (changes requested):

bucket/networkmanager.json
Ash258: https://github.com/lukesampson/scoop/blob/master/schema.json

bucket/networkmanager.json
diggit: By binary, I understood  `*.exe`. So guessing by you reaction, you meant to remove shortcut without `.exe` suffix?
I am long term linux user and now I am forced to use to use Windows, but catching up is not easy...

diggit: Now I get it, you meant that "bin" line.

bucket/networkmanager.json
diggit: Fixed according to schema in bc6268e.

bucket/networkmanager.json
diggit: Yeah, it's not great, but I did not have time to fix that manifest (yet). IDK how to handle exe installers and there does not seem to be official archive version of installer.

diggit: Only ASP.NET core and non desktop runtime (CLI only apps) seems to have zip archives. https://dotnet.microsoft.com/download/dotnet/5.0 . Looking into their contents, it should be possible to strip down sdk zip from non runtime components, but I can't see any easy way with scoop.

Ash258: https://github.com/lukesampson/scoop-extras/blob/master/bucket/windowsdesktop-runtime.json

diggit: Well, that is unfortunate name. Yeah it is inherited from downloaded archive name but I'd still search for `dotnet` when I need to instal dotnet runtime. That is also reason why I was not aware of it.

diggit: Would you be OK with new [dotnet-desktop-runtime manifest](https://gist.github.com/diggit/211b218071e759ed6c4b4402db256baf)? It has working auto update, it is installed systemlessly (in scoop apps) and IMO has better name.

diggit: To overcome absence of archive for dotnet dekstop runtime, it uses SDK archive and removes all not necessary contents in install script. tested with NETworkManager and no problems occurred.

Ash258: No. You still need to download 200mb instead of like 50, which does not make sense.

Also all runtimes will be installed as vcredists and current desktop

diggit: Well, so what should I put into suggest section for this manifest?

Ash258: Updated windowsdesktop-runtime with version 5

image

PR #5541, opened 2021-01-23
mousejiggler: Update to version 2.0.13

cerebrate:
Update to 2.0 release.

### DIFF
--- a/bucket/mousejiggler.json
+++ b/bucket/mousejiggler.json
@@ -1,10 +1,18 @@
 {
     "version": "2.0.9",
-    "description": "A very simple piece of software whose sole function is to \"fake\" mouse input to Windows, and jiggle the mouse pointer back and forth.",
+    "description": "A very simple piece of software whose sole function is to \"fake\" mouse input to Windows, and jiggle the mouse$
     "homepage": "https://github.com/cerebrate/mousejiggler/",
     "license": "MS-PL",
-    "url": "https://github.com/cerebrate/mousejiggler/releases/download/2.0.9/MouseJiggler.zip",
-    "hash": "1123c73a786fd2c79d629a40e9d42a2858a1207fe19c944245d725e08b652dc8",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/cerebrate/mousejiggler/releases/download/2.0.12/MouseJiggler-x64.zip",
+            "hash": "1984401ad9b9bdb0fe906713b220a0f0472ad399a599c6cb87d7ed5d1461df30"
+        },
+        "32bit": {
+            "url": "https://github.com/cerebrate/mousejiggler/releases/download/2.0.12/MouseJiggler-x86.zip",
+            "hash": "f862f1c7fcc44472b64ac6b1c09b7d09391fe99b466a91d632f71432ca2b5d45"
+        }                                                                                                                               
+    },
     "bin": "MouseJiggler.exe",
     "shortcuts": [
         [
@@ -14,6 +22,13 @@
     ],
     "checkver": "github",
     "autoupdate": {
-        "url": "https://github.com/cerebrate/mousejiggler/releases/download/$version/mousejiggler.zip"
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/cerebrate/mousejiggler/releases/download/$version/mousejiggler-x64.zip"
+            },
+            "32bit": {
+                "url": "https://github.com/cerebrate/mousejiggler/releases/download/$version/mousejiggler-x86.zip"
+            }
+        }
     }
 }
--- a/bucket/mousejiggler.json
+++ b/bucket/mousejiggler.json
@@ -1,10 +1,10 @@
 {
-    "version": "2.0.7",
+    "version": "2.0.9",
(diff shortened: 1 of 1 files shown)


bucket/mousejiggler.json
```diff
@@ -1,19 +1,20 @@
 {
-    "version": "1.8.42",
+    "version": "2.0.7",
     "description": "A very simple piece of software whose sole function is to \"fake\" mouse input to Windows, and jiggle the mouse pointer back and forth.",
-    "homepage": "https://github.com/cerebrate/mousejiggler",
+    "homepage": "https://github.com/cerebrate/mousejiggler/",
```
Ash258, line 4: Why do you "fix" something, which is not broken??

image

bucket/mousejiggler.json
```diff
@@ -1,19 +1,20 @@
...
     "license": "MS-PL",
-    "url": "https://github.com/cerebrate/mousejiggler/releases/download/1.8.42/mousejiggler-1.8.42-net.zip",
-    "hash": "368df214d40bc26d8961ccdefcd43dd9303471c4338cfca13681ff0cbd1fd526",
-    "bin": "MouseJiggle.exe",
+    "url": "https://github.com/cerebrate/mousejiggler/releases/download/2.0.7/MouseJiggler.zip",
```
Ash258, line 6: It is 64bit only

image

Ash258 (changes requested):

bucket/mousejiggler.json
cerebrate: Just because it works doesn't mean it's not broken. 🙂

That URL points to a (directory-like) collection of resources, not a (file-like) single resource, so it should have a trailing slash. This is, admittedly, a fairly pedantic interpretation of web standards that only affects relative URLs when parsing and most web servers are sloppy enough to let me get away with omitting the /, and yet. It's still wrong.

bucket/mousejiggler.json
cerebrate: Fixed in 2.0.9.

cerebrate:
Fixed _properly_ in 2.0.12.

bucket/mousejiggler.json
```diff
@@ -1,19 +1,34 @@
 {
-    "version": "1.8.42",
+    "version": "2.0.9",
```
Ash258, line 2: 2.0.13

Ash258 (changes requested):
32bit version on 64bit

bucket/mousejiggler.json
cerebrate: Fixed the version numbers.

Wrt the error shown, as I understand it it's possible to install the .NET 5 runtime on x64 without the x86 runtime included. If it wasn't included, that message is expected.

I retested here with a fresh x64 VM, a default .NET 5 Desktop runtime install, and the x86 download, and don't get it.

Ash258: So there is dotnet dependency and nowhere in manifest it is mentioned??

cerebrate: There isn't a package for it to depend on, and since the .NET 5 runtime has to be installed globally, it wouldn't make a good Scoop package anyway.

(In theory I could build MJ as a self-contained app that would eliminate the dependency, but that turns a 1.4 MB executable into a 126 MB executable, and no-one wants that.)

Ash258: https://raw.githubusercontent.com/lukesampson/scoop-extras/master/bucket/windowsdesktop-runtime.json

cerebrate: Huh. That did not show up when I searched for it. Curious.

On the other hand, that's also for 3.1, not 5.

I can make one for 5, readily enough, but I've got another consideration to bring up. The reason I'm setting up a Scoop install for MJ is because of an issue one of my users brought up that they can't install it using Chocolatey if they don't have admin rights.

Looking at that package, it requires admin rights to install. Seems to me that adding a specific dependency to a .NET 5 desktop runtime package that explicitly requires admin rights would block installation for people who don't have admin rights but do have the .NET 5 desktop runtime already, which is problematic. Thoughts?

Ash258: It will be updated to 5 next weekend, since it is needed by multiple manifests.

So rather shipped invalid application, which needs specific dependency to work and not notify user by any mean, so they will struggle with it and opening issues about it, instead of suggesting it??

cerebrate: Well, if you run it on a machine without the .NET 5 desktop runtime installed, you get a dialog telling you to install it complete with a link to the download site. Seems friendlier than telling people for whom it would otherwise work (since they already have the runtime) that they need to install one again via Scoop first so that it'll recognize the dependency. But as you like.

Next question: I'm looking at these dependencies, and I'm not seeing a way to indicate which version of a package you depend on?

cerebrate: ...okay, no, wait. You can only depend on the latest manifest version, because the latest manifest version is the only one in the bucket.

Which means that you _will_ need a separate manifest to install the .NET 5 runtime, because .NET Core-and-now-.NET runtimes aren't backwards compatible across major version numbers, and so updating the windowsdesktop-runtime package to point at a non-.NET Core 3 runtime would break everything currently depending on it.

PR #5548, opened 2021-01-25
php(-nts)-xdebug: Update to version 3.0.2-8.0

### LINKED ISSUE #5480: php-xdebug not working
Hello!

I am trying to install `php-xdebug` on my machine running win10 and php8. After installing I have all the required configuration set up (at least, I think so), but when running `php -v` it tells me that it is unable to load the module.

But, if I download xdebug directly, the "php_xdebug-3.0.2-8.0-vs16-x86_64.dll" it does indeed work.

I guess this is a mis-match between php8, VS16 and php-xdebug?

Is it possible to specify VS version when running `scoop install php-xdebug` somehow?

VenatusSimpleX:
Same as what #5343 had attempted to do but was reverted due to some fields not changed.

Fixes #5480


silverqx:
> Same as what #5343 had attempted to do but was reverted due to some fields not changed.
> 
> Fixes #5480

Thank you that was my bad, I forgot to update checkver and autoupdate fields 😕, anyway, we are good now, at least for a while 🤔😀

PR #5628, opened 2021-02-06
scoop-completion: Add version 0.2.3

alexchapar:
This can be use to enable scoop auto-completion command for PowerShell

### DIFF
--- a/bucket/scoop-completion.json
+++ b/bucket/scoop-completion.json
@@ -3,17 +3,18 @@
     "description": "A Scoop tab completion module for PowerShell",
     "homepage": "https://github.com/Moeologist/scoop-completion",
     "license": "MIT",
-    "url": "https://psg-prod-eastus.azureedge.net/packages/scoop-completion.0.2.3.nupkg",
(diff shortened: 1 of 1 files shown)


alexchapar:
Hi @Ash258, can you give me some pointers about what I should review ?

issaclin32 (approved):

rashil2000:
I feel this should be a part of Scoop itself, like we can put this in the `supporting/` folder

issaclin32:
> I feel this should be a part of Scoop itself, like we can put this in the `supporting/` folder

I agree. But we can keep this as a seperate app as a test bed, before eventually implemented in *Scoop*.
(such is the case of `boost`/`Intel tbb`, and c++ standards)

Also, reviewing/merging of *Scoop* core is somewhat pretty slow. Because some members are too busy, and the *Scoop* core really needs some testing before putting it online. That is why I am trying to put something into `UpdateTracker` before integrating into *Scoop*.

rashil2000:
Sure, seems good to me.

Btw scoop-completion hasn't been in active development, but I've been using it for more than a year without any issues.

PR #5655, opened 2021-02-10
switchhosts: Update to version 3.5.7

DoveBoy:
switchhosts: Update to version 3.5.7


Ash258:
Prerelease version will not be used, unless there is available stable release

PR #5669, opened 2021-02-11
powertoys: Add installer script and autoupdate

7crabs:
- add script to extract MSI from .exe installer and files from the MSI
- add autoupdate
- update version to 0.31.2


Ash258:
Duplicate of #5391

7crabs:
I read #5391 first and figured the problem is that powertoys' installer make changes to the system.
So this added script avoid running the installer. As with the previous manifest, it only extracts files from MSI .
Did that way become deprecated?

PR #5670, opened 2021-02-11
musescore: Actually update to 3.6.2.548021803

PrOF-kk:
Wrong tag name leads to a 404 from Github


MunsterPlop:
autoupdate urls also need to be fixed #5686

PR #5683, opened 2021-02-14
Add wireguard

simnalamburt:
It's based on @raisercostin's wireguard.json

Wireguard is quite awesome VPN software. I think it deserves to be in scoop-extras!

##### Reference
- https://www.wireguard.com/
- https://github.com/raisercostin/raiser-scoop-bucket/blob/21521bf/wireguard.json


Ash258:
https://github.com/ScoopInstaller/Main/issues/397

PR #5703, opened 2021-02-17
Create freecad-pre.json

pandaero:
Please check that it is correct: 
- normal freecad release is .exe installer
- pre-release is .7z archive
 
First contribution ever. Not sure about actual installation commands.


Ash258:
https://github.com/ScoopInstaller/Versions

PR #5716, opened 2021-02-18
marktext-portable: Add version 0.16.3

ouyangwanxi:


Ash258 (changes requested):
Again with no migration of data.

https://github.com/lukesampson/scoop-extras/pull/5523#pullrequestreview-570053144

PR #5726, opened 2021-02-20
milkman: Add version 5.1.0

kinosang:
Milkman is an Extensible Request/Response Workbench.
https://github.com/warmuuh/milkman

### DIFF
--- a/bucket/milkman.json
+++ b/bucket/milkman.json
@@ -1,0 +1,31 @@
+{
+    "version": "5.1.0",
+    "description": "An Extensible Request/Response Workbench",
+    "homepage": "https://github.com/warmuuh/milkman",
+    "license": "MIT",
+    "url": [
+        "https://github.com/warmuuh/milkman/releases/download/5.1.0/milkman-dist-win64-bin.zip",
+        "https://github.com/warmuuh/milkman/releases/download/5.1.0/milkman-dist-plugins-bin.zip"
+    ],
+    "hash": [
+        "a332f698efe992f9da49fa36b9b96f3bd03e3a4f8238056ef486703a16e5d45b",
+        "83d1db252d35de8f2a987e55a9ad01e4f23706b728b1d8188b59b2f362defec3"
+    ],
+    "extract_dir": [
+        "milkman-5.1.0",
+        "milkman-5.1.0"
+    ],
+    "bin": "milkman.exe",
+    "shortcuts": [
+        [
+            "milkman.exe",
+            "Milkman"
+        ]
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/milkman.json
```diff
@@ -0,0 +1,31 @@
...
+    "homepage": "https://github.com/warmuuh/milkman",
+    "license": "MIT",
+    "url": [
+        "https://github.com/warmuuh/milkman/releases/download/5.1.0/milkman-dist-win64-bin.zip",
+        "https://github.com/warmuuh/milkman/releases/download/5.1.0/milkman-dist-plugins-bin.zip"
+    ],
+    "hash": [
+        "a332f698efe992f9da49fa36b9b96f3bd03e3a4f8238056ef486703a16e5d45b",
+        "83d1db252d35de8f2a987e55a9ad01e4f23706b728b1d8188b59b2f362defec3"
+    ],
```
Ash258, line 13: 64bit only

bucket/milkman.json
```diff
@@ -0,0 +1,31 @@
...
+            "milkman.exe",
+            "Milkman"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/warmuuh/milkman"
+    },
+    "autoupdate": {
+        "url": "https://github.com/warmuuh/milkman/releases/download/$version/milkman-dist-win64-bin.zip"
+    }
```
Ash258, line 30:
```suggestion
    "checkver": "github"
```

bucket/milkman.json
```diff
@@ -0,0 +1,31 @@
...
+        "83d1db252d35de8f2a987e55a9ad01e4f23706b728b1d8188b59b2f362defec3"
+    ],
+    "extract_dir": [
+        "milkman-5.1.0",
+        "milkman-5.1.0"
+    ],
+    "bin": "milkman.exe",
+    "shortcuts": [
+        [
+            "milkman.exe",
```
Ash258, line 21: Use bat
image

Ash258 (changes requested):

bucket/milkman.json
kinosang: @Ash258 the second one is plugins, not 32bit application.

Ash258: So the users will be able to run random plugins jars, but not the main application because it will not work on their platform?? Seems like good idea.

kinosang: @Ash258 Milkman only have 64bit binaries. If scoop does not accept 64bit only applications, I'll close this PR.

Ash258: https://github.com/lukesampson/scoop/blob/master/schema.json

PR #5736, opened 2021-02-21
sqlitestudio: make config local and persist

nuno-andre:
This PR makes SQLiteStudio keep its config locally, as stated in [its wiki](https://github.com/pawelsalawa/sqlitestudio/wiki/Configuration_directory_location#portable-configuration), and persists the settings folder.


Ash258 (changes requested):
- Never use aliases
- No data migration from previous config location
- Incosistent format
- Files has to be precreated
- It is nonsense to precreate persist dir

PR #5756, opened 2021-02-25
Add x plane sdk

pandaero:
added latest X-Plane SDK


Ash258:
This is not application. Some random c files will not be added to the offical bucket

PR #5757, opened 2021-02-25
qtdesigner: Add version 5.14.1

nuno-andre:


bucket/qtdesigner.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "5.14.1",
+    "description": "Qt tool for designing and building GUIs with Qt Widgets",
+    "homepage": "https://doc.qt.io/qt-5/qtdesigner-manual.html",
+    "license": "GPL-3.0",
+    "url": "https://files.pythonhosted.org/packages/3c/c5/f010b4d1f39993fefeb124df98f7b279f1a0339179f998976a5ab2e8da15/PyQt5Designer-5.14.1-py3-none-win32.whl",
+    "hash": "9d22b4f0f47d81859f253850f2a2d66e29011b4a380541b6a891e035a0914769",
+    "installer": {
+        "script": "Expand-7zipArchive $dir\\PyQt5Designer-$version-py3-none-win32.whl -ExtractDir QtDesigner -Removal"
+    },
```
Ash258, line 10: rewrite to 7z and use extract_dir

bucket/qtdesigner.json
```diff
@@ -0,0 +1,36 @@
...
+    ],
+    "bin": "qml.exe",
+    "checkver": {
+        "url": "https://pypi.org/pypi/PyQt5Designer/json",
+        "jsonpath": "$.urls[?(@.url =~ /.*win32\\.whl/')].url",
+        "regex": "^https://(.*?PyQt5Designer-([\\d\\.]*?)-.*?)$",
+        "replace": "$2"
+    },
+    "autoupdate": {
+        "url": "https://$match1",
```
Ash258, line 29: Find different source. This is horrible.

bucket/qtdesigner.json
```diff
@@ -0,0 +1,36 @@
...
+            "linguist.exe",
+            "Qt Designer\\Qt Linguist"
+        ]
+    ],
+    "bin": "qml.exe",
+    "checkver": {
+        "url": "https://pypi.org/pypi/PyQt5Designer/json",
+        "jsonpath": "$.urls[?(@.url =~ /.*win32\\.whl/')].url",
+        "regex": "^https://(.*?PyQt5Designer-([\\d\\.]*?)-.*?)$",
+        "replace": "$2"
```
Ash258, line 26: Find more relevant source

bucket/qtdesigner.json
```diff
@@ -0,0 +1,36 @@
...
+        [
+            "designer.exe",
+            "Qt Designer\\Qt Designer"
+        ],
+        [
+            "linguist.exe",
+            "Qt Designer\\Qt Linguist"
+        ]
+    ],
+    "bin": "qml.exe",
```
Ash258, line 21: Bin before shortcuts

Ash258 (changes requested):

bucket/qtdesigner.json
nuno-andre: It doesn't extract the archive. The extension, maybe?

Ash258: **Rewrite to 7z** and use extract_dir

bucket/qtdesigner.json
nuno-andre: That is the source. They package the tool and release it via PyPI.

Ash258: > They package the tool and release it via PyPI.

Source of this claim?? There is no such information in docs

nuno-andre:
The schema is too restrictive with `autoupdate.url`. It waits for a uri, but the version with `"url": "$match1"` installs without problems.

Ash258:
No it is not. Using such matches is not good idea in autoupdate as it leads to random URLs. That is why this PR will most likely will be closed until better source of installation will be found.

Ash258 (changes requested):
Add 64bit version

bucket/qtdesigner.json
nuno-andre: if there is an order for fields, it should be in the schema, or at least documented

nuno-andre:
There's no 64bit version.
There's no other source AFAIK.

Ash258:
...

image

Ash258:
Going to close this.

As soon as this will be available on their download page it could be added again. 

https://download.qt.io/official_releases/

nuno-andre:
That's not the Qt Designer version but the wheel's, which is irrelevant for this case, because there's no Python application involved.

image

nuno-andre:
What a shame of the maintainer.

Ash258:
It is not listed in official releases on their site. There is no mention how to install it in docs. And only random pip package available on internet. This cannot be official.

The project link on pypi lead ot random pyqt.site in chinese.
Author do not have email, which could be even similar to the qt organization

image

Hence this is not official.

nuno-andre:
I agree it's not the most polished page in PyPI, but it's that of a component of the main package.

This is the project
https://github.com/altendky/pyqt-tools
https://pypi.org/project/pyqt5-tools/

It's not an official Qt build. Qt packages Qt Designer in Qt Creator, this is an only Qt Designer build.

Issue #2535, opened 2019-07-30
calibre-portable path error


danalec:
"Path to Calibre Portable too long. Must be less than 59 characters"
![calibre-portable-installer-3 46 0_Ucpo3YmBGo](https://user-images.githubusercontent.com/2339131/62143701-dfd41e00-b2c6-11e9-98b1-bb1a580faed0.png)

linsui:
See https://www.mobileread.com/forums/showthread.php?t=142913

Ash258:
#1765

7wells:
Hi, I (still) have the same problem. Maybe removing the portable version of Calibre from this bucket is better than letting users run into this problem and repeating the same comments (like me, blush). Or does it work for somebody?

See there:
https://github.com/lukesampson/scoop-extras/issues/1765

linsui:
Yep, you can install scoop to e.g. `D:/scoop` .

7wells:
Ah, but that involves placing scoop and all apps to another and shorter path. If no other drive is available, then maybe scoop could be installed directly to C:\scoop? Still, without any further explanation in the install process of the portable Calibre version, it will confuse more users. Maybe the json file could be amended to offer some help, e.g. by proposing a shorter installation path? Is it possible to have only Calibre installed somewhere else while keeping everything else as per scoop standard?

linsui:
> Is it possible to have only Calibre installed somewhere else while keeping everything else as per scoop standard?

You can write a custom manifest and install it to anywhere you like.

7wells:
Indeed I could (try) doing it, but most of the users might not know how. Please accept my apologies if I wrongly assumed that the apps offered in this bucket work "out of the box" (understanding quite well that the maintainer/contributors do this all in their free time, for which I am very grateful). I just thought that many average users might unnecessarily stumble over this. If this remains as it is, it is perfectly fine with me, and I am still happy with scoop and this bucket. Currently, I use the "normal" Calibre from this bucket without any problems. 👍

DOFfactory:
Same issue here with the latest version of scoop

rashil2000:
With calibre or calibre-normal?

gigberg:
update: you could set windows long path option by regedit. then calibre portable work properly.
https://learn.microsoft.com/zh-cn/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell

VitaliyAT:
[...]

PR #5762, opened 2021-02-26
calibre: bump to 5.12.0

adnaanbheda:
bump to 5.12.0, since 5.11.0 doesn't have download links anymore


adnaanbheda:
@Ash258

adnaanbheda:
also, I'm still facing this #2535, so there's no point in keeping the portable version here in scoop

PR #5763, opened 2021-02-26
ibmcloud-cli: Update urls, remove unused shims

se35710:


Ash258:
Change it rather in Main https://github.com/ScoopInstaller/Main/blob/master/bucket/ibmcloud-cli.json

This manifest will be deprecated in favor of Main bucket in near future.

PR #5791, opened 2021-03-04
vieb: Add version 4.0.0

run2cmd:
I'd like to add very promising browser [Vieb](https://vieb.dev/). I use that day to day and would be great to have Scoop packages for it. 

I have tested installation and update process with my own test bucket under https://github.com/run2cmd/scoop-test-bucket

### DIFF
--- a/bucket/vieb.json
+++ b/bucket/vieb.json
@@ -1,7 +1,7 @@
 {
     "version": "4.1.0",
     "description": "Vim Inspired Electron Browser",
-    "homepage": "https://vieb.dev/",
+    "homepage": "https://vieb.dev",
     "license": "GPL-3.0-or-later",
     "architecture": {
         "64bit": {
--- a/bucket/vieb.json
+++ b/bucket/vieb.json
@@ -1,6 +1,6 @@
 {
     "version": "4.0.0",
-    "description": "Vieb is the Vim Inspired Electron Browser",
+    "description": "Vim Inspired Electron Browser",
     "homepage": "https://vieb.dev/",
     "license": "GPL-3.0-only",
     "url": "https://github.com/Jelmerro/Vieb/releases/download/4.0.0/Vieb.4.0.0.exe",
@@ -15,7 +15,6 @@
         "github": "https://github.com/Jelmerro/Vieb"
     },
     "autoupdate": {
-        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb.$version.exe",
-        "extract_dir": "vieb-$version-x64"
+        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb.$version.exe"
     }
 }
--- a/bucket/vieb.json
+++ b/bucket/vieb.json
@@ -1,13 +1,13 @@
 {
-    "version": "4.0.0",
-    "description": "Vim Inspired Electron Browser",
+    "version": "4.1.0",
+    "description": "Vieb is the Vim Inspired Electron Browser",
     "homepage": "https://vieb.dev/",
-    "license": "GPL-3.0-only",
-    "url": "https://github.com/Jelmerro/Vieb/releases/download/4.0.0/Vieb.4.0.0.exe",
-    "hash": "d5f46c3fb5a8a379866f4a97cd2ad5622fc6430465623e291aefb0dd1f8e6e4a",
+    "license": "GPL-3.0-or-later",
+    "url": "https://github.com/Jelmerro/Vieb/releases/download/4.1.0/Vieb-4.1.0-win.zip",
+    "bin": "Vieb.exe",
     "shortcuts": [
         [
-            "Vieb.4.0.0.exe",
+            "Vieb.exe",
             "Vieb"
         ]
     ],
@@ -15,6 +15,6 @@
         "github": "https://github.com/Jelmerro/Vieb"
     },
     "autoupdate": {
-        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb.$version.exe"
+        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb-$version-win.zip"
     }
 }
--- a/bucket/vieb.json
+++ b/bucket/vieb.json
@@ -1,0 +1,21 @@
+{
+    "version": "4.0.0",
+    "description": "Vieb is the Vim Inspired Electron Browser",
+    "homepage": "https://vieb.dev/",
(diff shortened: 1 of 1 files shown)


bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "4.0.0",
+    "description": "Vieb is the Vim Inspired Electron Browser",
+    "homepage": "https://vieb.dev/",
+    "license": "GPL-3.0-only",
```
Ash258, line 5: invalid SPX identifier

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
...
+            "Vieb.4.0.0.exe",
+            "Vieb"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/Jelmerro/Vieb"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb.$version.exe",
+        "extract_dir": "vieb-$version-x64"
```
Ash258, line 19: ???

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "4.0.0",
+    "description": "Vieb is the Vim Inspired Electron Browser",
```
Ash258, line 3: Do not mention applicaiton name

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
...
+    "homepage": "https://vieb.dev/",
+    "license": "GPL-3.0-only",
+    "url": "https://github.com/Jelmerro/Vieb/releases/download/4.0.0/Vieb.4.0.0.exe",
+    "hash": "d5f46c3fb5a8a379866f4a97cd2ad5622fc6430465623e291aefb0dd1f8e6e4a",
+    "shortcuts": [
+        [
+            "Vieb.4.0.0.exe",
```
Ash258, line 10: Having autoupdate and version in shortcut does not make sense and will lead to failed installation after autoupdate

Ash258 (changes requested):

bucket/vieb.json
run2cmd: This is what I found on https://spdx.org/licenses/GPL-3.0-only.html. What identifier I should put there?

Ash258: Sorry my bad. I overlooked it.

bucket/vieb.json
run2cmd: This is self execute binary. I'll talk with developer if we can have file without version in name.

Ash258: <url>#/view.exe

bucket/vieb.json
run2cmd: Removed.

bucket/vieb.json
run2cmd: Changed.

bucket/vieb.json
Jelmerro: I want to mention that it should actually be https://spdx.org/licenses/GPL-3.0-or-later.html

bucket/vieb.json
Jelmerro: Would it be ok if the zip is versioned, but the exe inside it isn't? For example `Vieb-4.1.0-win.zip` containing `Vieb.exe` inside at the root of the zip?

Ash258: Does not make sense to waste time with different packaging. As already said. Just use URL rewrite `<url>#/view.exe`

run2cmd: I have tested both the rewrite option and with zip file. Rewrite was waring however when using Vieb from single exe I had issues with creating shortcuts of my own for example on task bar. I did not experience those issues with zip file. Sine it's already created I used that one.

bucket/vieb.json
run2cmd: Updated license to GPL-3.0-or-later

bucket/vieb.json
Ash258: Again

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "4.1.0",
+    "description": "Vieb is the Vim Inspired Electron Browser",
+    "homepage": "https://vieb.dev/",
+    "license": "GPL-3.0-or-later",
+    "hash": "1c958e4cba47248213f4b9ede909ea6f57cf24967e953fceb489b5342b1f6f88",
```
Ash258, line 6: Hash is checked after the URL is downloaded. Does not make sense to have the hash above URL

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
...
+    "url": "https://github.com/Jelmerro/Vieb/releases/download/4.1.0/Vieb-4.1.0-win.zip",
```
Ash258, line 7: This has to be 64bit version

bucket/vieb.json
```diff
@@ -0,0 +1,21 @@
...
+        [
+            "Vieb.exe",
+            "Vieb"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/Jelmerro/Vieb"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Jelmerro/Vieb/releases/download/$version/Vieb-$version-win.zip"
```
Ash258, line 19: 64bit

bucket/vieb.json
Ash258, line 7: What does portable means? Where are application data and configurations stored?

image

Ash258 (changes requested):

bucket/vieb.json
run2cmd: Fixed

bucket/vieb.json
run2cmd: Added.

bucket/vieb.json
run2cmd: Added.

bucket/vieb.json
run2cmd: The initial configuration is not needed. Vieb reads `%APPDATA%\Vieb\viebrc` file if its created.

bucket/vieb.json
run2cmd: Fixed

run2cmd:
@Ash258 Anything else you'd like me to improve about this PR? Do you want me to squash changes, the commit history in this PR is a bit messy.

Ash258 (approved):

PR #5815, opened 2021-03-09
obs-studio-plugin-virtualcam: Add version 2.0.5

alexchapar:
obs-studio plugin to add virtualcam


alexchapar:
A better solution has been proposed by @MunsterPlop with the PR #6018. 
This PR is not longer useful. :wink:

PR #5846, opened 2021-03-11
rpcview: Add version 0.3.0.67

L4ys:
A free and powerful tool to explore and decompile all RPC functionalities present on a Microsoft system


Ash258:
Does not make sense to use automatic, latest build of appveryor.

Come back when there will be valid releaes cycle

L4ys:
Appveryor is the only officially release channel provided by RpcView
so does scoop currently not accept these software that released with Appveryor?

Ash258:
- Downloading latest successful build does not make sense.
- There could be dozens of them daily, which will only spam users.
- Appveyor artifacts are purged after some time

PR #5852, opened 2021-03-12
rtss: Update to version 7.3.1

stasjok:
There is an OverlayEditor plugin in version 7.3.0 which saves overlays to `Plugins\Client\Overlays\` (even if you choose another directory). I've added it to persist.
Checkver was disabled in commit https://github.com/lukesampson/scoop-extras/commit/d1fa5192c3268bd9e3d130413de0521d0a532a0c. I didn't touch it.

### DIFF
--- a/bucket/rtss.json
+++ b/bucket/rtss.json
@@ -1,6 +1,6 @@
 {
     "##": "Renaming the downloaded zip to 'sourceforge.net.zip' is a hack for the referer header. See dl()-function in \\lib\\install.ps1",
-    "version": "7.2.3",
+    "version": "7.3.1",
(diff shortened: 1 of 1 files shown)


Fooughhy:
This update also fixes an issue with an annoying popup I get during boot, "Some system components cannot be hooked right now" which I believe simply was an incompatability issue with an update to Windows 10.

Would appreciate if this could be merged 😄

bucket/rtss.json
```diff
@@ -31,8 +31,12 @@
             "RivaTuner Statistics Server"
         ]
     ],
-    "persist": "Profiles",
-    "autoupdate": {
```
pratikpc, line 35: Why did you remove autoupdate?

PR #5859, opened 2021-03-17
wpsoffice: Modify checkver.url

sinloss:
To fetch the newest version. 

As the version on the page `https://pc.wps.com` has lagged behind a lot, a newer solution is needed.


sinloss:
Kindly request a faster merge of this due to the real need of using this one... 😄

PR #5864, opened 2021-03-19
winssh-pageant: Add version 1.2

ndbeals:
https://github.com/ndbeals/winssh-pageant

### DIFF
--- a/bucket/winssh-pageant.json
+++ b/bucket/winssh-pageant.json
@@ -13,13 +13,11 @@
             "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
         }
     },
-    "extract_dir": "",
     "bin": "winssh-pageant.exe",
-    "shortcuts": [
-        [
-            "winssh-pageant.exe",
-            "..\\Startup\\winssh-pageant"
-        ]
+    "notes": [
+        "This application is meant to run in the background upon system startup. Do that by running the following commands:",
+        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
+        "$Shortcut.Save()"
     ],
     "checkver": "github",
     "autoupdate": {
--- a/bucket/winssh-pageant.json
+++ b/bucket/winssh-pageant.json
@@ -3,6 +3,11 @@
     "description": "Pageant to Windows OpenSSH agent bridge",
     "homepage": "https://github.com/ndbeals/winssh-pageant",
     "license": "BSD-3-Clause",
+    "notes": [
+        "This application is meant to run in the background upon system startup. Do that by running the following commands:",
+        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
+        "$Shortcut.Save()"
+    ],
     "architecture": {
         "64bit": {
             "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_amd64.zip",
@@ -14,11 +19,6 @@
         }
     },
     "bin": "winssh-pageant.exe",
-    "notes": [
-        "This application is meant to run in the background upon system startup. Do that by running the following commands:",
-        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
-        "$Shortcut.Save()"
-    ],
     "checkver": "github",
     "autoupdate": {
         "architecture": {
@@ -32,7 +32,7 @@
             "32bit": {
                 "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_386.zip",
                 "hash": {
-                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version",
                     "regex": "386:$sha256"
                 }
             }
--- a/bucket/winssh-pageant.json
+++ b/bucket/winssh-pageant.json
@@ -3,7 +3,7 @@
     "description": "Pageant to Windows OpenSSH agent bridge",
     "homepage": "https://github.com/ndbeals/winssh-pageant",
     "license": "BSD-3-Clause",
-    "notes": "This application is meant to run in the background upon system startup. If you want to fully use it, copy shortcut into Start Menu Startup folder",
+    "notes": "This application is meant to run in the background upon system startup. If you want to use it, copy shortcut into Start Menu Startup folder",
     "architecture": {
         "64bit": {
             "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_amd64.zip",
--- a/bucket/winssh-pageant.json
+++ b/bucket/winssh-pageant.json
@@ -1,0 +1,43 @@
+{
+    "version": "1.2",
+    "description": "Pageant to Windows OpenSSH agent bridge",
+    "homepage": "https://github.com/ndbeals/winssh-pageant",
+    "license": "BSD-3-Clause",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_amd64.zip",
+            "hash": "a9bef9d134e872e60fd1d0950ce29f73b98af4802c9e93987a4ecafe8debf814"
+        },
+        "32bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_386.zip",
+            "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
+        }
+    },
+    "extract_dir": "",
+    "bin": "winssh-pageant.exe",
+    "shortcuts": [
+        [
+            "winssh-pageant.exe",
+            "..\\Startup\\winssh-pageant"
+        ]
+    ],
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_amd64.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "regex": "amd64:$sha256"
+                }
+            },
+            "32bit": {
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_386.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "regex": "386:$sha256"
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


bucket/winssh-pageant.json
```diff
@@ -0,0 +1,43 @@
...
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_386.zip",
+            "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
+        }
+    },
+    "extract_dir": "",
+    "bin": "winssh-pageant.exe",
+    "shortcuts": [
+        [
+            "winssh-pageant.exe",
+            "..\\Startup\\winssh-pageant"
```
Ash258, line 21: Does not make sense.

Manipulating with startup items (or any system configuraion) is awesome example of practice what scoop users are trying to avoid.

Remove whole shortcut property

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,43 @@
...
+        "64bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_amd64.zip",
+            "hash": "a9bef9d134e872e60fd1d0950ce29f73b98af4802c9e93987a4ecafe8debf814"
+        },
+        "32bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_386.zip",
+            "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
+        }
+    },
+    "extract_dir": "",
```
Ash258, line 16:
```suggestion
```

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,43 @@
...
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "regex": "amd64:$sha256"
+                }
+            },
+            "32bit": {
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_386.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "regex": "386:$sha256"
```
Ash258, line 38: Use parsing of the table rather than some text, which will be most likely removed in future

Ash258 (changes requested):

bucket/winssh-pageant.json
ndbeals: I will do this.

But this is not a tool you run when you need it (like youtube-dl or wireshark). It is a background program that's meant to run alongside the ssh-agent, very similar to [ssh-agent-wsl](https://github.com/lukesampson/scoop-extras/blob/master/bucket/ssh-agent-wsl.json). There's no direct .bashrc equivalent for outside wsl, should I use a similar `post_install` script to tell the user to create a shortcut into the appropriate startup folder?

Ash258: Use notes

bucket/winssh-pageant.json
ndbeals: Do you refer to the table of hashes on the release description?

If so, that table are hashes of the executable themselves, because the go binary is reproducible, so users may check it themselves. 

I can add another table of the zip file hashes to the release description, but wouldn't it be using regex to match it anyways? I guess I'm confused what you mean by parse the table?

bucket/winssh-pageant.json
ndbeals: Done, thanks for the direction :)

bucket/winssh-pageant.json
ndbeals: I'm retrieving the hash with regex just like [anaconda3](https://github.com/lukesampson/scoop-extras/blob/master/bucket/anaconda3.json) and [accesser](https://github.com/lukesampson/scoop-extras/blob/master/bucket/accesser.json) are. I can remove the hash checking if you'd like.

ndbeals (commented):
Changes implemented in different commit.

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,41 @@
...
+        "32bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_386.zip",
+            "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
+        }
+    },
+    "bin": "winssh-pageant.exe",
+    "notes": [
+        "This application is meant to run in the background upon system startup. Do that by running the following commands:",
+        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
+        "$Shortcut.Save()"
```
Ash258, line 20: Use only verbal instructions to copy the shortcut file. This command will never be executed by sanity people.

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,41 @@
...
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_amd64.zip",
+            "hash": "a9bef9d134e872e60fd1d0950ce29f73b98af4802c9e93987a4ecafe8debf814"
+        },
+        "32bit": {
+            "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v1.2/winssh-pageant-v1.2_386.zip",
+            "hash": "0591f0c4475dba8fbfacc7ffe5fcb77ac384008d884976b386b4f25fa894e3a8"
+        }
+    },
+    "bin": "winssh-pageant.exe",
+    "notes": [
```
Ash258, line 17: Notes are essential information for users. Does not make sense to hide them in manifest.

Put it above architecture property

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,41 @@
...
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_amd64.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
+                    "regex": "amd64:$sha256"
+                }
+            },
+            "32bit": {
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_386.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
```
Ash258, line 35:
```
                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version",
```

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,41 @@
...
+        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
+        "$Shortcut.Save()"
+    ],
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/ndbeals/winssh-pageant/releases/download/v$version/winssh-pageant-v$version_amd64.zip",
+                "hash": {
+                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version/",
```
Ash258, line 28:
```suggestion
                    "url": "https://github.com/ndbeals/winssh-pageant/releases/tag/v$version",
```

Ash258 (changes requested):

bucket/winssh-pageant.json
ndbeals: I'm doing the exact same thing https://github.com/lukesampson/scoop-extras/blob/master/bucket/ssh-agent-wsl.json is doing, echoing a command that users can run (or add to an autoran script, same thing). Why is that utility allowed to do it?

bucket/winssh-pageant.json
ndbeals: I'm doing the exact same thing https://github.com/lukesampson/scoop-extras/blob/master/bucket/ssh-agent-wsl.json is doing, echoing a command that users can run (or add to an autoran script, same thing). Why is that utility allowed to do it?

Ash258: it is different to run simple command with 2 parameters instead of some advanced powershell specific command. Lots of users are using cmd wrapper to run scoop.

Add notes about manually copying the shortcut file.

ndbeals: Lots of other manifests (filezilla_server, grepwin, eagleget_portable, smarttaskbar) have notes telling users to run scripts in `$dir` (sometimes even as administrator). What I'm doing by outputting a 3 line command is no different. I'd say it's even more transparent than running a file.

ndbeals: I'll just include a .ps1 script that does it then, just like filezilla_server is doing.

Ash258: It is different...

There is huge different between powershell and cmd. You cannot suggest running powershell commands to cmd users.

ndbeals: But it's fine, I'll add notes about manual shortcuts.

bucket/winssh-pageant.json
```diff
@@ -0,0 +1,42 @@
...
+    "version": "1.2",
+    "description": "Pageant to Windows OpenSSH agent bridge",
+    "homepage": "https://github.com/ndbeals/winssh-pageant",
+    "license": "BSD-3-Clause",
+    "notes": [
+        "This application is meant to run in the background upon system startup. Add a shortcut to %APPDATA%\\Microsoft\\Windows\\Start Menu\\Startup that points to $dir\\winssh-pageant.exe",
+        "Alternatively, you can run the following snippet in powershell:",
+        "$Shortcut = (New-Object -comObject WScript.Shell).CreateShortcut($env:APPDATA+'\\Microsoft\\Windows\\Start Menu\\Programs\\Startup\\winssh-pageant.lnk')","$Shortcut.TargetPath = '$dir\\winssh-pageant.exe'",
+        "$Shortcut.Save()"
+    ],
```
Ash258, line 11: It is not good idea to use %appdata% as it is cmd only and it ignores global installation options

```suggestion
    "notes": "This application is meant to run in the background upon system startup. If you want to fully use it, copy shortcut into Start Menu Startup folder",
```

Ash258 (changes requested):

bucket/winssh-pageant.json
ndbeals: > it ignores global installation options

That is the intention. This is a tool that works with the users ssh-agent. Adding this as a startup item to other users that may not even use an ssh-agent is wrong.

Ash258: Theses intentions are nowhere mentioned. It is not handled. Noone will know this.

ndbeals: Anyone using an ssh-agent (for which this tool only works for) will know this by default. That is how ssh-agent works. The intent of this tool, to be a bridge to the windows ssh agent is in the description. I have updated the notes anyways.

bucket/winssh-pageant.json
shawnz: I don't think that's true that there is a huge difference (PowerShell has shipped with the OS for over a decade now).

But still, what about advising the user to use schtasks.exe instead to set the logon action?

ndbeals:
Got my answer.

ndbeals:
Why would you waste a week of both our time to review a manifest you never were going to accept in the first place???

ndbeals:
I would have really appreciated being told to not waste my time a week ago when I opened this PR, instead of the frustrating and ridiculous back-and-forth you put me through.

bucket/winssh-pageant.json
ndbeals: > I don't think that's true that there is a huge difference (PowerShell has shipped with the OS for over a decade now).

Not to mention the fact that the _official install instructions for scoop itself_ has the user downloading and executing a powershell script straight from the internet. (A La `curl <url> | sh` infamy). I realized I must have got Ash258 on a bad day or something, because half of the things he didn't want in this manifest exist in numerous other manifests in this bucket.


> But still, what about advising the user to use schtasks.exe instead to set the logon action?

Ultimately I don't really care how scoop tells users to auto-start this tool, that could do just fine.

I'm not going to spend any more time on this, the inconsistent standards in this repo has seen to that. If you want to use scoop to install this, you can find a manifest here: https://gist.github.com/ndbeals/08fd5f1eed011bce2e7bba06da0e5314

Issue #5877, opened 2021-03-20
idea-ultimate@2020.3.3: hash check failed


Weidows:
❯ scoop update idea-ultimate
idea-ultimate: 2020.3.2 -> 2020.3.3
Updating one outdated app:
Updating 'idea-ultimate' (2020.3.2 -> 2020.3.3)
Downloading new version
Loading ideaIU-2020.3.3.win.zip from cache
Checking hash of ideaIU-2020.3.3.win.zip ... ERROR Hash check failed!        
App:         extras/idea-ultimate
URL:         https://download.jetbrains.com/idea/ideaIU-2020.3.3.win.zip     
First bytes: 50 4B 03 04 0A 00 00 08
Expected:    419196bea9995c0577192e72aca1b62669bcd0d55dbfc93ee67928225a4be8a4
Actual:      6f3a7dcb6aa4c1d3f9dc99883aa3b7dbc0a19d5a46d3c0965dc989e2f6c905aa

PR #5914, opened 2021-03-27
idea-ultimate@2020.3.3 Hash fix

Weidows:
Five issues were mistakenly checked and closed before.

- #5895 
- #5891 
- #5882 
- #5877 
- #5876


Ash258:
The hash is correct. It is china only problem

PR #5917, opened 2021-03-28
Add manifest for discord-delete

cedws:


bucket/discord-delete.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "1.3.1",
+    "description": "Tool to delete Discord message history",
+    "homepage": "https://github.com/adversarialtools/discord-delete",
+    "license": "GPL-3.0-or-later",
```
Ash258, line 5: Github licenses are not explicitly `-or-later`. Use `-only`

bucket/discord-delete.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "1.3.1",
+    "description": "Tool to delete Discord message history",
```
Ash258, line 3: > Be warned that using this tool could result in the termination of your account,

I do not like expose this tool to scoop users.

Ash258 (changes requested):

bucket/discord-delete.json
cedws: No comment. It's just how the tool works. I see no reason why this should be excluded from the bucket.

Ash258: /unsubscribe

Allowing users to install the tool, which will terminate their accounts without prior notice is dangerous and does not make sense.

I will not be the one who will merge this and be responsible for this.

cedws: That's not what it says. It says that it *could* result in termination, and to date, this has never happened to a user of the tool.
It's a disclaimer to warn that it is not risk-free.

cedws:
@Ash258 Could you kindly label this for review from another maintainer? If not, this should be closed.

PR #5919, opened 2021-03-28
qbittorrent-portable: Update to version 4.3.4.1

littleli:
- manual update
- version 4.3.4.1 is not enlisted so checkver cannot find it


littleli:
Why is this closed? It seems to be a fix for the previous release.

qbittorrent

littleli:
> Why is this closed? It seems to be a fix for the previous release.
> 
> qbittorrent

I see your reply in the other thread. Please ignore my comment.

PR #5920, opened 2021-03-28
qbittorrent: Update to version 4.3.4.1

littleli:
- manual update
- version 4.3.4.1 is not enlisted so checkver cannot find it


Ash258:
This version was taken down two times, it officially do not exists, so there is no reason to update to it. Not talking about the Windows defender is blocking this version.
It will not be updated until it is officially communicated.

image

littleli:
Thanks. Got it! This is just annoying, because app wants to intstall the update :-(

Issue #5270, opened 2020-12-08
[Request] Add OpenToonz


AbedSHP:
**Software for the production of 2D animation.**
Based on the software "Toonz", developed by Digital Video S.p.A. in Italy, OpenToonz has been customized by Studio Ghibli, and used for the creation of its works for many years. Dwango has launched the OpenToonz project in cooperation with Digital Video and Studio Ghibli

https://opentoonz.github.io/e/
https://opentoonz.github.io/e/download/opentoonz.html

flavio20002:
Hi, I've tried to add it, but it's very difficult. The installer creates regex and a lot of directories in the filesystem. I will work on it in the next days

AbedSHP:
hi, why?
OpenToonzSetup.exe /?
OpenToonzSetup.exe /VERYSILENT /DIR="Simple"

PR #5930, opened 2021-03-29
Add Opentoonz 1.4.0

flavio20002:
#5270 OpenToonz is a 2D animation software published by DWANGO

### DIFF
--- a/bucket/opentoonz.json
+++ b/bucket/opentoonz.json
@@ -1,0 +1,30 @@
+{
+    "version": "1.4.0",
+    "description": "OpenToonz is a 2D animation software published by DWANGO",
+    "homepage": "https://opentoonz.github.io",
+    "license": "BSD-3-Clause",
+    "url": "https://github.com/opentoonz/opentoonz/releases/download/v1.4.0/OpenToonzSetup.exe",
+    "hash": "8b87dc4ac37a1f4acb5f4523a5db2309d2447f317cc03cd71c4c2a77e4f9cd4f",
+    "pre_install": "Write-Host 'Installing OpenToonz. This can take up to 5 minutes on an HDD.' -ForegroundColor Magenta",
+    "installer": {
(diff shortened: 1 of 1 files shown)


bucket/opentoonz.json
```diff
@@ -0,0 +1,30 @@
...
+    "url": "https://github.com/opentoonz/opentoonz/releases/download/v1.4.0/OpenToonzSetup.exe",
+    "hash": "8b87dc4ac37a1f4acb5f4523a5db2309d2447f317cc03cd71c4c2a77e4f9cd4f",
+    "pre_install": "Write-Host 'Installing OpenToonz. This can take up to 5 minutes on an HDD.' -ForegroundColor Magenta",
+    "installer": {
+        "script": "Invoke-Expression \"$dir\\$fname /DIR=$dir /SP- /VERYSILENT /NOICONS | Out-Null \" "
+    },
+    "uninstaller": {
+        "file": "unins000.exe",
+        "args": "/VERYSILENT"
+    },
```
Ash258, line 15: Use innosetup

Ash258 (changes requested):

flavio20002:
@Ash258 I've tried with innosetup and is not working. The installer creates a regex and a folder in "c:/Opentoonz Stuff" that is not replicable with innosetup.

bucket/opentoonz.json
flavio20002: I've tried with innosetup and is not working. The installer creates a regex entry and a folder in "c:/OpentoonzStuff" that is not replicable with innosetup. I will try more

Ash258: > regex entry

Does not make sense.

Then it is clear that this is not suitable for any of official buckets. If this "application" create some garbage folder named `OpentoonzStuff` on C drive it will not be accepted.

PR #5971, opened 2021-04-03
shadowsocks: Update to version 4.4.0.185

davidxuang:
Fix build number syntax in archive name

### DIFF
--- a/bucket/shadowsocks.json
+++ b/bucket/shadowsocks.json
@@ -1,10 +1,10 @@
 {
-    "version": "4.3.0.0",
+    "version": "4.4.0.0",
     "description": "A secure socks5 proxy, designed to protect your Internet traffic.",
     "homepage": "https://github.com/shadowsocks/shadowsocks-windows",
     "license": "GPL-3.0-only",
-    "url": "https://github.com/shadowsocks/shadowsocks-windows/releases/download/4.3.0.0/Shadowsocks-4.3.0.0.zip",
-    "hash": "sha512:0154a548cd317252e7ff350ce2e0bb8bd85b25163b86ffb70cb605b211074438de514a804d18a550225e066c6ea93702a81e457feddd8f2fe8b5ca33e53c37db",
+    "url": "https://github.com/shadowsocks/shadowsocks-windows/releases/download/4.4.0.0/Shadowsocks-4.4.0.185.zip",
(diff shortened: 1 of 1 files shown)


bucket/shadowsocks.json
```diff
@@ -27,12 +27,15 @@
         "pac.txt",
         "user-rule.txt"
     ],
-    "checkver": "github",
+    "checkver": {
+        "url": "https://api.github.com/repos/shadowsocks/shadowsocks-windows/releases/latest",
+        "regex": "(?s)\"tag_name\":\"([\\d.]+)\".+Shadowsocks-(?<main>[\\d.]+)\\.(?<build>\\d+).zip"
```
Ash258, line 32:
```suggestion
        "github": "https://github.com/shadowsocks/shadowsocks-windows",
        "regex": "Shadowsocks-([\\d.]+)\\.zip"
```

bucket/shadowsocks.json
```diff
@@ -27,12 +27,15 @@
...
+    },
     "autoupdate": {
-        "url": "https://github.com/shadowsocks/shadowsocks-windows/releases/download/$version/Shadowsocks-$version.zip",
+        "url": "https://github.com/shadowsocks/shadowsocks-windows/releases/download/$version/Shadowsocks-$matchMain.$matchBuild.zip",
```
Ash258, line 35:
```suggestion
        "url": "https://github.com/shadowsocks/shadowsocks-windows/releases/download/$matchHead.0/Shadowsocks-$version.zip",
```

Ash258 (changes requested):

bucket/shadowsocks.json
davidxuang: Well, after checking the tag list, I don't think that we can assume that the last number is always 0.

Ash258: I can. It is broken since version 4.3.0.0. After that there were no different naming.

bucket/shadowsocks.json
```diff
@@ -1,10 +1,10 @@
 {
-    "version": "4.3.0.0",
+    "version": "4.4.0.0",
```
davidxuang, line 2:
```suggestion
    "version": "4.4.0.185",
```

Issue #5479, opened 2021-01-10
ungoogled-chromium crashes on everypage


hronro:
ungoogled-chromium version: 87.0.4280.67-1

https://user-images.githubusercontent.com/15059605/103122329-b6394500-46ba-11eb-8ef2-7edc41a102e6.mp4

foisonocean:
Hi @Ash258, I just notice you have upgraded `ungoogled-chromium` in 86fe373ec84c4b0197925ab9a34707e3b8319689.

Does ungoogled-chromium works for your? I install `87.0.4280.67-1` and `87.0.4280.141-1` via scoop, neither of them works.

Downloading zip from https://github.com/tangalbert919/ungoogled-chromium-binaries/releases directly works for me thought.

pdeak:
Yea, this still doesn't work for me either, crash on every page.

1hiking:
Can report this also happens to me :/

ikenichiro:
To comment what I know, executing from the "current" directory will crash upon startup.
However, executing within the latest version number directory (e.g. directory "89.0.4389.90-1") will have no problem.
I think it has something to do with the sandbox feature, since you can execute the chrome.exe by using "--no-sanbox" argument within the "current" directory.

Unfortunately I do not have a solution to fix this issue, but I hope this information (and a workaround?) will help people solve this issue.

PR #5975, opened 2021-04-05
Ungoogled-chromium-portable: Update to version 89.0.4389.114

ikenichiro:
Autoupdate has been also modified and updated.
This app is basically the portable version for ungoogled-chromium, so it will temporary provide an alternative workaround for #5479.

### DIFF
--- a/bucket/ungoogled-chromium-portable.json
+++ b/bucket/ungoogled-chromium-portable.json
@@ -1,19 +1,19 @@
 {
     "##": "Check chromium.woolyss.com for different versions of Chromium releases.",
-    "version": "85.0.4183.83",
+    "version": "89.0.4389.114",
     "description": "Browser aiming for safer, faster, and more stable way for all users to experience the web.",
     "homepage": "https://www.chromium.org",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


pratikpc:
Closing in favour of https://github.com/ScoopInstaller/Extras/pull/7282

ikenichiro:
Sorry I couldn't come around to fix after proposing this PR

pratikpc:
@ikenichiro no worries.

You can always help us fix another issue if you find them.

For example, not all Chromium packages are still at their latest versions.

ikenichiro:
@pratikpc
My initial motivation of this PR was from #5479  but with the ungoogled-chromium-portable working, ( #7282 ) I don't have enough motivation to provide fixes to the original ungoogled-chromium.
I know that the package ungoogled-chromium is outdated (as you said, not all Chromium packages are their latest) and I think the repository needs to be changed to Eloston/ungoogled-chromium and download binaries from, https://ungoogled-software.github.io/ungoogled-chromium-binaries/
But as I reported in the issues #5479 , some kind of work around may still be needed to be done for chrome.exe to work in the "current" directory.

PR #5987, opened 2021-04-06
TortoiseHg for mercurial

tukanos:
Adding a TortoiseHg GUI client for mercurial versioning system


bucket/tortoisehg.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "5.7.1",
+    "description": "TortoiseHg. A set of graphical tools and a shell extension for the Mercurial distributed revision control system.",
+    "homepage": "https://tortoisehg.bitbucket.io/",
+    "license": "GNU General Public License",
```
Ash258, line 5: Invalid SPDX identifier

bucket/tortoisehg.json
```diff
@@ -0,0 +1,30 @@
...
+        },
+        "32bit": {
+            "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/tortoisehg-5.7.1-x86.msi",
+            "hash": "02192542a79b8bc4b312f087cd71242ae6a04fe6c47dcf3d6f78e2cd29e725a8"
+        }
+    },
+    "extract_dir": "PFiles\\TortoiseHg",
+    "bin": [
+        "thgw.exe"
+    ],
```
Ash258, line 19: Add all relevant shortcuts and bins

bucket/tortoisehg.json
```diff
@@ -0,0 +1,30 @@
...
+    "checkver": {
+        "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/",
+        "re": "tortoisehg-([\\d\\.]+)-(x64|x86).msi"
```
Ash258, line 22:
```suggestion
        "regex": "tortoisehg-([\\d.]+)-x[468]{2}\\.msi"
```

bucket/tortoisehg.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "shortcuts": [
+        [
+            "thgw.exe",
+            "TortoiseHg Workbench"
+        ]
+    ]
```
Ash258, line 29: add autoupdate

Ash258 (changes requested):
Add reg files to manually register all context menus

bucket/tortoisehg.json
tukanos: > Add reg files to manually register all context menus
Is there a way for scoop to differentiate among different languages in the context menu?  The context menu comes in 24 different language mutations.

Everything else is clear and I'll fix it and send patches for that.

tukanos: One more question.  I would like to detect if `hg.exe` is already in path.  If yes, I would like to put the `env` tortoisehg path after the current `hg.exe` path.  Is that possible?

tukanos (commented):
The registry files have correct CRLF, but are UTF16LE with BOM encoded so probably that is the reason why the check failed.

bucket/tortoisehg.json
tukanos: https://github.com/lukesampson/scoop-extras/pull/5987/commits/7e19af42eb4a017ed0b194f7dc7ff19d0c1bb4f6

bucket/tortoisehg.json
tukanos: https://github.com/lukesampson/scoop-extras/pull/5987/commits/f4a29088629e53a3f4e11ee0526d35485c61fce8

bucket/tortoisehg.json
tukanos: Added shortcuts and bins - https://github.com/lukesampson/scoop-extras/pull/5987/commits/58189921f751fde217ec817f16b23ad1a0937c6b

tukanos:
@Ash258 
I think your PowerShell CRLF test is incorrect.  Your test fails here:
```
at <ScriptBlock>, C:\projects\scoop\test\Import-File-Tests.ps1: line 68
68:             throw "The following files do not end with a newline: `r`n`r`n$($badFiles -join "`r`n")"
```

The correct test for the end lines should be along these lines (the file is the registry file I'm trying to push):

`$crlf = [System.IO.File]::ReadAllBytes('C:\prg\scoop\missing_software\scripts\tortoisehg\thg_cmenu_uk.reg')`
Then the test that it continas both cr and lf would be:
`($crlf -contains 10) -And ($crlf -contains 13)`

You could do it the powershell only way - but it is like 100x slower:

`(Get-Content C:\prg\scoop\missing_software\scripts\tortoisehg\thg_cmenu_uk.reg -Raw) -match "\r\n$"`

bucket/tortoisehg.json
```diff
@@ -18,12 +18,42 @@
     },
     "extract_dir": "PFiles\\TortoiseHg",
     "bin": [
-        "thgw.exe"
+        "thg.exe",
+        "thgw.exe",
+        "hg.exe",
+        "Pegeant.exe",
```
tukanos, line 24: Must be Pageant.exe instread of Pegeant.exe!

bucket/tortoisehg.json
```diff
@@ -0,0 +1,74 @@
...
+            "lib\\TortoisePlink.exe",
+            "Command-line connection utility (based on PuTTY Plink)"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/",
+        "regex": "tortoisehg-([\\d.]+)-x[468]{2}\\.msi"
+    },
+    "autoupdate": {
+    "note": "Thanks for using autoupdate, please test your updates!",
```
rashil2000, line 64: Is this required?

bucket/tortoisehg.json
```diff
@@ -0,0 +1,74 @@
...
+    "architecture": {
+        "64bit": {
+            "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/tortoisehg-$version-x64.msi"
+        },
+        "32bit": {
+            "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/tortoisehg-$version-x86.msi"
+        }
+    }
```
rashil2000, line 72:
```suggestion
        "architecture": {
            "64bit": {
                "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/tortoisehg-$version-x64.msi"
            },
            "32bit": {
                "url": "https://www.mercurial-scm.org/release/tortoisehg/windows/tortoisehg-$version-x86.msi"
            }
        }
```

indentation

rashil2000 (changes requested):
Add note on how to use the reg files

bucket/tortoisehg.json
tukanos: I don't see any benefit by adding the indentation.  The `architecture` tag is correctly indented.

bucket/tortoisehg.json
tukanos: You mean the note?

rashil2000: Yes

bucket/tortoisehg.json
rashil2000: It is a nested object, it needs to be indented. 
image

tukanos: I know that and I agree.  The point I'm trying to make is that I already see it indented.

image

rashil2000: Do you not notice the second last and third last closing bracket? They are in the same column.

rashil2000:
What are the reg files there for? How will a user use them?

bucket/tortoisehg.json
tukanos: @rashil2000 ah, sorry about did not notice :).  Thank you.

tukanos:
@rashil2000 Registry files are translations of the context menu TortoiseHG - each file represents one language.  If you right click on a repository or a file in a repository you have a context menu which is translated into the language Windows is using.

You do not need to do anything with the registry files they will be installed automagically.  These files were extracted from the original msi file.

rashil2000:
I see. But if they are already present in the msi, why do we need to put them in the scripts folder? The manifest does not specify how to use these files in the scripts folder, and I assume the msi will have no knowledge of them because it's located outside the application folder.

tukanos:
@rashil2000 haha :), good point.  I have created them during the msi package disassembly and use them in the first versions.  However, you are right that they are not used in the current version so I can safely remove them.

The point about the  PowerShell CRLF test not being accurate is still valid thou (https://github.com/ScoopInstaller/Extras/pull/5987#issuecomment-826597566)

rashil2000:
> The point about the PowerShell CRLF test not being accurate is still valid thou (#5987 (comment))

I see. Can you open as issue in ScoopInstaller/Scoop for this (after checking for any duplicates if they exist)?

tukanos:
> > The point about the PowerShell CRLF test not being accurate is still valid thou (#5987 (comment))
> 
> I see. Can you open as issue in ScoopInstaller/Scoop for this (after checking for any duplicates if they exist)?

I will check out the ScoopInstaller/Scoop.  I find it weird that registry files are used within scoop and the preferred encoding UCS-2 LE BOM (UTF-16) does not work correctly.  The previous ANSI encoding was moved to UTF-16 when the registry version was updated from version 4.0 -> 5.0.  The UTF-8, UTF-8 BOM will work in most cases, but the correct and preferred encoding is UCS-2 LE BOM (UTF-16).  For more you can see this [gist](https://gist.github.com/SalviaSage/8eba542dc27eea3379a1f7dad3f729a0).

rashil2000:
The latest version is 6.0.0. Also, you need to use `"reverse": true` in the checkver object.

tukanos:
> 

The 6.0.0 version was released yesterday so I have not updated the package yet :).  It is not even on the project page yet [tortoisehg](https://tortoisehg.bitbucket.io/download/index.html).

I see the `"reverse": true` for the last occurrence.  Will add it now.

tukanos:
@rashil2000 I have done the update to 6.0.0 and the `"reverse": true`.  Now it should be complete.

tukanos:
@rashil2000 Thank you for your patience and merging!

rashil2000:
No worries :)

PR #5992, opened 2021-04-08
add miktex

ShoofLLC:
This manifest adds miktex, the installation shows a window but it works well, alternatively using the CLI tool would work too but I haven't tested that.


Ash258:
https://github.com/ScoopInstaller/Main/blob/master/bucket/latex.json

PR #5996, opened 2021-04-09
fat32format-cli: Add version 1.07

kriswilk:


Ash258:
The name of application is not `fat32format-cli`. Naming manifest like that does not make sense

1. Move this PR into main bucket
2. Rename the manifest to original name without -cli prefix

kriswilk:
> The name of application is not `fat32format-cli`. Naming manifest like that does not make sense
> 
> 1. Move this PR into main bucket
> 2. Rename the manifest to original name without -cli prefix

Sorry. I created 2 PRs for two sibling apps from the same author. One is a CLI tool (fat32format.exe) and the other is the GUI equivalent (guiformat.exe). I was trying to keep these very closely-related apps similarly related in scoop. That's why I named them "fat32format-cli" and "fat32format-gui".

But I do understand, and the CLI app is consistent with the main bucket so I'll give it a go there...thanks for the guidance.

PR #5997, opened 2021-04-09
fat32format-gui: Add version 1.01

kriswilk:


bucket/fat32format-gui
```diff
@@ -0,0 +1,15 @@
...
+    "description": "A GUI FAT32 format tool that supports volumes greater than 32 GB.",
+    "homepage": "http://ridgecrop.co.uk/",
+    "license": "Unknown",
+    "url": "http://ridgecrop.co.uk/guiformat.exe",
+    "hash": "647fb4f5108af632c3d52fec34934922c50c70585697504e92fb80b3b7d05ee3",
+    "bin": "guiformat.exe",
+    "shortcuts": [
+        [
+            "guiformat.exe",
+            "fat32format-gui"
```
Ash258, line 12:
```suggestion
            "FAT32 Format"
```

bucket/fat32format-gui
```diff
@@ -0,0 +1,15 @@
+{
+    "version": "1.01",
+    "description": "A GUI FAT32 format tool that supports volumes greater than 32 GB.",
+    "homepage": "http://ridgecrop.co.uk/",
+    "license": "Unknown",
+    "url": "http://ridgecrop.co.uk/guiformat.exe",
+    "hash": "647fb4f5108af632c3d52fec34934922c50c70585697504e92fb80b3b7d05ee3",
+    "bin": "guiformat.exe",
```
Ash258, line 8:
```suggestion
```

Ash258 (changes requested):
`git mv bucket/fat32format-gui bucket/fat32format-gui.json`

kriswilk:
> `git mv bucket/fat32format-gui bucket/fat32format-gui.json`

Sorry. 🤦‍♂️

Thanks for the other suggestions. Still new to scoop and didn't know that "bin" could be omitted in this case.

kriswilk:
On second thought, I re-added the shim for "guiformat.exe" since it's still nice to be able to launch the tool from the command line.

Ash258:
No it is not. If there are no command line arguments available it does not make sense to add it.

kriswilk:
> No it is not. If there are no command line arguments available it does not make sense to add it.

OK, fair enough. I've removed it. But I have launched it more than a few times from my terminal that way.

bucket/fat32format-gui.json
```diff
@@ -0,0 +1,14 @@
+{
+    "version": "1.01",
+    "description": "A GUI FAT32 format tool that supports volumes greater than 32 GB.",
+    "homepage": "http://ridgecrop.co.uk/",
+    "license": "Unknown",
+    "url": "http://ridgecrop.co.uk/guiformat.exe",
+    "hash": "647fb4f5108af632c3d52fec34934922c50c70585697504e92fb80b3b7d05ee3",
```
Ash258, line 7:
```suggestion
    "homepage": "http://ridgecrop.co.uk",
    "license": "Unknown",
    "architecture": {
        "64bit": {
            "url": "http://ridgecrop.co.uk/guiformat-x64.Exe#/guiformat.exe",
            "hash": "13d045f582d559c830f2e30f835ac56803844825fe6152d2d5f78367b19edb9a"
        },
        "32bit": {
            "url": "http://ridgecrop.co.uk/guiformat.exe",
            "hash": "647fb4f5108af632c3d52fec34934922c50c70585697504e92fb80b3b7d05ee3"
        }
    },
```

Ash258 (changes requested):

kriswilk:
Thought of adding the 64-bit version but was hesitant since the app author states:

> There is an x64 build of Guiformat here. However the x86 build works fine on both x86 and x64 and you should use that.

No idea why that recommendation was made but I was trying to respect it. Personally, I agree it makes sense to offer both.

Ash258 (approved):

PR #6017, opened 2021-04-11
tomcat: Persist conf and webapps folders

### LINKED ISSUE #6010: Tomcat: conf and webapps directories are not persisted
The `conf` and `webapps` directories in Apache Tomcat are not persisted. This means you will lose custom configuration files and web apps after upgrading - permanently after a `cleanup`. In comparison the Apache installation does persist `htdocs` and `conf`.

MunsterPlop:
* closes #6010

### DIFF
--- a/bucket/tomcat.json
+++ b/bucket/tomcat.json
@@ -22,6 +22,10 @@
         "CATALINA_HOME": "$dir",
         "CATALINA_BASE": "$dir"
     },
+    "persist": [
+        "conf",
+        "webapps"
+    ],
     "checkver": {
         "url": "https://www.apache.org/dist/tomcat/tomcat-10/?C=M;O=D",
         "regex": "v([\\d.]+)"


pratikpc:
@MunsterPlop I don't use Tomcat but could you add some link to docs which mention these are to be persisted?

MunsterPlop:
@pratikpc I am not an active Tomcat user either but looking at:

[Apache Tomcat 10 Configuration Reference](https://tomcat.apache.org/tomcat-10.0-doc/config/index.html)

and

[Application Developer's Guide](https://tomcat.apache.org/tomcat-10.0-doc/appdev/deployment.html)

It looks like Tomcat uses its installation directory as a base for its config files by default,
so I believe it makes sense to persist these. Please correct me if I'm wrong.

pratikpc (approved):

issaclin32 (approved):

PR #6029, opened 2021-04-13
cherrytree: Update to version 0.99.44.0

### LINKED ISSUE #5145: Cherrytree autoupdate broken

shaqo88:
- Closes #5145


Ash258 (changes requested):
Configuration files were changed. Persisting needs to be reconfigured and migrated

shaqo88:
> Configuration files were changed. Persisting needs to be reconfigured and migrated

Can you explain what I need to do?

shaqo88:
Can anyone help with this?

> Configuration files were changed. Persisting needs to be reconfigured and migrated

issaclin32:
I updated the manifest myself, because the PR has not been active in a while.

rashil2000:
I think the `*_portable.7z` does have a portable config. See: 

image

issaclin32:
> I think the `*_portable.7z` does have a portable config. See:
> 
> image

Thanks. It looks like if we put `config.cfg` in `$dir`, then `$LocalAppData` will not be used.
Sorry I didn't check carefully.

rashil2000 (approved):

rashil2000:
No worries :)

PR #6049, opened 2021-04-16
flutter: remove dependencies and add path

zoritle:
flutter support desktop and web, which is not depend on android sdk and jdk.
According to official guide, the recommended way to set up android develop environment is using android studio.
Using shim for flutter.bat will cause error of vscode extension `unable to find flutter.bat`.

### DIFF
--- a/bucket/flutter.json
+++ b/bucket/flutter.json
@@ -3,14 +3,17 @@
     "description": "Google’s mobile app SDK for crafting high-quality native interfaces on iOS and Android",
     "homepage": "https://flutter.dev",
     "license": "BSD-3-Clause",
-    "depends": [
-        "android-sdk",
-        "java/adopt8-hotspot"
-    ],
     "suggest": {
         "Visual Studio Code with Flutter Extension": [
             "vscode",
             "vscode-portable"
+        ],
+        "Android SDK and JDK":[
+            "android-sdk",
+            "java/adoptopenjdk-hotspot"
+        ],
+        "Android Studio":[
+            "android-studio"
         ]
     },
     "url": [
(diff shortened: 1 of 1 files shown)


zoritle:
Sorry for my mistake,leaving a trailing comma.

bucket/flutter.json
```diff
@@ -27,11 +30,7 @@
         "Write-Host 'Some licenses need to be accepted before developing. It is recommended to do by running ''flutter doctor --android-licenses''.' -ForegroundColor Yellow",
         "flutter doctor"
     ],
-    "bin": [
-        "bin\\flutter.bat",
-        "flutter-dev-setup.ps1"
-    ],
-    "env_add_path": "bin\\cache\\dart-sdk\\bin",
+    "env_add_path": ["bin","bin\\cache\\dart-sdk\\bin"],
```
pratikpc, line 33: @zoritle can you separate the two PRs?

Replacing depends with suggest would make it good to go very quickly but I think we would like more discussion when it comes to env_add_path and bin removals

bucket/flutter.json
```diff
@@ -3,14 +3,17 @@
     "description": "Google’s mobile app SDK for crafting high-quality native interfaces on iOS and Android",
     "homepage": "https://flutter.dev",
     "license": "BSD-3-Clause",
-    "depends": [
-        "android-sdk",
```
pratikpc, line 7: Also add android-clt

bucket/flutter.json
```diff
@@ -3,14 +3,17 @@
...
-        "java/adopt8-hotspot"
-    ],
     "suggest": {
         "Visual Studio Code with Flutter Extension": [
             "vscode",
             "vscode-portable"
+        ],
+        "Android SDK and JDK":[
+            "android-sdk",
+            "java/adoptopenjdk-hotspot"
```
pratikpc, line 13: Separate JDK and Android suggestion

bucket/flutter.json
```diff
@@ -3,14 +3,17 @@
...
-        "android-sdk",
-        "java/adopt8-hotspot"
-    ],
     "suggest": {
         "Visual Studio Code with Flutter Extension": [
             "vscode",
             "vscode-portable"
+        ],
+        "Android SDK and JDK":[
+            "android-sdk",
```
pratikpc, line 12: Also add android-clt to Android suggestion

pratikpc:
@zoritle there seem to be conflicts.

I have made some suggested changes for suggests.

Could you create 2 new PRs, one for env and another for changes from depend to suggest?

Thank you!!!

PR #6071, opened 2021-04-21
nimbletext@3.1.0.33266: Adopt HTTPS urls

kriswilk:
This app currently fails to install. The manifest is broken in a few ways:

- Download URL no longer supports http. Must use https.
- Current downloadable version is 3.1.0.33266, not 3.1.0.17465 because....
- The version notes page used for checkver/autoupdate is broken (for quite a while), so useless.

This PR fixes the above issues.

### DIFF
--- a/bucket/nimbletext.json
+++ b/bucket/nimbletext.json
@@ -1,9 +1,9 @@
 {
-    "version": "3.1.0.17465",
+    "version": "3.1.0.33266",
     "description": "Text manipulation and code generation tool.",
     "homepage": "https://nimbletext.com/",
     "license": "Unknown",
-    "url": "http://nimbletext.com/Download/NimbleText.exe",
+    "url": "https://nimbletext.com/Download/NimbleText.exe",
     "hash": "710ee5e8b449aca2b54c45c426f3e8316da31a78463cae373b40f92a161380f6",
     "bin": "NimbleText.exe",
     "shortcuts": [
@@ -11,12 +11,5 @@
             "NimbleText.exe",
             "NimbleText"
         ]
-    ],
-    "checkver": {
-        "url": "http://nimbletext.com/Version/Notes",
-        "regex": "Version ([\\d.]+)"
-    },
-    "autoupdate": {
-        "url": "http://nimbletext.com/Download/NimbleText.exe"
-    }
+    ]
 }
--- a/bucket/nimbletext.json
+++ b/bucket/nimbletext.json
@@ -1,7 +1,7 @@
 {
     "version": "3.1.0.33266",
-    "description": "Text manipulation and code generation tool.",
-    "homepage": "https://nimbletext.com/",
+    "description": "Text manipulation and code generation tool",
+    "homepage": "https://nimbletext.com",
     "license": "Unknown",
     "url": "https://nimbletext.com/Download/NimbleText.exe",
     "hash": "710ee5e8b449aca2b54c45c426f3e8316da31a78463cae373b40f92a161380f6",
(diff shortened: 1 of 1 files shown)


kriswilk:
Addendum: More specifically, the [version notes](https://nimbletext.com/version/notes) page actually DOES show the latest version number but not at the top of the page, so the checker regex will never find it. Until this malformed page is fixed, autoupdate needs to be disabled.

bucket/nimbletext.json
```diff
@@ -1,22 +1,15 @@
...
         ]
-    ],
-    "checkver": {
-        "url": "http://nimbletext.com/Version/Notes",
-        "regex": "Version ([\\d.]+)"
-    },
-    "autoupdate": {
-        "url": "http://nimbletext.com/Download/NimbleText.exe"
-    }
+    ]
```
Ash258, line 14:
```suggestion
    ],
    "checkver": {
        "url": "https://nimbletext.com/Version",
        "regex": "([\\d.]+),\\d+",
        "reverse": true
    }
```

bucket/nimbletext.json
```diff
@@ -1,22 +1,15 @@
 {
-    "version": "3.1.0.17465",
+    "version": "3.1.0.33266",
     "description": "Text manipulation and code generation tool.",
     "homepage": "https://nimbletext.com/",
```
Ash258, line 4:
```suggestion
    "description": "Text manipulation and code generation tool",
    "homepage": "https://nimbletext.com",
```

Ash258 (changes requested):

bucket/nimbletext.json
kriswilk: That URL for checkver is new to me, is it published somewhere by the author? The previous manifest used

https://nimbletext.com/version/notes

Second, BOTH sources of version information list 3.1.0.17465 as the top/bottom of the list, out of order with the actual released file, version 3.1.0.33266.

I'm waiting for a response from the author about what the heck is going on with the version history but the only actual file that you can download is 3.1.0.33266. That's why I suggested removing the version check/autoupdate, at least temporarily.

What am I missing here?

secretGeek: NimbleText author here -- yeh this is my bad. I'll let you know when I've fixed it (a lot on my plate at this minute.........)

Incidentally, "https://nimbletext.com/Version" is the better page to use for checking for updates (it's what NimbleText itself uses.)

This mistake hasn't broken nimbletext's check for updates -- by chance -- because the dev build # that was written into that Version file just happened to be a bit smaller number than the released version number.... (ends with .17465, smaller than .33266)  and nimbletext gets the last line from 

    https://nimbletext.com/Version

...and doesn't just check if it's different to itself, but is it a "larger" number.

Huge thanks to Kris for informing me!

secretGeek: Also re "That URL for checkver is new to me" -- yes it is undocumented, but if it helps the "scoops" of this world, i'll put a comment on the version/notes page that helps people such as yourself to find the secret best place to find the latest version.

kriswilk: @secretGeek Thanks for the quick response! I figured it was probably just a simple mistake.

@Ash258, you've suggested leaving the `checkver` block in place (with the preferred version history URL)...shouldn't it be removed until the version history is properly formatted? I thought it would be safest to remove it and restore the functionality once the author lets us know it's fixed.

secretGeek: ok - i think it's fixed, but please check it's good for you there. 
cheers, lb

kriswilk: > ok - i think it's fixed, but please check it's good for you there.

Looks good, thanks for the fix!

kriswilk:
OK, the author has fixed the version history. This PR now reflects the preferred version history URL, fixes the download link (to HTTPS), and we're on the correct version number so all should be good to go, I think!

PR #6083, opened 2021-04-23
Adds Blender LTS version 2.83.13  

vncastanheira:
Adds an option for the long-term support version of Blender.


Ash258:
Add it to https://github.com/ScoopInstaller/Versions

PR #6105, opened 2021-04-26
quicklook: Persist UserData folder

kriswilk:
The portable version of QuickLook now (actually, already with 3.6.10) stores user data in the `UserData` folder. Hence, it should be persisted.

Source: https://github.com/QL-Win/QuickLook/wiki/Differences-Between-Distributions

### DIFF
--- a/bucket/quicklook.json
+++ b/bucket/quicklook.json
@@ -11,6 +11,7 @@
             "QuickLook"
         ]
     ],
+    "persist": "UserData",
     "checkver": {
         "github": "https://github.com/QL-Win/QuickLook"
     },


Ash258 (changes requested):
Enabling persist is not enough. This will break the configuration for all users. Data migration to persist from system wide location is required

kriswilk:
A similar situation occurred when VS Code began offering "true" portable mode. I like how the "vscide-portable" manifest handles this transition by alerting the user and letting them decide whether or not to migrate settings.

I don't like the idea of *automatically* moving or copying settings from AppData.

Ash258:
Do it automatically or close PR

kriswilk:
A portable application should never interfere with the files of an "installed" version. The settings stored in AppData should not be assumed to be the ones that a portable install by scoop should use. The user may have installed the software themselves and it is unwise to move/copy them without consideration.

That is why vscode-portable does not automatically migrate settings, isn't it?

If you absolutely will not accept the more conservative, basic `persist`, then I will close this PR and file an issue instead. Because currently anyone upgrading this app using scoop will have their settings reset every single time the version changes.

Looking forward to your feedback.

Ash258:
Every manifest is doing the migration when possible. There is no reason to assume that user has some random installation from installer.

Breaking everyone installation and forcing them to take manual action is not acceptable.

kriswilk:
I understand the desire to streamline the user experience but have always preferred leaving the user in control of ambiguous situations. That said, if there is an established precedent for such migrations (they can't be that common??), I'll consider revising my PR.

Can you provide an example (or two?) of a manifest with a forced migration for reference? More generally-speaking, are there any documented conventions or best practices for manifests in scoop-extras?

kriswilk:
In addition to my last message, here's a situation I don't know how to handle:

What if a user has already upgraded to 3.6.10 or 3.6.11? These users would now be operating with new settings data within the scoop apps folder. Do we look for data in AppData and move it to the persistent folder, overwriting their active settings? Do we instead prefer the settings in scoop and ignore AppData (which may be the settings they actually want)?

This is kind of the definition of an edge case, which is why I wanted to let users sort it out, but if you can provide some guidance, I can force it one way or another and give the user notice of what has been done...

kriswilk:
I'd like to work on this PR before another release of the application to prevent further headaches for current users upgrading (who will lose their settings/plugins every time).

Is there a way to know if an upgrade is from version 3.6.9 or earlier? In that case I would be fine just copying from AppData to scoop's persisted UserData folder. But anyone on 3.6.10 or 3.6.11 will already have portable settings and plugins installed which shouldn't be replaced, even if AppData has settings in it.

I still don't know how this edge case is any different from how the vscode-portable manifest works?

PR #6145, opened 2021-04-30
amazon-workspaces: remove hash

FilipDeVos:
Amazon doesn't publish a hash so it doesn't make sense to keep it
in the file.


Ash258:
That is not how it works.

PR #6156, opened 2021-05-02
imdone: Add version 1.23.2

### LINKED ISSUE #7170: Do non-free-to-use (paid) apps belong to extras?
Triggered by @rashil2000 in [6156](https://github.com/ScoopInstaller/Extras/issues/6156#issuecomment-953852711_):

> I don't think there are any paid applications on Scoop

Do those applications belong to extras? What acceptance criteria shall be applied to freemium and non-free-to-use applications? And.. what happens when a once free to use application ceases to be so... so many questions.

A quick search turned out that there are paid applications in`extras`:

- [FileBot](https://github.com/ScoopInstaller/Extras/blob/master/bucket/filebot.json), via #2708.
- [wavebox10-pro](https://github.com/ScoopInstaller/Extras/blob/master/bucket/wavebox10-pro.json), via #3605
- [renamer](https://github.com/ScoopInstaller/Extras/blob/master/bucket/renamer.json), see #7189 

Though there are several PRs and requests where the concern has been raised which may suggest a discussion/clarification is due to avoid unconsistent criteria by our beloved maintainers (ie: FileBot is in via #2708, registry-workshop is not via #2630) :

- PR
  - imdone, #6156
  - registry-workshop, #5063; which was on #2630
- Requests
  - Glasswire, #6985

Further on, it was previously suggested (2019, by @rasa on #2564) to create a separate bucket (`nonfree`, `paid`). At the time @issaclin32 followed up about that (same #2564) but likes like it had not been done because of lack of permissions to create the repo under the ScoopInstaller org. Maybe this has changed after #4456.

This alternative is consistent with other open source practices, like Debian's non-free repository. It would then have to be decided if that bucket become also "well known" to scoop as `extras` is (specially if FileBot is moved to that eventual new bucket).

So, briefly:

- Do paid apps (trial versions, freemium software, etc) belong to `extras`?
- If not, shall `nonfree` exist inside ScoopInstaller?
- If yes, shall it be made well known to scoop as is `extras` to reduce friction (specialy for users of filebot and/or wavebox10-pro which would then have to be moved)

telenieko:
Simple and powerful kanban board built on top of plain text markdown files or code.

Regarding the criteria for inclussion in the main bucket: this is not a developer tool, and it is GUI based which excludes it from main.

Hence, because the first criteria is not met, this is proposed for inclusion in extras.

The project does not publish hashes, if this is an absolute requirement for inclusion into extras please say so I can open an issue there.

Licensing is not clear, yet,  PR kept as draft until https://github.com/imdone/imdone/issues/109 is closed.

Closes #7170

### DIFF
--- a/bucket/imdone.json
+++ b/bucket/imdone.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.17.4",
+    "version": "1.17.6",
     "description": "Simple and powerful kanban board built on top of plain text markdown files or code.",
     "homepage": "https://imdone.io/",
     "license": "Propietary",
@@ -15,8 +15,8 @@
     ],
     "architecture": {
         "64bit": {
-            "url": "https://imdone.io/downloads/imdone-1.17.4.exe#/dl.7z",
-            "hash": "d6e69b83ae4879c4a78f8d3d699d5efa6a904f8a3bbcadfe5d71795898bc1b0c",
+            "url": "https://imdone.io/downloads/imdone-1.17.6.exe#/dl.7z",
+            "hash": "2601f99fbb3bfd7ad53790f8b2e85ee9b146e4fa889340f864aad0b5ea164dbf",
             "installer": {
                 "script": "Expand-7zipArchive \"$dir\\$PLUGINSDIR\\app-64.7z\" \"$dir\""
             }
--- a/bucket/imdone.json
+++ b/bucket/imdone.json
@@ -1,0 +1,36 @@
+{
+    "version": "1.17.4",
+    "description": "Simple and powerful kanban board built on top of plain text markdown files or code.",
+    "homepage": "https://imdone.io/",
+    "license": "Propietary",
+    "extract_dir": "$PLUGINSDIR",
+    "pre_install": "Get-ChildItem \"$dir\" -Exclude 'app-64.7z' | Remove-Item -Force -Recurse",
+    "post_install": "Remove-Item \"$dir\\app-64.7z\" -Force -ErrorAction SilentlyContinue",
+    "bin": "imdone.exe",
+    "shortcuts": [
+        [
+            "imdone.exe",
+            "imdone"
+        ]
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://imdone.io/downloads/imdone-1.17.4.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


telenieko:
The license has been confirmed as propietary.

Ash258:
Proprietary means nothing when there is no license agreement, which user has to accept. Until there is no URL with the license file it is useless.

telenieko:
> Proprietary means nothing when there is no license agreement, which user has to accept. Until there is no URL with the license file it is useless.

Shall I write "Unknown" like in https://github.com/lukesampson/scoop-extras/blob/master/bucket/7tt.json ?

bucket/imdone.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": "1.22.2",
+    "description": "Simple and powerful kanban board built on top of plain text markdown files or code.",
+    "homepage": "https://imdone.io/",
+    "license": {
+        "identifier": "Unknown",
```
rashil2000, line 6:
```suggestion
        "identifier": "Proprietary",
```

rashil2000 (changes requested):
Modify the manifest according to the order of the fields here - https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md

Also, does `imdone.exe` binary take any commandline arguments?

rashil2000:
@issaclin32 could you please have a go at it?

issaclin32:
> @issaclin32 could you please have a go at it?

The manifest works fine.
*Persist* is not needed because config file is at `$Env:AppData\imdone`.

I am okay with adding this. It is up to your choice.

rashil2000:
> I am okay with adding this. It is up to your choice.

I'm okay with this too 👍

> image
> 
> * They have a portable release too. Use that.
> * Modify the manifest according to the order of the fields here - https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md
> * Remove `bin` field

Is the portable release functional?

issaclin32:
> Is the portable release functional?

Yep. It works totally fine. Also the config file location is the same. I think the only difference is that portable version will not register itself in "installed programs".

issaclin32 (approved):

rashil2000:
Great, thanks so much!

PR #6188, opened 2021-05-08
qt-creator: Add CDB debuger support

wasphin:


wasphin:
Hi @Ash258, does the failure mean that the ``autoupdate`` does not support urls? How to fix this? Please help, thanks.

PR #6210, opened 2021-05-13
gimp: Update to version 2.10.24

sdellenb:
Supersedes #4879

Regarding https://github.com/lukesampson/scoop-extras/pull/4879#pullrequestreview-506076351, there was no change since the 32bit install-script section was added in #1966, so 32bit arch probably broke long ago.


JesseRacine:
I have a question here about the Gimp installation. I am new to Scoop and just started using it recently. I noticed that Gimp was out of date, and I started to look at the install manifest but I got confused by all the "twain" related stuff in the manifest (i.e. lines 14 and 15).

I took a look at the installer for Gimp 2.10.24, extracted all the files, and I only found one twain.exe file in there. So for that reason I don't understand what all the renaming of different twain files is about.

2ion:
Can we get this merged? As with the previous bumps, the version bump is unrelated to the 32bit install issues.

stmax82:
Could we get this merged please? That 32 bit problem is unrelated to the update.. installation of the 64 bit version works fine. Who still uses 32 bits anyway? :p

sdellenb:
Updated to the latest GIMP version 2.10.28 (2021-09-16)

32bit installer fix in #6598 not part of this PR (because I can't test it).

PR #6217, opened 2021-05-15
Create AddressBook.json

Fatpandac:


Ash258:
Applications without English homepage are not accepted in official buckets

PR #6294, opened 2021-06-02
Update sublime-text.json

fiammanda:
Change version format so that scoop can recognize update.


Ash258:
Same schema was used for version 3. Since there is no semver versioning for sublime text it does not make sense to change it.

PR #6300, opened 2021-06-03
ilspy: Update to version 7.0.0.6488

TFisherInComm:
ILSpy was woefully out of date


Ash258 (changes requested):
- Checkver, autoupdate cannot work

TFisherInComm:
It was broken well before I got here, I can just remove it I guess

PR #6309, opened 2021-06-06
Creat AddressBook.json

Fatpandac:


Ash258:
Does not make sense to add project, which was created few days back and it does not have any stars, interactions, reputation and not verifiable manual builds. Easily infectable release artefacts.

Issue #4573, opened 2020-08-12
Zoom launching problem version 5.2.0 (with a temporary fix)


yiwenleo:
Zoom has trouble launching from the browser after installing by scoop.
It shows "Need new app to open zoommtg link" and direct to MS Store
The temp solution is to run  "Zoom_launcher.exe" under installation folder 
(`C:\Users\***\scoop\apps\zoom\5.2.0\Zoom_launcher.exe`)

FriendlyNeighborhoodShane:
I am not even able to launch zoom (5.4.7) from the executable. Zoom.exe says that the application was unable to start correctly.

Zoom_launcher.exe actually starts installing itself somewhere else.

FriendlyNeighborhoodShane:
Nevermind, figured it out.

For Windows versions older than 10, you have to extract all the dlls in `win10rt.7z` (should be in the app driectory) to the app directory.

brechtm:
> The temp solution is to run "Zoom_launcher.exe" under installation folder

That seems to install Zoom to the `C:\Users\<username>\AppData\Roaming\Zoom`, so you're no longer running the scope-managed version.

brechtm:
You can register the scoop-managed Zoom for zoomtg:// URLs by importing this into the registry:

```reg
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\SOFTWARE\Classes\zoommtg]
"URL Protocol"=""
@="Zoom (Application Name)"

[HKEY_CURRENT_USER\SOFTWARE\Classes\zoommtg\shell]

[HKEY_CURRENT_USER\SOFTWARE\Classes\zoommtg\shell\open]

[HKEY_CURRENT_USER\SOFTWARE\Classes\zoommtg\shell\open\command]
@="\"C:\\Users\\<USERNAME>\\scoop\\apps\\zoom\\current\\Zoom.exe\" --url=\"%1\""
````

Replace `<USERNAME>` with your username, save this as _zoommtg.reg_ and double-click to import it into the registry.

PR #6347, opened 2021-06-18
zoom: register/unregister zoommtg:// protocol handler

brechtm:
- add registry keys post-install
- remove these during uninstallation

Fixes: #4573
Related: lukesampson/scoop#3944, lukesampson/scoop#3412

### DIFF
--- a/bucket/zoom.json
+++ b/bucket/zoom.json
@@ -20,6 +20,19 @@
         "Expand-7zipArchive \"$dir\\Zoom.msi\" -Removal",
         "Remove-Item \"$dir\\Install*\""
     ],
+    "post_install": [
+        "Echo \"Registering zoommtg:// protocol handler\"",
+        "$regpath = \"HKCU:\\SOFTWARE\\Classes\\zoommtg\"",
+        "New-Item -Path $regpath -Value \"Zoom (Application Name)\" -Force | Out-Null",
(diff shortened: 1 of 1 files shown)


Ash258 (dismissed):
- Do not use `Echo`
- Global installation is not respected
- Lots of code repetition

brechtm:
Thanks for the quick review!

>     * Do not use `Echo`

Is there an alternative, or shouldn't it print anything?
 
>     * Global installation is not respected

Are you referring to writing to HKCU? That's user-specific, and Zoom is basically broken without setting up the URL protocol handler.

>     * Lots of code repetition

Partly addressed in second commit. Is it possible to avoid repeating the registry path in the uninstaller?

brechtm:
@Ash258 Friendly reminder to answer my questions. Thanks 🙏

brechtm:
Or perhaps @r15ch13?

altro3:
Hi! Want to ask about this pull request. Are you planning to merge it (or answer to questions)? Can't use zoom from scoop now without this fix.

issaclin32:
I think writing to *HCKU* is totally fine. This does not require admin rights, so it is unlikely to cause trouble during install/uninstall.

issaclin32 (approved):

PR #6367, opened 2021-06-24
openboard@1.6.1: Hash fix

liangjihua:


Ash258:
Does not make sense to open hash fixes manually, when there is bot who do it. Stop breaking the flow, when you have no extra information to give about the hash change

PR #6385, opened 2021-06-26
goldendict: Add persist

mickmetalholic:


Ash258 (changes requested):
- Where is documentation?
- This will break everyoneone's installation
- Do the migration from old locatin

PR #6411, opened 2021-06-30
updated powertoys version to 0.41.3

jackitsdotcom:


jackitsdotcom:
Only exe file available, no MSI provided by repo

IndraGunawan:
Hi, just wanted to check-in about this PR since it's been a bit

fn5:
> Hi, just wanted to check-in about this PR since it's been a bit

Should be good to go!

MarkKoz (changes requested):
The autoupdate URL also needs to be changed, since it seems the suffix changes from Win10 to amd64. Also, there should probably be separate links specified for 32-bit and 64-bit systems.

MarkKoz:
This is actually totally broken because of the change from an exe to an msi. Scoop just extracts MSIs rather than actually trying to install them. I get the following error:

```
Installing 'openvpn' (2.5.3) [64bit]
Loading OpenVPN-2.5.3-I601-amd64.msi from cache
Checking hash of OpenVPN-2.5.3-I601-amd64.msi ... ok.
Extracting OpenVPN-2.5.3-I601-amd64.msi ... done.
Running pre-install script...
Running installer... error.
ERROR Exception calling "Start" with "0" argument(s): "The system cannot find the file specified"
```

Even if the manifest is simplified to just allow contents to be extracted, OpenVPN won't work because it's not properly installed (missing registry entries, TAP drive not installed, etc).

PR #6461, opened 2021-07-08
playnite: Update to version 8.14

mgziminsky:
As of version 8.12, the author is no longer distributing the updates through the main site. The latest version does include a github release, but the previous 2 don't and the original update url is still being used and will likely be more reliable, so I left that as-is.

See https://playnite.link/forum/thread-613-post-3925.html#pid3925

### DIFF
--- a/bucket/playnite.json
+++ b/bucket/playnite.json
@@ -25,7 +25,8 @@
         "browsercache",
         "cache",
         "ExtensionsData",
-        "library"
+        "library",
+        "Themes"
     ],
     "checkver": {
         "github": "https://github.com/JosefNemec/Playnite"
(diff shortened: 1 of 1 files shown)


Lockszmith:
Came here after making a similar modification on my end, before creating a new pull request.
@mgziminsky already did the hard work.

mgziminsky:
Why did you remove Themes from persist again? Is there some reason you don't want that persisted?

Lockszmith:
> Why did you remove Themes from persist again? Is there some reason you don't want that persisted?

I'm thinking these two entries are the reason, it's a different type of modification.

https://github.com/lukesampson/scoop-extras/issues/4919
https://github.com/lukesampson/scoop-extras/pull/5088

so discussion should be moved to those items.

Not sure I completely understand, but I think Ash258 (bless his soul, he probably deals with a lot of crap from us users of application he doesn't use) stated that some automated test failed, and he never received any feedback.

PR #6468, opened 2021-07-09
Installs PEP-397 along with python, autoupdates registry entries.

JKorbelRA:
Rationale:
In the standard scoop Main python package, py.exe is not installed.
In the standard scoop Main python package, a manual action is needed for the newly installed python to be found by the py.exe launcher.

### DIFF
--- a/bucket/python-reg.json
+++ b/bucket/python-reg.json
@@ -1,0 +1,108 @@
+{
+    "version": "3.9.6",
+    "description": "A programming language that lets you work quickly and integrate systems more effectively. Includes py Python Launcher and updates registry automatically for it.",
+    "homepage": "https://www.python.org/",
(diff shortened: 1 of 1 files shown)


Ash258:
This is definitelly not for extras. Automatic manipulation with users registry are not what scoop is for. Use personal bucket for this.

JKorbelRA:
`This is definitely not for extras. Automatic manipulations with users registry are not what scoop is for. Use a personal bucket for this.`

Did that already (main python distribution is unusable with virtual environments). Unlike scoop, chocolatey for example does allow me to use py.exe (pep-397) conveniently including pep-514 (standard way to register python installation in windows registry). Would you have any philosophical guideline for main / extra repos, so that my future pull requests and your future time aren't in vain? Thanks.

PR #6469, opened 2021-07-09
V2RayN: Update manifest to support Xray-core

### LINKED ISSUE #5259: add v2rayN that uses Xray-core.

maokwen:
Fixes #5259
New depends: [main/xray](https://github.com/ScoopInstaller/Main/blob/master/bucket/xray.json)


bucket/v2rayn.json
```diff
@@ -16,6 +19,9 @@
         "if (!(Test-Path \"$persist_dir\\config.json\")) {",
         "    Copy-Item \"$(appdir v2ray $global)\\current\\config.json\" \"$dir\" | Out-Null",
         "}",
+        "if (!(Test-Path \"$persist_dir\\config.json\")) {",
```
Ash258, line 22: Does not make sense. The config from v2ray will be always overrided with the one from xray

bucket/v2rayn.json
```diff
@@ -24,6 +30,15 @@
         "            New-Item -ItemType HardLink -Path \"$dir\" -Name $name -Target \"$(appdir v2ray $global)\\current\\$name\" | Out-Null",
         "        }",
         "    }",
+        "}",
+        "foreach ($form in @('*.exe')) {",
```
Ash258, line 34: Same as previous comment.

Ash258 (changes requested):

bucket/v2rayn.json
MaoKwen: The executable file name from xray is different from the one from v2ray.

bucket/v2rayn.json
MaoKwen: If there is already a config form v2ray, the second configuration will not be copied.

bucket/v2rayn.json
```diff
@@ -16,6 +19,9 @@
         "if (!(Test-Path \"$persist_dir\\config.json\")) {",
         "    Copy-Item \"$(appdir v2ray $global)\\current\\config.json\" \"$dir\" | Out-Null",
         "}",
+        "if (!(Test-Path \"$persist_dir\\config.json\") and (Test-Path \"$(appdir xray $global)\\current\\config.json\")) {",
```
Ash258, line 22: This will be always true, since xray is automatically installed

Ash258 (changes requested):

bucket/v2rayn.json
MaoKwen: But If the user does not manually add config.json for xray, it wont be generated

Ash258: That file should be persisted in xray, otherwise it does not make sense as after xray update and then v2ray update it will not be working.

MaoKwen: I removed it.

issaclin32 (approved):

PR #6476, opened 2021-07-11
streamlink: Update 2.2.0 config path

404NetworkError:
See streamlink/streamlink#3766


Ash258 (changes requested):
So lets break everyone configuration without migration or anything to reflect the change?

issaclin32:
Replaced by #7361.

PR #6477, opened 2021-07-11
fork: Update to version 1.64.1

404NetworkError:
The version check fails because the RELEASE file hasn't been updated to reflect the new versions.  This is a valid version because the download URL works, and the [latest installer version](https://git-fork.com/update/win/ForkInstaller.exe) is 1.64.1 when checking the details of the Fork.exe file inside the installer.


404NetworkError:
Closed due to cf04a8e

PR #6483, opened 2021-07-12
Fixed hash code - openboard.json

mehran-njz:
the hash code is incorrect and makes error while installing. I have changed it to the correct sha256 hash code


Ash258:
Duplicate of #6409

PR #5224, opened 2020-12-02
archwsl: Prevents installation outside Scoop management

yuk7:
Prevents users from accidentally performing an installation outside of Scoop's management.
If the installation is performed outside Scoop's management, the data will be created in a location other than the persist directory.
In that case, the installation will be destroyed when the update is performed.

PR #6505, opened 2021-07-19
manjarowsl: Add version 20211101

sitiom:
Reopening from #6089 because I accidentally deleted my fork.

Manifest derived from [archwsl](https://github.com/lukesampson/scoop-extras/blob/master/bucket/archwsl.json) and #5224.


bucket/manjarowsl.json
```diff
@@ -0,0 +1,62 @@
...
+        "   $installable = $false",
+        "}",
+        "if ($null -eq (Get-Command 'wslconfig' -ErrorAction SilentlyContinue)) {",
+        "    warn 'WSL appears not to be enabled!'",
+        "    warn 'Run ''Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux'''",
+        "    warn 'from an elevated PowerShell. Restart your computer when prompted and reinstall ManjaroWSL.'",
+        "    $installable = $false",
+        "}",
+        "if ($installable) {",
+        "    $installable = !(Invoke-ExternalCommand \"$dir\\Manjaro.exe\" 'isregd')",
```
rashil2000, line 31: Put this line in the next if block itself.

bucket/manjarowsl.json
```diff
@@ -0,0 +1,62 @@
...
+        "}",
+        "if ($installable) {",
+        "    warn 'DO NOT KILL THE PROCESS. Installation is still running!'",
+        "    Copy-Item \"$dir\\Manjaro.exe\" \"$persist_dir\\data\\Manjaro.exe\"",
+        "    $res = Invoke-ExternalCommand \"$persist_dir\\data\\Manjaro.exe\" 'install', \"$dir\\rootfs.tar.gz\"",
+        "    Remove-Item \"$persist_dir\\data\\Manjaro.exe\" -Force",
+        "    if(!$res) { error 'ManjaroWsl installation failed!'; return }",
+        "}",
+        "Remove-Item \"$dir\\rootfs.tar.gz\" -Force"
```
rashil2000, line 40: Shouldn't this line be in the above if block too?

rashil2000 (changes requested):

bucket/manjarowsl.json
sitiom: No, I don't think it should be. There is a check in the first if block that changes `$installable` (also just copied that from #5224 so it should be suggested there too).

But I do think there is a better way to write that.

rashil2000:
There are certain differences in the `post_install` of ArchWSL and your manifest (towards the end). Why is that?

rashil2000:
I see that #5224 hasn't been merged yet. Let me review that first.

rashil2000 (approved):

rashil2000:
@sitiom The checkver is failing (note that the version string is different from ArchWSL)

sitiom:
> @sitiom The checkver is failing (note that the version string is different from ArchWSL)

My bad, didn't push the latest version earlier

issaclin32 (approved):
Tested. LGTM.

PR #6522, opened 2021-07-25
Add ffcss v0.2.0

SaifAqqad:


bucket/ffcss.json
```diff
@@ -0,0 +1,32 @@
+{
+    "version": "0.2.0",
+    "description": "A CLI to apply and configure FirefoxCSS themes.",
+    "homepage": "https://github.com/ewen-lbh/ffcss",
+    "license": "GPL-3.0",
```
Ash258, line 5: Invalid SPDX identifier

bucket/ffcss.json
```diff
@@ -0,0 +1,32 @@
...
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/ewen-lbh/ffcss/releases/download/v0.2.0/windows-64bit.zip",
+            "hash": "F8CAE5F1A35C95E3A06F18C6711B0C092F554DF3FDDF747ABB2DC1D5E565F59F"
```
Ash258, line 9: Use checkver utility to provide proper hashes

bucket/ffcss.json
```diff
@@ -0,0 +1,32 @@
...
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/ewen-lbh/ffcss/releases/download/v$version/windows-64bit.zip"
+            },
+            "32bit": {
+                "url": "https://github.com/ewen-lbh/ffcss/releases/download/v$version/windows-32bit.zip"
+            }
+        }
```
Ash258, line 30: Add hash extraction from checksums file

bucket/ffcss.json
```diff
@@ -0,0 +1,32 @@
...
+        },
+        "32bit": {
+            "url": "https://github.com/ewen-lbh/ffcss/releases/download/v0.2.0/windows-32bit.zip",
+            "hash": "CCD39E7D39EAE226B861D31DED292B17A178EF85A5F0E9BF4766ADBBADC126BA"
+        }
+    },
+    "post_install": [
+        "New-Item -Path \"$env:userprofile\\.config\\\" -Name 'ffcss' -ItemType 'directory' -Force | Out-Null",
+        "Copy-Item -Path \"$dir\\themes\" -Destination \"$env:userprofile\\.config\\ffcss\" -Recurse -Force | Out-Null"
+    ],
```
Ash258, line 19: Why? This should be done automatically by the application

Willingly flooding userprofile with files is not what is acceptable here

Ash258 (changes requested):

bucket/ffcss.json
SaifAqqad: The application does not do that automatically.
Do you think displaying a note saying: `You need to copy the themes folder to userprofile` is a better alternative?
or should I add an uninstaller script that deletes them?

SaifAqqad: The application will creates the `user\.config\ffcss\themes` folder on the first run anyway, this just copies the default themes manifests to that folder

Ash258: /unsubscribe

I really do not like this approach. The installation/usage does not make any sense.

Considering this confusing installation steps:

image

- Why it needs to be executed before copying the themes?
	- When you clearly state that it does not work without it.
- Why I cannot configure from where the themes should be used when there is config file?
	- Global installations will make it unusable for all other users, except the one who installed it.

I am not going to take part in the merging of applications, which will not work in widely used installations. Not talking about the application is not known by that many users that it should be part in official bucket.

PR #6528, opened 2021-07-28
add noMeiryoUI

entr0pia:
No!! MeiryoUI is Windows system font setting tool on Windows 8.1/10.


Ash258:
Application without English homepage and documentation sites are not accepted in any official bucket.

PR #6557, opened 2021-08-03
waifu2x-ncnn-vulkan: Add version 20210521

ItsBasi:
A Vulkan API based waifu2x, usable via the command-line.


Ash258:
Does not make sense to add inactive fork of application, which is already in the bucket

PR #6560, opened 2021-08-04
Update picgo.json to 2.3.0-beta.7

chengnan049:
This version adds support for 32-bit Windows, but I don't know how to write it : (


Ash258:
There is no reason to prefer beta over stable version

PR #6565, opened 2021-08-05
vnote: add checkver field

zhoujin7:
https://github.com/lukesampson/scoop-extras/issues/6321

Vnote.json is missing checkver field. 
There is a new version of the vnote (2.10 --> 3.5.1)

### DIFF
--- a/bucket/vnote.json
+++ b/bucket/vnote.json
@@ -21,6 +21,9 @@
             "VNote"
         ]
     ],
+    "checkver": {
+        "github": "https://github.com/vnotex/vnote"
+    },
     "autoupdate": {
         "architecture": {
             "64bit": {


bucket/vnote.json
```diff
@@ -21,6 +21,9 @@
             "VNote"
         ]
     ],
+    "checkver": {
+        "github": "https://github.com/vnotex/vnote"
+    },
     "autoupdate": {
         "architecture": {
             "64bit": {
```
Ash258, line 29: Why do you propose random change, which cannot work?? There are no portable binaries. That is the reason why the autoupdate was removed...

image

Ash258 (changes requested):

PR #6567, opened 2021-08-05
audacity: Update to version 3.0.3

### LINKED ISSUE #6546: audacity@3.0.2: Download failed
Seems like the current mirror https://fossies.org, has removed the older build of 3.0.2 and replaced it with the newer one 3.0.3.

petlyh:
- Closes #6546

Update audacity to version 3.0.3 and add autoupdate

### DIFF
--- a/bucket/audacity.json
+++ b/bucket/audacity.json
@@ -15,12 +15,7 @@
             "extract_dir": "audacity-win-3.0.3-32bit"
         }
     },
-    "bin": [
-        [
-            "Audacity.exe",
-            "audacity"
-        ]
-    ],
+    "bin": "Audacity.exe",
     "shortcuts": [
         [
             "Audacity.exe",
@@ -30,7 +25,7 @@
     "persist": "Portable Settings",
     "checkver": {
         "url": "https://fossies.org/windows/misc/",
-        "regex": "[Aa]udacity-win-([\\d.]+)-64bit\\.zip\""
+        "regex": "[Aa]udacity-win-([\\d.]+)-32bit\\.zip\""
     },
     "autoupdate": {
         "architecture": {
--- a/bucket/audacity.json
+++ b/bucket/audacity.json
@@ -1,6 +1,6 @@
 {
     "version": "3.0.3",
-    "description": "An easy-to-use, multi-track audio editor and recorder.",
+    "description": "An easy-to-use, multi-track audio editor and recorder",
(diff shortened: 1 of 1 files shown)


cybernoid23:
If I am not mistaken, there is a slight error in the checkver regex. Shouldn't it be `"[Aa]udacity-win-([\\d.]+)-64bit\\.zip"`
 instead of `"[Aa]udacity-win-([\\d.]+)-64bit\\.zip\""` ?

petlyh:
The escaped quotation mark does seem to be unnecessary.

This is the HTML that the regex looks for:
`<A HREF="audacity-win-3.0.3-64bit.zip"></A></TD><TD> <A HREF="audacity-win-3.0.3-64bit.zip/" TITLE="Downloads, contents, member browsing &amp; more ..."><B>audacity-win-3.0.3-64bit.zip</B></A>`

With the current regex, only the first instance of the filename is matched.
By removing the quotation mark, all three instances will be matched.
However, as scoop uses the first match in the case of multiple matches
(unless reverse is used), the result seems to be the same either way.

I am not sure if it should be changed. Any thoughts?

bucket/audacity.json
```diff
@@ -1,20 +1,50 @@
...
+            "hash": "76fa4a81fba8e7b9ca705547e00e1cabcc6577ec1de8e131f342fc2fb7ef75a0",
+            "extract_dir": "audacity-win-3.0.3-32bit"
+        }
+    },
+    "bin": [
+        [
+            "Audacity.exe",
+            "audacity"
+        ]
+    ],
```
Ash258, line 23:
```suggestion
    "bin": "Audacity.exe",
```

bucket/audacity.json
```diff
@@ -1,20 +1,50 @@
...
-            "audacity.exe",
+            "Audacity.exe",
             "Audacity"
         ]
     ],
     "persist": "Portable Settings",
     "checkver": {
         "url": "https://fossies.org/windows/misc/",
-        "regex": "[Aa]udacity-win-([\\d.]+)-32bit\\.zip\""
+        "regex": "[Aa]udacity-win-([\\d.]+)-64bit\\.zip\""
```
Ash258, line 33: Not needed change

```suggestion
        "regex": "[Aa]udacity-win-([\\d.]+)-32bit\\.zip\""
```

Ash258 (changes requested):

PR #6584, opened 2021-08-08
foxit-reader: Deprecate manifest in favour of foxit-pdf-reader

gitolicious:
Foxit has renamed most of their products, see https://www.foxit.com/.

> Foxit has changed some product names. To learn more click here.

"Foxit Reader" is now called "Foxit PDF Reader". How are changes like this handled usually?
- create a separate manifest `foxit-pdf-reader.json`
- rename the manifest file
- just keep using the old manifest

### DIFF
--- a/bucket/foxit-reader.json
+++ b/bucket/foxit-reader.json
@@ -6,6 +6,7 @@
         "identifier": "Freeware",
         "url": "https://www.foxitsoftware.com/pdf-reader/eula.html"
     },
+    "notes": "Deprecated. Use extras/foxit-pdf-reader",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
1. Revert everything that you have done
2. Add `"note": "Deprecated. Use extras/foxit-pdf-reader"`
3. Come back when done

gitolicious:
Thanks, that sounds like a reasonable approach. It's done.

Ash258 (changes requested):
Move foxit-pdf-reader manifest into separate PR

gitolicious:
Sorry, I misunderstood. Done and added #6585

issaclin32 (approved):

PR #6631, opened 2021-08-15
openocd: Update to version 0.11.0-1

EzekielDaun:
Hello! I notice that the old [url](http://www.freddiechopin.info/en) no longer update. This PR is about this [git-repo](https://github.com/xpack-dev-tools/openocd-xpack) which I found from [OpenOCD homepage](http://openocd.org/pages/getting-openocd.html).

I'm not sure if "checkver" and "autoupdate" could work properly. Could someone check them? Sorry about that.


Ash258:
/unsubscribe

Manifest is fine. Will not be merge by me as these random unofficial distributions are not something I would like to see in official buckets

PR #6654, opened 2021-08-18
imhex: Update to version 1.8.1

DoveBoy:


Ash258 (changes requested):
Adjust previously persisted files to new format

DoveBoy:
> Adjust previously persisted files to new format

Please merge

rpinolf (changes requested):
1.10.1 is the actual release. Extract_dir "portable" is no longer in the new zip package so should be probably "".

PR #6728, opened 2021-08-28
cupscale: Add version 1.39.0f1

Optinux:


Ash258:
/unsubscribe

Manifest is working now.

I do not think random fork based on 20 more random forks should be added in official bucket.

bucket/cupscale.json
```diff
@@ -0,0 +1,33 @@
...
+        "github": "https://github.com/n00mkrad/cupscale",
+        "regex": "/Cupscale\\.([\\w.]+)\\.zip"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/n00mkrad/cupscale/releases/download/$majorVersion.$minorVersion/Cupscale.$version.zip"
+            }
+        },
+        "extract_dir": "Cupscale $version"
```
pratikpc, line 31: Usually extract_dir is present within the architecture block itself

Check https://github.com/ScoopInstaller/Extras/blob/master/bucket/staxrip.json for reference

rashil2000: This isn't really a requirement. (the global extract_dir is outside the arch block too, so this is fine)

pratikpc: Usually Scoop manifests keep both extract_dir and URL within the same blocks so we could do it for standardization.

rashil2000: If something is arch-specific, then only it should go inside the arch block. This is the standard to be followed. Otherwise it's common for both archs, so it should be outside.

pratikpc: But there is a single arch in this case.

Sane as Staxrip and a lot of other applications

rashil2000: Staxrip has the string `x64` in its extract_dir, so it's inside arch block.

rashil2000 (approved):

PR #6729, opened 2021-08-29
add goneovim

bandithedoge:


bucket/goneovim.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "v0.4.12",
+    "description": "Neovim GUI written in Golang, using a Golang qt backend",
+    "url": "https://github.com/akiyosi/goneovim/releases/download/v0.4.12/Goneovim-v0.4.12-win64.zip",
```
pratikpc, line 4: Could you add URL and Hash to architecture 64bit block?

This ensures the users using 32 bit are not inconvenienced

bucket/goneovim.json
```diff
@@ -0,0 +1,22 @@
...
+            "goneovim.exe",
+            "Goneovim"
+        ]
+    ],
+    "suggest": {
+        "Neovim": "neovim"
+    },
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://github.com/akiyosi/goneovim/releases/download/$version/Goneovim-$version-win64.zip"
```
pratikpc, line 20: Could you add URL to architecture 64bit block?

pratikpc: Check https://github.com/ScoopInstaller/Extras/blob/master/bucket/staxrip.json for example

pratikpc: Check https://github.com/ScoopInstaller/Extras/blob/master/bucket/staxrip.json for example

bucket/goneovim.json
pratikpc: Check https://github.com/ScoopInstaller/Extras/blob/master/bucket/staxrip.json for example

bucket/goneovim.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "v0.4.12",
```
pratikpc, line 2: Version here usually does not have the word v.

Please fix that.

It should be 0.4.12 only

bucket/goneovim.json
pratikpc: Further URL should have v$version I believe

PR #6730, opened 2021-08-29
Firefox*: persist 'distribution' folder

HUMORCE:
related:
https://support.mozilla.org/en-US/kb/customizing-firefox-using-policiesjson
https://github.com/mozilla/policy-templates/blob/master/README.md

### DIFF
--- a/bucket/firefox-developer.json
+++ b/bucket/firefox-developer.json
@@ -30,6 +30,7 @@
             "Firefox Developer Edition"
         ]
     ],
+    "persist": "distribution",
     "checkver": {
         "url": "https://aus5.mozilla.org/update/6/Firefox/60.0/_/WINNT_x86_64-msvc-x64/en-US/aurora/_/_/_/_/update.xml",
         "xpath": "/updates/update/patch/@URL",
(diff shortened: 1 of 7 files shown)


bucket/firefox-developer.json
```diff
@@ -30,6 +30,7 @@
             "Firefox Developer Edition"
         ]
     ],
+    "persist": "distribution",
```
pratikpc, line 33: Should we persist profile here as well and in the others?

I noticed firefox.json persists profile

HUMORCE: this PR for persist 'distribution' only.

issaclin32: `profile` folder only exists in portable version. It is designated in shortcut argument in the manifest.

https://github.com/ScoopInstaller/Extras/blob/a9301ccd584b3f5b20d2fe03fcdf9c4f2cfab0a9/bucket/firefox-portable.json#L21

issaclin32 (approved):

pratikpc (approved):

PR #6740, opened 2021-08-30
figma: Add version 104.1.0

joe733:


rashil2000:
The URL is not changing, so autoupdate is not required.

bucket/figma.json
```diff
@@ -0,0 +1,25 @@
...
+            "FigmaSetup.exe",
+            "Figma"
+        ]
+    ],
+    "checkver": {
+        "url": "https://www.figma.com/downloads",
+        "regex": ">v([\\d.]+)\\s+\\("
+    },
+    "autoupdate": {
+        "url": "https://desktop.figma.com/win/FigmaSetup.exe"
```
pratikpc, line 23: Does it not support different URLs for specific versions?

That sometimes helps with archiving.

rashil2000: They just keep updating the setup file and reuploading it. Cygwin does this too.

pratikpc: Yes but if there is a way to get versioned setups, it might make it easier for those who want to use older versions via scoop install <app>@<version>

A check should certainly be made once to find out if they provide versioned releases as well.

bucket/figma.json
```diff
@@ -0,0 +1,25 @@
+{
+    "version": "101.6",
+    "description": "Figma connects everyone in the design process so teams can deliver better products, faster.",
```
pratikpc, line 3: The description sounds like an advert. It doesn't help me understand what Figma does

bucket/figma.json
```diff
@@ -0,0 +1,25 @@
...
+    "homepage": "https://www.figma.com/",
+    "license": {
+        "identifier": "Shareware",
+        "url": "https://www.figma.com/summary-of-policy/"
+    },
+    "url": "https://desktop.figma.com/win/FigmaSetup.exe",
+    "hash": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+    "bin": "FigmaSetup.exe",
```
pratikpc, line 11: What does FigmaSetup exe do?

If its a setup, shouldn't it already be extracted?

Could you share screenshots?

bucket/figma.json
joe733: I guess it's all in there. "Figma somehow connects designers to develop better designs, faster". It's more than just a design tool, but maybe you can suggest a better one..

> PS: No affiliations to Figma, just a simple user here 👋🏻

bucket/figma.json
joe733: - It installs the designing tool into system.
- Yup, it's an executable not an archive (afaik they don't provide one).

image

bucket/figma.json
rashil2000: "A vector graphics editor and prototyping tool focusing on use in user interface and user experience design, with an emphasis on real-time collaboration"

bucket/figma.json
pratikpc: Could you run Scoop Install and then open the shortcut and show us what happens?

Scoop supports extraction from a lot of EXE based installers as well

joe733: image

is this what you're looking for?

niheaven:
@joe733 Please check if the new manifest works.

joe733:
@niheaven yes it does. (Tested after uninstalling old one)

image

PR #6786, opened 2021-09-05
Pyzo: Add version 4.11.7

maphew:
Review warranted as this is my first attempt at adding a package to Scoop. Local install of 64bit was successful on my machine with:

    scoop install .\bucket\pyzo.json

### DIFF
--- a/bucket/pyzo.json
+++ b/bucket/pyzo.json
@@ -7,8 +7,7 @@
         "64bit": {
             "url": "https://github.com/pyzo/pyzo/releases/download/v4.11.2/pyzo-4.11.2-win64.zip",
             "hash": "435cfb5ee617c30dac5fc7880a41c0046c3375067e7b24da3e7e0f121099ec6a",
-            "extract_dir": "pyzo-4.11.2",
-            "extract_to": "pyzo-4.11.2-win64"
+            "extract_dir": "pyzo-4.11.2"
         },
         "32bit": {
             "url": "https://github.com/pyzo/pyzo/releases/download/v4.11.2/pyzo-4.11.2-win32.zip",
--- a/bucket/pyzo.json
+++ b/bucket/pyzo.json
@@ -1,18 +1,18 @@
 {
-    "version": "4.11.2",
+    "version": "4.11.4",
     "description": "The Interactive editor for scientific Python",
     "homepage": "https://pyzo.org/",
     "license": "BSD-3-Clause",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/pyzo/pyzo/releases/download/v4.11.2/pyzo-4.11.2-win64.zip",
-            "hash": "435cfb5ee617c30dac5fc7880a41c0046c3375067e7b24da3e7e0f121099ec6a",
-            "extract_dir": "pyzo-4.11.2"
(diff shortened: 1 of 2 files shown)


bucket/pyzo.json
```diff
@@ -0,0 +1,44 @@
...
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/pyzo/pyzo/releases/download/v$version/pyzo-$version-win64.zip",
+                "extract_dir": "pyzo-$version",
+                "extract_to": "pyzo-$version-win64"
+            },
+            "32bit": {
+                "url": "https://github.com/pyzo/pyzo/releases/download/v$version/pyzo-$version-win32.zip",
+                "extract_dir": "pyzo-$version",
+                "extract_to": "pyzo-$version-win32"
```
pratikpc, line 40: Why is extract-to required?

pratikpc (changes requested):

bucket/pyzo.json
maphew: I don't that it is necessary. I added extract-to following the samples I studied. One thought is that since extract-dir on it's own is same for 32 and 64bit ensuring extract-to uses different names would allow side-by-side installs. I don't see a situation where I would do this myself though. I'm happy to remove it if this if it isn't standard practice.

pratikpc: It is not

maphew:
I believe the requested changes have been met. Is there a method to float this PR up for a re-check?

niheaven:
Please run checkver to update Pyzo to 4.11.4 and check its installation again.

maphew:
Hmmm. I'm not sure what to do with this:
~~~
» .\bin\checkver.ps1 -App Pyzo
pyzo: The remote server returned an error: (429) too many requests.
URL https://github.com/pyzo/pyzo/releases/latest is not valid
~~~

Going to the url in a browser works okay.

maphew:
Waiting a few minutes and then running `checkver -App Pyzo -Update` worked, result in commit 843d71a. After that bare `checkver -App Pyzo` worked also.

maphew:
thanks @niheaven . like this in 9d0d581?

niheaven:
Right👍🏻, I'll try to install it later 'cos I'm not with my PC.

niheaven:
64bit 4.11.4 is broken, isn't it?

maphew:
Oh! That's wierd. Running the pyzo shim does nothing. Running `%userprofile%\\scoop\apps\pyzo\current\pyzo.exe` yields:

>The application has failed to start because its side-by-side configuration is incorrect. Please see the application event log or use the command-line sxstrace.exe tool for more detail.

I manually downgraded pyzo.json to v4.11.2 and ran `scoop install bucket\pyzo.json` --> Pyzo shim and pyzo itself run correctly. 

I'll keep digging.

maphew:
Upstream confirmed this is an issue specific to v4.11.4, see the issue linked above). A new release is coming, pending confirmation the solution works for Mac. 

I rolled the json back to v4.11.2 and tested that it works locally. Apologies for not doing a full test with v4.11.4. This is one I could have caught before involving you.

niheaven:
Aha, so please wait until Pyzo update to 4.11.5 and then you could update the manifest.

maphew:
ok, seems to be all good now.

PR #6821, opened 2021-09-09
vale: Add version 2.10.6

arvikon:


bucket/vale.json
```diff
@@ -0,0 +1,19 @@
+{
+    "version": "2.10.6",
+    "description": "A syntax-aware linter for prose built with speed and extensibility in mind.",
+    "homepage": "https://docs.errata.ai/vale/about/",
+    "license": "MIT",
+    "url": "https://github.com/errata-ai/vale/releases/download/v2.10.6/vale_2.10.6_Windows_64-bit.zip",
```
Ash258, line 6: Clearly 64bit only

bucket/vale.json
```diff
@@ -0,0 +1,19 @@
...
+    "hash": "5ad04b54a61d8acf9cb636329721918fca407a0d072177c9b2638e91804805ac",
+    "bin": "vale.exe",
+    "checkver": {
+        "github": "https://github.com/errata-ai/vale"
+    },
+    "autoupdate": {
+        "url": "https://github.com/errata-ai/vale/releases/download/v$version/vale_$version_Windows_64-bit.zip",
+        "hash": {
+            "url": "$baseurl/vale_$version_checksums.txt",
+            "regex": "$sha256\\s+$basename"
```
Ash258, line 16: I do not beleive this is needed

Ash258 (changes requested):

bucket/vale.json
arvikon: Thx, updated  with instructions for 64-bit architecture only.

bucket/vale.json
arvikon: Makes sense. Removed regex.

Ash258 (approved):

PR #2730, opened 2019-08-24
[ADD] Syncplay 1.6.4a

anodium:
Client/server to synchronize media playback on mpv/VLC/MPC-HC/MPC-BE on many computers

https://syncplay.pl/

PR #6824, opened 2021-09-09
syncplay: Add version 1.6.9

Duckle29:
- Closes #2730

syncplay allows far away friends to sync their media players

### DIFF
--- a/bucket/syncplay.json
+++ b/bucket/syncplay.json
@@ -1,0 +1,21 @@
+{
+    "version": "1.6.9",
+    "description": "Solution to synchronize video playback across multiple instances of mpv, VLC, MPC-HC, MPC-BE and mplayer2 over the Internet.",
+    "homepage": "https://syncplay.pl/",
+    "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


bucket/syncplay.json
```diff
@@ -0,0 +1,21 @@
+{
+    "version": "1.6.9",
+    "description": "Solution to synchronize video playback across multiple instances of mpv, VLC, MPC-HC, MPC-BE and mplayer2 over the Internet.",
+    "homepage": "https://syncplay.pl/",
+    "license": "Apache-2.0",
+    "url": "https://github.com/Syncplay/syncplay/releases/download/v1.6.9/Syncplay-Portable-Client-1.6.9.zip",
```
Ash258, line 6: Why `portable`?

What does it mean? What needs to be persisted?

Duckle29: Their page says the non-portable version stores connection settings in appdata

Duckle29: Added persist for the client settings

PR #6826, opened 2021-09-09
add envyupdate

fyr77:
EnvyUpdate is a small update checker for NVidia GPUs.


Ash258:
Same as https://github.com/lukesampson/scoop-extras/pull/6310#issuecomment-860171782

PR #6841, opened 2021-09-11
librehardwaremonitor: Add version 0.8.8

ronhuang:
Manifest is adapted from openhardwaremonitor.


Ash258:
/unsubsribe

Adding forks of actively maintained applications is not something that should be done in the Extras/Main buckets.

PR #6842, opened 2021-09-11
rpg-cli: Add version 1.0.0

ronhuang:
rpg-cli is a minimalist computer RPG written in Rust. Its command-line interface can be used as a cd replacement where you randomly encounter enemies as you change directories.


Ash258:
https://github.com/Calinou/scoop-games

PR #6858, opened 2021-09-14
fclones: Update to version 0.15.0

dyphire:


Ash258:
https://github.com/lukesampson/scoop-extras/blob/master/bucket/fclones.json

PR #6864, opened 2021-09-15
fclones: Update to version 0.15.0

dyphire:


Ash258:
What is the point of openint 20 same PRs? And then close them after 1 hour?

dyphire:
> What is the point of openint 20 same PRs? And then close them after 1 hour?

sorry,If this is causing trouble,it will be closed and don't open again

PR #6868, opened 2021-09-16
coretemp@1.7.1: Persist adjustment

issaclin32:
Since version 1.7, **CoreTemp** will crash on startup if encounters an empty config file (`CoreTemp.ini`).

This fixes the problem.

### DIFF
--- a/bucket/coretemp.json
+++ b/bucket/coretemp.json
@@ -16,14 +16,19 @@
             "hash": "5b71c1109d8cacc3763ba4dd470d6ec5f426c69a4b675121d1557e9b54e2c0db"
         }
     },
-    "pre_install": "if (!(Test-Path \"$dir\\CoreTemp.ini\")) { New-Item \"$dir\\CoreTemp.ini\" | Out-Null }",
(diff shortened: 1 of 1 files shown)


Ash258 (changes requested):
Revert everything what you have done and change pre_install block to:

```json

    "pre_install": [
        "if (!(Test-Path \"$persist_dir\\CoreTemp.ini\")) {",
        "    Set-Content \"$dir\\CoreTemp.ini\" @('[General]', 'AutoUpdateCheck=0;') -Encoding 'Ascii'",
        "}"
    ],
```

image

Ash258:
@issaclin32 Did you tested on clean environment after the newest changes?

issaclin32:
> @issaclin32 Did you test on clean environment after the newest changes?

I found the problem on a new PC, and re-produced the problem on another new PC.
The CPUs are `Intel i7-9700F` and `AMD R9 5950X`.

However I cannot re-produce the problem on a VM (not sure about the reason)

Ash258:
I really do not understand what you are trying to say.

Did it work with this fixed manifest or not?

PR #6869, opened 2021-09-17
fix picgo

huangnauh:


Ash258:
By default it will be in chinese. Does not make sense to update to this version

PR #6872, opened 2021-09-18
firefox-portable: Add scripts to set as default browser

### LINKED ISSUE #4090: set firefox as default browser
I just find Windows 10 does not support non-installed programs in **System - Default apps**. If you use Google Chrome and choose "set as default" in Chrome's setting, it will write itself into registry and you can find Chrome in **System - Default apps**. While if one installs Firefox using scoop and click on "set as default" in Firefox setting, one will be navigated into the **System - Default apps**, like . But Firefox does not write itself into registry and you can not find Firefox in **Choose an application** dialog.

I find one solution [here](https://www.winhelponline.com/blog/register-firefox-portable-with-default-programs-in-vista/). It works but it is not open source. Anyone has other ideas? I guess a post-installed script may help.

QingKongBaiYu:
fix #4090 .

### DIFF
--- a/bucket/firefox-portable.json
+++ b/bucket/firefox-portable.json
@@ -3,14 +3,31 @@
     "description": "Popular open source web browser",
     "homepage": "https://www.mozilla.org/firefox/",
     "license": "MPL-2.0",
+    "notes": "Set Firefox as default browser by running: \"$dir\\firefox-install.reg\"",
     "architecture": {
         "64bit": {
-            "url": "https://archive.mozilla.org/pub/firefox/releases/92.0/win64/en-US/Firefox%20Setup%2092.0.exe#/dl.7z",
-            "hash": "sha512:ca0855eddfd4c9d0d0116f8993fb3a1e947cd7d91b391d7913fbe36b91f238a5025fcb7129234a7dc07763540d33b5233b592737574db7999f29df2c5c21cf0f"
+            "url": [
+                "https://archive.mozilla.org/pub/firefox/releases/92.0/win64/en-US/Firefox%20Setup%2092.0.exe#/dl.7z",
+                "https://raw.githubusercontent.com/lukesampson/scoop-extras/master/scripts/firefox-portable/firefox-install.reg",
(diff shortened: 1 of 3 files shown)


bucket/firefox-portable.json
```diff
@@ -21,6 +38,28 @@
...
     ],
+    "post_install": [
+        "if (Test-Path \"$dir\\firefox-install.reg\") {",
+        "  $fpath = \"$dir\".Replace('\\', '\\\\')",
+        "  $content = Get-Content \"$dir\\firefox-install.reg\"",
+        "  $content = $content.Replace('<path>', $fpath)",
+        "  if ($global) {",
+        "    $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "  }",
+        "  $content | Set-Content -Path \"$dir\\vscode-install-context.reg\"",
```
Ash258, line 49: nonsense

bucket/firefox-portable.json
```diff
@@ -21,6 +38,28 @@
...
+        "  $content | Set-Content -Path \"$dir\\firefox-install.reg\"",
+        "}",
+        "if (Test-Path \"$dir\\firefox-uninstall.reg\") {",
+        "  $fpath = \"$dir\".Replace('\\', '\\\\')",
+        "  $content = Get-Content \"$dir\\firefox-uninstall.reg\"",
+        "  $content = $content.Replace('<path>', $fpath)",
+        "  if ($global) {",
+        "    $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "  }",
+        "  $content | Set-Content -Path \"$dir\\vscode-uninstall-context.reg\"",
```
Ash258, line 59: nonsense

Ash258 (changes requested):

Ash258:
/unsubscribe

I do not think more scripts like this should be introduced in official buckets.

QingKongBaiYu:
> /unsubscribe
> 
> I do not think more scripts like this should be introduced in official buckets.

sorry for the mistake!

PR #6874, opened 2021-09-19
quicklook: persist 'UserData' directory

HUMORCE:
persist **UserData** directory for keeping the installed plugins.

project doc.:
https://github.com/QL-Win/QuickLook/wiki/Differences-Between-Distributions

### DIFF
--- a/bucket/quicklook.json
+++ b/bucket/quicklook.json
@@ -11,6 +11,7 @@
             "QuickLook"
         ]
     ],
+    "persist": "UserData",
     "checkver": {
         "github": "https://github.com/QL-Win/QuickLook"
     },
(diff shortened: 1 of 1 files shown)


Ash258 (dismissed):
No data migration from appdata

HUMORCE:
i considered for, but the ​application no longer reads any data from %appdata% since 2020-09-20 (version: 3.6.10)

if users have installed an older version before, `%APPDATA%\pooi.moe\QuickLook\` may contain older versions of plugins or no longer need anymore.

the directory currently contains the plugins and setting value(LastUpdateTicks). users can easily feel plugins are installed or not. is this really necessary?

HUMORCE:
migration for configfiles only, avoid installing old or unneeded plugins for users.

`%APPDATA%\pooi.moe\QuickLook` will be deleted after migration.

issaclin32:
> migration for configfiles only, avoid installing old or unneeded plugins for users.
> 
> `%APPDATA%\pooi.moe\QuickLook` will be deleted after migration.

I think we should avoid deleting folders, because these kind of operations is likely to cause some un-recoverable problems.

How about check persist folder before migration?
`"if (!(Test-Path \"$persist_dir\\UserData\") -and Test-Path \"$env:APPDATA\\pooi.moe\\QuickLook\") {"`

HUMORCE:
actually, i dont think package manager should manage data migration.

fixed in this PR: https://github.com/ScoopInstaller/Extras/pull/6105

PR #6883, opened 2021-09-22
termius: Update to version 7.21.2

igitur:


bucket/termius.json
```diff
@@ -21,8 +21,8 @@
         ]
     ],
     "checkver": {
-        "url": "https://autoupdate.termius.com/win/latest.yml",
-        "regex": "version:\\s+([\\d.]+)"
+        "url": "https://docs.termius.com/changelog/desktop",
+        "regex": "<span data-offset-key=\"[^\"]+\">\\s*(\\d+\\.\\d+\\.\\d+) -"
```
Ash258, line 25: Test before applying

```suggestion
        "regex": "\">\\s*([\\d.]+)\\s+-\\s+\\d+"
```

Ash258 (changes requested):

bucket/termius.json
igitur: that regex is wrong

Ash258: > Regex is wrong

Lie.

image

bucket/termius.json
```diff
@@ -21,11 +21,11 @@
         ]
     ],
     "checkver": {
-        "url": "https://autoupdate.termius.com/win/latest.yml",
-        "regex": "version:\\s+([\\d.]+)"
+        "url": "https://autoupdate.termius.com/windows/latest.yml",
```
igitur, line 24: How did you find this URL?

PR #6894, opened 2021-09-24
powertoys: Update to version 0.45.0

beerpsi:
The manifest now includes checkver and autoupdate. A pre-install script is also included to check if .NET Core is installed.


jedieaston:
BTW, it doesn't seem to work on a machine without .NET Core installed (Windows 11 Insider Preview, maybe that has something to do with it).

PR #6916, opened 2021-09-28
youtubedownloader: Add version 1.8.3

KaranKad:
Added new application

### DIFF
--- a/bucket/youtubedownloader.json
+++ b/bucket/youtubedownloader.json
@@ -1,0 +1,33 @@
+{
+    "version": "1.8.3",
+    "description": "YoutubeDownloader is an application that lets you download videos from YouTube.",
+    "homepage": "https://github.com/Tyrrrz/YoutubeDownloader",
+    "license": "GPL-3.0-or-later",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Tyrrrz/YoutubeDownloader/releases/download/1.8.3/YoutubeDownloader.zip",
+            "hash": "9F40318E4CEE331924F4FD593428BC0BCDD05C696F475E48EE3B1542AF15CC1C"
+        }
+    },
+    "pre_install": [
+        "if(!(Test-Path \"$persist_dir\\Settings.dat\")) { New-Item \"$dir\\Settings.dat\" | Out-Null }",
+        "if (!(Test-Path \"$persist_dir\\Settings.dat\")) {",
+        "    Set-Content \"$dir\\Settings.dat\" -Value '{ \"IsAutoUpdateEnabled\": false }' -Encoding Ascii",
+        "}"
+    ],
+    "persist": "Settings.dat",
+    "shortcuts": [
+        [
+            "YoutubeDownloader.exe",
+            "YoutubeDownloader"
+        ]
+    ],
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/Tyrrrz/YoutubeDownloader/releases/download/$version/YoutubeDownloader.zip"
+            }
+        }
+    }
+}
--- a/bucket/youtubedownloader.json
+++ b/bucket/youtubedownloader.json
@@ -10,7 +10,6 @@
         }
     },
     "pre_install": [
-        "if(!(Test-Path \"$persist_dir\\Settings.dat\")) { New-Item \"$dir\\Settings.dat\" | Out-Null }",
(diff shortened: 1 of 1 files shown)


bucket/youtubedownloader.json
```diff
@@ -0,0 +1,33 @@
+{
+    "version": "1.8.3",
```
Ash258, line 2: Why would you add basically dead project, when there are lots of maintained alternatives?

image

bucket/youtubedownloader.json
```diff
@@ -0,0 +1,33 @@
...
+    "license": "GPL-3.0-or-later",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Tyrrrz/YoutubeDownloader/releases/download/1.8.3/YoutubeDownloader.zip",
+            "hash": "9F40318E4CEE331924F4FD593428BC0BCDD05C696F475E48EE3B1542AF15CC1C"
+        }
+    },
+    "pre_install": [
+        "if(!(Test-Path \"$persist_dir\\Settings.dat\")) { New-Item \"$dir\\Settings.dat\" | Out-Null }",
+        "if (!(Test-Path \"$persist_dir\\Settings.dat\")) {",
```
Ash258, line 14: What??

Why two checks for the same thing?

Ash258 (changes requested):

bucket/youtubedownloader.json
KaranKad: Even though this project is in maintenance mode it's not dead yet and there are no alternatives to it right now because it's core is not as same as youtube-dl or any other forks of it.

Ash258: > there are no alternatives to it

Name one feature, which is not covered by other youtube downlaoders already in extras

KaranKad: Unlike youtube-dl and its forks YouTube downloads don't get throttled, that's what I've noticed.

bucket/youtubedownloader.json
```diff
@@ -0,0 +1,32 @@
+{
+    "version": "1.8.3",
+    "description": "YoutubeDownloader is an application that lets you download videos from YouTube.",
```
issaclin32, line 3:
```suggestion
    "description": "Download videos from YouTube.",
```
We tend to keep the description shorter, so that we can fit more information in a short paragraph.

issaclin32:
I personally think that we should include as many applications as we can. Some applications may not receive further updates, but people are still using them.

Take [Pietty](https://github.com/ScoopInstaller/Extras/blob/59b802f3e6b9b9b52c3c632d449afb48f54f023c/bucket/pietty.json) and [LF2](https://github.com/Calinou/scoop-games/blob/master/bucket/lf2.json) for example. Both of them have not receive updates for more than 5 years, but they are both very well-known softwares in the Chinese-speaking region, and still work properly on *Windows 11*.

I think we can accept old softwares, unless there are some reasons to reject, for example:

- There are no manuals written in English, which makes it un-maintainable, such as [UniKey](https://github.com/ScoopInstaller/Main/pull/2572).
- The developer does not want users to download without registering an account, providing e-mail, etc.
- Frequent updates without traceable update logs, which makes it hard to maintain.

rashil2000:
If a package is popular enough (many users, reviews in recent times), then there's no problem in adding it.

stevehipwell:
I'd also add that if there are multiple maintainers who understand the docs language(s) then having docs in English might not be required?

Valinor:
>> I personally think that we should include as many applications as we can. Some applications may not receive further updates, but people are still using them.

I am agree with that. (even is you not ask my opinion, I support you ;) )

>> I think we can accept new packages, unless there are some reasons to reject, for example:

>>  * There are no manuals written in English, which makes it un-maintainable, such as [UniKey](https://github.com/ScoopInstaller/Main/pull/2572).
> I'd also add that if there are multiple maintainers who understand the docs language(s) then having docs in English might not be required?
Hum, maybe in this case, it would be more relevant to add a new bucket (main-de / main-ch / main-it / main-es / main-fr or main-notEN/...)

> * The developer does not want users to download without registering an account, providing e-mail, etc.
> * Frequent updates without traceable update logs, which makes it hard to maintain.

For youtubedownloader, it's in maintenance mode, it's ok to add it. But a project with no maintenance can potentially be dangerous (no security fix), or I am a little to extremist?

Again, I specific bucket can be more relevant for "dead application" (youtubedownloader is not a dead application)

issaclin32 (approved):

PR #6948, opened 2021-10-01
Update openocd to 0.11.0

EzekielDaun:
As in #6631, this PR is now from the [official mirror](https://github.com/openocd-org/openocd).


pratikpc:
openocd removed from extras

Was present in main as well.

PR #7074, opened 2021-10-25
sharik: Add version 3.1

sitiom:


bucket/sharik.json
```diff
@@ -0,0 +1,27 @@
+{
+    "version": "3.1",
+    "description": "An open-source, cross-platform solution for sharing files via Wi-Fi or Mobile Hotspot",
```
issaclin32, line 3:
```suggestion
    "description": "Sharing files via Wi-Fi or Mobile Hotspot.",
```
*open-source* or *cross-platform* does not matter too much for end-users.
We tend to keep the description shorter, so that we can fit more information in a short paragraph.

niheaven (approved):

issaclin32 (approved):

PR #7085, opened 2021-10-27
hwmonitor: shortcut capitalization

HUMORCE:


Valinor:
Ok, I understand, I am not completly agree (I still prefer lowercase even in shortcut), but this is legit. Thanks for the clarification @rashil2000 and @HUMORCE .

Issue #5845, opened 2021-03-11
qBittorrent doesn't install as portable by default


escape0707:
As https://github.com/lukesampson/scoop-extras/pull/2014#issuecomment-483700572 said, keeping the non-portable version is only in case the portable implementation cause some problems. But after these years the portable implementation seems pretty solid to me.

Shall we remove the `qbittorrent-portable` and change `qbittorrent` to be installed as portable by default just like other apps (vlc, mpv, etc)?

PR #7173, opened 2021-10-29
Merge qbittorrent-portable into qbittorrent

escape0707:
Fix https://github.com/ScoopInstaller/Extras/issues/5845

### DIFF
--- a/bucket/qbittorrent.json
+++ b/bucket/qbittorrent.json
@@ -6,11 +6,6 @@
         "identifier": "GPL-2.0-only",
         "url": "https://github.com/qbittorrent/qBittorrent/blob/master/COPYING"
     },
-    "notes": [
-        "qBittorrent now supports Portable Mode! Please move the following directories:",
-        "From \"$env:APPDATA\\qBittorrent\" to \"$env:USERPROFILE\\scoop\\persist\\qbittorrent\\profile\\config\"",
-        "From \"$env:LOCALAPPDATA\\qBittorrent\" to \"$env:USERPROFILE\\scoop\\persist\\qbittorrent\\profile\\data\""
-    ],
     "architecture": {
         "64bit": {
             "url": "https://downloads.sourceforge.net/project/qbittorrent/qbittorrent-win32/qbittorrent-4.3.9/qbittorrent_4.3.9_x64_setup.exe#/dl.7z",
@@ -22,6 +17,15 @@
         }
     },
     "pre_install": "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst.exe\" -Force -Recurse",
+    "post_install": [
+        "if ( (!(Test-Path \"$persist_dir\\profile\")) -and (Test-Path \"$env:APPDATA\\qBittorrent\") ) {",
+        "    Write-Host \"Scoop is migrating qbittorrent to use portable mode by default.\" -ForeGroundColor Yellow",
+        "    Write-Host \"For details, see: https://github.com/ScoopInstaller/Extras/pull/7173\" -ForeGroundColor Yellow",
+        "    Write-Host \"Copying files from APPDATA to Scoop directory...\" -ForeGroundColor Yellow",
+        "    Copy-Item \"$env:APPDATA\\qBittorrent\\*\" \"$persist_dir\\profile\\config\" -Force | Out-Null",
+        "    Copy-Item \"$env:LOCALAPPDATA\\qBittorrent\\*\" \"$persist_dir\\profile\\data\" -Force | Out-Null",
+        "}"
+    ],
     "bin": "qbittorrent.exe",
     "shortcuts": [
         [
--- a/bucket/qbittorrent.json
+++ b/bucket/qbittorrent.json
@@ -6,6 +6,11 @@
         "identifier": "GPL-2.0-only",
         "url": "https://github.com/qbittorrent/qBittorrent/blob/master/COPYING"
     },
+    "notes": [
+        "qBittorrent now supports Portable Mode! Please move the following directories:",
+        "From \"$env:APPDATA\\qBittorrent\" to \"$env:USERPROFILE\\scoop\\persist\\qbittorrent\\profile\\config\"",
+        "From \"$env:LOCALAPPDATA\\qBittorrent\" to \"$env:USERPROFILE\\scoop\\persist\\qbittorrent\\profile\\data\""
+    ],
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


issaclin32:
> I think leave too many appX and appX-portable is ugly and inconsistent for scoop's package management, too.

> I always want to install apps as portable as possible when using scoop, and for these apps I always have to do a specific search to verify I'm actually installing in a portable way.

I totally agree with that. We should apply the same criteria to all packages as we can. This makes end-users easier to understand what's going on, and reduces chores/hassles for maintainers.

On the other hand, @pratikpc 's concern is legit. I think migrating to *portable* version is a minor issue, as long as we keep the files/config intact. However, it would be pretty unpleasant to run `scoop update *` and **see your unfinished torrents / config got reset** (and it is very possible that people will send issues to complain about that)

I think the better way is to add (permanent)**migration script** in the manifest, such as:
```
(in post_install)
# This migrates config file from $Env:Appdata to $dir. For details, see: https://github.com/ScoopInstaller/Extras/pull/7173
if ( (!(Test-Path "$persist_dir\profile")) -and (Test-Path "$env:APPDATA\qBittorrent") ) {
    Write-Host "Copying files from APPDATA to Scoop directory" -ForeGroundColor Yellow
    Copy-Item "$Env:APPDATA\qBittorrent" "$persist_dir\profile\config" -Force | Out-Null
    Copy-Item "$Env:LOCALAPPDATA\qBittorrent" "$persist_dir\profile\data" -Force | Out-Null
}
```
*** This requires further testing
Also see: https://github.com/ScoopInstaller/Scoop/wiki/Pre--and-Post-install-scripts

issaclin32:
Asking for more opinions:
@pratikpc @stevehipwell @rashil2000 @Valinor

escape0707:
As of migration script, I've used things like this: https://github.com/chawyehsu/dorado/blob/1ea4d5f555e95f154afe957787b2a516dcd48f2b/bucket/clash-for-windows.json#L25-L43

And it works well. Maybe we can learn the way to do it from there.

But I'm not using Windows and Scoop right now, so it might take sometime for me to test the result.

escape0707:
Could someone review / test this? Or I can test it tomorrow when it is easier for me to reach a Windows machine.

rashil2000:
The migration script by @issaclin32 seems like a pretty good way of letting users know of about the migration. I'm with it.

bucket/qbittorrent.json
```diff
@@ -16,14 +16,24 @@
             "hash": "01487a0e2594a5065e4d780eb012dcd0dafadc218d1b6aba69528bd6ede6afb5"
         }
     },
-    "pre_install": "Remove-Item \"$dir\\`$PLUGINSDIR\" -Force -Recurse",
+    "pre_install": "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst.exe\" -Force -Recurse",
```
issaclin32, line 19:
```suggestion
    "pre_install": [
        "Remove-Item \"$dir\\`$PLUGINSDIR\" -Force -Recurse",
        "if ( (!(Test-Path \"$persist_dir\\profile\")) -and (Test-Path \"$env:APPDATA\\qBittorrent\") ) {",
        "    Write-Host \"Scoop is migrating qbittorrent to use portable mode by default.\" -ForeGroundColor Yellow",
        "    Write-Host \"For details, see: https://github.com/ScoopInstaller/Extras/pull/7173\" -ForeGroundColor Yellow",
        "    Write-Host \"Copying files from APPDATA to Scoop directory...\" -ForeGroundColor Yellow",
        "    New-Item \"$persist_dir\\profile\\config\", \"$persist_dir\\profile\\data\" -ItemType Directory | Out-Null",
        "    Copy-Item \"$env:APPDATA\\qBittorrent\\*\" \"$persist_dir\\profile\\config\" -Force | Out-Null",
        "    Copy-Item \"$env:LOCALAPPDATA\\qBittorrent\\*\" \"$persist_dir\\profile\\data\" -Force | Out-Null",
        "}"
    ],
```
`uninst.exe` does not exist in *portable* version.

The migration script should be in *pre_install*. Otherwise `$persist_dir\profile` will be created before running the script. (sorry I was wrong)

escape0707: Sorry for the late response, you are totally right sir.

issaclin32 (approved):

escape0707:
Oh my god, I just found out that the proposed change accepted doesn't include removal or the won't be reached post_install script...

Also, the deletion of the `qbittorrent-portable.json` seems to be missed during my rebase and edit attempt.

Sorry for being sloppy at checking everything up. I'll follow up with a new PR to complement that.

kidyudiqy:
The migration script doesn't migrate the data properly, at least on my install -- my config got "reset" and I "lost" my torrent list.
At first I thought qBittorrent isn't playing nice with the persisted folder, since it's a junction, but after playing around with it, I noticed that the actual path for the config and data are `profile/qBittorrent/config` and `profile/qBittorrent/data` instead of just `profile/config` and `profile/data`. So I simply moved the folders copied by the migration script to the correct location, and now my config and torrent list is back.
Since the persisted folder is `profile` I think only the migration script needs to be changed(?)

pratikpc:
@escape0707 could you please fix this?

escape0707:
> The migration script doesn't migrate the data properly, at least on my install -- my config got "reset" and I "lost" my torrent list.
> 
> At first I thought qBittorrent isn't playing nice with the persisted folder, since it's a junction, but after playing around with it, I noticed that the actual path for the config and data are `profile/qBittorrent/config` and `profile/qBittorrent/data` instead of just `profile/config` and `profile/data`. So I simply moved the folders copied by the migration script to the correct location, and now my config and torrent list is back.
> 
> Since the persisted folder is `profile` I think only the migration script needs to be changed(?)

This should have been noticed and resolved by https://github.com/ScoopInstaller/Extras/pull/7259 .

kidyudiqy:
@escape0707 Tested the fix on several machines, it works as intended now.

issaclin32:
At first glance I thought `qbitrorrent-portable` should be a minor issue (the users will figure out), but I changed my mind after seeing [the comments](https://github.com/ScoopInstaller/Extras/pull/7256#issuecomment-962478238) from @pratikpc.

I think we can have some sort of "standard procedure" to deal with this kind of issues, such as:
```
(1) Check if the merge is viable: Is there? any difference in function between portable and non-portable (such as the case of `xampp`). Also, if the non-portable version is preferred by most users, then we should make the package an exception (exception to *Scoop*'s policy of `as portable as we can`)
(2) Write migration script for the (old)non-portble package and the portable one named `xxx-portable`.
(3) Check and make sure that every persist file is **intact** before merging any PRs.
(4) Create a non-portable package in the *nonportabe* bucket later.
```

Issue #7009, opened 2021-10-12
Bizhawk@2.6.3 two persist folders are missing


satorope:
Two Bizhawk persist folders are missing.

NDS
melon (to save NDS firmware)

HUMORCE:
Confirmed

PR #7184, opened 2021-10-29
bizhawk: Persist 'NDS' and 'Melon' folders

HUMORCE:
close https://github.com/ScoopInstaller/Extras/issues/7009


pratikpc:
@HUMORCE  could you share some documentation confirming they are supposed to be persisted and compatible between versions?

HUMORCE:
only work for latest version is well

pratikpc:
@HUMORCE no I mean like for example with WinSCP, translations are separate based on versions and 3.5.1 translation is not assured to be compatible with 3.6

So wanted to verify if we face the same issue here? Like no docs say this issue exists right?

If no docs say this, we can merge it now

HUMORCE:
in the example you mentioned, winscp translations is not managed via scoop.
users need to manage these manually.

issaclin32 (approved):

PR #7195, opened 2021-10-30
Update cpu-z.json

RavenMacDaddy:
- Updates json to v1.98

### DIFF
--- a/bucket/cpu-z.json
+++ b/bucket/cpu-z.json
@@ -3,7 +3,7 @@
     "description": "System information software",
     "homepage": "https://www.cpuid.com/softwares/cpu-z.html",
     "license": "Freeware",
-    "url": "http://download.cpuid.com/cpu-z/cpu-z_1.98-en.zip",
+    "url": "https://download.cpuid.com/cpu-z/cpu-z_1.98-en.zip",
     "hash": "faea7cc753bb655f64121f32932f622f9064f8ce080de1f25d5e69d1e1a69474",
     "architecture": {
         "64bit": {
@@ -36,7 +36,8 @@
         }
     },
     "persist": "cpuz.ini",
+    "checkver": "Version ([\\d.]+) for windows",
     "autoupdate": {
-        "url": "http://download.cpuid.com/cpu-z/cpu-z_$version-en.zip"
+        "url": "https://download.cpuid.com/cpu-z/cpu-z_$version-en.zip"
     }
 }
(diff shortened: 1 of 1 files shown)


rashil2000:
`checkver` is missing

RavenMacDaddy:
Sadly it's always been missing.

Personally I don't understand how regex works.

Hopefully someone has expertise on that and can improve the manifest.

bucket/cpu-z.json
```diff
@@ -1,10 +1,10 @@
 {
-    "version": "1.96.1",
+    "version": "1.98",
     "description": "System information software",
     "homepage": "https://www.cpuid.com/softwares/cpu-z.html",
     "license": "Freeware",
-    "url": "http://download.cpuid.com/cpu-z/cpu-z_1.96-en.zip",
-    "hash": "c8445a136b2a63a356e558dfc6e500d625859f05afe36e75391e94f407382f61",
+    "url": "http://download.cpuid.com/cpu-z/cpu-z_1.98-en.zip",
```
pratikpc, line 6: Can we switch the URLs to https as well?

Both here and in autoupdate?

rashil2000:
You can specify `"checkver": "Version ([\\d.]+) for windows"`

pratikpc:
@rashil2000 I will create a PR for the same

Just checkver present.

Based on offline tests, it works

bucket/cpu-z.json
```diff
@@ -36,7 +36,8 @@
         }
     },
     "persist": "cpuz.ini",
+    "checkver": "Version ([\\d.]+) for windows",
     "autoupdate": {
-        "url": "http://download.cpuid.com/cpu-z/cpu-z_$version-en.zip"
+        "url": "https://download.cpuid.com/cpu-z/cpu-z_$version-en.zip"
     }
 }
```
pratikpc, line 43: I believe we need a new line at the end of file

RavenMacDaddy: There aren't any new brackets or errors on preview, shouldn't be needed normally in that case - or do you just mean an empty line?

pratikpc: Empty line

checkver.ps1 adds one to the end of file.

But I noticed a lot of files don't have them.

Let me check with /verify once

rashil2000 (approved):

bucket/cpu-z.json
pratikpc: Needs new line @RavenMacDaddy by the looks of it

RavenMacDaddy: Looking at both commits in edit mode there seems to be an empty line at the end, does there need to be two?

PR #7199, opened 2021-10-31
Fix bug of copying xnview.ini to wrong location if the persist directory was not created beforehand

CrendKing:
My [previous attempt](https://github.com/ScoopInstaller/Extras/pull/7181) to persist XnView MP missed a case, that when the app's persist directory was not existed beforehand, the .ini just get copied to the root directory.

This attempt uses the "persist" section to achieve two points in one stone: 1) guarantee the persist directory exist. 2) copy xnview.ini to the app dir. Thanks to this, we can simplify the "pre_install" section to just create empty new file.

Also added the trailing "\" in the uninstaller script to make sure this problem no longer happens.

### DIFF
--- a/bucket/xnviewmp.json
+++ b/bucket/xnviewmp.json
@@ -14,15 +14,9 @@
         }
     },
     "extract_dir": "XnViewMP",
-    "pre_install": [
-        "if (Test-Path \"$persist_dir\\xnview.ini\") {",
-        "    Copy-Item \"$persist_dir\\xnview.ini\" \"$dir\"",
-        "} else {",
-        "    New-Item \"$dir\\xnview.ini\" | Out-Null",
-        "}"
-    ],
+    "pre_install": "New-Item \"$persist_dir\\xnview.ini\" -ErrorAction SilentlyContinue",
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):

issaclin32 (approved):

mossglen90:
This still causes a problem when the persist directory does not exist. When persist directory does not exist, the pre_install silently fails since it cannot create the xnview.ini file (as $persist_dir itself does not exist). The following persist creates a directory junction named xnview.ini which prevents the program from creating the xnview.ini as a file. The pre_install command needs to be changed to:
`New-Item -Type Directory -Force -Path \"$persist_dir\" -ErrorAction SilentlyContinue; New-Item \"$persist_dir\\xnview.ini\" -ErrorAction SilentlyContinue`

CrendKing:
I did this before https://github.com/ScoopInstaller/Main/pull/2936, so some improvement there is not here. I don't think we can `-Force` unconditionally because an update would reset the content.

Update PR: https://github.com/ScoopInstaller/Extras/pull/7488

PR #7200, opened 2021-10-31
diagram-designer: Add version 1.29.5

issaclin32:
Supercedes #6975

[Diagram Designer](https://logicnet.dk/DiagramDesigner/) is a **vector graphics editor** for creating flowcharts, UML class diagrams, illustrations and slide shows.

### DIFF
--- a/bucket/diagram-designer.json
+++ b/bucket/diagram-designer.json
@@ -3,6 +3,7 @@
     "description": "Vector graphics editor for creating flowcharts, UML class diagrams, illustrations and slide shows",
     "homepage": "https://logicnet.dk/DiagramDesigner/",
     "license": "Freeware",
+    "notes": "Template palettes can be downloaded from 'https://github.com/meesoft/DiagramDesigner/tree/master/TemplatePalettes'",
     "url": "https://www.fosshub.com/Diagram-Designer.html/DiagramDesignerSetup.1.29.5.msi",
     "hash": "381c696c81ab57faaf40ce690c760f549056fb1af299ebf550c08dae5f379bf7",
     "shortcuts": [
--- a/bucket/diagram-designer.json
+++ b/bucket/diagram-designer.json
@@ -1,0 +1,18 @@
+{
+    "version": "1.29.5",
+    "description": "Vector graphics editor for creating flowcharts, UML class diagrams, illustrations and slide shows",
+    "homepage": "https://logicnet.dk/DiagramDesigner/",
+    "license": "Freeware",
+    "url": "https://www.fosshub.com/Diagram-Designer.html/DiagramDesignerSetup.1.29.5.msi",
+    "hash": "381c696c81ab57faaf40ce690c760f549056fb1af299ebf550c08dae5f379bf7",
+    "shortcuts": [
+        [
+            "DiagramDesigner.exe",
+            "Diagram Designer"
+        ]
+    ],
+    "checkver": "Diagram Designer ([\\d.]+)",
+    "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/diagram-designer.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "1.29.5",
+    "description": "Vector graphics editor for creating flowcharts, UML class diagrams, illustrations and slide shows",
+    "homepage": "https://logicnet.dk/DiagramDesigner/",
+    "license": "Freeware",
+    "url": "https://www.fosshub.com/Diagram-Designer.html/DiagramDesignerSetup.1.29.5.msi",
```
pratikpc, line 6: Is it 32 bit?

issaclin32: It is a 32-bit application. However, if I put the url under *architecture -> 32bit*, it will cause an error when installing under 64bit systems. (`ERROR 'xxx' doesn't support 64bit architecture!`)

I personally think that we can discard 32-bit support, as there are few PCs running under 32-bit, but that requires further discussion with the maintainers.

```
(In VS2019 developer command prompt)
C:\> dumpbin /headers D:\scoop\apps\diagram-designer\current\DiagramDesigner.exe
Microsoft (R) COFF/PE Dumper Version 14.29.30133.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file D:\scoop\apps\diagram-designer\current\DiagramDesigner.exe

PE signature found

File Type: EXECUTABLE IMAGE

FILE HEADER VALUES
             14C machine (x86)
               3 number of sections
        2A425E19 time date stamp Sat Jun 20 06:22:17 1992
               0 file pointer to symbol table
               0 number of symbols
              E0 size of optional header
            818F characteristics
                   Relocations stripped
                   Executable
                   Line numbers stripped
                   Symbols stripped
                   Bytes reversed
                   32 bit word machine
```

pratikpc: No no

If its 32 bit, this scope is correct

bucket/diagram-designer.json
```diff
@@ -0,0 +1,19 @@
...
+    "shortcuts": [
+        [
+            "DiagramDesigner.exe",
+            "Diagram Designer"
+        ]
+    ],
+    "checkver": "Diagram Designer ([\\d.]+)",
+    "autoupdate": {
+        "url": "https://www.fosshub.com/Diagram-Designer.html/DiagramDesignerSetup.$version.msi"
+    }
```
pratikpc, line 18: On https://www.fosshub.com/Diagram-Designer.html/, when you press the Signature button you can get access to the SHA256 hash of the file.

Would it be possible for us to extract that hash using existing tooling in Scoop?

That is if Diagram Designer is expected to get updates. By the looks of their GitHub repo it doesn't seem like that is going to ever be a case

issaclin32: Currently it is implemented in *Scoop*'s main program (*FossHub mode*). Therefore we don't need to download the file while running *autoupdate* for files hosted on FossHub.
https://github.com/ScoopInstaller/Scoop/blob/092005046454d94d141d5c68fdbdb4c4a1229ae9/lib/autoupdate.ps1#L222.

I would say it is better to extend the feature and enable users to make use of it **in the manifest**, but this requires further update on *Scoop*'s main program.
I also wish that child paths (e.g. `some_app\x86\bin`) can be applied on *extract_dir* through all means of extraction including *innounp*, *lessmsi*, *uniextract2*, etc. , but that is another topic.

pratikpc: I had no idea!!!

pratikpc (approved):

PR #7201, opened 2021-10-31
grepWin: Add checkver for automated updates

JamesCullum:
Noticed that grepWin didn't get updated for quite a while, probably due to missing checkver.
Wanted to automate this using the Github repository with a small change - this should pick up the correct hashes at next update cycle.

### DIFF
--- a/bucket/grepwin.json
+++ b/bucket/grepwin.json
@@ -52,7 +52,7 @@
         [
             "grepWin.exe",
             "grepWin",
-			"/portable"
+            "/portable"
         ]
     ],
     "checkver": {
--- a/bucket/grepwin.json
+++ b/bucket/grepwin.json
@@ -52,6 +52,9 @@
             "grepWin"
         ]
     ],
+    "checkver": {
+        "github": "https://github.com/stefankueng/grepWin"
+    },
     "autoupdate": {
         "architecture": {
             "64bit": {
--- a/bucket/grepwin.json
+++ b/bucket/grepwin.json
@@ -58,10 +58,10 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/stefankueng/grepWin/releases/download/$version/grepWin-x64-$version_portable.exe#/grepWin.exe"
+                "url": "https://github.com/stefankueng/grepWin/releases/download/$version/grepWin-$version-x64.msi"
             },
             "32bit": {
-                "url": "https://github.com/stefankueng/grepWin/releases/download/$version/grepWin-$version_portable.exe#/grepWin.exe"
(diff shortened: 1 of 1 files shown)


pratikpc (approved):

pratikpc (changes requested):
Hey we need to update the autoupdate URLs as well

https://github.com/stefankueng/grepWin/releases/download/2.0.8/grepWin-2.0.8_portable.exe#/grepWin.exe no longer exists

JamesCullum:
Thanks for the good comment!

The newest release has this URL: https://github.com/stefankueng/grepWin/releases/download/2.0.8/grepWin-x64-2.0.8_portable.zip
The file inside is called grepWin-x64-2.0.8_portable.exe

So I assume that the correct autoupdate URL would need to be https://github.com/stefankueng/grepWin/releases/download/$version/grepWin-x64-$version_portable.zip#/grepWin-x64-$version_portable.exe, right?

pratikpc:
No no

How can a ZIP file become an exe?

JamesCullum:
I see - the `#` then stands for renaming, I assumed it was for extraction.

I now simply used the MSI - let me know if this works for you 👍

pratikpc:
@JamesCullum please update the versions offline and check once if installation takes place properly.

Check if all persist based configurations work as well and share screenshots for all steps

Scoop usually prefers portable files so we are more likely to update to portable ZIP files.

JamesCullum:
Tested and updated - grepWin persist nowadays via registry.
The MSI is just a ZIP with the portable files, but packed with more translations.

Valinor:
> Tested and updated - grepWin persist nowadays via registry. The MSI is just a ZIP with the portable files, but packed with more translations.

No, I currently test it, and msi record grepwin.ini in AppData/Roaming/...
The portable version record the grepwin.ini in the current directory.

JamesCullum:
Good hint, didn't know that! Will take a look at the portable version then again!

JamesCullum:
Okay, I'm just renaming it via script now.

@Valinor I noticed that even the portable version in 2.0.8 doesn't automatically store the ini in the same folder. Its actually now only decided by the flag `/portable` flag, which I added to the shim. Maybe give it a try and let me know if it works 👍

Valinor:
@pratikpc Can you validate this push (approuve your "requested changes")? Seem that the next commit correct your point

pratikpc:
@Valinor @JamesCullum could some screenshots be shared on this working with both 32 and 64 bit?

The manifest currently looks fine to me and passes verify too

Valinor:
image

image

32bits and 64bits works for me.

PR #7205, opened 2021-11-01
add context menu handlers for lockhunter

phanirithvij2000:
Added context menu handlers (shellex) for Lockhunter.

There is a [portable installer](https://lockhunter.com/download.htm) on the website but that doesn't provide the required shellex dll files `LHShellExt32.dll, LHShellExt64.dll` to add to the context menu, so we should keep in mind that in the future, this script shouldn't download the portable version but the full installer version that is currently being downloaded.

Tested locally on my Windows 10 64bit system

```sh
scoop install lockhunter
reg import "%USERPROFILE%\scoop\apps\lockhunter\current\install-context.reg"
# Works now from context menu
scoop uninstall lockhunter # will automatically unregister the registry from uninstall-context.reg
```

Works with `sudo install lockhunter -g` globally as well.


phanirithvij2000:
Because this is an innosetup scoop is trying to extract .reg files too.
How do I prevent that?
```
Extracting install-context.reg ... ERROR Exit code was 1!
Failed to extract files from C:\Users\Rithvij\scoop\apps\lockhunter\3.4.3\install-context.reg.
```

bucket/lockhunter.json
```diff
@@ -3,8 +3,17 @@
     "description": "Delete files blocked by something you do not know.",
     "homepage": "https://lockhunter.com/index.htm",
     "license": "Unknown",
-    "url": "https://lockhunter.com/assets/exe/lockhuntersetup_3-4-3.exe",
-    "hash": "2c0d52dca3e5ce9cfc8062cb72c7235e3c6ff650242fcb7d46a892d602dd3dd1",
+    "depends": "sudo",
+    "url": [
+        "https://lockhunter.com/assets/exe/lockhuntersetup_3-4-3.exe",
+        "https://raw.githubusercontent.com/phanirithvij2000/Extras/lockhunter-shellex/scripts/lockhunter/install-context.reg",
```
phanirithvij2000, line 9: Will change this to `ScoopInstaller/Extras/master` once I get the script working locally

rashil2000:
> Also, I'm tinkering with scoop for the first time, so I have a question does $dir mean %Userprofile%%\scoop\apps\lockhunter\current in this case?

Yes

bucket/lockhunter.json
rasa: Why not add them as https://github.com/ScoopInstaller/Extras/tree/master/scripts/lockhunter/install-context.reg and
https://github.com/ScoopInstaller/Extras/tree/master/scripts/lockhunter/uninstall-context.reg

and them change
`$dir\\install-context.reg`
to
`$scoopdir/apps/scoop/current/scripts/lockhunter/install-context.reg`
per https://github.com/ScoopInstaller/Scoop/wiki/Pre--and-Post-install-scripts#variables

That makes more sense them unnecessarily downloading files that are already in the repo.

bucket/lockhunter.json
```diff
@@ -2,19 +2,83 @@
...
-            "pre_install": "Get-ChildItem \"$dir\" '*,2.*' | Rename-Item -Path { $_.Fullname } -NewName { $_.Basename.TrimEnd(',2') + $_.Extension }"
-        }
+    "notes": "Add Lockhunter as context menu option by running: reg import \"$dir\\install-context.reg\"",
+    "pre_install": [
+        "If ($architecture -eq '64bit') {",
+        "   Get-ChildItem \"$dir\" '*,1.*' | Rename-Item -Path { $_.Fullname } -NewName { $_.Basename.TrimEnd(',1') + $_.Extension }",
+        "} Else {",
+        "   Get-ChildItem \"$dir\" '*,2.*' | Rename-Item -Path { $_.Fullname } -NewName { $_.Basename.TrimEnd(',2') + $_.Extension }",
+        "}",
+        "$url = \"https://raw.githubusercontent.com/ScoopInstaller/Extras/master/scripts/lockhunter/install-context.reg\"",
```
rasa, line 17: Why not add them as https://github.com/ScoopInstaller/Extras/tree/master/scripts/lockhunter/install-context.reg and
https://github.com/ScoopInstaller/Extras/tree/master/scripts/lockhunter/uninstall-context.reg

and them change
`$dir\\install-context.reg`
to
`$scoopdir/apps/scoop/current/scripts/lockhunter/install-context.reg`
per https://github.com/ScoopInstaller/Scoop/wiki/Pre--and-Post-install-scripts#variables

That makes more sense them unnecessarily downloading files that are already in the repo.

phanirithvij2000: I think you mean from `$scoopdir\scoop\buckets\extras\scripts\lockhunter`?

Yeah didn't think of that, so just copy those 2 files from the bucket scripts directory to `$dir`.

After that should I be checking the sha256 for these?

phanirithvij2000: done

bucket/lockhunter.json
```diff
@@ -2,19 +2,71 @@
     "version": "3.4.3",
     "description": "Delete files blocked by something you do not know.",
     "homepage": "https://lockhunter.com/index.htm",
-    "license": "Unknown",
+    "license": "Freeware",
```
phanirithvij2000, line 5: From the website, https://lockhunter.com/download.htm license is https://lockhunter.com/eula.rtf which claims it is a freeware

bucket/lockhunter.json
```diff
@@ -14,20 +14,8 @@
...
-        "   error $err",
-        "   abort $(new_issue_msg $app $bucket \"hash check failed\")",
-        "}",
-        "$url = \"https://raw.githubusercontent.com/ScoopInstaller/Extras/master/scripts/lockhunter/uninstall-context.reg\"",
-        "do_dl $url \"$dir\\uninstall-context.reg\"",
-        "$ok, $err = check_hash \"$dir\\uninstall-context.reg\" \"38443a27b65dda6554a8470261064bc312b202ed81c356ef363b10b089aa5790\" $(show_app $app $bucket)",
-        "if(!$ok) {",
-        "   error $err",
-        "   abort $(new_issue_msg $app $bucket \"hash check failed\")",
-        "}"
```
rasa: `$scoopdir\\scoop\\buckets` can be shortened to `$bucketsdir`

phanirithvij2000: Done, also cleaned up a few commits.

phanirithvij2000: @rasa requesting a re-review

bucket/lockhunter.json
phanirithvij2000: @pratikpc I agree with liking to the license but there are two things to consider here, the actual official license URL downloads the file which might not be very convenient for many users, the other thing is I remember reading somewhere (maybe on the main repo wiki) that only a few strings are valid for this "license" field, I might be wrong.

Also separate PR makes sense

bucket/lockhunter.json
phanirithvij2000: @issaclin32 @rasa can one of you review this, I can confirm it works locally. But I'm not sure why I don't see the github actions check on any commit for this PR anywhere.

bucket/lockhunter.json
pratikpc: @phanirithvij2000 you can have an object with identifier and url in License as well.

Personally I would support using this approach here because the user should have an ability to read the license even if it downloads a file.

phanirithvij2000: The setup does come with that eula.rtf file downloaded so maybe having `"license": "$dir\\eula.rtf"` or if that isn't valid adding a note "The license is located at $dir\\eula.rtf" should suffice I think.

rasa (approved):
Untested, but lgtm

PR #7207, opened 2021-11-01
x64dbg: Update to 2021-10-31_10-53

AntonOks:
Seems like
1. releases are updated very frequent sometimes _AND_
2. only the very last one is kept available on https://github.com/x64dbg/x64dbg/releases

which may mean, the manifest should be changed to download from https://sourceforge.net/projects/x64dbg/files/snapshots/ in the future?!?


rashil2000:
If you can make that change too, it'd be great!

rashil2000 (approved):

Issue #5960, opened 2021-04-02
[Request] Add SDL_Image


boy-in-blue:
SDL2 is already included in this bucket. So, it would be convenient to have SDL_Image to go along with it.

https://www.libsdl.org/projects/SDL_image/

boy-in-blue:
And fonts: https://www.libsdl.org/projects/SDL_ttf/

g-berthiaume:
Hi!
I want to re-energize this issues.

SDL is great for teaching game dev on windows.
With the support of this package, student could just use scoop to get started.
```
scoop install pck-config sdl2 sdl-image
```

rashil2000:
Hey @g-berthiaume!

Would you like to make a PR to add your package?

g-berthiaume:
Hi @rashil2000. I can certainly try!

rashil2000:
Yes! 
See here https://scoop-docs.now.sh/docs/concepts/App-Manifests.html#a-simple-example

The docs contain many links to examples - be sure to check them out.

g-berthiaume:
I'm working on it.
While doing so, I made an update to SDL2: https://github.com/ScoopInstaller/Extras/pull/7202

g-berthiaume:
Progress 
- [ ] Adding sdl2_image https://github.com/ScoopInstaller/Extras/pull/7209
- [ ] Adding sdl2_ttf

PR #7209, opened 2021-11-02
sdl2-image: Add version 2.0.5

g-berthiaume:
This PR add the SDL2_image library to scoop; addressing this [issue](https://github.com/ScoopInstaller/Extras/issues/5960).
SDL2_image is a complement to the already supported [SDL2](https://www.libsdl.org/) library.

My thanks for the creators of the [sdl2.json](https://github.com/ScoopInstaller/Extras/blob/master/bucket/sdl2.json) app manifest.   
This PR was inspired from their work.


rashil2000:
Autoupdate missing

Also, please add a newline at the end

bucket/sdl2_image.json
```diff
@@ -0,0 +1,123 @@
...
+            "$was_in_path, $newpath = strip_path $env:PKG_CONFIG_PATH $pc_dir",
+            "if($was_in_path) { $env:PKG_CONFIG_PATH = $newpath }",
+            "$was_in_path, $newpath = strip_path $env:CMAKE_PREFIX_PATH $cm_dir",
+            "if($was_in_path) { $env:CMAKE_PREFIX_PATH = $newpath }"
+        ]
+    },
+    "checkver": {
+        "url": "https://www.libsdl.org/projects/SDL_image/",
+        "regex": "SDL2_image-([\\d.]+)\\.zip"
+    }
```
issaclin32, line 122:
```suggestion
    },
    "autoupdate": {
        "url": [
            "https://www.libsdl.org/projects/SDL_image/release/SDL2_image-devel-$version-VC.zip",
            "https://www.libsdl.org/projects/SDL_image/release/SDL2_image-$version.zip"
        ]
    }
```
This should work after https://github.com/ScoopInstaller/Scoop/pull/3518 is merged into master branch.

issaclin32:
I think this package should be in [the main bucket](https://github.com/ScoopInstaller/Main).
Also, please rename the file to `sdl2-image.json`, following the pattern of existing packages.

bucket/sdl2_image.json
g-berthiaume: Thanks for your suggestion, I will add it.

g-berthiaume:
> I think this package should be in the main bucket.

I kinda agree. 
That said, the SDL2 package is already in the Extra bucket. 
I don't know what was the rationale for putting it there.

@r15ch13 has made the earliest commit on the `sdl2.json` file.   
Maybe he/she would know.

g-berthiaume:
Everything should be all right.
Not sure what to think about the Extra vs Main bucket.

rashil2000:
The installer scripts are quite elaborate here, as opposed to the Main bucket's criteria. I think it should stay in Extras.

rashil2000:
@niheaven wouldn't this require https://github.com/ScoopInstaller/Scoop/pull/3518?

issaclin32:
>  wouldn't this require [ScoopInstaller/Scoop#3518](https://github.com/ScoopInstaller/Scoop/pull/3518)?

Yes, you are right, but I think it is okay to use it now.

This will just cause *autoupdate* to fail (thus the package will not be updated before *Scoop#3518* is in *master*.

We can choose not to add *autoupdate* this time, but we still need to update it manually. So it does not make any difference.

rashil2000 (approved):

issaclin32:
> I kinda agree. That said, the SDL2 package is already in the Extra bucket. I don't know what was the rationale for putting it there.

This has to do with earlier criteria of the Main bucket. I think these criteria are abolished after late 2019, but some CLI packages still remain in the Extras bucket.

```
For an app to be acceptable for the main bucket, it should be:

A reasonably well-known and widely used developer tool
The latest stable version of the program
The full version i.e. not a trial version
A fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)
A non-GUI tool
```
I tend to add the package to *Extras* for now. I think it is better to move all the CLI packages into *Main*, but this requires further discussion.

issaclin32 (approved):

niheaven:
> > I kinda agree. That said, the SDL2 package is already in the Extra bucket. I don't know what was the rationale for putting it there.
> 
> This has to do with earlier criteria of the Main bucket. I think these criteria are abolished after late 2019, but some CLI packages still remain in the Extras bucket.
> 
> ```
> For an app to be acceptable for the main bucket, it should be:
> 
> A reasonably well-known and widely used developer tool
> The latest stable version of the program
> The full version i.e. not a trial version
> A fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)
> A non-GUI tool
> ```
> 
> I tend to add the package to _Extras_ for now. I think it is better to move all the CLI packages into _Main_, but this requires further discussion.

Emm, lots of job 😄  But we can move whatever we meet.

PR #7217, opened 2021-11-02
Add postinstall to remove autoupdate file

Valinor:
When an update is present and scoop is late, a popup propose to update bitwarden. But this installation will end in program files.... So not really what we want here.

The post install delete the file used to check if there is an update. So now, should use scoop to update. No more popup

### DIFF
--- a/bucket/bitwarden.json
+++ b/bucket/bitwarden.json
@@ -14,6 +14,7 @@
         }
     },
     "bin": "Bitwarden.exe",
+    "post_install": "Remove-Item \"$dir\\resources\\app-update.yml\" -Recurse",
     "shortcuts": [
         [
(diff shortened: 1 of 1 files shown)


bucket/bitwarden.json
```diff
@@ -14,6 +14,7 @@
         }
     },
     "bin": "Bitwarden.exe",
+    "post_install": "Remove-Item \"$dir\\resources\\app-update.yml\" -Recurse",
```
rashil2000, line 17: Recurse flag is not required for one file?

Valinor: Right

rashil2000 (approved):

PR #7221, opened 2021-11-02
Add putty-cac

Dreista:
Add PuTTY CAC with CAPI and PKCS support.

### DIFF
--- a/bucket/putty-cac.json
+++ b/bucket/putty-cac.json
@@ -5,11 +5,11 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


rashil2000:
Use the tag name instead of `master` in the URLs, otherwise people won't be able to download a previous version.

Dreista:
Nice catch. Fixed!

niheaven:
Please move this PR to https://github.com/ScoopInstaller/Versions

Dreista:
https://github.com/ScoopInstaller/Versions/pull/303

PR #7225, opened 2021-11-03
fix(script): use `scoop prefix` to correctly find scoop's path

niheaven:


pratikpc:
@niheaven could we do the same for all the scripts?

niheaven:
Indeed, we should. `scoop which scoop` gives me nothing here. I'll make some commit to my PR.

pratikpc (approved):

Issue #6293, opened 2021-06-02
[Request] Update ILSpy to 7.1


rubenprins:
https://github.com/lukesampson/scoop-extras/blob/70b44fa847b46cc8846e6bf54a3ab4927dbccfb7/bucket/ilspy.json#L6

Latest version is [7.1.0.6543](https://github.com/icsharpcode/ILSpy/releases/tag/v7.1).

PR #7230, opened 2021-11-03
ilspy: update to 7.1.0.6543

pratikpc:
add missing autoupdate

close https://github.com/ScoopInstaller/Extras/issues/6293

@rubenprins could you check this once?

I personally don't use ILSPY

### DIFF
--- a/bucket/ilspy.json
+++ b/bucket/ilspy.json
@@ -1,10 +1,10 @@
 {
-    "version": "6.2.1.6137",
+    "version": "7.1.0.6543",
     "description": ".NET assembly browser and decompiler",
     "homepage": "http://ilspy.net",
     "license": "MIT",
-    "url": "https://github.com/icsharpcode/ILSpy/releases/download/v6.2.1/ILSpy_binaries_6.2.1.6137.zip",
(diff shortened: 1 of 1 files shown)


rubenprins:
@pratikpc looks like the right version & download URL 👍

pratikpc:
@rubenprins could you install ILSPY from this file and check if it works?

Or if changes need to be made?

Because even if we have correct URLs, we might need to make other changes

rubenprins:
@pratikpc Installing the app definition from this PR works:
```
scoop install https://raw.githubusercontent.com/ScoopInstaller/Extras/b260437ef4730a39d3741f524d15938a36f5cd50/bucket/ilspy.json
```
So, no need for additional changes, I guess.

pratikpc:
@rubenprins after that you were able to use it without issues as well right?

Try some production workloads with it once

rubenprins:
@pratikpc Yes, the app works fine this way. In fact, I normally update the app in exactly the same way as Scoop does: extract the zip file into some user folder, and run ILSpy.exe. User settings are stored in your AppData.

issaclin32 (approved):

PR #7241, opened 2021-11-04
capframex: Add extract dir

404NetworkError:
Currently breaks with `Can't shim 'CapFrameX.exe': File doesn't exist.` because the files are now in a folder.

### DIFF
--- a/bucket/capframex.json
+++ b/bucket/capframex.json
@@ -8,6 +8,7 @@
     },
     "url": "https://github.com/CXWorld/CapFrameX/releases/download/v1.6.6/CapFrameX_v1.6.6_Portable.zip",
     "hash": "f889bd33d52c05172f79d18579cc2249b3e1c750a6ec4daef735e840f637608b",
+    "extract_dir": "CapFrameX 1.6.6 Portable",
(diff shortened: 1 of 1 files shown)


bucket/capframex.json
```diff
@@ -8,6 +8,7 @@
     },
     "url": "https://github.com/CXWorld/CapFrameX/releases/download/v1.6.6/CapFrameX_v1.6.6_Portable.zip",
     "hash": "f889bd33d52c05172f79d18579cc2249b3e1c750a6ec4daef735e840f637608b",
+    "extract_dir": "CapFrameX 1.6.6 Portable",
```
rashil2000, line 11: You need to add this to autoupdate as well

rashil2000 (changes requested):

rashil2000:
Is this a 32bit application or 64bit?

bucket/capframex.json
pratikpc: @rashil2000  personally I feel we should merge it now and fix it with a new PR because this is definitely a breaking change and we should first minimize damage for existing users before fixing autoupdate and setting it to x64 while we wait for the next release

PR #7250, opened 2021-11-05
nirlauncher: Update via nlp2json.py

rasa:


rashil2000:
The version doesn't need to be updated?

rasa:
> The version doesn't need to be updated?

Good question! The version is updated by the Excavator. This update just freshens the .exes that are created when the app is installed.

rashil2000 (approved):

PR #7253, opened 2021-11-06
Update q-dir.json - correcting version to 9.97 

RavenMacDaddy:
When running the app it's currently reporting v9.97 which also coincides with the information on the app's homepage, hence why I'm updating that part of the manifest.


rashil2000:
You will need to update the hashes too.

PR #7255, opened 2021-11-06
chromium: Update to version 95.0.4638.69-r920003-1

lunarwhite:
Update [chromium](https://github.com/macchrome/winchrome/releases/) version 85.0.4183.83 to 95.0.4638.69.


pratikpc:
@lunarwhite can you fix the checkver as well?

lunarwhite:
> @lunarwhite can you fix the checkver as well?

@pratikpc checkver fixed

pratikpc:
@lunarwhite checkver and autoupate are both broken

Please test offline first

https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#using-checkverps1-to-query-and-autoupdate

Would recommend running with a ForceUpdate parameter.

pratikpc:
@lunarwhite converted it to draft because to bring it back to working condition, it would take a lot of effort.

> https://github.com/macchrome/winchrome/releases/download/v95.0.4638.69-r920003-Win64/ungoogled-chromium-95.0.4638.69-1_Win32.7z

The URL has 3 separate parameters to extract

First:-
1. 95.0.4638.69
2. 920003
3. -1. If you look into autoupdate, it was set as -2 there which is how I knew this PR is broken.
4. To fix this, you would have to look at /releases/latest instead as URL

The regex here will be very complicated. If confused, look into https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate

We had a similar issue in Gimp once which had the -1 and -2 endings. But it was only 95.0.4638.69-1 type issue there. Yours is a bit more complicated

If you need help, do tell. Fixing Chrome and Chromium is very important for us.

lunarwhite:
OK, I will work on it. Thx for your guidance.

bucket/chromium.json
```diff
@@ -31,21 +31,22 @@
...
-        "regex": "v([\\d.]+-r\\d+)-Win(64|32)"
+        "regex": "/v(?<version>[\\d.]+)-r(?<build>[\\d]+)-Win64/ungoogled-chromium-(?<version>[\\d.]+)-1_Win(64|32)\\.7z",
+        "replace": "${version}"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/macchrome/winchrome/releases/download/v$version-r$matchBuild-Win64/ungoogled-chromium-$version-2_Win64.7z",
-                "extract_dir": "ungoogled-chromium-$version-2_Win64"
+                "url": "https://github.com/macchrome/winchrome/releases/download/v$matchVersion-r$matchBuild-Win64/ungoogled-chromium-$matchVersion-1_Win64.7z",
```
pratikpc, line 40: use $version not $matchVersion

bucket/chromium.json
```diff
@@ -31,21 +31,22 @@
     ],
     "checkver": {
         "github": "https://github.com/macchrome/winchrome",
-        "regex": "v([\\d.]+-r\\d+)-Win(64|32)"
+        "regex": "/v(?<version>[\\d.]+)-r(?<build>[\\d]+)-Win64/ungoogled-chromium-(?<version>[\\d.]+)-1_Win(64|32)\\.7z",
+        "replace": "${version}"
```
pratikpc, line 35: Let the final version string be 95.0.4638.69-r920003-1 for this one.

Safer. 

@rashil2000 @issaclin32 your views?

pratikpc (changes requested):

bucket/chromium.json
issaclin32: I agree. This will help us to keep track of updates (and hotfixes) better.

pratikpc:
@lunarwhite it works properly when you install both 32 and 64 bit versions right?

lunarwhite:
> @lunarwhite it works properly when you install both 32 and 64 bit versions right?

Yeah, I just finished testing. It worked both 32 and 64 bit.

pratikpc:
@lunarwhite LGTM but we need some more help from you

Could you do the same for https://github.com/ScoopInstaller/Extras/blob/master/bucket/ungoogled-chromium-portable.json?

They both use the same binaries

lunarwhite:
Sure, I’ll do it as soon as possible:)

rashil2000 (approved):

pratikpc (approved):

pratikpc:
Merging this

The ungoogled-chromium-portable could be a fix in a new PR.

PR #7256, opened 2021-11-06
Fix up forgotten delete oprations in #7173

escape0707:
https://github.com/ScoopInstaller/Extras/pull/7173#issuecomment-962463851

Sorry for the additional confusion. Orz. @issaclin32


issaclin32 (approved):

issaclin32:
No Worries. It was also part of my problem, I didn't check carefully enough.

pratikpc:
Hey it looks like data from qBittorrent portable is not being moved to qBittorrent.

This would break it for qBitTorrent-portable users as not only have they lost a manifest but they might have to manually copy paste their persisted data from one persist directory to another.

Could we please restore qBitTorrent-portable till the data move from qBitTorrent-portable persist to qBitTorrent's persist is completed?

escape0707:
I'll test it now.

escape0707:
@pratikpc Thanks for mentioning this overlooked point. I'll submit a fix now.

PR #7273, opened 2021-11-09
netron: Update homepage 

### LINKED ISSUE #7247: netron: update homepage to https://netron.app/
https://github.com/ScoopInstaller/Extras/blob/master/bucket/netron.json

We need to update the homepage to https://netron.app/

This is a good issue to fix for new users or existing users who wish to better their understanding of Scoop.

We would definitely help out.

infomaniac50:
Closes #7247

### DIFF
--- a/bucket/netron.json
+++ b/bucket/netron.json
@@ -1,7 +1,7 @@
 {
     "version": "5.3.4",
     "description": "Visualizer for deep learning and machine learning models",
-    "homepage": "https://github.com/lutzroeder/Netron",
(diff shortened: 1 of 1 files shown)


rashil2000:
You will need to update the checkver property too, since it no longer knows where the repo is

infomaniac50:
Autoupdate is correct. I assume the bot stops on the first error.

rashil2000:
Great, thanks for the fix!

PR #7284, opened 2021-11-10
neovim-qt: Add version 0.2.16.1

rashil2000:
Neovim GUI client - like https://github.com/ScoopInstaller/Extras/blob/master/bucket/fvim.json

### DIFF
--- a/bucket/fvim.json
+++ b/bucket/fvim.json
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/yatli/fvim",
     "license": "MIT",
     "suggest": {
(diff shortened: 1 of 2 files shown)


issaclin32 (approved):

niboan:
But why?
The same version of neovim-qt is already included in the neovim app, which is required by this app.
What does this app add, that neovim does not already provide?

From the neovim-qt Github page:
> Neovim Qt is bundled with all releases of Neovim on Windows.

rashil2000:
This is for people who might have installed neovim from other sources (or locally built it, which was my use case).

Issue #2821, opened 2019-09-08
Un-deprecate Nightlies bucket?


rasa:
In my opinion, anything that has a version of “nightly”, or that updates daily, or more often, should live in the (currently unused) [nightlies](https://github.com/ScoopInstaller/Nightlies) bucket. I would think the following would be candidates to move there:

#### Main bucket:
1. [docker-nightly](https://github.com/ScoopInstaller/Main/search?q=docker-nightly&type=Commits)
1. [glslang-nightly](https://github.com/ScoopInstaller/Main/search?q=glslang-nightly&type=Commits)
1. [jar2app](https://github.com/ScoopInstaller/Main/search?q=jar2app&type=Commits)
1. [jx](https://github.com/ScoopInstaller/Main/search?q=jx&type=Commits)
1. [ffmpeg-nightly](https://github.com/ScoopInstaller/Main/search?q=ffmpeg-nightly&type=Commits)

#### Extras bucket:
1. [cheat-engine](https://github.com/lukesampson/scoop-extras/search?q=cheat-engine&type=Commits)
1. [firefox-nightly](https://github.com/lukesampson/scoop-extras/search?q=firefox-nightly&type=Commits)
1. [jdownloader](https://github.com/lukesampson/scoop-extras/search?q=jdownloader&type=Commits)
1. [lightshot](https://github.com/lukesampson/scoop-extras/search?q=lightshot&type=Commits)
1. [oneget](https://github.com/lukesampson/scoop-extras/search?q=oneget&type=Commits)
1. [rust-msvc-nightly](https://github.com/lukesampson/scoop-extras/search?q=rust-msvc-nightly&type=Commits)
1. [rust-nightly](https://github.com/lukesampson/scoop-extras/search?q=rust-nightly&type=Commits)
1. [uplay](https://github.com/lukesampson/scoop-extras/search?q=uplay&type=Commits)
1. [vscode-insiders](https://github.com/lukesampson/scoop-extras/search?q=vscode-insiders&type=Commits)
1. [vscode-insiders-portable](https://github.com/lukesampson/scoop-extras/search?q=vscode-insiders-portable&type=Commits)
1. [watchman-nightly](https://github.com/lukesampson/scoop-extras/search?q=watchman-nightly&type=Commits)

_Originally posted by @rasa in https://github.com/lukesampson/scoop-extras/pull/2814#issuecomment-528957813_

niheaven:
[...]

PR #7285, opened 2021-11-10
wezterm-nightly: Add manifest

rashil2000:
Daily builds of Wezterm emulator.

Manifest inspired by https://github.com/ScoopInstaller/Extras/blob/master/bucket/rust-nightly.json

### DIFF
--- a/bucket/wezterm-nightly.json
+++ b/bucket/wezterm-nightly.json
@@ -1,0 +1,25 @@
+{
+    "version": "nightly",
+    "description": "GPU-accelerated terminal emulator (nightly builds)",
+    "homepage": "https://github.com/wez/wezterm",
+    "license": "Apache-2.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/wez/wezterm/releases/download/nightly/WezTerm-windows-nightly.zip"
+        }
+    },
+    "installer": {
+        "script": [
+            "$foldername = (Get-Item $dir\\WezTerm-windows-*).Name",
+            "Move-Item $dir\\$foldername\\* $dir",
+            "Remove-Item $dir\\$foldername"
+        ]
+    },
+    "bin": "wezterm.exe",
+    "shortcuts": [
+        [
+            "wezterm-gui.exe",
+            "WezTerm"
+        ]
+    ]
+}


pratikpc:
@rashil2000 I personally think we should not add more nightlies to extras.

Not till excavator issues are fixed at least.

rashil2000:
Nightlies are not affected by Excavator, Excavator skips these manifests. So no issues there.

pratikpc:
@rashil2000 so how are these updated then?

This for example cannot update

pratikpc:
@rashil2000 I do believe we should add new nightlies to https://github.com/ScoopInstaller/Nightlies only.

rashil2000:
They don't update. Manifests like rust-nightly do not update - they just download the latest nightly available at the time of installation. Because we can't keep track of non-versioned hashes daily.

rashil2000:
The Nightlies bucket has been deprecated :)

pratikpc:
> They don't update. Manifests like rust-nightly do not update - they just download the latest nightly available at the time of installation.

That does not sound like something I can agree with.

This would mean a Scoop user would have to manually perform re-installs whenever new nightlies come out and would have to manually update.

@rashil2000 if this is the case, as far as I can see, with excavator not running on nightlies bucket, we should move this PR there.

pratikpc:
@rashil2000 in that case we should bring it back to supported state right?

It's not like it would involve much effort to maintain or support.

Nightlies is till present in bucket.json so am not sure how deprecated it is.

https://github.com/ScoopInstaller/Scoop/blob/master/buckets.json#L5

rashil2000:
> This would mean a Scoop user would have to manually perform re-installs whenever new nightlies come out and would have to manually update.

If a user is downloading a nightly build, it's a fair assumption that they are aware of this.

Regarding the nightlies bucket, you can refer to the discussion here - https://github.com/ScoopInstaller/Extras/issues/2821

If we are to revive it, we will need to transfer 8 manifests from the Main bucket and 3 (+1 this one) manifests from the Extras bucket there

pratikpc:
> If we are to revive it, we will need to transfer 8 manifests from the Main bucket and 3 (+1 this one) manifests from the Extras bucket there

Or we could just add new PRs there while letting the old PRs remain here on account of age.

> If a user is downloading a nightly build, it's a fair assumption that they are aware of this.

I don't think that would be the case.

pratikpc:
But if we add it to a nightlies bucket and make it clear how nightlies work currently in Scoop, that would be a lot more clearer.

rashil2000:
> But if we add it to a nightlies bucket and make it clear how nightlies work currently in Scoop, that would be a lot more clearer.

Yeah that will be better

niheaven:
Maybe consider add it to https://github.com/ScoopInstaller/Versions if there is a `wezterm` already in `extras`?

rashil2000:
Isn't the Versions bucket for older and frozen packages?

I don't feel a nightly would fit there.

niheaven:
📦 A Scoop bucket for alternative versions of apps.

rashil2000:
I think we can aim for a more granular approach to buckets:

Versions - frozen versions, older versions, previews, derived apps

Nightlies - apps that update daily

PR #7297, opened 2021-11-10
gimp: Fix broken regex in installer script

kriswilk:
The installer script in this manifest has a broken regex. It currently leaves a broken path to the bundled python interpreter. This PR corrects the regex, at least for the latest version of GIMP's file.

This change has been tested locally.


rashil2000:
> This change has been tested locally.

With both a clean install and an existing install/upgrade, right?

kriswilk:
> > This change has been tested locally.
> 
> With both a clean install and an existing install/upgrade, right?

Clean install. The affected file is not persisted, so both a clean install and update (whether forced or not) will extract a fresh copy from the app payload. The old version simply remains in the old version folder or the _xxxxx.old folder in the case of a forced update.

That said, I just gave a forced update a try and it does work fine...

rashil2000:
Great, thanks for the fix!

PR #7303, opened 2021-11-11
impregnate: Add version 1.2.2

### LINKED ISSUE #7290: [Request] Impregnate
## Package Request
Add Impregnate (https://github.com/Cumcord/Impregnate).

### Information
Impregnate is the official installer for the Cumcord Discord client modification.

**Name:**
Impregnate

**Description:**
Impregnate is the official installer for the Cumcord Discord client modification.

**Homepage:**
https://cumcord.com/#install
https://github.com/Cumcord/Impregnate/releases

**Download link:**
https://github.com/Cumcord/Impregnate/releases/latest

https://github.com/Cumcord/Cumcord repo has 61 stars.

Cr3atable:
This pull request adds Impregnate (https://github.com/Cumcord/Impregnate), the installer for Cumcord (https://cumcord.com/).

Closes #7290

### DIFF
--- a/bucket/impregnate.json
+++ b/bucket/impregnate.json
@@ -1,28 +1,18 @@
 {
-    "version": "1.2.2",
+    "version": "1.2.3",
     "description": "A standalone installer for Cumcord, a Discord client modification that allows you to interface with Discord's internals via plugins.",
     "homepage": "https://github.com/Cumcord/Impregnate",
     "license": "MIT",
     "suggest": {
-        "discord": [
-            "discord",
-            "discord-ptb",
-            "discord-canary"
-        ]
+        "Discord": "extras/discord"
     },
     "architecture": {
         "64bit": {
-            "url": "https://github.com/Cumcord/Impregnate/releases/download/v1.2.2/impregnate.exe",
-            "hash": "ae00db169c649428da2bc7baea1acaaa0290c95fb1de9852f3cb9c63a5e4b337"
+            "url": "https://github.com/Cumcord/Impregnate/releases/download/v1.2.3/impregnate.exe",
+            "hash": "138cde1b90549f94a1f81adac82f765b1addc6208bebc750e210c36ea8516745"
         }
     },
-    "bin": "impregnate.exe",
-    "shortcuts": [
-        [
-            "impregnate.exe",
-            "Impregnate"
-        ]
-    ],
+    "shortcuts": [["impregnate.exe", "Impregnate"]],
     "checkver": "github",
     "autoupdate": {
         "architecture": {
--- a/bucket/impregnate.json
+++ b/bucket/impregnate.json
@@ -12,7 +12,12 @@
             "hash": "138cde1b90549f94a1f81adac82f765b1addc6208bebc750e210c36ea8516745"
         }
     },
-    "shortcuts": [["impregnate.exe", "Impregnate"]],
+    "shortcuts": [
+        [
+            "impregnate.exe",
+            "Impregnate"
+        ]
+    ],
     "checkver": "github",
     "autoupdate": {
         "architecture": {
--- a/bucket/impregnate.json
+++ b/bucket/impregnate.json
@@ -1,6 +1,6 @@
 {
     "version": "1.2.2",
-    "description": "A standalone installer for Cumcord.",
+    "description": "A standalone installer for Cumcord, a Discord client modification that allows you to interface with Discord's internals via plugins.",
     "homepage": "https://github.com/Cumcord/Impregnate",
     "license": "MIT",
     "architecture": {
--- a/bucket/impregnate.json
+++ b/bucket/impregnate.json
@@ -3,6 +3,13 @@
     "description": "A standalone installer for Cumcord, a Discord client modification that allows you to interface with Discord's internals via plugins.",
     "homepage": "https://github.com/Cumcord/Impregnate",
     "license": "MIT",
+    "suggest": {
+        "discord": [
+            "discord",
+            "discord-ptb",
+            "discord-canary"
+        ]
+    },
     "architecture": {
         "64bit": {
             "url": "https://github.com/Cumcord/Impregnate/releases/download/v1.2.2/impregnate.exe",
--- a/bucket/impregnate.json
+++ b/bucket/impregnate.json
@@ -1,22 +1,31 @@
 {
-    "version": "1.2.2",
-    "description": "A standalone installer for Cumcord.",
-    "homepage": "https://github.com/Cumcord/Impregnate",
-    "license": "Apache-2.0",
-    "url": "https://github.com/Cumcord/Impregnate/releases/download/v1.2.2/impregnate.exe",
-    "hash": "AE00DB169C649428DA2BC7BAEA1ACAAA0290C95FB1DE9852F3CB9C63A5E4B337",
-    "bin": "impregnate.exe",
-    "shortcuts": [
-        [
-            "impregnate.exe",
-            "Impregnate"
-        ]
-    ],
-    "checkver": {
-        "url": "https://github.com/Cumcord/Impregnate/releases/latest",
-        "regex": "v(\\d\\.\\d+\\.\\d+),"
-    },
-    "autoupdate": {
-        "url": "https://github.com/Cumcord/Impregnate/releases/download/$version/impregnate.exe"
+  "version": "1.2.2",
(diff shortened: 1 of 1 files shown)


bucket/impregnate.json
```diff
@@ -0,0 +1,22 @@
...
+    "hash": "AE00DB169C649428DA2BC7BAEA1ACAAA0290C95FB1DE9852F3CB9C63A5E4B337",
+    "bin": "impregnate.exe",
+    "shortcuts": [
+        [
+            "impregnate.exe",
+            "Impregnate"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/Cumcord/Impregnate/releases/latest",
```
pratikpc, line 16: I would recommend following checkver: github format

bucket/impregnate.json
```diff
@@ -0,0 +1,22 @@
...
+        "regex": "v(\\d\\.\\d+\\.\\d+),"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Cumcord/Impregnate/releases/download/$version/impregnate.exe"
```
pratikpc: Is it 32 or 64 bit?

bucket/impregnate.json
```diff
@@ -0,0 +1,22 @@
...
+    "bin": "impregnate.exe",
+    "shortcuts": [
+        [
+            "impregnate.exe",
+            "Impregnate"
+        ]
+    ],
+    "checkver": {
+        "url": "https://github.com/Cumcord/Impregnate/releases/latest",
+        "regex": "v(\\d\\.\\d+\\.\\d+),"
```
pratikpc: Is this needed?

rashil2000:
Is the application 32bit or 64bit?

Cr3atable:
> Is the application 32bit or 64bit?

64 bit.

bucket/impregnate.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "1.2.2",
+    "description": "A standalone installer for Cumcord.",
+    "homepage": "https://github.com/Cumcord/Impregnate",
+    "license": "Apache-2.0",
```
pratikpc: License mentioned is MIT

https://github.com/Cumcord/Impregnate/blob/288f37d550218bdabc8dc4395c811a916d1170e9/LICENSE

pratikpc:
Would recommend renaming to organisation name-package name.json

rashil2000:
> > Is the application 32bit or 64bit?
> 
> 64 bit.

In that case, put the URL and hash in architecture block. And update the `autoupdate` section too

bucket/impregnate.json
pratikpc: Could you check if this application works with checkver.ps1 offline once with ForceUpdate parameter

https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#using-checkverps1-to-query-and-autoupdate

Cr3atable:
> > > Is the application 32bit or 64bit?
> > 
> > 
> > 64 bit.
> 
> In that case, put the URL and hash in architecture block. And update the `autoupdate` section too

Resolved

pratikpc:
@Cr3atable that won't work

pratikpc (changes requested):
Correct license

Autoupdate is incorrect.

Regex probably is both incorrect and unnecessary

Will work with checkver: github as well

pratikpc:
Can you upload a video or screenshots of what happens when you install the tool for the first time?

Is this an installer for another tool for example?

lexisother:
https://user-images.githubusercontent.com/54973868/141304245-dabed893-62b9-4f68-a73f-0f49a8a34952.mp4

pratikpc:
Thanks for sharing but this looks like an installer for another tool rather than the application itself.

I am not very aware of how Discord works but this app does not look very portable to me.

Views @rashil2000?

bucket/impregnate.json
```diff
@@ -0,0 +1,27 @@
+{
+    "version": "1.2.2",
+    "description": "A standalone installer for Cumcord.",
```
pratikpc, line 3: I would also recommend expanding on the definition and making it clear what the tool this installs does.

pratikpc:
@lexisother can we not install the tool this installs directly?

Instead of using another installer tool for the same purpose?

lexisother:
I'm unsure how that would work. The installation logic isn't exactly simple, and is best done by a tool such as this one.
See: https://github.com/Cumcord/Impregnate/blob/de15a565984f18f046fe75a6b1be8ba352778926/src/managerView.go#L33-L51

pratikpc:
@lexisother this looks like it could be doable with PowerShell

Maybe with an installer and uninstaller script. Scoop supports those.

But am still concerned about the portability of this solution. And what the update process would look like.

Cr3atable:
> @lexisother this looks like it could be doable with PowerShell
> 
> Maybe with an installer and uninstaller script. Scoop supports those.
> 
> But am still concerned about the portability of this solution. And what the update process would look like.

This software installs a modification for a non-portable app. The installer is portable, the app it installs to isn't. That's why scoop would not be a good fit for managing the installation or uninstallation of it.

rashil2000:
I'm not much familiar with this application, but it looks like it installs mods in the Discord PC client?

@pratikpc we can't have manifests for each individual mod. But we can have one for a mod installer app.

lexisother:
Any news? @pratikpc

pratikpc (commented):
Manifest looks fine.

I would like to know what happens when:-
1. Discord installed via Scoop receives an update
2. This ideally would change directory structure
3. Does this mean this setup ends up being completely broken and requires re-runs?
4. If true, add it to notes

I would recommend
1. Suggest block for Discord
2. Changing name of package to org name-app name or app name-setup because the app is far more recognized. (Evidence would be the difference in GitHub Stars).
Based on info received, this is an installer for an app
3. Also need to add notes detailing that the setup needs to be rerun after each and every update as it is not a very portable application.

I am still unsure if this should be added to Scoop due to a lack of portability. In my opinion, this is more suited for the NP bucket but if other maintainers choose to add, I personally have no issues

lexisother:
> Discord installed via Scoop receives an update
Nothing, for this program injects into the Discord `AppData` folder, not the directory where scoop installs Discord.

> This ideally would change directory structure
Directory structure of what?

> Does this mean this setup ends up being completely broken and requires re-runs?
No, refer to the first quote reply.

> I would suggest:
> Suggest block for Discord
On it.

> Changing name of package to org name-app name or app name-setup
Example? I don't follow.

> Also need to add notes detailing that the setup needs to be rerun after each and every update as it is not a very portable application.
False.

bucket/impregnate.json
```diff
@@ -0,0 +1,34 @@
...
+            "discord-canary"
+        ]
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Cumcord/Impregnate/releases/download/v1.2.2/impregnate.exe",
+            "hash": "ae00db169c649428da2bc7baea1acaaa0290c95fb1de9852f3cb9c63a5e4b337"
+        }
+    },
+    "bin": "impregnate.exe",
```
rashil2000, line 19: Does the binary take any commandline arguments? If not, remove this line.

bucket/impregnate.json
```diff
@@ -0,0 +1,34 @@
...
+    "version": "1.2.2",
+    "description": "A standalone installer for Cumcord, a Discord client modification that allows you to interface with Discord's internals via plugins.",
+    "homepage": "https://github.com/Cumcord/Impregnate",
+    "license": "MIT",
+    "suggest": {
+        "discord": [
+            "discord",
+            "discord-ptb",
+            "discord-canary"
+        ]
```
rashil2000, line 11:
```suggestion
        "Discord": "extras/discord"
```

rashil2000 (changes requested):
Update manifest to version 1.2.3

bucket/impregnate.json
```diff
@@ -1,28 +1,18 @@
...
         }
     },
-    "bin": "impregnate.exe",
-    "shortcuts": [
-        [
-            "impregnate.exe",
-            "Impregnate"
-        ]
-    ],
+    "shortcuts": [["impregnate.exe", "Impregnate"]],
```
rashil2000, line 15: Don't do this, this is hard to read. Revert to as it was before.

rashil2000 (changes requested):

PR #7311, opened 2021-11-12
update ungoogled-chromium to 95.0.4638.69-1

nuudul:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/ungoogled-chromium.json
+++ b/bucket/ungoogled-chromium.json
@@ -1,5 +1,5 @@
 {
-    "version": "89.0.4389.114-1",
+    "version": "95.0.4638.69-1",
     "description": "Modifications to Google Chromium for removing Google integration and enhancing privacy, control, and transparency.",
     "homepage": "https://github.com/Eloston/ungoogled-chromium",
     "license": {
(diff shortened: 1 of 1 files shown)


rashil2000:
There's a problem in autoupdate

pratikpc:
@rashil2000 @nuudul has changed the URL and repo from which ungoogled-chromium is being downloaded.

pratikpc:
@rashil2000 further, the fork they have switched to has zero stars

https://github.com/Nifury/ungoogled-chromium-binaries/

pratikpc:
@rashil2000 the fork is mentioned on the official website but by the looks of it older versions are not available and the repo is probably liable to be changed multiple times which would require updating checkver and autoupdate multiple times and it will remain in a broken state for a long long time

@nuudul could you please share your views regarding the ungoogled-chromium-portable package?

rashil2000:
The new repo is mentioned on their website - https://ungoogled-software.github.io/ungoogled-chromium-binaries/releases/windows/64bit/95.0.4638.69-1

For the time being we can switch to what the website says - otherwise this will stay out of date

pratikpc (changes requested):
@nuudul you will have to update both checkver and autoupdate.

You further will have to extract the .1 manually from the URL present

pratikpc:
> otherwise this will stay out of date

The moment they change repos it becomes outdated too. For example they just now changed to the current one

@rashil2000 also their release schedule is very very slow if you look at the directory listing

https://ungoogled-software.github.io/ungoogled-chromium-binaries/releases/windows/64bit/

After April with v89, this is their first release.

@nuudul could you share your opinion on this?

https://github.com/ScoopInstaller/Extras/blob/master/bucket/ungoogled-chromium-portable.json

If https://github.com/macchrome/winchrome/ is okay for example, we could use that. They also for example have v94, v93 etc.

Issue #7096, opened 2021-10-28
Replace google-chrome series broken installer


Lutra-Fs:
Opened as @rashil2000 requested. The affiliated pull request is #7238 now,

issaclin32:
I vote up for adapting @Lutra-Fs 's solution.

This may bring some issues. But it is better than a **obviously broken** one.

We can consider switching to Chrome's official API later, or make our own version if needed.

Another option is to involve @Lutra-Fs 's repo as an affiliation of Scoop, such as the case of [games bucket](https://github.com/Calinou/scoop-games), [nerd-fonts bucket](https://github.com/matthewjberger/scoop-nerd-fonts) or [Java bucket](https://github.com/ScoopInstaller/Java) (previously a repo of *se35710*). Just need to add a few maintainers from Scoop as collaborators.

Lutra-Fs:
> Another option is to involve @Lutra-Fs 's repo as an affiliation of Scoop, such as the case of games bucket, nerd-fonts bucket or Java bucket (previously a repo of se35710). Just need to add a few maintainers from Scoop as collaborators.
If there is anything I can help with, just ping me. My repo is now a fork of Sukka (the original API provider) one with the fix of the problem. It is not a scoop bucket. If the repo needs to be a scoop bucket, we will need to add the relevant part.

issaclin32:
> If the repo needs to be a scoop bucket, we will need to add the relevant part.

Thank you for your great help.
The repo does not have to become a bucket. I just want our maintainers to have write access, 
so that we can fix the problem sooner (even if you are busy).

Lutra-Fs:
> > If the repo needs to be a scoop bucket, we will need to add the relevant part.
> 
> Thank you for your great help. The repo does not have to become a bucket. I just want our maintainers to have write access, so that we can fix the problem sooner in the future (even when you are busy).

Who should be added as a collaborator？I can do it now.

issaclin32:
> Who should be added as a collaborator？I can do it now.

I would suggest the following:
@rasa @niheaven -- They are the early members of *Scoop*. They may not respond quickly but are more experienced.
@stevehipwell @rashil2000 @pratikpc @issaclin32 -- Currently most active members

pratikpc:
Can we not use https://chromeenterprise.google/browser/download/#windows-tab?

Why do we need to use 3rd party sites?

Using the official sites is just safer tbh

Jogai:
[...]

PR #7317, opened 2021-11-13
ryzen-controller: Add version 2.5.4

### LINKED ISSUE #7261: [Request] Ryzen Controller
### Information

**Name:**
Ryzen Controller

**Description:**
Ryzen Controller is an open source utility that unleashes the power of AMD Ryzen™ laptops with an easy-to-use GUI. Tweaks can be saved as custom presets which can be enabled whenever you need extra power - or disabled when you want to keep things cool.

**Homepage:**
https://ryzencontroller.com/
https://gitlab.com/ryzen-controller-team/ryzen-controller

**Download link(s):**
https://gitlab.com/ryzen-controller-team/ryzen-controller/-/releases

gitlab 164 star

issaclin32:
close #7261

[Ryzen Controller](https://ryzencontroller.com/) is an open source utility that unleashes the power of AMD Ryzen™ laptops with an easy-to-use GUI. Tweaks can be saved as custom presets which can be enabled whenever you need extra power - or disabled when you want to keep things cool.

**NOTES**:
* This package has to be *updated manually* because the download URL is based on [Job IDs](https://gitlab.com/ryzen-controller-team/ryzen-controller/-/jobs/1710820449/artifacts/browse/dist/win/). The newly proposed *Update Handler* (#7096) may solve the problem.

* persist is not needed: config file is stored at `$Env:AppData\ryzen-controller`

### DIFF
--- a/bucket/ryzen-controller.json
+++ b/bucket/ryzen-controller.json
@@ -23,7 +23,14 @@
         ]
     ],
     "checkver": {
-        "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/tags?sort=name_desc&format=atom",
-        "regex": "ryzen-controller/-/tags/([\\d.]+)"
+        "url": "https://scoopinstaller.github.io/UpdateTracker/ryzen-controller/latest.xml",
+        "regex": "<version>([\\d.]+)</version><url>(?<url>.*?)</url>"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "$matchUrl#/dl.7z"
+            }
+        }
     }
 }
--- a/bucket/ryzen-controller.json
+++ b/bucket/ryzen-controller.json
@@ -5,7 +5,7 @@
     "license": {
         "identifier": "CC0-1.0",
         "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/blob/master/LICENSE.md"
-    }
+    },
     "architecture": {
         "64bit": {
             "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/jobs/1710820449/artifacts/raw/dist/win/Ryzen%20Controller%20Setup%202.5.4.exe#/dl.7z",
--- a/bucket/ryzen-controller.json
+++ b/bucket/ryzen-controller.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.54",
+    "version": "2.5.4",
     "description": "Utility for tweaking AMD Ryzen laptops.",
     "homepage": "https://ryzencontroller.com/",
     "architecture": {
--- a/bucket/ryzen-controller.json
+++ b/bucket/ryzen-controller.json
@@ -2,6 +2,10 @@
     "version": "2.5.4",
     "description": "Utility for tweaking AMD Ryzen laptops.",
     "homepage": "https://ryzencontroller.com/",
+    "license": {
+        "identifier": "CC0-1.0",
+        "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/blob/master/LICENSE.md"
+    }
     "architecture": {
         "64bit": {
             "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/jobs/1710820449/artifacts/raw/dist/win/Ryzen%20Controller%20Setup%202.5.4.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


issaclin32:
Not sure why PR check for `checkver` is failing.
```
.\checkver ryzen-controller
ryzen-controller: 2.5.4
```

bucket/ryzen-controller.json
```diff
@@ -0,0 +1,26 @@
...
+        "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\"",
+        "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall Ryzen Controller.exe\" -Force -Recurse | Out-Null"
+    ],
+    "shortcuts": [
+        [
+            "Ryzen Controller.exe",
+            "Ryzen Controller"
+        ]
+    ],
+    "checkver": "(?sm)Windows 10.*?Version ([\\d.]+) via GitLab"
```
pratikpc, line 25: You can use this URL

https://gitlab.com/ryzen-controller-team/ryzen-controller/-/tags?sort=name_desc&format=atom

Maybe use XMLPath or something?

issaclin32: Thanks for the info.

I tried using XPath, but it seems that the XML cannot be handled properly by *Scoop*. I think it is related to this line: `<feed xmlns="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">`

I am now using *regex* instead.
Will investigate about the bug later.

log:
```
MethodInvocationException: D:\scoop\apps\scoop\current\bin\checkver.ps1:234
Line |
 234 |                  $nsmgr.AddNamespace('xmlns', $_.Value)
     |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "AddNamespace" with "2" argument(s): "Prefix "xmlns" is reserved for use by XML."

ryzen-controller: couldn't find '/feed/entry[1]/title' in https://gitlab.com/ryzen-controller-team/ryzen-controller/-/tags?sort=name_desc&format=atom
```

bucket/ryzen-controller.json
```diff
@@ -0,0 +1,29 @@
...
+    "version": "2.5.4",
+    "description": "Utility for tweaking AMD Ryzen laptops.",
+    "homepage": "https://ryzencontroller.com/",
+    "license": {
+        "identifier": "CC0-1.0",
+        "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/blob/master/LICENSE.md"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://gitlab.com/ryzen-controller-team/ryzen-controller/-/jobs/1710820449/artifacts/raw/dist/win/Ryzen%20Controller%20Setup%202.5.4.exe#/dl.7z",
```
pratikpc, line 11: How do you plan to extract 1710820449?

issaclin32: That is why we need to implement `UpdateTracker`.

This package (and many other packages) has to be updated manually. However, in this case, it can be solved by a simple mechanism such as the case of `FossHub`'s hash.

pratikpc: I personally am not very supportive of manual updates.

issaclin32: > I personally am not very supportive of manual updates.

I agree. Manual updates cost a lot of unnecessary labor. I think we should implement more mechanism to eliminate manual updates as much as possible.

However I don't think it is a reason **not to** add a package. I don't think our solutions has to be "perfect" right from the beginning. We can gradually improve that. ^_^

pratikpc: We can merge it later when we support it on stable as well

issaclin32: > We can merge it later when we support it on stable as well

Then let's hold this PR until it is finished.

issaclin32:
The [update tracker](https://github.com/ScoopInstaller/UpdateTracker/blob/master/ryzen-controller/run.py) has been implemented ([log](https://github.com/ScoopInstaller/UpdateTracker/runs/4346344998?check_suite_focus=true)). I will check if more packages can be automatically updated in this manner.

niheaven:
Should this one's autoupdate be temporally removed since it breaks CI check?

PR #7318, opened 2021-11-13
Update Brave Browser version

cristianrz:
Installing `brave` installs a 2 year old version of version where autoupdate is broken which is a potential security risk. Updated Brave to 1.31.87 which is the last version I found that can be installed as portable.

### DIFF
--- a/bucket/brave.json
+++ b/bucket/brave.json
@@ -1,8 +1,11 @@
 {
     "version": "1.31.87-77",
     "description": "Secure, Fast & Private Web Browser with Adblocker",
-    "homepage": "https://github.com/portapps/brave-portable",
-    "checkver": "github",
+    "homepage": "https://brave.com",
+    "checkver": {
+        "url": "https://github.com/portapps/brave-portable",
+        "regex": "/releases/tag/(?:v|V)?([\\d.-]+)"
+    },
     "license": {
         "identifier": "Freeware,BSD-3-Clause,GPL-3.0-only,...",
         "url": "https://github.com/brave/brave-browser/blob/master/LICENSE"
--- a/bucket/brave.json
+++ b/bucket/brave.json
@@ -2,10 +2,6 @@
     "version": "1.31.87-77",
     "description": "Secure, Fast & Private Web Browser with Adblocker",
     "homepage": "https://brave.com",
-    "checkver": {
-        "url": "https://github.com/portapps/brave-portable",
-        "regex": "/releases/tag/(?:v|V)?([\\d.-]+)"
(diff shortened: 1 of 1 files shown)


rashil2000:
- Do not remove autoupdate
- Add a checkver. Checkver was missing, that's why the manifest didn't get updated
- Use the official GitHub release. Do not use other URLs

cristianrz:
@rashil2000 brave's official releases do not include a 7z/zip file and the *.exe installer does not allow changing the install location. Is it a no-go that it is not brave's github? Portapps is a recognised portable apps website.

I will check the other 2 issues.

rashil2000:
How about https://github.com/brave/brave-browser/releases/download/v1.32.99/BraveBrowserStandaloneSilentSetup.exe? Scoop has the ability to extract exe files too.

cristianrz:
Just gave it a try and it won't work, it extracts a file named '102' (same with the BraveBrowserSetup.exe). Manually browsing the file with 7-zip doesn't seem to show any files that can be trivially extracted, any other ideas?

Latest commit should work with the portapps file if that's acceptable.

rashil2000:
Using Portapps is fine if the official releases are not extractable.

Some more issues in the manifest:
- Don't change the homepage. brave.com is fine.
- The URL should be in arch 64bit block. For autoupdate also.
- Checkver: github will not work. Use this:
```json
"checkver": {
    "url": "https://github.com/portapps/brave-portable",
    "regex": "/releases/tag/(?:v|V)?([\\d.-]+)"
}
```

cristianrz:
Used "github" because of https://scoop-docs.vercel.app/docs/concepts/App-Manifest-Autoupdate.html#special-cases

Fixed it anyway with your suggestions. Hopefully it's alright now.

rashil2000:
> Used "github" because of https://scoop-docs.vercel.app/docs/concepts/App-Manifest-Autoupdate.html#special-cases
> 
> Fixed it anyway with your suggestions. Hopefully it's alright now.

That works only when your homepage is a github repo. Here it is brave.com.

Also the default regex for github won't be valid here, because the version string contains a hyphen.

rasa:
We may want to just move https://github.com/rasa/scoops/blob/master/bucket/brave.json to Extras, as it sets the proper persist folders, etc.

rashil2000:
> We may want to just move https://github.com/rasa/scoops/blob/master/bucket/brave.json to Extras, as it sets the proper persist folders, etc.

2f2042d98e4ccd765b989d3a944760659bafb06d

PR #7332, opened 2021-11-15
Fix old path in vscode-insiders.json that was blocking installation

lediur:
This PR corrects some old script URLs that pointed to the previous repos maintained by `lukesampson` to the current repo, which was causing installation to fail with a 404.

Might be worth doing a find-replace on all of the JSONs in this repository and making the same fix.


rashil2000:
It is worth noting that the 404 was not due to the GitHub URL, but due the slash in path. GitHub automatically redirects links to current repos, so technically it shouldn't be a problem.

PR #7373, opened 2021-11-20
Update firefox.json: v94.0.2

RavenMacDaddy:
- Updates to version 94.0.2.
- Not fetched by checkver since Mozilla hasn't updated their release notes, but the news broke on the Neowin software page on nov 18th and it was added to the archive page used in the manifest in the evening of nov 19th.

### DIFF
--- a/bucket/firefox.json
+++ b/bucket/firefox.json
@@ -1,16 +1,16 @@
 {
-    "version": "94.0.1",
+    "version": "94.0.2",
     "description": "Popular open source web browser.",
     "homepage": "https://www.mozilla.org/firefox/",
     "license": "MPL-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://archive.mozilla.org/pub/firefox/releases/94.0.1/win64/en-US/Firefox%20Setup%2094.0.1.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


rashil2000:
> Not fetched by checkver since Mozilla hasn't updated their release notes, but it was added to the archive page used in the manifest in the evening of nov 19th.

Should we switch to the archive page (https://archive.mozilla.org/pub/firefox/releases/) in checkver then?

RavenMacDaddy:
That does sound like a good idea.

Do you mind fixing that change before merging?

rashil2000:
I don't :)

Would you like to try that? In this PR itself.

RavenMacDaddy:
Sure, just make it as a repository admin correction.

RavenMacDaddy:
To be clear, I spent a couple of hours previously to try and wrap my head around regex, but I can't understand it. 

If changing the checkver to the suggested URL requires that knowledge, which I suppose it does looking at the current regex code, it would be great if someone is able to make that correction in another commit based on mine as a baseline.

rashil2000:
Try this:

```json
"checkver": {
        "url": "https://archive.mozilla.org/pub/firefox/candidates/",
        "regex": "pub/firefox/candidates/([\\d.]+)-candidates",
        "reverse": true
},
```

PR #7375, opened 2021-11-20
touchcursor: Add version 1.7.1

g-berthiaume:
[Touchcursor](https://martin-stone.github.io/touchcursor) is a hidden gem for windows programming productivity.

> **TouchCursor** lets you use the home keys as cursor keys – in all Windows programs – keeping your fingers in the best position for fast typing. It's free and open-source.

Let me know if you see something to change.

### DIFF
--- a/bucket/touchcursor.json
+++ b/bucket/touchcursor.json
@@ -1,6 +1,6 @@
 {
     "version": "1.7.1",
-    "description": "TouchCursor lets you use the home keys as cursor keys – in all Windows programs – keeping your fingers in the best position for fast typing.",
+    "description": "Lets you use the home keys as cursor keys.",
     "homepage": "https://martin-stone.github.io/touchcursor/",
     "license": "GPL-3.0-only",
     "url": "https://github.com/martin-stone/touchcursor/releases/download/v1.7.1/TouchCursor-1.7.1.zip",
@@ -12,12 +12,5 @@
             "touchcursor.exe",
             "touchcursor"
         ]
-    ],
-    "checkver": {
-        "url": "https://martin-stone.github.io/touchcursor/",
-        "regex": "Version ([\\d.]+)"
-    },
-    "autoupdate": {
-        "url": "https://github.com/martin-stone/touchcursor/releases/download/v$version/TouchCursor-$version.zip"
-    }
+    ]
 }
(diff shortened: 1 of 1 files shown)


rashil2000:
Last release in 2010??? This is very out of date.

g-berthiaume:
> Last release in 2010??? This is very out of date.

I'm not sure what you mean by that.

This is a compiled C++ program using as only dependency the Win32 api; As long as windows exist, this software will work.
To me, this is what the epitome of good software looks like.

rashil2000:
I'm not saying it's bad.

But software that isn't patched regularly can be vulnerable. Putting it in an official bucket exposes users to risks.

g-berthiaume:
Ah ok. I see what you mean.

Looking at the repo, the author seems to be still active.
https://github.com/martin-stone/touchcursor/issues

That said, it's true it has had update in years.

g-berthiaume:
Your call. Your the maintainer.
I just though people would like, it so I created a PR. :)

rashil2000:
Let me ask some more maintainers about it. @issaclin32 @pratikpc thoughts?

bucket/touchcursor.json
```diff
@@ -0,0 +1,23 @@
...
+    "extract_dir": "TouchCursor",
+    "bin": "touchcursor.exe",
+    "shortcuts": [
+        [
+            "touchcursor.exe",
+            "touchcursor"
+        ]
+    ],
+    "checkver": {
+        "url": "https://martin-stone.github.io/touchcursor/",
```
pratikpc, line 17: Why not checkver: github?

pratikpc (changes requested):

bucket/touchcursor.json
g-berthiaume: I can certainly do that if you prefer the manifest to parse github releases.

issaclin32: Actually we can just remove `checkver` and `autoupdate`. The software has not been updated for a few years, so we don't need to check it. This saves resources (i.e. number of connections) for out *Excavator* bot.

issaclin32:
@rashil2000 I terms of security, I am fine with this package.
There are [source code](https://github.com/martin-stone/touchcursor/tree/master/touchcursor) available. We can see it only leverages Windows API to do its tasks.

If it is a software that involves inbound connections, e.g. `apache`, `filezilla-server`, `mercury`, `megacmd`. I will definitely say no. But for this one, I think it is okay.

rashil2000:
Okay, all good from me!

issaclin32 (approved):

PR #7380, opened 2021-11-21
featherwallet: Add version 1.0.1

shelmor:
Feather: a free Monero desktop wallet
https://featherwallet.org/
https://docs.featherwallet.org/guides/portable-mode

### DIFF
--- a/bucket/featherwallet.json
+++ b/bucket/featherwallet.json
@@ -3,6 +3,7 @@
     "description": "A Monero SPV Wallet",
     "homepage": "https://featherwallet.org/",
     "license": {
+        "identifier": "BSD-3-Clause",
         "url": "https://git.featherwallet.org/feather/feather/raw/branch/master/LICENSE"
     },
     "architecture": {
--- a/bucket/featherwallet.json
+++ b/bucket/featherwallet.json
@@ -1,0 +1,39 @@
+{
+    "version": "1.0.1",
+    "description": "A Monero SPV Wallet",
+    "homepage": "https://featherwallet.org/",
+    "license": {
+        "url": "https://git.featherwallet.org/feather/feather/raw/branch/master/LICENSE"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://featherwallet.org/files/releases/win/feather-1.0.1-win.zip",
+            "hash": "aa5f0a38d6258af2deb01c5c5c35acbc1e1311882357e7996c3022bff48932fe"
+        }
+    },
+    "pre_install": [
+        "Move-Item \"$dir\\feather*.exe\" \"$dir\\feather.exe\"",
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):

bucket/featherwallet.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": "1.0.1",
+    "description": "A Monero SPV Wallet",
+    "homepage": "https://featherwallet.org/",
+    "license": {
+        "url": "https://git.featherwallet.org/feather/feather/raw/branch/master/LICENSE"
```
issaclin32, line 6:
```suggestion
        "identifier": "BSD-3-Clause",
        "url": "https://git.featherwallet.org/feather/feather/raw/branch/master/LICENSE"
```
It is actually a [BSD 3-Clause license](https://opensource.org/licenses/BSD-3-Clause)

pratikpc: Also we don't need the URL as well

bucket/featherwallet.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://featherwallet.org/files/releases/win/feather-1.0.1-win.zip",
+            "hash": "aa5f0a38d6258af2deb01c5c5c35acbc1e1311882357e7996c3022bff48932fe"
+        }
+    },
+    "pre_install": [
+        "Move-Item \"$dir\\feather*.exe\" \"$dir\\feather.exe\"",
```
pratikpc, line 15: What is the default name after unzipping?

feather*.exe could cause issues later if they release something else as well

pratikpc: So if possible we should try to fix this to the best of our abilities.

bucket/featherwallet.json
issaclin32: > Also we don't need the URL as well

I tend to keep the URL, because this saves our time to find the license if something happens in the future like:
(1) Someone make a complaint that license is not correct
e.g. `LICENSE` exists in the repo because some library has it, but the software applies another rule.
e.g. The author of the PR mis-read the license as some generic one (like MIT), and our maintainers didn't find out.
(2) License changed by the developer.

I don't think it should be enforced to keep the URL, though.

bucket/featherwallet.json
issaclin32, line 15:
```suggestion
        "Rename-Item \"$dir\\feather-$version.exe\" \"$dir\\feather.exe\"",
```
In response to @pratikpc's comment

shelmor: this is actually how i approached it originally but didn't know i was allowed to use $version in there. thanks!

issaclin32 (approved):

PR #7388, opened 2021-11-23
scratch: use winrar to extract nsis sfx

### LINKED ISSUE #7338: Scratch install don't work anymore (last update)
## Bug Report

**Scratch**

### Current Behaviour
image

### Expected Behaviour

Correct installation

### Additional context/output

Tested the downloaded version. It's not an autoextractible .exe like previously. It's a classic .exe which is not extractible with 7zip.

### System details

**Windows version:** [e.g. 7, 8, 10]

**OS architecture:** [e.g. 32bit, 64bit]

**PowerShell version:** [output of `"$($PSVersionTable.PSVersion)"`]

phanirithvij:
* added winrar as a depenceny
* fixes #7338
* This is a workaround

### DIFF
--- a/bucket/scratch.json
+++ b/bucket/scratch.json
@@ -6,11 +6,15 @@
         "identifier": "Freeware",
         "url": "https://scratch.mit.edu/info/faq"
     },
-    "url": "https://downloads.scratch.mit.edu/desktop/Scratch%203.27.0%20Setup.exe#/dl.7z",
+    "depends": "winrar",
+    "url": "https://downloads.scratch.mit.edu/desktop/Scratch%203.27.0%20Setup.exe#/setup.exe",
     "hash": "343135e697db089a7ce6e8736950c4693b418d7cbc41151e33a45cac4ecc89db",
     "pre_install": [
-        "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-32.7z\" \"$dir\"",
-        "Remove-Item \"$dir\\`$*\", \"$dir\\Uninstall*\" -Recurse"
+        "$Args = @\"",
+        "x -IBCK \"$dir\\setup.exe\" \"$dir\"",
+        "\"@",
+        "Start-Process -Wait winrar.exe -NoNewWindow -ArgumentList $Args",
+        "Remove-Item \"$dir\\setup.exe\""
     ],
     "shortcuts": [
         [
--- a/bucket/scratch.json
+++ b/bucket/scratch.json
@@ -27,6 +27,6 @@
         "regex": "refs/tags/v([\\d.]+)"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


issaclin32:
My concern is that `"depends": "winrar"` can bring some confusing situations for end-users.

Most people have their WinRAR installed through other means. If we silently install another WinRAR without asking the user, people may complain about it.

I tend to add a new package named `winrar-helper`, which is almost indentical to current `winrar`, but does not add any bin/shortcuts. We can always keep `winrar-helper` updated, so it is better than detecting WinRAR installations in the script.

phanirithvij:
Where is `winrar-helper` located I don't see it in this bucket, is it yet to be created? If so I will make that fix in a new PR.

issaclin32:
> Where is `winrar-helper` located I don't see it in this bucket, is it yet to be created? If so I will make that fix in a new PR.

It will be created in another PR. However I cannot decide this on my own, so I will wait for other maintainers' opinions in that PR. It will be merged in a few days. Please give us a little bit of time.

phanirithvij:
I understand.
I was thinking as an alternative, we can download winrar to a temp directory in this scratch.json manifest `pre_install` step itself and use it to extract this nsis sfx.
But I agree that a helper in the bucket is more useful for keeping winrar up to date and also for other packages which might require winrar.

bucket/scratch.json
```diff
@@ -6,11 +6,15 @@
         "identifier": "Freeware",
         "url": "https://scratch.mit.edu/info/faq"
     },
-    "url": "https://downloads.scratch.mit.edu/desktop/Scratch%203.27.0%20Setup.exe#/dl.7z",
+    "depends": "winrar",
```
phanirithvij, line 9:
```suggestion
    "depends": "winrar-helper",
```

bucket/scratch.json
```diff
@@ -6,11 +6,15 @@
...
+    "url": "https://downloads.scratch.mit.edu/desktop/Scratch%203.27.0%20Setup.exe#/setup.exe",
     "hash": "343135e697db089a7ce6e8736950c4693b418d7cbc41151e33a45cac4ecc89db",
     "pre_install": [
-        "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-32.7z\" \"$dir\"",
-        "Remove-Item \"$dir\\`$*\", \"$dir\\Uninstall*\" -Recurse"
+        "$Args = @\"",
+        "x -IBCK \"$dir\\setup.exe\" \"$dir\"",
+        "\"@",
+        "Start-Process -Wait winrar.exe -NoNewWindow -ArgumentList $Args",
```
phanirithvij, line 16:
```suggestion
        "Start-Process -Wait \"$(appdir winrar-helper $global)\\current\\winrar.exe\" -NoNewWindow -ArgumentList $Args",
```

phanirithvij (commented):
Once #7390 is reviewed these changes can be made

issaclin32 (approved):

PR #7389, opened 2021-11-23
winrar-helper: add a helper script for scoop

phanirithvij:
Adds a winrar helper script as per the suggestion https://github.com/ScoopInstaller/Extras/pull/7388#issuecomment-976313692
- Copied from winrar.json, removed bin/shortcuts/notes, edited description/homepage
- This is required for `scratch` refer #7338 #7388.

PR #7390, opened 2021-11-23
winrar: Add 'helper' package

issaclin32:
Will be used in #7388. (as a workaround for the extraction failure mentioned in #7338)

This adds an **WinRAR** "helper" package, which is identical to `winrar` package, but without bin/shortcut.

This is an alternative to `"depends:extras/winrar"`, as it can avoid some confusing situations for end-users.
Most people have their WinRAR installed through other means. If we silently install another WinRAR without asking the user, people may complain about it.
(Why there are 2 WinRARs? Why my WinRAR suddenly turn into English rather than my localized version?)

With this helper package, we can extract the files using *WinRAR* without affecting the users, and always keep this "helper" WinRAR updated.

### DIFF
--- a/bucket/winrar-helper.json
+++ b/bucket/winrar-helper.json
@@ -1,0 +1,32 @@
+{
+    "version": "6.02",
+    "description": "WinRAR helper package (only used for extracting installers such as NSIS)",
+    "homepage": "https://rarlab.com/",
+    "license": {
+        "identifier": "Shareware",
+        "url": "https://www.win-rar.com/gtb_priv.html?&L=0"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://www.rarlab.com/rar/winrar-x64-602.exe#/dl.7z",
+            "hash": "d41ed4b4de255bee35f93372d023203c9a43694ef88a759ad61b41dfbd0f345d"
+        },
+        "32bit": {
+            "url": "https://www.rarlab.com/rar/wrar602.exe#/dl.7z",
+            "hash": "f3c32238f23c09f989902644df19e0c1156a8ee9aab552e9c39e869e42c5a71f"
+        }
+    },
+    "pre_install": "if (!(Test-Path \"$persist_dir\\rarreg.key\")) { New-Item \"$dir\\rarreg.key\" | Out-Null }",
+    "persist": "rarreg.key",
+    "checkver": "WinRAR and RAR ([\\d.]+) release",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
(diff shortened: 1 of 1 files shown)


phanirithvij:
I opened a duplicate one #7389, please feel free to close one of these two.

issaclin32:
Waiting for opinions from other maintainers.

Valinor:
> Will be used in #7388. (as a workaround for the extraction failure mentioned in #7338)
> 
> This adds an **WinRAR** "helper" package, which is identical to `winrar` package, but without bin/shortcut.
> 
> This is an alternative to `"depends:extras/winrar"`, as it can avoid some confusing situations for end-users. Most people have their WinRAR installed through other means. If we silently install another WinRAR without asking the user, people may complain about it. (Why there are 2 WinRARs? Why my WinRAR suddenly turn into English rather than my localized version?)
> 
> With this helper package, we can extract the files using _WinRAR_ without affecting the users, and always keep this "helper" WinRAR updated.

This is a great improvement, this silent use of winrar. What about the licence (shareware). Can this be a problem?

issaclin32:
> What about the licence (shareware). Can this be a problem?

The license problem is somewhat unsettled.
WinRAR **explicitly** says it is a Shareware. However, it does not put any restrictions on free users, except for a message box.

I think we can leave this for discussion in another issue/PR.

rashil2000:
The idea seems fine to me, but won't https://github.com/ScoopInstaller/Scoop/pull/3502 be a better and more robust solution for this? Then we don't need to depend on anything.

issaclin32:
> The idea seems fine to me, but won't [ScoopInstaller/Scoop#3502](https://github.com/ScoopInstaller/Scoop/pull/3502) be a better and more robust solution for this? Then we don't need to depend on anything.

Yeah. It should be a better solution. But we will still need this `winrar-helper` package, since some NSIS installer now needs WinRAR for extraction.

I tend to add `"depends": "winrar-helper"` and use *pre_install* script as a quick fix. We can delete it from the manifest after built-in unpacker is settled and merged into *master*.

niheaven:
`Invoke-ExternalCommand winrar.exe @('x', '-ibck', '"F:\Downloads\Scratch 3.27.0 Setup.exe"', 'F:\Downloads\scratch\')`

Emm, then what's wrong with Scratch's installer?

niheaven:
I'll add it as fallback in https://github.com/ScoopInstaller/Scoop/pull/3502

issaclin32:
Base on the comments, I assume this package is ready to go.

rashil2000:
OK from me 👍

rashil2000 (approved):

PR #7397, opened 2021-11-24
Add GeoGebra version 6.0.676.0

tech189:
Relates to https://github.com/ScoopInstaller/Versions/pull/323#issuecomment-976392477 where we discussed adding the current version of GeoGebra.

One caveat to the manifest - checkver will need to be edited once the next major release arrives (7.0) due to the way it gets the latest version from the website. The website also has "5-0-676-0" as the latest version for the 6.0 release (!!) so my regex should be able to adapt once they fix that (I sent them an email about it).

### DIFF
--- a/bucket/geogebra.json
+++ b/bucket/geogebra.json
@@ -1,0 +1,48 @@
+{
+    "version": "6.0.676.0",
+    "description": "A dynamic mathematics software for education that brings together geometry, algebra, spreadsheets, graphing, statistics and calculus.",
+    "homepage": "https://www.geogebra.org/",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://www.geogebra.org/license"
+    },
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):

tech189:
> The website also has "5-0-676-0" as the latest version for the 6.0 release (!!) so my regex should be able to adapt once they fix that (I sent them an email about it).

They replied and unfortunately will not be fixing it because they only support official installations directly from their website. But I think it should be okay to keep this package because it can also be found in [Winget](https://github.com/microsoft/winget-pkgs/tree/master/manifests/g/GeoGebra), [Chocolatey](https://community.chocolatey.org/packages/geogebra-classic.install), and [Homebrew](https://formulae.brew.sh/cask/geogebra).

rashil2000:
No worries. Thanks for following up, @tech189 :)

PR #7400, opened 2021-11-24
trilium: Add persistent data

### LINKED ISSUE #4822: trilium: add Persistent data
Locations in scoop-extras: https://github.com/lukesampson/scoop-extras/blob/master/bucket/trilium.json

Project page: https://github.com/zadam/trilium

Wiki: https://github.com/zadam/trilium/wiki/Data-directory#changing-the-location-of-data-directory

shelmor:
- Closes #4822 

1. Update to the latest version (not sure why it wasn't already..)
2. Add persistent data as per [their wiki](https://github.com/zadam/trilium/wiki/Data-directory#changing-the-location-of-data-directory) and attempt to migrate existing configuration from AppData\Roaming to $persist_dir

Also I feel like renaming the existing `trilium-data` folder in AppData\Roaming to `trilium-data.old` or something like that to prevent possible confusion. It's not a big deal though. Thoughts?

### DIFF
--- a/bucket/trilium.json
+++ b/bucket/trilium.json
@@ -16,6 +16,17 @@
             "Trilium"
         ]
     ],
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\trilium-data\") -and (Test-Path \"$env:APPDATA\\trilium-data\")) {",
+        "    info \"Copying old '$env:APPDATA\\trilium-data' to '$persist_dir\\trilium-data'\"",
+        "    ensure \"$persist_dir\\trilium-data\" | Out-Null",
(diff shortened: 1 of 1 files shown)


rashil2000:
> Also I feel like renaming the existing trilium-data folder in AppData\Roaming to trilium-data.old or something like that to prevent possible confusion. It's not a big deal though. Thoughts?

If both \<package folder\>/trilium-data and %AppData%/trilium-data exist, which one will trilium prefer?

shelmor:
> If both <package folder>/trilium-data and %AppData%/trilium-data exist, which one will trilium prefer?

That will depend on the `TRILIUM_DATA_DIR` environment variable, hence I added `env_set` to the manifest.

rashil2000:
Okay, thanks for the info. Then I think no need to rename %AppData%/trilium-data.

Issue #7412, opened 2021-11-26
ImageGlass is not updated to the latest version


eugenesvk:
## Bug Report

**Package Name:** imageglass

### Current Behaviour

ImageGlass's version is outdated at `8.2.6.6`

### Expected Behaviour

ImageGlass should be updated to version `8.3.11.21`

### Possible Solution

I'll submit a PR that updates the manifest

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.0-preview.8

PR #7413, opened 2021-11-26
imageglass: Update to version 8.3.11.21

eugenesvk:
Used `checkver.ps1` ​from [autoupdate guide](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate) to update imageglass

Fixes: #7412

### DIFF
--- a/bucket/imageglass.json
+++ b/bucket/imageglass.json
@@ -1,14 +1,14 @@
 {
-    "version": "8.2.6.6",
+    "version": "8.3.11.21",
     "description": "A lightweight, versatile image viewer",
     "homepage": "https://imageglass.org",
     "license": "GPL-3.0-only",
     "notes": "If this app doesn't work maybe you need to clean '$dir\\igconfig.xml'.",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


eugenesvk:
@rashil2000 by the way, as far as I understand, the lack of autoupate is related to GitHub rate limiting, however some apps (like this one) also host binaries on their websites (https://imageglass.org/release/download/imageglass-kobe-8-3-11-21-79), would it make sense to go (back) to checking the website? Or is there a fallback option in the manifest for an alternative link or something?
Or is it not worth the hassle before the rate limit issue is resolved?

rashil2000:
If you can update the checkver to use an alternative source (like the website you mentioned), it would be great!

eugenesvk:
Well, I don't want to replace the link (websites can be brittle too, change format etc., so besides the rate limit issue, are actually worse than GitHub), but from a brief look at the app [manifest spec](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests) I gather there is actually no option for multiple source urls, **is that so**?

rashil2000:
For checker, only one URL is allowed.

Issue #7415, opened 2021-11-26
Update qimgv and qimgv-video to 1.0.2


eugenesvk:
## Bug Report

**Package Name:** qimgv and qimgv-video

### Current Behaviour

Versions of qimgv and qimgv-video are outdated at `0.9.1 `

### Expected Behaviour

Versions of qimgv and qimgv-video should be updated to `1.0.2 `

### Additional context/output

The dev doesn't host 32bit versions for the new versions (though mentions he'll post them later), so auto-updated fails since it fails to download 32bit versions for hash generation

### Possible Solution

I'll submit a PR that updates the manifest for the x64 versions while leaving the old ones as is

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.0-preview.8

PR #7416, opened 2021-11-26
qimgv and qimgv-video: Update to version 1.0.2 (x64 only)

eugenesvk:
Used `checkver.ps1` ​from [autoupdate guide](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate) to update qimgv and qimgv-video

### Additional context/output

As I mentioned in the issue, the dev doesn't host 32bit versions for the new versions (though mentions he'll post them later), so `checkver.ps1` auto-update fails on the full manifest since it fails to download 32bit versions for hash generation
I've left the 32-bit version as is (old `0.9.1`), only updated the 64-bit version

Fixes: #7415

### DIFF
--- a/bucket/qimgv-video.json
+++ b/bucket/qimgv-video.json
@@ -1,12 +1,12 @@
 {
-    "version": "0.9.1",
+    "version": "1.0.2",
     "description": "Fast, configurable and easy to use image viewer with video support",
     "homepage": "https://github.com/easymodo/qimgv",
     "license": "GPL-3.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/easymodo/qimgv/releases/download/v0.9.1/qimgv-x64_0.9.1-video.zip",
-            "hash": "26c679a572159b5e405b213c6ea433a8a284fa9d7b79ae90bb3763bd04efd59d"
+            "url": "https://github.com/easymodo/qimgv/releases/download/v1.0.2/qimgv-x64_1.0.2-video.zip",
+            "hash": "5476268f5f237fbb4dbd840ac626dc63f54007ab6d79a130e31dafeba35142ed"
         },
         "32bit": {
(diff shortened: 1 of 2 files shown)


rashil2000:
The version difference between 32bit and 64bit URLs seems a bit odd to me. It will definitely cause the CI to fail.

When will the author release 32bit versions? If 32bit version is discontinued, we can remove it from the manifest.

eugenesvk:
I'd remove them from the manifest, but then I saw a message:
  > 32bit builds will be uploaded later

— that's a [1.0.0](https://github.com/easymodo/qimgv/releases/tag/v1.0.0) release from Sep 23, 2021, two minor versions back — and decided to err on a safer side and leave the 32-bit version. No clue about author's intent, will ask him.
Meanwhile, do you collect any stats on what is actually downloaded, how many users still need the 32-bit version?

rashil2000:
I'm afraid we have no way of knowing that. Scoop does not collect _any_ information.

eugenesvk:
Ok, let's wait for the author's response then https://github.com/easymodo/qimgv/issues/366
(or we can always remove 32bit and wait till someone complains ;) )

rashil2000:
We can wait. I've added the upstream label.

eugenesvk:
The dev responded
> right.. i forgot about it
> expect for v1.0.3

I guess we should wait till 1.0.3 and then do a proper update then?

rashil2000:
Sure, update this PR itself, when 1.0.3 comes out.

eugenesvk:
well, it seems the 32bit was removed in this commit after all https://github.com/ScoopInstaller/Extras/commit/0daffab609f6c0a354ffde6174544ba5d1bd23b3

rashil2000:
Oh, didn't notice. 

We can add 32bit later, if and when it gets released

PR #7427, opened 2021-11-27
languagetool-electron: Add version 1.0.4

### LINKED ISSUE #7186: (Request) add LanguageTools
PackageIdentifier: Languagetools
PackageVersion: 1.0.2
PackageLocale: en-US
Publisher: LanguageTools
PackageName: LanguageTools
License: open source
ShortDescription: open source spellcheck app
Installers:

Architecture: x64
InstallerType: exe
InstallerUrl: https://languagetool.org/download/windows-app/LanguageTool-1.0.2.exe
ManifestType: singleton
ManifestVersion: 1.0.0

shelmor:
- Closes #7186


rashil2000:
What is the difference between the 2 variants?

shelmor:
as far as i understand the original one doesn't have a binary (uses jar files) and require java runtime installed. this one is a compiled windows binary and thus can be run without java

rashil2000:
I think it will be better to just replace the languagetool.json then, since it can run on its own.

shelmor:
i think so as well, but it might not be an option for someone who already uses the java version and is cool with it. so i went for a separate manifest, that way people will have a choice and imo having a choice is always good. 
what we can do is replace the existing one with this and rename the old one to something like `languagetool-java` to keep both manifest for those in need.

rashil2000:
> what we can do is replace the existing one with this and rename the old one to something like languagetool-java to keep both manifests for those in need.

Okay, please do that

rashil2000 (approved):

issaclin32 (approved):

PR #7437, opened 2021-11-29
sysinternals: fix disk2vhd in 64-bit architecture  installation

### LINKED ISSUE #7431: disk2vhd.exe doesn't work on 64 bit windows 10 (need disk2vhd64.exe also)
## Bug Report

**Package Name:** sysinternals

### Current Behaviour

disk2vhd fails to launch on 64bit windows x86-64 architectures

### Expected Behaviour

disk2vhd64.exe should be installed / launched on 64 bit systems rather than disk2vhd.exe which is the 32 bit version

### Additional context/output

when running `disk2vhd.exe` or `disk2vhd.ps1` the following error popup is shown:

```
---------------------------
Disk2Vhd
---------------------------
Error launching 64-bit version: The system cannot find the file specified.
---------------------------
OK   
---------------------------
```

### Possible Solution

It seems that the disk2vhd64.exe should be installed instead of or in addition to disk2vhd.exe for 64 bit windows.

### System details

**Windows version:** [e.g. 7, 8, 10]
Window 10 21H2

**OS architecture:** [e.g. 32bit, 64bit]
64bit

**PowerShell version:**  $PSVersionTable.PSVersion

Major  Minor  Patch  PreReleaseLabel BuildLabel
-----  -----  -----  --------------- ----------
7      2      0

Same error happens with native PS 5.1

Windows Terminal

Lutra-Fs:
Closes #7431

### DIFF
--- a/bucket/sysinternals.json
+++ b/bucket/sysinternals.json
@@ -59,7 +59,11 @@
                 "ctrl2cap.exe",
                 "Dbgview.exe",
                 "Desktops.exe",
-                "disk2vhd.exe",
+                [
+                    "disk2vhd64.exe",
+                    "disk2vhd"
+                ],
+                "disk2vhd64.exe",
                 [
                     "diskext64.exe",
                     "diskext"
@@ -330,7 +334,7 @@
                     "/accepteula"
                 ],
                 [
-                    "Disk2vhd.exe",
+                    "Disk2vhd64.exe",
                     "SysInternals/Disk2vhd - Create VHD files from online disk",
                     "/accepteula"
                 ],


bucket/sysinternals.json
```diff
@@ -59,7 +59,11 @@
                 "ctrl2cap.exe",
                 "Dbgview.exe",
                 "Desktops.exe",
-                "disk2vhd.exe",
+                [
+                    "disk2vhd64.exe",
+                    "disk2vhd"
+                ],
+                "disk2vhd64.exe",
```
rashil2000, line 66:
```suggestion

```

This is not required then

rashil2000 (changes requested):

bucket/sysinternals.json
Lutra-Fs: It is an alias to make sure that if the user calls disk2vhd the 64bit ver will be called. I can see the other program in this suite do the same thing.

PR #7448, opened 2021-12-01
autohotkey: add version 2.0-beta.3

### LINKED ISSUE #7446: [Request] AutoHotkey v2
### Information

This is a beta version of the next big upgrade to the already included package AutoHotkey, it's now in beta (syntax is stabilized) and is prominently referenced on their main download page.
I'll submit a PR shorty with a manifest that installs this side-by-side with v1 given that it's highly beneficial to have both on your system due to the fact that you'll encounter/use scripts using either v1 or v2 syntax variants for a while

**Name:** AutoHotkey v2

**Description:** The ultimate automation scripting language for Windows

**Homepage:** https://www.autohotkey.com/

**Download link(s):** https://www.autohotkey.com/download/ahk-v2.zip, or [direct to beta3](https://www.autohotkey.com/download/2.0/AutoHotkey_2.0-beta.3.zip)

**Some indication of popularity/repute:** Popular enough to already be in Scoop :) (~5k GitHub stars)

eugenesvk:
Adds AutoHotkey v2 beta

I've copied&pasted the main AutoHotkey manifest, updated all the links to version 2 (with a nice bonus of not relying on GitHub) and tested via both installing `2.0-beta.2` and `2.0-beta.3` and autoupdating `2.0-beta.2` manifest to `2.0-beta.3` with checkver

To avoid conflict with v1 I've added a rename rule `AutoHotkey` `32`/`64` `.exe`→`autohotkey2.exe` to install a single executable without installing the original `32`/`64`, so you could have 

- `autohotkey` `32`/`64` `.exe` and `autohotkey.exe` from `autohotkey.json` manifest and
- `autohotkey2.exe` from this manifest,

though I guess in some distant future it should just override the first version and manifests be merged into one. Or maybe not since it's worthwile keeping both versions installed as you'll see scripts using both syntaxes

Closes #7446

### DIFF
--- a/bucket/autohotkey2.json
+++ b/bucket/autohotkey2.json
@@ -1,0 +1,36 @@
+{
+    "version": "2.0-beta.3",
+    "description": "The ultimate automation scripting language for Windows.",
+    "homepage": "https://www.autohotkey.com/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://www.autohotkey.com/download/2.0/AutoHotkey_2.0-beta.3.zip",
+    "hash": "8f646a06d9b7e84b5f9195a3f114781e036965c216a6cc3e66cef1f83bc660b0",
+    "architecture": {
+        "64bit": {
+            "bin": [
+                [
+                    "autohotkey64.exe",
+                    "autohotkey2"
+                ]
+            ]
+        },
+        "32bit": {
+            "bin": [
+                [
+                    "autohotkey32.exe",
+                    "autohotkey2"
+                ]
+            ]
+        }
+    },
+    "checkver": {
+        "url": "https://www.autohotkey.com/download/2.0/version.txt",
+        "regex": "([\\d.]+[-a-z]*[\\d.]+)"
+    },
+    "autoupdate": {
+        "url": "https://www.autohotkey.com/download/2.0/AutoHotkey_$version.zip",
+        "hash": {
+            "url": "https://www.autohotkey.com/download/2.0/AutoHotkey_$version.zip.sha256"
+        }
+    }
+}


rashil2000 (approved):

issaclin32:
[autohotkey.com](https://www.autohotkey.com/) says version 1.0 is deprecated.
I think we should just update the *autohotkey* package to v2, and move v1 to the *versions* bucket.

Then, for *bin*, use `autohotkey` instead of `autohotkey2`.

*shortcuts* is missing.

eugenesvk:
> [autohotkey.com](https://www.autohotkey.com/) says version 1.0 is deprecated.

yeah, this is the old-old version, but it isn't used, current AutoHotkey manifest uses version `1.1`, which is very much alive
The new version is `2.0`, which entered beta

> _shortcuts_ is missing.

so it is for `autohotkey.json`, and that's supposed to be like that since you can't really launch a portable AHK and expect it to do anything, it just shows a help file, you must supply a path to script (not sure, but I think only AHK installed to `Program Files` automatically looks for a script in user's documents). So a shortcut that doesn't do anything useful isn't useful

eugenesvk:
> For _bin_, I think use `autohotkey` is better than `autohotkey2`.

but then you wouldn't be able to install both, right?, which is of a **significant value** during the transition between syntaxes.

Now, I'd personally use `autohotkey.exe` for version 2 as you suggest (or even `ahk.exe` for simpler cli typing when you need to check a script or two) and then rename the current manifest with version 1.1 to `autohotkey1.exe`, but given that most of people are still using version 1.1 not sure that'd be the best approach for them.
What do you think?

rashil2000:
I vote for autohotkey2, as many people might not currently know about it because of it's beta status

issaclin32:
> but then you wouldn't be able to install both, right?, which is of a significant value during the transition between syntaxes.

I see your point. Using `autohotkey2` is probably the best option for now.

issaclin32 (approved):

PR #7460, opened 2021-12-03
Add SolveSpace version 3.0

### LINKED ISSUE #7456: [Request] solvespace
### Information

**Name:** solvespace

**Description:** SOLVESPACE is a free (GPLv3) parametric 3d CAD tool.

**Homepage:** <https://solvespace.com/>

**Download link(s):** <https://github.com/solvespace/solvespace/releases>

GitHub 1.9k stars.

It releases single executable binary so it can be packed easily.

tech189:
Closes #7456

No architecture section because app is 32-bit.

### DIFF
--- a/bucket/solvespace.json
+++ b/bucket/solvespace.json
@@ -1,0 +1,19 @@
+{
+    "version": "3.0",
+    "description": "Parametric 2D/3D CAD tool",
+    "homepage": "https://github.com/solvespace/solvespace",
+    "license": "GPL-3.0",
+    "url": "https://github.com/solvespace/solvespace/releases/download/v3.0/SolveSpace.exe",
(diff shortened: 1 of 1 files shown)


rashil2000:
- Anything to persist?
- Does the binary take any commandline arguments? If no, then we can remove the bin field and keep only the shortcuts field.

tech189:
- Nothing to persist because the app saves config in the registry on Windows, there's an issue discussing using JSON here: https://github.com/solvespace/solvespace/issues/1136
- I think it does have a cli (see https://github.com/solvespace/solvespace/blob/0e0b0252e23dd5bd4ae82ababcc54c44aee036d6/src/platform/entrycli.cpp), but I couldn't get it to work, it only seems to expect a filename to open, maybe @demolg knows better

rashil2000 (approved):

issaclin32 (approved):

PR #7462, opened 2021-12-04
transmission: Add CLI, update checkver

se35710:
All of the command line utilities was missing from the manifest.

`checkver`  used Github to check for version change, but that is not reliable, when getting the hash from an external file, as they can update at different times. Change to use to the version time in the file where we get the hash for a more robust update approach.

### DIFF
--- a/bucket/transmission.json
+++ b/bucket/transmission.json
@@ -21,7 +21,12 @@
         [
             "transmission-qt.exe",
             "transmission"
-        ]
+        ],
+        "transmission-create.exe",
+        "transmission-daemon.exe",
+        "transmission-edit.exe",
+        "transmission-remote.exe",
+        "transmission-show.exe"
     ],
     "shortcuts": [
         [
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

b4imetu:
Just curious about why `checkver` needs to be modified.
The original form is more clean and pretty in my opinion.

issaclin32:
> Just curious about why checkver needs to be modified.
I think the new one is slightly better.

What the author meant is, the **Github page** can be updated slower or faster than `https://raw.githubusercontent.com/transmission/transmission.github.io/master/includes/js/constants.js`, which can cause us to fetch the wrong hash.

Since we are already getting hash values from `https://raw.githubusercontent.com/transmission/transmission.github.io/master/includes/js/constants.js`, we can as well check version from the page.

b4imetu:
Thank you for answering my confusion! @issaclin32

issaclin32:
> Is there any particular reason to get the hash value in `autoupdate`?

It can save bandwidth because we don't have to download the whole file and calculate hash.
This is particularly useful for some large apps like `libreoffice` or `cuda`.

b4imetu:
Thank you! That makes sense!

PR #7465, opened 2021-12-04
chore(ci): Use Windows Actions to avoid SSL errors

rashil2000:
The SSL errors on running commands such as `/verify` do not occur on Windows runners. See here - https://github.com/rashil2000/Extras/pulls

PR #7468, opened 2021-12-05
Fix Wireshark autoupdate and update version to 3.6.0

akrabi:
The autoupdate functionality in the Wireshark manifest is broken due to a filename change which now include "32" suffix to the  downloadable file.
This PR fixes the manifest and updates Wireshark to the latest version.


rashil2000:
The verification run failed with SSL errors again - https://github.com/ScoopInstaller/Extras/runs/4421177397?check_suite_focus=true

We need to merge #7465 ASAP

PR #7488, opened 2021-12-08
XnViewMP: Fix xnview.ini creation and persistence in certain cases

CrendKing:
Closes #7199

Back porting relevant parts from https://github.com/ScoopInstaller/Main/pull/2936

### DIFF
--- a/bucket/xnviewmp.json
+++ b/bucket/xnviewmp.json
@@ -15,8 +15,12 @@
     },
     "extract_dir": "XnViewMP",
     "pre_install": [
-        "if (!(Test-Path \"$persist_dir\\xnview.ini\") -or !(Get-Item \"$persist_dir\\xnview.ini\").Length) {",
-        "   New-Item \"$persist_dir\\xnview.ini\" -ItemType File -Force | Out-Null",
+        "# --- If xnview.ini exists as a folder, then remove it ---",
+        "if (Test-Path \"$persist_dir\\xnview.ini\" -PathType Container) {",
+        "    Remove-Item \"$persist_dir\\xnview.ini\" -Force -Recurse",
+        "}",
+        "if (!(Test-Path \"$persist_dir\\xnview.ini\")) {",
+        "   New-Item \"$dir\\xnview.ini\" -ItemType File | Out-Null",
         "}"
     ],
     "uninstaller": {
--- a/bucket/xnviewmp.json
+++ b/bucket/xnviewmp.json
@@ -14,9 +14,13 @@
         }
     },
     "extract_dir": "XnViewMP",
-    "pre_install": "New-Item \"$persist_dir\\xnview.ini\" -ErrorAction SilentlyContinue",
+    "pre_install": [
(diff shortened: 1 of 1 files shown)


rashil2000:
@mossglen90 can you confirm if this fixes your issue?

issaclin32:
@rashil2000 This current manifest will create `$dir\xnview.ini` as a folder in a clean install, which is wrong.
This can fix the problem if that is the case.

Modifying `pre_install` to `New-Item $dir\xnview.ini` would be simpler, but this will not fix problems for users installed the package before, unless they purge persist with `-p`.

issaclin32 (approved):

bucket/xnviewmp.json
```diff
@@ -14,9 +14,13 @@
         }
     },
     "extract_dir": "XnViewMP",
-    "pre_install": "New-Item \"$persist_dir\\xnview.ini\" -ErrorAction SilentlyContinue",
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\xnview.ini\") -or !(Get-Item \"$persist_dir\\xnview.ini\").Length) {",
+        "   New-Item \"$persist_dir\\xnview.ini\" -ItemType File -Force | Out-Null",
```
issaclin32, line 19:
```suggestion
        "   New-Item \"$dir\\xnview.ini\" -ItemType File -Force | Out-Null",
```

CrendKing: Whether we create `xnview.ini` in the perist_dir or dir doesn't matter. `"persist": "xnview.ini"` takes care of linking it to dir anyways. So it's just a matter of style.

issaclin32: @CrendKing I added more details in the comment. Please check if this explains or not.

CrendKing: You are correct. However, the `pre_install` script happens between (1) and (2), which guarantees a file named "xnview.ini" exists in persist_dir with the PR script. Since the file exists, (2) will always happen and the first part of (4) will never happen. Right?

issaclin32: Your script works fine, but I think this is easier to understand (and maintain):
```
    "pre_install": [
        "# --- If xnview.ini exists as a folder, then remove it ---",
        "if (Test-Path \"$persist_dir\\xnview.ini\" -PathType Container) {",
        "    Remove-Item \"$persist_dir\\xnview.ini\" -Force -Recurse",
        "}",
        "if (!(Test-Path \"$persist_dir\\xnview.ini\")) {",
        "   New-Item \"$dir\\xnview.ini\" -ItemType File | Out-Null",
        "}"
    ],
```

If there is something wrong, please tell me.

CrendKing: If you guys want to fix the xnview.ini directory situation, feel free to do that.

The `!(Get-Item \"$persist_dir\\xnview.ini\").Length` check is indeed not needed for this PR. I'll remove.

issaclin32 (approved):

mossglen90:
@rashil2000 Confirmed. This fixes the problem. Both directory and file are created if they do not exist. If they do exist the existing .ini is linked.

rashil2000 (approved):

PR #7491, opened 2021-12-08
Swapped yt-dlg to the maintained fork by @oleksis

brian6932:
The current youtube-dl-gui release is very out of date, with the last release being from 2017. The maintained fork by @oleksis supports yt-dlp and has a lot more additional features, with the last release being yesterday.


rashil2000:
This is a 64bit application. Please use the architecture block.

Remove the empty newline.

PR #7520, opened 2021-12-11
Update msiafterburner.json: Corrects version in the manifest

RavenMacDaddy:
- Updates version described in the manifest to the current which is 4.6.4.16255

### DIFF
--- a/bucket/msiafterburner.json
+++ b/bucket/msiafterburner.json
@@ -40,7 +40,7 @@
     "persist": "Profiles",
     "checkver": {
         "url": "https://www.guru3d.com/files-details/msi-afterburner-beta-download.html",
-        "regex": ">Download\\s+([\\d.]+)\\s+Stable\\s+\\(Final)",
+        "regex": ">Download\\s+([\\d.]+)\\s+Stable\\s+\\(Final)\\",
         "reverse": true
     }
 }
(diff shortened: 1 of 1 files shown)


rashil2000:
Checkver is failing. See the Actions log.

RavenMacDaddy:
Where's the GIF when I need it...

I should've expected it when it didn't catch it on its own.

I was assuming since it already had the checkver and everything that it was just the versioning getting muddled up, let's see...

RavenMacDaddy:
Well, the URL is the same, and what regex is based on seems unchanged, including the versioning; went from 4.6.3 to 4.6.4.

From what I saw, what was missing was one of the parentheses at the end, so that's what I changed.

See if you notice anything else wrong with the regex.

rashil2000:
The closing parenthesis needs to be escaped as well

RavenMacDaddy:
Like that?

rashil2000:
Before the parenthesis, not after

RavenMacDaddy:
How about now? Anything else that makes it fail?

rashil2000:
The issue is that this regex matches only the 4.6.4 part, and not the 16255 part. The 16255 part cannot be found anywhere on the webpage. 

We can change the version to "4.6.4" only and remove the 16255 part.

RavenMacDaddy:
You're right, hang on.

RavenMacDaddy:
There

rashil2000:
We can add the full version later when we find a good source

PR #7534, opened 2021-12-14
vcredist: Update to vcredist2022

issaclin32:
This updates the package and adds **vcredist 2022** support.

**vcredist 2022** provides runtime support for VC 2015, 2017, 2019 and 2022. Therefore we only need to replace the 2019 package with the 2022 one.

https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170
> Visual Studio 2015, 2017, 2019, and 2022
This table lists the latest supported English (en-US) Microsoft Visual C++ Redistributable packages for Visual Studio 2015, 2017, 2019, and 2022. The latest supported version has the most recent implemented C++ features, security, reliability, and performance improvements. It also includes the latest C++ standard language and library standards conformance updates. We recommend you install this version for all applications created using Visual Studio 2015, 2017, 2019, or 2022.

By the way, @ScoopInstaller/maintainers do you think package should be **renamed** to `vcredist-all`? I think it would be less confusing, because users may think package `vcredist` means the latest version of *vcredist*.

### DIFF
--- a/bucket/vcredist.json
+++ b/bucket/vcredist.json
@@ -1,5 +1,5 @@
 {
-    "version": "14.28.29914.0",
+    "version": "14.31.30818.0",
     "description": "Microsoft Visual C++ Redistributables bundle for all major versions",
     "homepage": "https://www.visualstudio.com/downloads/",
     "license": {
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):

niheaven:
Please add a separate vcredist2022 manifest.

For the name, current one is clear enough IMO.

issaclin32:
> Please add a separate vcredist2022 manifest.

[vcredist2022](https://github.com/ScoopInstaller/Extras/blob/master/bucket/vcredist2022.json) is already in the bucket.

Issue #7399, opened 2021-11-24
haskellx@7.8.3: hash check failed


TonyTheAce:
```
App:         extras/haskellx
URL:         https://www.haskell.org/cabal/release/cabal-install-1.20.0.2/cabal-i386-unknown-mingw32.tar.gz
First bytes: 3C 68 74 6D 6C 3E 0D 0A
Expected:    7f8add65c98214ee76757d9dfd35f0e18f7799d04343a50593c9f6ff86ffe5a1
Actual:      3c63ce537d469e44cf4a0c7ed54c9e623f8edfdb27803a19f352107d77c8ca79
```

HUMORCE:
the manifest need to remake and out of date.

HUMORCE:
`extras/haskellx` is outdated, for install haskell toolchains:

```
scoop install main/haskell
scoop install main/haskell-cabal
### haskell-stack: 
scoop install main/stack
```

PR #7544, opened 2021-12-15
haskellx: Deprecate manifest

HUMORCE:
Close https://github.com/ScoopInstaller/Extras/issues/7399


rashil2000:
You need to remove it from the bucket folder too.

PR #7552, opened 2021-12-18
Create autohdr.json: For Codectory's AutoHDR project

RavenMacDaddy:
This is a manifest for the project AutoHDR by Codectory [here](https://github.com/Codectory/AutoHDR).

I wanted to create a pre-install section to persist UserSettings.json as a hardlink, but I didn't manage to do it by taking inspiration from [flycast.json](https://github.com/Calinou/scoop-games/blob/master/bucket/flycast.json) in the Scoop Games bucket and do something similar to:  

"pre_install": "if (!(Test-Path \"$persist_dir\\UserSettings.json\")) New-Item -ItemType File \"$dir\\UserSettings.json\" | Out-Null",

It's just throwing errors - so someone who knows better is most welcome to figure that one out with me.

Apart from that it should work just fine.

### DIFF
--- a/bucket/autohdr.json
+++ b/bucket/autohdr.json
@@ -13,6 +13,7 @@
             "hash": "745ffd001d189b6d701647d76b2fef9318561eb83b4153035bf90cea2b36bb4a"
         }
     },
+    "pre_install": "if (!(Test-Path \"$persist_dir\\UserSettings.json\")) { New-Item -ItemType File \"$dir\\UserSettings.json\" | Out-Null }", 
     "bin": "AutoHDR.exe",
     "shortcuts": [
         [
@@ -30,5 +31,6 @@
                 "url": "https://github.com/Codectory/AutoHDR/releases/download/$version/Release_AutoHDR_$version_x86.zip"
             }
         }
-    }
+    },
+    "persist": "UserSettings.json"
 }
--- a/bucket/autohdr.json
+++ b/bucket/autohdr.json
@@ -14,7 +14,6 @@
         }
     },
     "pre_install": "if (!(Test-Path \"$persist_dir\\UserSettings.json\")) { New-Item -ItemType File \"$dir\\UserSettings.json\" | Out-Null }", 
-    "bin": "AutoHDR.exe",
     "shortcuts": [
         [
             "AutoHDR.exe",
--- a/bucket/autohdr.json
+++ b/bucket/autohdr.json
@@ -1,8 +1,8 @@
 {
-    "homepage": "https://github.com/Codectory/AutoHDR",
-    "description": "Create profiles for your displays and apps to automatically switch the resolution, refresh rate and HDR state according to the app you're using",
     "version": "1.7.19",
-    "license": "https://github.com/Codectory/AutoHDR/blob/main/LICENSE",
+    "description": "Create profiles for your displays and apps to automatically switch the resolution, refresh rate and HDR state according to the app you're using",
+    "homepage": "https://github.com/Codectory/AutoHDR",
+    "license": "GPL-3.0-only",
     "architecture": {
         "64bit": {
             "url": "https://github.com/Codectory/AutoHDR/releases/download/1.7.19/Release_AutoHDR_1.7.19_x64.zip",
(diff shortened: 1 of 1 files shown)


rashil2000:
What errors does it show?

RavenMacDaddy:
+ if (!(Test-Path "$persist_dir\UserSettings.json")) New-Item -ItemType …
+                                                  ~
Missing statement block after if ( condition ).

rashil2000:
The statement after the if condition needs to be under curly braces, that's what the error is saying. See docs: https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-if?view=powershell-7.2

RavenMacDaddy:
I've tried to create the corresponding code with different combinations, but I can't figure it out.

Do you know what the code must be to be functional?

rashil2000:
```
"pre_install": "if (!(Test-Path \"$persist_dir\\UserSettings.json\")) { New-Item -ItemType File \"$dir\\UserSettings.json\" | Out-Null }", 
```

Try installing with this.

RavenMacDaddy:
Such a simple mistake on my part, thanks @rashil2000.

rashil2000:
No worries :)

Can you confirm if persistence works correctly? With both an update and uninstall.

RavenMacDaddy:
For me it's working even though I stress-tested it by reinstalling a couple of times, the persisted json stays intact and the app reads the settings from there.

Of course hardlinking always creates a more complicated scenario to verify things are working (even deleting the file will link it to the recycle bin and so on), but I could successfully:

1. Back up my json file with all of my settings.
2. Delete all the app/persist data in Scoop.
3. Install the app again.
4. Copy my backed up json file to the persist folder.
5. ... And run the app which reads my corresponding settings just fine.

RavenMacDaddy:
Just upped the version to 1.7.20 in my own repo to make Scoop install "a new version" and sure enough the app is linked to current and the json stays hardlinked to the new folder.

rashil2000:
Okay, thanks for the detailed checkup! Just a couple of minor comments.

RavenMacDaddy:
No worries, I prefer things to be thoroughly worked through myself, see if verify checks out. Hopefully I didn't miss curly braces or data for autoupdate or architectures.

bucket/autohdr.json
```diff
@@ -0,0 +1,36 @@
+{
+    "homepage": "https://github.com/Codectory/AutoHDR",
+    "description": "Create profiles for your displays and apps to automatically switch the resolution, refresh rate and HDR state according to the app you're using",
+    "version": "1.7.19",
+    "license": "https://github.com/Codectory/AutoHDR/blob/main/LICENSE",
```
rashil2000, line 5:
```suggestion
    "version": "1.7.19",
    "description": "Create profiles for your displays and apps to automatically switch the resolution, refresh rate and HDR state according to the app you're using",
    "homepage": "https://github.com/Codectory/AutoHDR",
    "license": "GPL-3.0-only",
```

bucket/autohdr.json
```diff
@@ -0,0 +1,36 @@
...
+            "url": "https://github.com/Codectory/AutoHDR/releases/download/1.7.19/Release_AutoHDR_1.7.19_x64.zip",
+            "hash": "13211a6cb0b6cc42754728cb991686582c5e6e5810e1af983bfd500c74681e96"
+        },
+        "32bit": {
+            "url": "https://github.com/Codectory/AutoHDR/releases/download/1.7.19/Release_AutoHDR_1.7.19_x86.zip",
+            "hash": "745ffd001d189b6d701647d76b2fef9318561eb83b4153035bf90cea2b36bb4a"
+        }
+    },
+    "pre_install": "if (!(Test-Path \"$persist_dir\\UserSettings.json\")) { New-Item -ItemType File \"$dir\\UserSettings.json\" | Out-Null }", 
+    "bin": "AutoHDR.exe",
```
rashil2000, line 17: Does the binary take any commandline arguments?

rashil2000 (changes requested):

bucket/autohdr.json
RavenMacDaddy: Not as far as I'm aware; nothing in the readme or any wiki available.

Think it's a simple app that reads config data .config and .json files in the same dir.

bucket/autohdr.json
RavenMacDaddy: Looks good to me.

bucket/autohdr.json
rashil2000: In that case we can remove the bin entry. The shortcut is sufficient.

RavenMacDaddy: Adjusted.

PR #7554, opened 2021-12-19
Added file type icon association support to VSCode and its variations

### LINKED ISSUE #2882: VS Code registers one file icon when installed via Scoop
# Environment
```none
Windows Build: 10.0.18362.0
Current Scoop Version: 3e55a709
```

# Steps to Reproduce
Assuming scoop is installed, vs code is not, and the extras bucket has been added:
1. Run `scoop install vscode`
1. Run `$dir\\vscode-install-context.reg` to add VS Code as a context menu option
1. In Explorer, create a folder with various file types supported by VS Code
1. For each file:
- Select "Open With..."
- Find VS Code in the list of applications OR if it's not there, select "Look for another app on this PC" and browse to the VS Code executable in scoop
- Check the "Always use this app to open '.file-ext' files" box

# Current Behavior
Each file type uses the same VS code icon in Windows explorer
image

# Expected Behavior
Each file type uses it's own icon, as described in the [October 2018 release notes](https://code.visualstudio.com/updates/v1_29#_updated-file-icons-per-file-type). This is the normal behavior when VS Code is installed normally rather than with Scoop.

MathewBlock:
Closes #2882

- Using the official VSCode [installer source](https://github.com/microsoft/vscode/blob/main/build/win32/code.iss) as a reference, this change recreates the necessary registry changes to implement file icon associations.
- File association change is presented as an optional registry file that the user is notified about after install (similar to existing context menu change).
- This change also reads directly from the scripts folder and writes the modified versions to the install directory, I've seen other packages do this and it appears to be the recommended way rather than pulling directly from raw.githubusercontent.com.


rashil2000:
Wow, thanks for doing this!

phanirithvij:
@rashil2000 as `scoop update vscode` doesn't work until the next vscode version update the only way to get this, for now, is to uninstall and install vscode again? 😢

PR #7573, opened 2021-12-21
(chore): Use portable configurations by default

### LINKED ISSUE #3479: gpg4win global shortcuts
When installing gpg4win via scoop, a system wide Kleopatra shortcut is created (`C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Kleopatra`), instead of the Scoop Apps folder.
Same for the desktop shortcut which goes to `C:\Users\Public\Desktop`.

This should be fixable by passing a file with the [/C=\<INIFILE\> option](https://www.gpg4win.de/doc/en/gpg4win-compendium_35.html), something like:

```ini
[gpg4win]
  inst_start_menu = false
  inst_desktop = false
```

and adding the shortcuts via scoops config.

### LINKED ISSUE #4377: [Bug] Firefox/thunderbird update switches profile
- I have installed Firefox 77.0.1.
- I updated it to 78.0.1 via the following command -->  $ scoop update *
I used this command because I had other apps to update and don't wanted update each one separately.
- After the process completed I ran Firefox from start menu and I saw: my history, open tabs and saved passwords has been lost.
- I don't know what happen but after a angry evening of this situation I remembered of profile manager builtin Firefox.
- I saw 5 new profiles and the oldest one is my profile.
- Do you know why updated Firefox created new profile?

### LINKED ISSUE #5479: ungoogled-chromium crashes on everypage

rashil2000:
This is done in order to fit in with Scoop's philosophy of portable applications. 

Closes #3479
Closes #4377 
Closes #5479

A side effect of this PR is reduced load on the Excavator.


rashil2000:
streamlink-portable.json is remaining, but the manifest is behind streamlink.json in version, so I'll look at it later.

rashil2000:
@issaclin32 any pointers on why ungoogled-chromium is giving errors?

rasa (approved):

rashil2000:
@issaclin32 do you think it'd be a good idea to add back the `vcredist20*` manifests as `vcredist20*-installer` (similar to [autohotkey-installer](https://github.com/ScoopInstaller/Extras/blob/master/bucket/autohotkey-installer.json))?

chawyehsu (changes requested):
I'm against this giant change. Scoop is a command-line installer, not the PortableApps.com. Although Scoop prefers portable apps, it's not mandatory and therefore there is no critial reason to break so many things in one pull request. 

This PR is a BREAKING CHANGE not what it called “chore”.

Please make one per PR and get enough reviews and user feedback if you insist on doing that.

rashil2000:
All the manifests (wherever required) have migration scripts in them, so it's not likely to break any configurations.

Migration scripts is the standard procedure for making a manifest portable, which has happened many times in the past here in this repo.

chawyehsu:
Most of those scripts from your changes do nothing about **data migration**, they are just for *making a manifest portable* as you said. Becasue of this change, users need to migrate their data from AppData or elsewhere *manually*, otherwise they will get a fresh installation since data are not migrated automatically.

Firefox users will “lose” their profile when they update Firefox to the portable one because there is no automatically profile migration.

Wouldn't you call this a breaking change? This is toally a breaking change imo.

rashil2000:
Firefox loses profile on each update anyways, see the linked issue in the top comment.

Same for Google Chrome - the non portable manifest crashes the browser (linked issue).

rashil2000:
That said, if there's room for improvement, I'll definitely make it.

chawyehsu:
**auto data migration** from non-portable to portable is required from my point of view. Otherwise, it's a breaking change.

issaclin32:
> @issaclin32 do you think it'd be a good idea to add back the `vcredist20*` manifests as `vcredist20*-installer` (similar to [autohotkey-installer](https://github.com/ScoopInstaller/Extras/blob/master/bucket/autohotkey-installer.json))?

I think `vcredist20*-np` (non-portable) would be better.

issaclin32 (approved):

niheaven:
`vcredist` is used only in installation. I don't know where the portable versions are used.

rashil2000:
@chawyehsu done, please review

chawyehsu:
vcredist manifests shouldn't be changed. They have to be installed as system-wide libraries for most apps that require vcredist to work.

> I don't know where the portable versions are used.

Perhaps @linsui knows, those extracted versions of vcredist were introduced in https://github.com/ScoopInstaller/Scoop/issues/3688. I don't think these portable vcredist could be used by other apps when they are not placed in app's directory.

rashil2000:
Okay, I'll revert the vcredist changes. The rest looks fine?

chawyehsu:
I'm reviewing the rest. It requires times for such a big PR, that's why I suggest separating them...

niheaven:
VCRedist is not used by scoop apps only, they're used system wide. DON'T make them portable, that's insignificance.

bucket/ungoogled-chromium.json
```diff
@@ -1,35 +1,47 @@
...
+    ],
     "shortcuts": [
         [
             "chrome.exe",
-            "Ungoogled Chromium"
+            "Chromium",
+            "--user-data-dir=\"$dir\\User Data\""
         ]
     ],
+    "persist": "User Data",
```
chawyehsu, line 28: Same as googlechrome.

bucket/googlechrome.json
```diff
@@ -19,13 +19,21 @@
...
+    ],
     "shortcuts": [
         [
             "chrome.exe",
-            "Google Chrome"
+            "Google Chrome",
+            "--user-data-dir=\"$dir\\User Data\""
         ]
     ],
+    "persist": "User Data",
```
chawyehsu, line 36: No `User Data` migration for googlechrome?

bucket/foobar2000.json
```diff
@@ -24,9 +18,17 @@
...
         ]
     },
+    "post_install": [
+        "if(Test-Path \"$persist_dir\\playlists\") {",
+        "    Rename-Item -Path \"$persist_dir\\playlists\" -NewName playlists-v1.4 -Force",
+        "    Get-ChildItem -Exclude \"profile\" -Path \"$persist_dir\" |",
+        "        Move-Item -Destination \"$persist_dir\\profile\" -Force",
+        "}"
+    ],
+    "persist": "profile",
```
chawyehsu, line 31: require data migration from `~\AppData\Roaming\foobar2000`

bucket/vscode.json
```diff
@@ -40,8 +40,15 @@
         "    }",
         "    $content | Set-Content -Path \"$dir\\$_.reg\"",
         "  }",
+        "}",
+        "if (!(Test-Path \"$dir\\data\\extensions\") -and (Test-Path \"$env:USERPROFILE\\.vscode\\extensions\")) {",
```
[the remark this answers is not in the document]
rashil2000: Okay, will add it

bucket/vscodium.json
```diff
@@ -38,8 +38,15 @@
         "    }",
         "    $content | Set-Content -Path \"$dir\\$_.reg\"",
         "  }",
+        "}",
+        "if (!(Test-Path \"$dir\\data\\extensions\") -and (Test-Path \"$env:USERPROFILE\\.vscode-oss\\extensions\")) {",
```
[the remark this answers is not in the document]
rashil2000: Will do

bucket/googlechrome.json
rashil2000: The non-portable manifest for Google Chrome stores data in the installation folder itself (as opposed to other non-portable manifests like Palemoon, which store in %AppData%), so no migration is needed here.

bucket/ungoogled-chromium.json
rashil2000: Same as the Google Chrome reply

bucket/foobar2000.json
rashil2000: Will look at it

bucket/googlechrome.json
chawyehsu: I got a fresh installation after running the portable update... I'm sure my User Data for the non-portable Google Chrome was in `~\AppData\Local\Google\Chrome\User Data`

rashil2000: The only difference between googlechrome.json and googlechrome-portable.json was the binary arguments and persist, which indicates the user data was being stored in the scoop\apps folder itself.

The %LocalAppData% folder is utilised by system-wide Google Chrome installation (which installs in C:\Program Files). I think this is due to the difference in the way the browser is installed, i.e. through running the installer, as opposed to Scoop which extracts the installer.

We should only migrate data from Scoop installations, and not system-wide installations (because they weren't installed through Scoop anyway).

chawyehsu: Not ture. I only have Chrome installed via scoop before.

For testing, I deleted all old chrome User Data, did a fresh installation by the current googlechrome.json, opened it, pinned a tab and modified some settings. I checked User Data was generated and saved in `~\AppData\Local\Google\Chrome\User Data`. Then I updated the manifest to this PR's and force updated it to the portable one. I got a fresh installation, no pinned tab, settings gone.

rashil2000: Okay, I'll add a migration for this too then.

What about Firefox and Thunderbird?

chawyehsu: I'm not sure how Firefox and Thunderbird profile work, but if they have issue #4377 of persisting profile, then I would keep them with no migration.

bucket/vscode.json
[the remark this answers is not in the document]
rashil2000: Added

bucket/googlechrome.json
rashil2000: Added

bucket/foobar2000.json
rashil2000: Added

rashil2000:
Changes in this PR:
- `arduino`: Migration script added.
- `foobar`: Migration script added.
- `googlechrome`, `ungoogled-chromium`: Migration script added.
- `palemoon`: Migration script added.
- `qtox`: Migration script added.
- `rainmeter`: Migration script added.
- `streamlink`: Migration script added.
- `uget`: Migration script added.
- `vscode`, `vscodium`: Migration script added.

.
- `firefox`, `firefox-esr`, `thunderbird`: Migration not possible, see linked issue.
- `gpg4win-portable`: Portable manifest non-functional. Removed.
- `vcredsit20XX-portable`: Portable manifests didn't do anything after extraction. Removed. (Note: the non-portable, i.e. installers are not modified).

chawyehsu (approved):

niheaven:
Could you please resolve the conflicts? @rashil2000

rashil2000:
Checks for ryzen-controller.json are failing, but I did not modify it in my PR

niheaven:
That's because of merging master. It may need to be checked.

rashil2000:
We can handle it in separate PR.

Fijxu:
This portable version of streamlink broke the integration with [Chatterino2](https://github.com/chatterino/chatterino2) which needs the .exe version of streamlink to work

image

rashil2000:
Does it always look for .exe file? The executable in portable version ends with .bat and it calls python within.

In the installer version too, the .exe file is just a python shim:

image

Fijxu:
> Does it always look for .exe file? The executable in portable version ends with .bat and it calls python within.
> 
> In the installer version too, the .exe file is just a python shim:
> 
> image

Yes, it needs a .exe version, i am using the d87473dd2f173360f2b8b0b77dce728e7bbb7793 commit of streamlink in scoop, but when i update streamlink to the current version of the bucket, the integration no longer works, it need to be a .exe, i already tried using the .bat version but doesn't work

rashil2000:
> it need to be a .exe, i already tried using the .bat version but doesn't work

I think you should report this bug to https://github.com/chatterino/chatterino2

Many people who don't have admin rights/write access to Documents folder, use the portable version of Streamlink. I think looking for .bat file along with .exe file should be a simple fix for Chatterino.

rdbisme:
This has been reverted 1 hour ago for Firefox, and broke my profile again. Can we please avoid switching setup every now and then and maybe put in place some kind of testing before pushing disruptive modifications to `master`?

rashil2000:
The latest change was non-destructive.

How will you test it anyway?

rdbisme:
Well, it was for me. It deleted the persisted profile and distribution folders, spinning up a brand new profile to be configured again. Luckily I had a backup somewhere else. 

Well, at very least I would have tested it locally against an already installed Firefox with a populated profile directory. I don't know all the bells and whistles of Scoop, but I guess it could be tested with some CI: checking for example that the Shortcut produced by Scoop correctly points to the persisted profile, that the persisted profile is correctly persisted and so on.

rashil2000:
> It deleted the persisted profile and distribution folders

These two folders were persisted, so they still exist in ~/scoop/persist/firefox folder. What got deleted (or rather, did not get linked in the new version) was the hardlinks to them (Scoop uses hardlinks to the persist directory), not the actual data. So all your settings are still present.

rdbisme:
Yes, indeed. But this needs time to be figured out. I need to go check where scoop stores persisted profiles. I need to restore the profile in the correct Firefox folder (that now is different). This takes time. What I'm arguing here is that it would be great to not having to do this every Firefox release :) 

What I expect here is that doing `scoop update firefox` will update Firefox keeping my profile. This is not the case right now.

rashil2000:
Ungoogled chromium's manifest (and all chromium-related browsers) has undergone a _lot_ of changes recently (independent of this PR), with version updates, source changes, portability etc. I would recommend going through the Git history of the manifest, and using the desired file from there instead of auto-downgrading.

tskimmett:
> Ungoogled chromium's manifest (and all chromium-related browsers) has undergone a _lot_ of changes recently (independent of this PR), with version updates, source changes, portability etc. I would recommend going through the Git history of the manifest, and using the desired file from there instead of auto-downgrading.

If you're referring to `ungoogled-chromium.json` in this repository, the last change before yours was April 2021 (our installation script was working then).

image


As for your recommendation, I don't exactly understand what you're suggesting when you say

>using the desired file from there instead of auto-downgrading

You did not answer the question of how to install a specific version of the `ungoogled-chromium` package (as far as I can tell this is no longer possible).

```
scoop install ungoogled-chromium@<version>
```

What can I put for `<version>`?

rashil2000:
Generating manifest from `scoop install ungoogled-chromium@<version>` command is never guaranteed to work, because the URL is autogenerated and the asset might not exist at the source. It's a very flaky method.

You can use this URL to install your required version https://github.com/ScoopInstaller/Extras/blob/8b7cc1be65402b20bffd17921c697927a9b9fe4b/bucket/ungoogled-chromium.json - note that I have not verified it, because as I said, it is possible that the download URL is stale and asset might not exist at the source.

tskimmett:
Using the old manifest json seems to work. I still think there is something off with the `ungoogled-chromium` manifest though.

rashil2000:
Please create a new issue describing any problems you are facing

brian6932:
@rashil2000 I have the same problem as @Fijxu, would it be possible to have an optional non-portable install? This makes using streamlink a lot more of a chore. I get that portable may be more desirable for scoop, but this is detrimental to it's users.

rashil2000:
Have the developers of chatterino responded?

brian6932:
There is no open issue, I could try making one, seems like a lot of work just to support something portable imo

zneix:
> Have the developers of chatterino responded? 

Hello, Chatterino has always had hardcoded name of the file that is required to be run as the streamlink process and on windows it's "streamlink.exe".


I'm not an expert when it comes to windows and don't want to sound ignorant, but is there anything stopping you from just using .exe files for portable versions of software? Besides, doesn't https://github.com/ScoopInstaller/Extras/pull/7573#issuecomment-1010699599 show a file that's named literally "streamlink.exe" or does this have some extra hidden extension or something else? 

In the end, I suppose that a patch to Chatterino giving user more freedom on what could be executed as streamlink child process might be a solution, we just didn't ever consider streamlink executable to be named anything else than "streamlink.exe".

rashil2000:
> I'm not an expert when it comes to windows and don't want to sound ignorant, but is there anything stopping you from just using .exe files for portable versions of software? 

That's a question for the streamlink devs ;) Technically, no. They can pack the python script in an .exe just fine.

> Besides, doesn't [#7573 (comment)](https://github.com/ScoopInstaller/Extras/pull/7573#issuecomment-1010699599) show a file that's named literally "streamlink.exe" or does this have some extra hidden extension or something else?

That screenshot is from the non-portable build.

> ...we just didn't ever consider streamlink executable to be named anything else than "streamlink.exe".

That does sound like oversight. On unix too, there are 2(?) types of 'executables' - ELF binaries and shell scripts (which you `chmod`). Same for Windows. `.bat` files (and any file with some extension that uses an interpreter, actually) are executables too.

In the end, I think I can add the non-portable version back (with a `-normal` suffix to differentiate it) if it is causing too many problems.

PR #7615, opened 2021-12-28
Added context menu items for Beyond Compare

### LINKED ISSUE #1645: How to bundle regkeys with beyondcompare
Beyond Compare is very useful due to the Explorer context options it adds (Select as "Left" for file compare), etc.

The program has an option to by default integrate itself into the Explorer shell menu, but it seems broken with the 64bit version of Beyond Compare.

The reason is that in the key `Computer\HKEY_CURRENT_USER\Software\Classes\CLSID\{57FA2D12-D22D-490A-805A-5CB48E84F12A}\InProcServer32` Beyond Compare by default adds the path to the `BCShellEx.dll` file, whereas `BCShellEx64.dll` is required instead.

Is it possible to bundle a regkey that users can double click to install BC4 to the explorer (similar to the functionality of TortoiseSVN)? Or to download two source files, since ScooterSoft distributes a fix for the issue here: https://www.scootersoftware.com/BC4ShellExFix.zip but it needs to have the path modified to support the scoop path.

MathewBlock:
Closes #1645 

* Added new reg files to associate Beyond Compare with the Windows explorer context menu based on content from the Beyond Compare support [website](https://www.scootersoftware.com/support.php?zz=kb_shellex).
* Modified installer script to notify the user about the new post install step and customize the reg file based on Scoop settings.

### DIFF
--- a/bucket/beyondcompare.json
+++ b/bucket/beyondcompare.json
@@ -6,6 +6,7 @@
         "identifier": "Shareware",
         "url": "https://www.scootersoftware.com/index.php?zz=kb_licensev4"
     },
+    "notes": "Add Beyond Compare as a context menu option by running: '$dir\\install-context.reg'",
     "architecture": {
         "64bit": {
             "url": "https://www.scootersoftware.com/BCompare-4.4.1.26165_x64.msi",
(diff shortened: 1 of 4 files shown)


rashil2000:
Thanks!

PR #7617, opened 2021-12-28
ci(test): Use `scoop prefix scoop`

niheaven:


rashil2000:
Where is the file Scoop-Bucket.Tests.ps1 used? Is it still required after https://github.com/ScoopInstaller/Extras/pull/1533? The tests are now directly imported from the SCOOP_HOME folder.

niheaven:
For local bucket test.

rashil2000:
How to run them locally?

rashil2000:
Also how is it different from bin/test.ps1?

niheaven:
Just cd in bucket dir and run `.\Scoop-Bucket.Tests.ps1`

rashil2000:
Ok, got it!

So would it be a good idea to move it in the bin folder (along with all other utilities)? With the name bin/bucket-test.ps1. Currently it's in root folder while all others are in bin folder

niheaven:
'cos I don't know if there are other build tools that use this file...

Make a test in `versions` and see if there are some bugs.

niheaven:
`bin\test.ps1` call `Scoop-Bucket.Tests.ps1`, and if we put `Scoop-Bucket.Tests.ps1` into `bin`, it doesn't work.

This needs test scripts refactoring, so leave the script outside and wait for later code review (about Scoop Core's test scripts).

rashil2000:
https://github.com/ScoopInstaller/Versions/commit/e4c6f8d5d196cfd39096fb762bf101a5ab0ef73a

rashil2000:
Oh, I made the commit, let me revert it

rashil2000 (approved):

niheaven:
There are some scripts that come before 2019-05-12 (at which time we separate main bucket from scoop core). So the entire test scripts should be reviewed again and see if they could be optimized.

PR #7626, opened 2021-12-30
Added optional script to update Discord settings to disable automatic…

### LINKED ISSUE #6596: discord@0.0.310: autoupdate leads to scoop manifest removal
This is possibly a duplicate, but I have not found a similar post.

Installing `discord` and running it triggers the auto-updater to run. After that, `scoop` shows the following:
```
PS C:\Users\User> scoop list
...
  discord download *failed*
...
```
```
PS C:\Users\User> scoop status
Scoop is up to date.
These app manifests have been removed:
    discord
These apps failed to install:
    discord
```

MathewBlock:
Closes #6596 

* Added an optional script that can stop Discord from performing automatic updates.
* Referenced a few online packages that do something similar: [NixOS](https://nixos.wiki/wiki/Discord#Discord_wants_latest_version) and [Flathub](https://github.com/flathub/com.discordapp.Discord/blob/master/disable-breaking-updates.py)
* I copy the disable script into the app directory so I can prompt the user to run it given `$dir` since the `"notes"` section of the manifest doesn't permit `$bucketsdir` which would allow me to point directly at the source of the script.

### DIFF
--- a/bucket/discord.json
+++ b/bucket/discord.json
@@ -6,6 +6,7 @@
         "identifier": "Freeware",
         "url": "https://discordapp.com/terms"
     },
+    "notes": "To avoid manifest removal due to internal updates, disable automatic updates by running: '$dir\\disable-auto-update.ps1'",
     "url": "http://dl.discordapp.net/apps/win/Discord-0.0.311-full.nupkg",
     "hash": "sha1:3928a4af8e249184f2d4212bd8c958adc8b2ff4f",
     "extract_dir": "lib\\net45",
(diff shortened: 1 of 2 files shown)


MathewBlock:
I should also mention ScoopInstaller/Scoop#4605 where there was some preliminary discussion about the best way to tackle this issue.

niheaven:
Download script just as [everything](https://github.com/ScoopInstaller/Extras/blob/master/bucket/everything.json) do.

MathewBlock:
> Download script just as [everything](https://github.com/ScoopInstaller/Extras/blob/master/bucket/everything.json) do.

It was my understanding that referencing `$bucketsdir` is now the preferred method over downloading another copy of the script that is already downloaded. Like [vscode](https://github.com/ScoopInstaller/Extras/blob/master/bucket/vscode.json) or [lockhunter](https://github.com/ScoopInstaller/Extras/blob/master/bucket/lockhunter.json) to name a couple.

rashil2000:
I also support the `$bucketsdir` method. It's cleaner because we don't have to pollute the manifest with the script's URLs and hashes.

rashil2000 (approved):

PR #5015, opened 2020-10-30
dxwnd: Add version 2.05.58

snowman:

PR #7629, opened 2021-12-30
dxwnd: Add version 2.05.78

issaclin32:
supersedes #5015. (see old PR for more details)

**dxwnd** is Window hooker to run fullscreen programs in window and much more.

NOTES:

* Based on the content of `$dir\tweaks.ini`, I don't think this needs to be persisted.
```
[FixD3DWindowFrame]
help="Tries to prevent D3D from rendering on the whole window surface including the window border. It activates a small trick that cause the program to render to a child modal surface within the main window borders."
[NoD3DWindowMove]
help="Do not try to update the window position and size on D3D rendering. In some cases, the window may be movable but the rendering area does not move with it. In this case, avoid moving the window at all. It is an experimental feature and doesn't always work."
```

* `$dir\vcda\readme.txt` says:
```
copy to the target program folder these file
	winmm.dll 
	winmm.ini 
copy there also these files from DxWnd folder
	dxwplay.dll
	libogg.dll
	libvorbis.dll
	libvorbisfile.dll
```

However, `dxwplay.dll` etc. are already in `$dir`. (I think the help text is not keep to date)
Also, it says copy `winmm.dll` and `winmm.ini` to target program folder, but pratically we can just add them to PATH.

### DIFF
--- a/bucket/dxwnd.json
+++ b/bucket/dxwnd.json
@@ -14,7 +14,8 @@
     ],
     "persist": [
         "dxwnd.ini",
-        "vcda\\winmm.ini"
+        "vcda\\winmm.ini",
+        "alt.dll"
     ],
     "env_add_path": "vcda",
     "checkver": {
--- a/bucket/dxwnd.json
+++ b/bucket/dxwnd.json
@@ -1,0 +1,29 @@
+{
+    "version": "2.05.78",
+    "description": "Window hooker to run fullscreen programs in window and much more.",
+    "homepage": "https://sourceforge.net/projects/dxwnd/",
+    "license": "GPL-3.0-or-later",
+    "url": "https://downloads.sourceforge.net/project/dxwnd/Latest%20build/v2_05_78_fx2_build.rar",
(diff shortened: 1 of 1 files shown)


rashil2000:
> Also, it says copy winmm.dll and winmm.ini to target program folder, but pratically we can just add them to PATH.

We should try to avoid PATH pollution. I don't think any other app uses the 2 files? If the Readme says to copy them, we can just do that in `post_install`.

issaclin32:
> > Also, it says copy winmm.dll and winmm.ini to target program folder, but pratically we can just add them to PATH.
> 
> We should try to avoid PATH pollution. I don't think any other app uses the 2 files? If the Readme says to copy them, we can just do that in `post_install`.

What I mean is, if we are trying to tweak some programs (for example, some old games) using **dxwnd**, we will have to copy `winmm.dll` and `winmm.ini` into the game folder. `env_add_path` can avoid the chore.

I agree that we should avoid `env_add_path` as much as possible, but this is the only solution I can think of.

rashil2000:
Oh, if some other programs require the DLL, then it's good 👍

PR #7642, opened 2022-01-01
(fix): move explorerplusplus and torrenttools manifest to bucket directory

Ay-355:
These manifests weren't added to `bucket/` for some reason, so they aren't found with a `scoop install` or `scoop info`. This PR just moves it to that directory.


rashil2000:
Ah, what a stupid mistake :(

Sorry, and thanks!

PR #7649, opened 2022-01-02
playnite: Update 9.12 -> 9.14

dragonmaus:
The download URL for 9.12 is no longer works; is there a reason we're not downloading the GitHub releases?

### DIFF
--- a/bucket/playnite.json
+++ b/bucket/playnite.json
@@ -1,10 +1,10 @@
 {
-    "version": "9.12",
+    "version": "9.14",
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

issaclin32:
I have switched it back to Github release. (https://github.com/ScoopInstaller/Extras/commit/7eed9325f8467a525616c96e6d1d4d303e80ac1d)

Looks like they removed Github release at some point (see #3248). That is why we are not using the Github one.

Issue #7292, opened 2021-11-10
chromium versions need to be updated


pratikpc:
1. Ungoogled Chromium currently works best thanks to @lunarwhite 
2. Chromium points to Ungoogled Chromium and has for quite a long time.
Ideally it should use https://github.com/Hibbiki/chromium-win64/releases/download/v95.0.4638.69-r920003/chrome.sync.7z based on what I can ascertain from https://chromium.woolyss.com/
3. chromium-dev is at 91.0 which is old
4. chromium-nosync is at version 67. I had to create a PR to deprecate.

Could you please help us update all packages?

PR #7664, opened 2022-01-04
chromium: Update to 96.0.4664.93-r929512 and switch to new URL

LazyGeniusMan:
Closes #XXXX 

Relates to #7292 

## Notes
- List of files can be found here: https://github.com/LazyGeniusMan/Sc00p-Test/runs/4697216768?check_suite_focus=true#step:3:3
- Should I also add bin & scripts that's used in `ungoogled-chromium` ?

### DIFF
--- a/bucket/chromium.json
+++ b/bucket/chromium.json
@@ -16,18 +16,26 @@
     },
     "extract_dir": "Chrome-bin",
     "bin": [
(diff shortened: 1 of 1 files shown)


rashil2000:
> Should I also add bin & scripts that's used in ungoogled-chromium ?

Yes

PR #7681, opened 2022-01-06
Include bin statements in p4v

Jestar342:

### DIFF
--- a/bucket/p4v.json
+++ b/bucket/p4v.json
@@ -12,13 +12,13 @@
             "hash": "6f423b7e58a92faa1b2a463c33209cb1c86a638dad2f1cddabd851d436e2cd96"
         }
     },
+    "extract_dir": "Perforce",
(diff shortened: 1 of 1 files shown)


rashil2000:
Reason?

Jestar342:
Without those bin statements, they aren't added to $Env:PATH, and can't be found from cli - so stuff like using p4merge as git's diff/merge tool fails to work.

rashil2000:
Do all of them support CLI arguments?

Jestar342:
Yes, they all do.

rashil2000:
Move them below the extract_dir field. 

What is the difference between p4.exe and p4v.exe?

Jestar342:
I believe p4 is the cli/tui client, whilst p4v is the visual client. Both accept cli args though, as you can start the visual client with connection details etc straight from the cli.

image

PR #7691, opened 2022-01-07
fix: Don't run Excavator in forks

rasa:
I can't freshen my fork from upstream via GitHub, see
https://github.com/rasa/scoop-extras/pull/1

Users can also turn off actions in their fork, but by then, their fork has already diverged. And they may want some actions.

### DIFF
--- a/.github/workflows/excavator.yml
+++ b/.github/workflows/excavator.yml
@@ -11,6 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@main
     - name: Excavate
+      if: github.repository == ‘ScoopInstaller/#xtras’
       uses: ScoopInstaller/GithubActions@main
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/excavator.yml
+++ b/.github/workflows/excavator.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@main
     - name: Excavate
-      if: github.repository == ‘ScoopInstaller/#xtras’
+      if: github.repository == ‘ScoopInstaller/Extras’
       uses: ScoopInstaller/GithubActions@main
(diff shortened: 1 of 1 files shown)


.github/workflows/excavator.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@main
     - name: Excavate
+      if: github.repository == ‘ScoopInstaller/#xtras’
```
niheaven, line 14: Is `#xtras` a typo?

issaclin32: Apostrophes (‘’) also looks weird. Normally should be `'ScoopInstaller/#xtras'`

niheaven: So why `#`...

rashil2000:
> Users can also turn off actions in their fork, but by then, their fork has already diverged. And they may want some actions.

When a fork is created, _scheduled_ actions are disabled by default (even if they have normally enabled Actions to run on forks). 

Since this is something a user needs to manually enable, I don't think we would need this.

niheaven:
But if I keep these forked repos up to date with official repos, the GitActions is enabled by default.

rashil2000:
> But if I keep these forked repos up to date with official repos, the GitActions is enabled by default.

Normal GitHub Actions are, but scheduled ones aren't.

.github/workflows/excavator.yml
rasa: Typo, sorry.

rasa:
> > Users can also turn off actions in their fork, but by then, their fork has already diverged. And they may want some actions.
> 
> When a fork is created, _scheduled_ actions are disabled by default (even if they have normally enabled Actions to run on forks).
> 
> Since this is something a user needs to manually enable, I don't think we would need this.

Agreed. Closng.

PR #7710, opened 2022-01-10
modernflyouts: Add version 0.9.3.0

### LINKED ISSUE #3142: [Request] Add AudioFlyout
Replace the Volume/SMTC UI with a custom one.
https://github.com/ADeltaX/AudioFlyout

issaclin32:
close #3142

[ModernFlyouts](https://github.com/ModernFlyouts-Community/ModernFlyouts) is a replacement for the built-in audio flyout UI, with a Modern Fluent Design.

Notes:
* *persist* is not needed because config file is at `$Env:LocalAppData\ModernFlyouts_Community`.
* ~*autoupdate* cannot work properly without `"replace": "${version}"` in *checkver*.~ This is probably a bug in Scoop core. 
**UPDATE**: The problem no longer happens in this manifest.
It happens when the 2nd matchgroup is named as *version*. (e.g. `"regex": "(\\w+)\\.ModernFlyouts_(?<version>[\\d.]+)(\\w+)\\.Msixbundle"`)
*checkver* will be fine, but in *autoupdate*, $version becomes an empty string.

Screenshots:

### DIFF
--- a/bucket/modernflyouts.json
+++ b/bucket/modernflyouts.json
@@ -28,11 +28,11 @@
         ]
     ],
     "checkver": {
+        "##": "Github release tag (e.g. 0.9.3) != full version (e.g. 0.9.3.0), which can cause error during extraction, thus 'regex' is needed.",
         "github": "https://github.com/ModernFlyouts-Community/ModernFlyouts",
-        "regex": "(\\w+)\\.ModernFlyouts_(?<version>[\\d.]+)(\\w+)\\.Msixbundle",
-        "replace": "${version}"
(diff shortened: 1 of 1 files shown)


rashil2000 (commented):
In `32669SamG.ModernFlyouts_0.9.3.0_neutral___pcy8vm99wrpcg`, the prefix and suffix are constants. My guess is they're the developer's Store account tag and Store application tag (or something similar). 

They have been constant since quite a few releases. So I think the GitHub checkver (no custom regex) will work fine.

rashil2000 (approved):

PR #7735, opened 2022-01-13
fatty: Add version 20210122

### LINKED ISSUE #4932: [Request] Add fatty
Fatty is a mintty based terminal with tabs

https://github.com/juho-p/fatty

issaclin32:
close #4932

[FaTTY](https://github.com/juho-p/fatty) is a Cygwin Terminal emulator **with tabs**.

Notes:
* FaTTY has not received any updates since ~2016~ Jan 2021. Therefore I think we can just compile from source, and stick with this version.
* The binary is compiled with Cygwin version 2.915 and the following packages:
cyg

Screenshots:

### DIFF
--- a/bucket/fatty.json
+++ b/bucket/fatty.json
@@ -1,0 +1,24 @@
+{
+    "version": "1.6",
+    "description": "Cygwin Terminal emulator with tabs",
+    "homepage": "https://github.com/juho-p/fatty",
+    "license": "GPL-3.0-or-later",
+    "depends": "cygwin",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


niheaven:
There's no tag since 2016, but the development is not stopped, so will you compile master branch instead of tag 1.6?

issaclin32:
They do have [some bug fixes](https://github.com/juho-p/fatty/commits/master) after version 1.6. (last update was  2021/1/22)
I just compiled from `master`, and use the last update date as "version".

tech189 (approved):
Tested, works for me

PR #7736, opened 2022-01-13
autoactions: Update the AutoHDR manifest

RavenMacDaddy:
The AutoHDR project is now named AutoActions to not be confused with the Windows 11 Auto HDR feature, hence this manifest update and correction is needed.

This PR also makes the version up-to-date based on the build available on creation.


rashil2000:
Move `persist` after shortcuts

PR #7738, opened 2022-01-14
keenwrite: Add version 2.5.1

rashil2000:

### DIFF
--- a/bucket/keenwrite.json
+++ b/bucket/keenwrite.json
@@ -1,0 +1,35 @@
+{
+    "version": "2.5.1",
+    "description": "Java-based desktop text editor with live preview, string interpolation, and formulas",
+    "homepage": "https://github.com/DaveJarvis/keenwrite",
+    "license": {
+        "url": "https://github.com/DaveJarvis/keenwrite/blob/master/LICENSE.md"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/DaveJarvis/keenwrite/releases/download/2.5.1/keenwrite.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


tech189:
Any idea why the program saves its files in the root of the Scoop directory by default?

bucket/keenwrite.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "2.5.1",
+    "description": "Java-based desktop text editor with live preview, string interpolation, and formulas",
+    "homepage": "https://github.com/DaveJarvis/keenwrite",
+    "license": {
+        "identifier": "Unknown",
```
issaclin32, line 6: I think `Unknown` implies that we cannot find License/EULA of the software, thus leave it unsettled.

Is `Proprietary` applicable to this case?
https://en.wikipedia.org/wiki/Proprietary_software
> While most proprietary software is distributed without the source code, some vendors distribute the source code or otherwise make it available to customers. For example, users who have purchased a license for the Internet forum software vBulletin can modify the source for their own site but cannot redistribute it. This is true for many web applications, which must be in source code form when being run by a web server. The source code is covered by a non-disclosure agreement or a license that allows

issaclin32:
> Any idea why the program saves its files in the root of the Scoop directory by default?

It depends on where the **current working directory** is.

When we start **keenwrite** via shortcut, the working directory will be `(SCOOP_DIR)\apps\keenwrite\current`.

If we start the app via command line (working directory: `D:\`), the file will be saved in `D:\`.

Not sure if there is any option available that can set the app to "ask saving location when saving a new file".

issaclin32 (approved):

Issue #3548, opened 2020-02-01
opentrack: Use portable configuration


Ash258:
https://github.com/opentrack/opentrack/wiki/portable-mode-for-USB-sticks

Migrate or add new `-portable` manifest.

PR #7760, opened 2022-01-17
opentrack: Update to version 2022.1.0

Lutra-Fs:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/opentrack.json
+++ b/bucket/opentrack.json
@@ -1,10 +1,10 @@
 {
-    "version": "2021.2.0",
+    "version": "2022.1.0",
     "description": "Tracks user's head movements and relaying the information to games and flight simulation software.",
     "homepage": "https://github.com/opentrack/opentrack",
     "license": "ISC",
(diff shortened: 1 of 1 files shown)


rashil2000:
An unrelated question: where are the settings/config stored for the portable version?

Lutra-Fs:
it is in Documents/opentrack-v2.3

Lutra-Fs:
if u create this file the settings will be stored in `installation path/ini` folder.

Lutra-Fs:
I will make an issue about the config store path in this repo.

rashil2000:
Issue is already present https://github.com/ScoopInstaller/Extras/issues/3548

We just need a PR to handle migration :)

Lutra-Fs:
Yes just found that. But if we change the config location, there will be some impact on the users who already installed it. I can do this later if we need it.

rashil2000:
Yes, that's what I mean by migration. See https://github.com/ScoopInstaller/Extras/blob/master/bucket/vscode.json#L44-L51 for example.

PR #7768, opened 2022-01-17
virt-viewer: Update to version 11.0-1.0 and add autoupdate

LazyGeniusMan:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/virt-viewer.json
+++ b/bucket/virt-viewer.json
@@ -34,10 +34,10 @@
     },
     "autoupdate": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


rashil2000:
You hardcoded the version in URLs in autoupdate

rashil2000:
The URL and hash are still pointing to 10.0-1.0

LazyGeniusMan:
> The URL and hash are still pointing to 10.0-1.0

sorry apparently it's because autoupdate schema is wrong so it's not updating correctly with checkver, fixed

PR #7780, opened 2022-01-18
rambox-pro: Fix exe name and download link

### LINKED ISSUE #7779: Rambox 2.0 filename change breaks update/download link for manifes
## Bug Report

**rambox-pro**

### Current Behaviour

Installs fine with current versions; however updates to, or trying to install 2.0.0 will fail. Upstream change to exe name has broken the manifest for version 2.0+

### Expected Behaviour
`scoop Install rambox-pro` or `scoop update rambox-pro` would install/update respectively

### Additional context/output
Rambox has dropped the "Pro" from the exe name, and now has added "-x64" to the file-name

### Possible Solution
Proposed fix in pull #7780

AyoungDukie:
Rambox has dropped the "Pro" from the exe name, and now has added "-x64" to the file-name

Closes #7779

### DIFF
--- a/bucket/rambox-pro.json
+++ b/bucket/rambox-pro.json
@@ -20,14 +20,14 @@
     "shortcuts": [
         [
             "Rambox.exe",
-            "RamboxPro"
+            "Rambox"
         ]
     ],
     "checkver": {
         "github": "https://github.com/ramboxapp/download"
     },
     "autoupdate": {
-        "url": "https://github.com/ramboxapp/download/releases/download/v$version/RamboxPro-$version-win.exe#/cosi.7z",
+        "url": "https://github.com/ramboxapp/download/releases/download/v$version/Rambox-$version-win-64.exe#/cosi.7z",
         "hash": {
             "url": "$baseurl/latest.yml",
             "regex": "sha512:\\s+$base64"
--- a/bucket/rambox-pro.json
+++ b/bucket/rambox-pro.json
@@ -6,14 +6,16 @@
         "identifier": "Proprietary",
         "url": "https://rambox.pro/#eula"
     },
-    "url": "https://github.com/ramboxapp/download/releases/download/v2.0.0/Rambox-2.0.0-win-x64.exe#/cosi.7z",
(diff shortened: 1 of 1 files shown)


niheaven:
You should change shortcut name and autoupdate.url as well.

rashil2000:
Also, the download name has `64` now. Does it still support 32bit? If not, remove it.

AyoungDukie:
I haven't personally confirmed 32-bit support (or lack-thereof), but I do understand and have updated the auto-update URL, and the shortcut name, accordingly.

rashil2000:
> I haven't personally confirmed 32-bit support (or lack-thereof), but I do understand and have updated the auto-update URL, and the shortcut name, accordingly.

Please check 32bit installationby running `scoop install <filename>.json -a 32bit`.

niheaven:
The manifest name should be changed to rambox too.

rashil2000:
There is already rambox.json - open source build of Rambox Pro

niheaven:
Rambox 2.0 has merged these two version into one package:

> In 2021, we decided to unify the CE and Pro versions in a single program to make the use and updates much easier.

So maybe this one should become `rambox.json` and existed one should be removed.

rashil2000:
The repo https://github.com/ramboxapp/community-edition is not archived as of now and had last release in August, I think we can keep it

niheaven:
The post and new release is released 2 days ago (https://rambox.medium.com/?p=4bdef5d6c3b8), so why not move original one to `Versions` and keep only one here? It's an upstream change, not a manifests reorganization.

AyoungDukie:
As a small note, the Rambox support team explained the following in an email:

>The idea is to move the users, in a few weeks, from the CE to our Basic plan in which they have the same features and a few more. The code and the app will keep free as it is, so users can continue to use the CE but without future updates. We will release one more CE version which users can migrate with a simple click to our latest version.

rashil2000:
Okay, so we will retire the manifest in the next CE release

niheaven:
> Okay, so we will retire the manifest in the next CE release

There're no CE release anymore, they all move to `Rambox`, both free and commercial version.

PR #7796, opened 2022-01-20
lunacy: Update to version 8.1.0.0

rodrigost23:
The regex used to find the .exe installer is no longer valid, as the link now says "Installer" instead of "here". So I updated it and ran checkver to update the manifest to the latest release.

### DIFF
--- a/bucket/lunacy.json
+++ b/bucket/lunacy.json
@@ -1,5 +1,5 @@
 {
-    "version": "7.1.0",
+    "version": "8.1.0.0",
     "description": "Free Graphic Design Software",
     "homepage": "https://icons8.com/lunacy",
     "license": {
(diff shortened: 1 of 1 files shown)


bucket/lunacy.json
```diff
@@ -22,7 +22,7 @@
     ],
     "checkver": {
         "url": "https://docs.icons8.com/release-notes/",
-        "regex": "LunacySetup_([\\d.]+)\\.exe\">here"
+        "regex": "LunacySetup_([\\d.]+)\\.exe\">Installer"
```
niheaven, line 25:
```suggestion
        "regex": "LunacySetup_([\\d.]+)\\.exe
```

What if using this regex?

rodrigost23:
I think it's ready now

niheaven (approved):

PR #7803, opened 2022-01-21
Add monokle 1.4.0

Slach:


bucket/monokle.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "1.4.0",
+    "description": "Monokle is your K8s best friend for creating, validating, debugging and managing manifests!",
+    "homepage": "https://github.com/kubeshop/monokle",
```
rashil2000, line 4:
```suggestion
    "homepage": "https://kubeshop.github.io/monokle",
```

bucket/monokle.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "1.4.0",
+    "description": "Monokle is your K8s best friend for creating, validating, debugging and managing manifests!",
```
rashil2000, line 3:
```suggestion
    "description": "K8s tool for creating, validating, debugging and managing manifests.",
```

bucket/monokle.json
```diff
@@ -0,0 +1,37 @@
...
+        "64bit": {
+            "url": "https://github.com/kubeshop/monokle/releases/download/v1.4.0/Monokle-win-1.4.0-x64.exe#/dl.7z",
+            "hash": "sha512:1c5537e5dbb1ea688bf6cd7544e53259a97825a12a27029befabc1fedd908b8439cf80017dd72c5f5ce5181694ba1915520d0930616898b24f2671e8f07b0303",
+            "pre_install": [
+                "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\"",
+                "Remove-Item \"$dir\\`$*\", \"$dir\\Uninstall*\" -Force -Recurse"
+            ]
+        }
+    },
+    "bin": "Monokle.exe",
```
rashil2000, line 16: Commandline arguments?

rashil2000 (changes requested):
Anything to persist?

bucket/monokle.json
Slach: no it's a UI only tool

Slach:
> Anything to persist?

Show we add `%USERPROFILE%\AppData\Roaming\monokle` to persist? How to do it?

bucket/monokle.json
rashil2000: Then `bin` should be removed

rashil2000:
> > Anything to persist?
> 
> Show we add `%USERPROFILE%\AppData\Roaming\monokle` to persist? How to do it?

This cannot be added to persist. Is there a way to configure this location?

Slach:
> This cannot be added to persist. Is there a way to configure this location?

Only if we can re-define `%APPDATA%` environment variable especially for `monokle.exe`
how to do it with scoop?

rashil2000 (changes requested):

bucket/monokle.json
```diff
@@ -0,0 +1,42 @@
...
+        [
+            "Monokle.exe",
+            "Monokle"
+        ]
+    ],
+    "depends": [
+        "kubectl",
+        "helm",
+        "kustomize"
+    ],
```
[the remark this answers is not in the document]
Slach: Yes, during startup these dependencies checks during monokle startup and warning show in logs and UI
If this utils missed, part of function is unavailable

I can move helm and kustomize into "suggest", but not sure it will right choice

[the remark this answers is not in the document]
rashil2000: Only put apps that are absolutely required in `depends`, i.e. without which Monokle will not function at all. Put the rest in `suggest`.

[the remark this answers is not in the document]
Slach: kubectl is necessary, `helm` and `kustomize` moved to `suggest`

bucket/monokle.json
Slach: bin removed

bucket/monokle.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "shortcuts": [
+        [
+            "Monokle.exe",
+            "Monokle"
+        ]
+    ],
+    "depends": [
+        "kubectl"
+    ],
```
rashil2000, line 24:
```suggestion
    "depends": "kubectl",
```

rashil2000 (changes requested):
Move depends and suggest after license

niheaven:
You could use https://github.com/ScoopInstaller/Extras/blob/master/bucket/webcopy.json as reference of persisting `$env:APPDATA`.

rashil2000:
> You could use https://github.com/ScoopInstaller/Extras/blob/master/bucket/webcopy.json as reference of persisting `$env:APPDATA`.

I find that method to be quite invasive, to be honest. If AppData is the only place an app stores data, we shouldn't interfere. Moreover, some programs do not play well with junctions/links.

Slach:
@rashil2000 @niheaven look https://github.com/ScoopInstaller/Extras/pull/7803/commits/2cf47e193be03729d68044b9e76f0fce0300ea9c
I checked, everything works fine with `junction`

niheaven:
> > You could use [`master`/bucket/webcopy.json](https://github.com/ScoopInstaller/Extras/blob/master/bucket/webcopy.json?rgh-link-date=2022-01-21T17%3A04%3A32Z) as reference of persisting `$env:APPDATA`.
> 
> I find that method to be quite invasive, to be honest (and against Scoop philosophy). If AppData is the only place an app stores data, we shouldn't interfere. Moreover, some programs do not play well with junctions/links.

To be honest, scoop shouldn't affect `$env:APPDATA`, since purging scoop by simple removing scoop root folder won't remove these items as well. This method should only be used when persistence is really important for the app.

rashil2000:
> This method should only be used when persistence is really important for the app.

Even in this case, the data is getting stored in AppData anyways, so we should not touch it, whether installed by Scoop or by other means.

PR #7806, opened 2022-01-21
gephi: add version 0.9.2

### LINKED ISSUE #7794: [Request] Gephi
### Information

**Name:** Gephi

**Description:** Gephi is the leading visualization and exploration software for all kinds of graphs and networks. Gephi is open-source and free.

**Homepage:** https://gephi.org/

**Download link(s):** https://github.com/gephi/gephi

**Some indication of popularity/repute:** 4.6k Stars

tech189:
- Closes #7794
- Nothing to persist as config stored in `C:\Users\username\AppData\Roaming\.gephi`
- Caveat: requires Java 1.8 to be installed - does NOT detect Java installed using any manifest in the Java bucket nor does it accept any higher Java version, should this be added as a note?
- Java issue would probably be fixed by https://github.com/ScoopInstaller/Java/issues/119 as at the moment Java manifests only add `JAVA_HOME` environment variable

### DIFF
--- a/bucket/gephi.json
+++ b/bucket/gephi.json
@@ -1,0 +1,31 @@
+{
+    "version": "0.9.2",
+    "description": "Visualization and exploration software for all kinds of graphs and networks",
+    "homepage": "https://github.com/gephi/gephi",
+    "license": "GPL-3.0-only",
+    "url": "https://github.com/gephi/gephi/releases/download/v0.9.2/gephi-0.9.2-windows.exe",
+    "hash": "cc8dbf806baa157fd34876c940bead660c30d160d391c0189826b5fe270678fb",
+    "innosetup": true,
+    "architecture": {
+        "64bit": {
+            "shortcuts": [
+                [
+                    "bin\\gephi64.exe",
+                    "Gephi"
+                ]
+            ]
+        },
+        "32bit": {
+            "shortcuts": [
+                [
+                    "bin\\gephi.exe",
+                    "Gephi"
+                ]
+            ]
+        }
+    },
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://github.com/gephi/gephi/releases/download/v$version/gephi-$version-windows.exe"
+    }
+}


rashil2000:
/verify

> should this be added as a note?

Add a suggest for Java 1.8 and a note that any other version won't work.

tech189:
Sorry I wasn't clear, I meant that it only works if Java 1.8 is installed from https://www.java.com/en/download/ (it probably only checks the registry), not through Scoop. So rather than a `suggest` should we put a note?

rashil2000:
Oh, okay. No suggest then. 

Add the URL for Java 1.8 in the note too.

tech189:
Actually reading up a bit more I discovered the `--jdkhome` flag and the `jdkhome=` config in `$dir\etc\gephi.conf` which could be modified on installation to point to `$env:JAVA_HOME`! I tested it works with Scoop's Java manifests of `oraclejre8` and `liberica8-jdk` (and maybe others) but definitely not `openjdk` because of the limitation of only Java 8 - for reference this is an issue being tracked in the Gephi repo: https://github.com/gephi/gephi/issues/2160

I will write an installer script to modify the conf file a bit later today.

tech189:
How can I get the `$env:JAVA_HOME` variable while running in `post_install`? It is empty on installation but contains the right path normally in my terminal.

rashil2000:
You can do a `Write-Host` in the post_install script to test the value of any variables. It will print to the terminal.

tech189:
@rashil2000 It's all ready now! The `post_install` checks if the conf file has got the `JAVA_HOME` set already and if not, it will add the current `JAVA_HOME` environment variable :)

bucket/gephi.json
```diff
@@ -0,0 +1,47 @@
...
+    "url": "https://github.com/gephi/gephi/releases/download/v0.9.2/gephi-0.9.2-windows.exe",
+    "hash": "cc8dbf806baa157fd34876c940bead660c30d160d391c0189826b5fe270678fb",
+    "innosetup": true,
+    "post_install": [
+        "$conf_file = \"$dir\\etc\\gephi.conf\"",
+        "$file_contents = Get-Content -path $conf_file -Raw",
+        "if (($file_contents -match \"#jdkhome\") -contains $true) {",
+        "   $replaced = $file_contents -replace \"#jdkhome=`\"\/path\/to\/jdk`\"\", \"jdkhome=`\"$env:JAVA_HOME`\"\"",
+        "   Set-Content -Path $conf_file -Value $replaced",
+        "   Write-Host \"Notes`n-----`nScoop has configured Gephi to run using your current JAVA_HOME - if that is not Java 8, you will need to modify $dir\\etc\\gephi.conf\"",
```
rashil2000, line 21: This line can be moved to `notes`, for consistency

rashil2000: Okay

PR #7810, opened 2022-01-22
Propertree: Add version 0.2022.01.22

hu3rror:

No persist file.


rashil2000:
I see python scripts, we should add `"depends": "python",`, no?

hu3rror:
> I see python scripts, we should add `"depends": "python",`, no?

You are right, it has been corrected, thank you!

hu3rror:
Why can't checkver pass after I add depends?🤔

rashil2000:
It fails because there's a new version available

PR #7812, opened 2022-01-22
KopiaUI: Add version 0.10.1

hu3rror:

Data is stored in `%appdata%/kopia-ui`

### DIFF
--- a/bucket/kopiaui.json
+++ b/bucket/kopiaui.json
@@ -1,0 +1,37 @@
+{
+    "version": "0.10.1",
+    "description": "Fast and secure open source backup.",
+    "homepage": "https://kopia.io/",
+    "license": "Apache-2.0",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


niheaven (approved):
You may also add kopia in `main` bucket.

rashil2000:
If you plan to add kopia to Main bucket remove the `"bin": "resources\\server\\kopia.exe",` line, otherwise it will conflict.

niheaven:
> If you plan to add kopia to Main bucket remove the `"bin": "resources\\server\\kopia.exe",` line, otherwise it will conflict.

Keep it here if someone wants to install only Kopia UI, IMO. shim overwriting will solve the conflict.

rashil2000:
But why would you need the shim to the server in a GUI app?

niheaven:
It is the client CLI, just packed in the UI. The hashsum is different from the one from cli package, though.

hu3rror:
I just added the CLI version of Kopia in Main.
And do I need to modify the manifest for kopiaUI？

PR #7819, opened 2022-01-23
cheat-engine: Undeprecate manifest

### LINKED ISSUE #7414: cheat-engine is removed from this bucket?
Will it be added back or it's not allowed any longer or something?

rashil2000:
Closes #7414

### DIFF
--- a/bucket/cheat-engine.json
+++ b/bucket/cheat-engine.json
@@ -1,13 +1,13 @@
 {
-    "version": "7.2-2020-11-04T09",
+    "version": "7.4",
     "description": "Tool for modifying/debugging single player games and applications.",
     "homepage": "https://cheatengine.org/index.php",
     "license": {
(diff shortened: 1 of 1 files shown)


Stzx:
Hey, I think we should remove this package, the full installer is no longer available directly on `github` since [Cheat Engine 7.1](https://github.com/cheat-engine/cheat-engine/releases/tag/7.1). The exe in the current file is just the downloader.

rashil2000:
Maybe we can use the URL that the downloader uses? It might be available in the source code.

Stzx:
I don't think it's included in the source code(maybe I'm missing something), and whether extracting the link to use it violates some kind of license, I don't know.

rashil2000:
Can you open an issue in their repo for advice? It would be best to ask them.

PR #7825, opened 2022-01-25
FastCopy: update to 4.0.2

CrendKing:
FastCopy is updated with a major version bump (v4). It introduces exciting features such as a command line app. It also no longer releasing a zip archive for download. Instead, there is a installer with "extraction" option. So we need to change the script a bit.

Here is my attempt. I tested locally without problem. One small question: when the manifest has "installer/script" specified, the installer file is always kept? Anyways, I explicitly remove `$dir\$fname` at the end.

### DIFF
--- a/bucket/fastcopy.json
+++ b/bucket/fastcopy.json
@@ -18,8 +18,8 @@
             "installer": {
                 "script": [
                     "Start-Process \"$dir\\$fname\" '/EXTRACT64' -Wait",
-                    "Get-ChildItem \"$dir\\FastCopy${version}_x64\" | Move-Item -Destination \"$dir\"",
-                    "Remove-Item \"$dir\\$fname\", \"$dir\\FastCopy${version}_x64\""
+                    "Get-ChildItem \"$dir\\FastCopy$version_x64\" | Move-Item -Destination \"$dir\"",
+                    "Remove-Item \"$dir\\$fname\", \"$dir\\FastCopy$version_x64\""
                 ]
             }
         },
@@ -47,6 +47,6 @@
     "persist": "FastCopy2.ini",
     "checkver": "FastCopy ver ([\\d.]+)",
     "autoupdate": {
-        "url": "https://fastcopy.jp/archive/FastCopy${version}_installer.exe"
+        "url": "https://fastcopy.jp/archive/FastCopy$version_installer.exe"
     }
 }
--- a/bucket/fastcopy.json
+++ b/bucket/fastcopy.json
@@ -27,8 +27,8 @@
             "installer": {
                 "script": [
                     "Start-Process \"$dir\\$fname\" '/EXTRACT32' -Wait",
-                    "Get-ChildItem \"$dir\\FastCopy${version}\" | Move-Item -Destination \"$dir\"",
-                    "Remove-Item \"$dir\\$fname\", \"$dir\\FastCopy${version}\""
+                    "Get-ChildItem \"$dir\\FastCopy$version\" | Move-Item -Destination \"$dir\"",
+                    "Remove-Item \"$dir\\$fname\", \"$dir\\FastCopy$version\""
                 ]
             }
         }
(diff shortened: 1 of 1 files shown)


rashil2000:
Does extraction with 7zip not work? You can also try innosetup if it doesn't.

CrendKing:
I tried [UniExtract](https://github.com/Bioruebe/UniExtract2/) and none of the extractors (I believe InnoSetup included) works.

CrendKing:
Maybe the owner would release a zip version sometime in the future and we can revert the change. But right now the old version is no longer downloadable, so I think we should update this ASAP.

CrendKing:
Also, we persist the "Log" directory, but FastCopy by default writes log to "FastCopy.log" in $dir. Should we either add that to persist as well, or just remove "Log" from persist? Do many apps persist log related stuff?

rashil2000:
Do the logs contain any user information/configuration? If not, we can remove it from persist

CrendKing:
Seems to only contain the name of operation, timestamp, list of file system objects, etc. No user config.

rashil2000:
Okay, we can remove it

CrendKing:
Removed. If this change upsets someone who archives their FastCopy logs, we can always add those back.

rashil2000:
Use `$version` instead of `${version}`

CrendKing:
> Use `$version` instead of `${version}`

You see, since we can't write something like `FastCopy$version_x64`, and the only place your comment applies is the 32bit installer script, for the sake of consistency I always brace the variables in double quoted string. The advantage is that I never need to spend that energy determining.

Updated anyways.

> Just to confirm, you have tested both 32bit and 64bit installations, right?

Yes. If you refer to that in the 32bit installation you still get a `FastEx64.dll`, that's the original's doing. Guess they hardcoded that name for loading.

rashil2000:
> > Use `$version` instead of `${version}`
> 
> You see, since we can't write something like `"FastCopy$version_x64"`, and the only place your comment applies is the 32bit installer script, for the sake of consistency I always brace the variables in double quoted string. The advantage is that I never need to spend that energy determining.
> 
> Updated anyways.

Variable substitution using `${}` does not work in Scoop. Check the action I just ran. You are using a fork where it does, but this is upstream. 

You need to replace all occurrences as suggested.

CrendKing:
Do you suggest to change the autoupdate line to `"url": "https://fastcopy.jp/archive/FastCopy$version_installer.exe"`?

rashil2000:
> Do you suggest to change the autoupdate line to `"url": "https://fastcopy.jp/archive/FastCopy$version_installer.exe"`? I guess Scoop has some magic to do replacement for that variable.

Yes, that will work.

> Should I revert the ones in the "script" part for consistency?

Use `$version` everywhere.

CrendKing:
> Use `$version` everywhere.

You 100% sure? Because if we do it verbatim, the script will fail to run. Error:
```
Get-ChildItem:
Line |
   2 |  Get-ChildItem "$dir\FastCopy$version_x64" | Move-Item -Destination "$ …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path 'D:\Scoop\apps\fastcopy\4.0.2\FastCopy' because it does not exist.
Remove-Item:
Line |
   3 |  Remove-Item "$dir\$fname", "$dir\FastCopy$version_x64"
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path 'D:\Scoop\apps\fastcopy\4.0.2\FastCopy' because it does not exist.
```
I updated it anyways. Please run a test yourself and change it if necessary. Thanks!

rashil2000:
A proper and robust way to handle this would be capturing the directory name first, using a variable.

```pwsh
$fc_name = Get-Item "$dir\FastCopy*"
Get-ChildItem $fc_name | Move-Item -Destination "$dir"
Remove-Item "$dir\$fname", $fc_name
```

This way, you don't have to worry about versions.

CrendKing:
Your code wouldn't work as-is, because the downloaded filename also starts with `FastCopy`, thus `Get-Item` will pick that file up. I used `Get-ChildItem -Directory` to work around it.

PR #7833, opened 2022-01-26
Update to version 2.3.1-beta.2

Gladtbam:


niheaven:
Please don't submit pre-release version to `Extras` except that app has not update its stable release for a long time and the pre-release is stable enough.

If you really need a `pre-release`, you could do a PR in `Versions` named e.g. picgo-beta for this app.

PR #7836, opened 2022-01-26
Fixed #7835 : Bump storageexplorer to azure-storageexplorer

### LINKED ISSUE #7835: [Request] Rename storageexplorer to azure-storageexplorer
### Description

In terms of being consistent with names, it might be thought about to give the `storageexplorer` the prefix `azure` like:

`azure-storageexplorer` or `azure-storage-explorer`

This would searching with `scoop` easier.

Anselmoo:
Bump `storageexplorer` to `azure-storageexplorer`

Closes #7835


rashil2000:
To prevent renaming multiple packages and cause as less disruption as possible, let's rename this to azurestorageexlorer.json instead.

Anselmoo:
> To prevent renaming multiple packages and cause as less disruption as possible, let's rename this to azurestorageexlorer.json instead.

I agree but then you also have to rename the `azure-iot-explorer`? I also took a look at `aws`-services and it seems you also separated this via -. 

Let me know about your thoughts.

rashil2000:
I'd like to break as much less workflows of people as possible. Presence/absence of hyphen can be ignored, they don't hamper discoverability. 

The `azure` suffix is useful, so I'm allowing it.

Anselmoo:
> I'd like to break as much less workflows of people as possible. Presence/absence of hyphen can be ignored, they don't hamper discoverability.
> 
> The `azure` prefix is useful, so I'm allowing it.

Make sense!

PR #7840, opened 2022-01-27
komorebi: Add version 0.1.8

sitiom:

There is still a problem with this manifest, `komorebi.exe` does not work when launched from the shim, although it works when launched directly. I hope someone here can help with that. 
Related issue: https://github.com/LGUG2Z/komorebi/issues/95

### DIFF
--- a/bucket/komorebi.json
+++ b/bucket/komorebi.json
@@ -24,7 +24,10 @@
         "Copy-Item $dir\\komorebi.exe $scoopdir\\shims\\komorebi.exe"
     ],
     "uninstaller": {
-        "script": "Remove-Item $scoopdir\\shims\\komorebi.exe"
+        "script": [
+            "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
+            "Remove-Item $scoopdir\\shims\\komorebi.exe"
+        ]
     },
     "bin": [
         "komorebi.exe",
--- a/bucket/komorebi.json
+++ b/bucket/komorebi.json
@@ -8,6 +8,9 @@
         "Run 'komorebic ahk-library' if you would like to generate an AHK helper library to use in your configuration.",
         "Once you have a configuration file in place, you can run 'komorebic start' to start the window manager."
     ],
+    "suggest": {
+        "autohotkey": "extras/autohotkey"
+    },
     "architecture": {
         "64bit": {
             "url": "https://github.com/LGUG2Z/komorebi/releases/download/v0.1.7/komorebi-0.1.7-x86_64-pc-windows-msvc.zip",
--- a/bucket/komorebi.json
+++ b/bucket/komorebi.json
@@ -1,0 +1,30 @@
+{
+    "version": "0.1.7",
+    "description": "A tiling window manager for Windows",
+    "homepage": "https://github.com/LGUG2Z/komorebi",
+    "license": "MIT",
+    "notes": [
+        "Run 'cp $original_dir\\komorebi.sample.ahk $env:USERPROFILE\\komorebi.ahk' to get started with the sample configuration.",
+        "Run 'komorebic ahk-library' if you would like to generate an AHK helper library to use in your configuration.",
+        "Once you have a configuration file in place, you can run 'komorebic start' to start the window manager."
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/LGUG2Z/komorebi/releases/download/v0.1.7/komorebi-0.1.7-x86_64-pc-windows-msvc.zip",
+            "hash": "dc7d22d7a3b3bae46c11f734c92e4c4c61a1eb5e3c344fd06cc4361b259067b5"
+        }
+    },
+    "pre_install": "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
+    "bin": [
+        "komorebi.exe",
+        "komorebic.exe"
+    ],
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/LGUG2Z/komorebi/releases/download/v$version/komorebi-$version-x86_64-pc-windows-msvc.zip"
+            }
+        }
+    }
+}
--- a/bucket/komorebi.json
+++ b/bucket/komorebi.json
@@ -17,7 +17,6 @@
             "hash": "dc7d22d7a3b3bae46c11f734c92e4c4c61a1eb5e3c344fd06cc4361b259067b5"
         }
     },
-    "pre_install": "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
     "post_install": [
         "# https://github.com/LGUG2Z/komorebi/issues/95",
         "Write-Host \"Copying 'komorebi' to shims folder.\"",
--- a/bucket/komorebi.json
+++ b/bucket/komorebi.json
@@ -1,5 +1,5 @@
 {
-    "version": "0.1.7",
+    "version": "0.1.8",
     "description": "A tiling window manager for Windows",
     "homepage": "https://github.com/LGUG2Z/komorebi",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


rashil2000:
Can you try changing the text in the associated `.shim` file to use the actual version number, instead of "current", and see if that works?

sitiom:
> Can you try changing the text in the associated `.shim` file to use the actual version number, instead of "current", and see if that works?

Doesn't work unfortunately

rashil2000:
Hmm. I can think of one method, but it's really ugly - copy the 2 files into $shimdir during installation, and delete them during uninstallation.

sitiom:
> > Doesn't work unfortunately
> 
> For both komorebi.exe and komorebic.exe?
> 
> Hmm. In that case, I can think of one method, but it's really ugly - copy the 2 .exe files into $shimdir during installation, and delete them during uninstallation.

`komorebic` does work, the problem is with `komorebi.exe` not launching properly.

> Hmm. In that case, I can think of one method, but it's really ugly - copy the 2 .exe files into $shimdir during installation, and delete them during uninstallation.

That's the solution in https://github.com/LGUG2Z/komorebi/issues/95#issuecomment-1022766335 and it does work (just replacing `komorebi.exe`)

rashil2000:
What is the error on launching komorebi.exe when installed from the manifest in this PR?

sitiom:
> What is the error on launching komorebi.exe when installed from the manifest in this PR?

Running `komorebi --help` produces an output:
```
komorebi 0.1.6
Jade Iqbal <jadeiqbal@fastmail.com>
A tiling window manager for Windows

USAGE:
    komorebi.exe [OPTIONS]

OPTIONS:
        --ffm        Allow the use of komorebi's custom focus-follows-mouse implementation
    -h, --help       Print help information
    -V, --version    Print version information
```
But running just `komorebi` returns nothing, even if the rust log is set to full (`$env:RUST_LOG = 'trace'; $env:RUST_BACKTRACE = 'full'`).

sitiom:
Pushed the workaround for the meanwhile.

bucket/komorebi.json
```diff
@@ -0,0 +1,39 @@
...
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/LGUG2Z/komorebi/releases/download/v0.1.7/komorebi-0.1.7-x86_64-pc-windows-msvc.zip",
+            "hash": "dc7d22d7a3b3bae46c11f734c92e4c4c61a1eb5e3c344fd06cc4361b259067b5"
+        }
+    },
+    "pre_install": "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
+    "post_install": [
+        "Write-Host \"Copying 'komorebi' to shims folder.\"",
+        "Copy-Item $dir\\komorebi.exe $scoopdir\\shims\\komorebi.exe -Force"
```
rashil2000, line 23:
```suggestion
        "Copy-Item $dir\\komorebi.exe $scoopdir\\shims\\komorebi.exe"
```

Why force?

bucket/komorebi.json
```diff
@@ -0,0 +1,39 @@
...
+    ],
+    "uninstaller": {
+        "script": "Remove-Item $scoopdir\\shims\\komorebi.exe"
+    },
+    "bin": [
+        "komorebic.exe"
+    ],
```
rashil2000, line 30:
```suggestion
    "bin": "komorebic.exe",
```

bucket/komorebi.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/LGUG2Z/komorebi/releases/download/v0.1.7/komorebi-0.1.7-x86_64-pc-windows-msvc.zip",
+            "hash": "dc7d22d7a3b3bae46c11f734c92e4c4c61a1eb5e3c344fd06cc4361b259067b5"
+        }
+    },
+    "pre_install": "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
+    "post_install": [
+        "Write-Host \"Copying 'komorebi' to shims folder.\"",
```
rashil2000, line 22: Add a comment here, linking to the GitHub issue

bucket/komorebi.json
```diff
@@ -0,0 +1,39 @@
+{
+    "version": "0.1.7",
+    "description": "A tiling window manager for Windows",
+    "homepage": "https://github.com/LGUG2Z/komorebi",
+    "license": "MIT",
+    "notes": [
+        "Run 'cp $original_dir\\komorebi.sample.ahk $env:USERPROFILE\\komorebi.ahk' to get started with the sample configuration.",
```
rashil2000, line 7:
```suggestion
        "Run 'copy $original_dir\\komorebi.sample.ahk $env:USERPROFILE\\komorebi.ahk' to get started with the sample configuration.",
```

rashil2000 (changes requested):

bucket/komorebi.json
sitiom: Actually, would it be good if I added `komorebi.exe` to `bin` so it can still be registered as a "shim"?

bucket/komorebi.json
sitiom: Putting in `Copy-Item` for completeness.

bucket/komorebi.json
rashil2000: Then you will need the force parameter while copying

bucket/komorebi.json
```diff
@@ -24,7 +24,10 @@
         "Copy-Item $dir\\komorebi.exe $scoopdir\\shims\\komorebi.exe"
     ],
     "uninstaller": {
-        "script": "Remove-Item $scoopdir\\shims\\komorebi.exe"
+        "script": [
+            "if (Get-Process -Name komorebi -ErrorAction SilentlyContinue) { komorebic stop }",
```
rashil2000, line 28: Now you can remove this line from pre_install

bucket/komorebi.json
sitiom: It works without the `-Force` parameter.

LGUG2Z:
Hey @rashil2000, I'm the maintainer of `komorebi` and I'm trying to figure out what is causing the behaviour described above. It seems like both binaries are compiled correctly and capable of execution, but the main event loop of `komorebi` is never entered into when it is executed as a shim.

The main event loop of `komorebi` [has an arg check as a prerequisite](https://github.com/LGUG2Z/komorebi/blob/master/komorebi/src/main.rs#L335) which basically enforces that if the binary is called with only one arg (ie. `komorebi.exe`), or if it is called with two args, a positive value must have been set for the `--ffm` flag (this is negative by default if the flag is not passed).

The only thing that I can imagine would cause this behaviour is if the shim is adding additional unexpected arguments when the binary is called.

I'm happy to make changes upstream to avoid any hacks on the scoop manifest if you can point me in the right direction!

LGUG2Z:
Thanks @rashil2000, I believe I have identified the issue.

The `komorebi.exe` in the `shims` folder starts a separate process with the same name which is not the 'real' `komorebi.exe`. This shim process remains running throughout the entire execution time of the real `komorebi.exe` process, which then causes [this check to ensure only one copy of the process is ever running at a time](https://github.com/LGUG2Z/komorebi/blob/e2ae9b1207102e1e303ea6d62f4439ac82fe91de/komorebi/src/main.rs#L344 to fail.

I will push a fix for this issue and cut a new release to allow the execution to continue if a shim process of the same name is running when this check is reached.

LGUG2Z:
On a side note, this check typically logs an error (see the code block linked above) to the console when the binary is executed directly- I think that the shim has been swallowing this error which has made it harder for end users (and me!) to identify the root cause of this issue.

LGUG2Z:
@sitiom https://github.com/LGUG2Z/komorebi/releases/tag/v0.1.8 should no longer have use the workaround with this release 🎉

sitiom:
Great, thanks for resolving the issue!

PR #7864, opened 2022-01-29
sdl2-ttf: Add version 2.0.18

criticic:
Added the sdl2_ttf library, which is used for font rendering.

Relates to #7867
Thanks to @g-berthiaume as this is based on sdl2-image manifest.


bucket/sdl2-ttf.json
```diff
@@ -0,0 +1,129 @@
...
+            "if($was_in_path) { $env:CMAKE_PREFIX_PATH = $newpath }"
+        ]
+    },
+    "checkver": {
+        "url": "https://github.com/libsdl-org/SDL_ttf/releases/download/latest/",
+        "regex": "SDL2_ttf-([\\d.]+)\\.zip"
+    },
+    "autoupdate": {
+        "url": [
+            "https://github.com/libsdl-org/SDL_ttf/releases/download/release-2.0.18/SDL2_ttf-devel-$version-VC.zip",
```
versedwildcat, line 125: Missed a $version for the update URLs
`https://github.com/libsdl-org/SDL_ttf/releases/download/release-$version/SDL2_ttf-devel-$version-VC.zip"`

issaclin32:
The manifest works fine for me, but I think this should be in [the main bucket](https://github.com/ScoopInstaller/Main).

Also, please use `checkver`->`regex` instead of `re`. 
`re` is now only kept for backward-compatibility.

rashil2000:
The manifest doesn't follow the 4th [criterion](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) for Main bucket.

criticic:
/verify

> Also, please use checkver->regex instead of re.
> re is now only kept for backward-compatibility.
Few examples here still use re, https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate, dont know from where it sourced though.

issaclin32:
> Few examples here still use re, https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate, dont know from where it sourced though.

Thank you for reporting. I have updated the page.

issaclin32:
> The manifest doesn't follow the 4th [criterion](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) for Main bucket.

I see.

PR #7871, opened 2022-01-30
vcredist2022: Update to version 14.31.31103.0

versedwildcat:

### DIFF
--- a/bucket/vcredist2022.json
+++ b/bucket/vcredist2022.json
@@ -20,7 +20,7 @@
         "$ec = @{",
         "    1638 = 'This product is already installed';",
         "    3010 = 'A restart is required to complete the installation';",
-        "}",
+        "} ",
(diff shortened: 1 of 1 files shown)


rashil2000:
Did you change the line endings of the file?

issaclin32:
There is an LF/CRLF issue, so I modified the file to fix it.

It is probably caused by Github's file upload. (which is quite an annoying problem)

If you use `git commit` and `git push` to upload the files, the CRLF line endings will all become LF. However, if you upload via Github's file upload, the line endings will be kept. (as what it is in the file, such as CRLF).

This will cause git to treat the local file is "modified" while it actually isn't, and cause git pulls from upstream unable to continue. Thus the bucket will be **unable** to update, unless we delete `vcredist2022.json`, or `scoop bucket rm extras` and add it back later.

issaclin32 (approved):

Issue #7867, opened 2022-01-29
[Request] SDL2


criticic:
Sort of Master Post related to SDL

### SDL2 (Base library)
https://www.libsdl.org/download-2.0.php
- [x] VC
- [ ] MinGW

### SDL_image (image)
https://www.libsdl.org/projects/SDL_image
- [x] VC
- [ ] MinGW

### SDL_ttf (font)
https://github.com/libsdl-org/SDL_ttf
- [ ] VC
- [ ] MinGW

### SDL_mixer (sound)
https://www.libsdl.org/projects/SDL_mixer/
- [ ] VC
- [ ] MinGW

## Bug fix
Currently the entire source code is downloaded, but we only need the package config file. So it would make sense to somehow just get the package config(eg. https://raw.githubusercontent.com/libsdl-org/SDL/main/sdl2.pc.in)

**Some indication of popularity/repute:** [GitHub stars/software reviews etc.]
https://www.wikiwand.com/en/List_of_games_using_SDL

rashil2000:
Related https://github.com/ScoopInstaller/Extras/pull/4259

criticic:
@rashil2000 how exactly should the mingw packages be added? 
Would making them available to pkg-config suffice? 
Does it need to be added to its own /bin /lib etc ?

rashil2000:
Sorry, I have no idea about this software, I won't be able to help :(

tresf:
> 

In short, yes, exposing to pkg-config is enough.  If scoop has since added a standardized way to do this, great.  At the time of the originating PR, none existed, so a [global environment variable was used](https://github.com/ScoopInstaller/Extras/pull/709/files#diff-cd78bc72628270670c8850c92e4905b28fd7a2be023dc43fcc6c50da2f84abc1R52) which pkg-config will pick up.

Of course, there are other ways the library can be found (e.g. `FindSDL2.cmake`) but this comes [with its own drama](https://stackoverflow.com/a/54006461/3196753). ❤️

rongcuid:
For whatever reason, pkg-config does not work even though the envs are set.

tresf:
> For whatever reason, pkg-config does not work even though the envs are set.

As long as `pkg-config` honors that variable, it should work.  My only recommendation is to echo it's value just before use and endure the path separators are valid for your operating system.

PR #7879, opened 2022-01-31
SDL2: Remove unnecessary cflags

criticic:
They were added in #709 by @tresf as a guess, but as far as I can tell pthread.h & asound.h are not available on windows. And thus caused library not found errors.

Relates to #7867


criticic:
https://github.com/libsdl-org/SDL/blob/main/configure.ac
Sdl2 is compiled for windows with winmm & threads support so we don't need thosel

PR #7893, opened 2022-02-02
TimeSlotTracker: Add version 1.3.23

zStruCat:
TimeSlotTracker is an application to register your time.


bucket/timeslottracker.json
```diff
@@ -0,0 +1,31 @@
+{
+    "version": "1.3.23",
+    "description": "TimeSlotTracker is an application to register your time.",
+    "homepage": "https://github.com/TimeSlotTracker/timeslottracker-desktop",
+    "license": "GPL-3.0-or-later",
+    "suggest": {
+        "JDK": [
+            "extras/oraclejdk",
+            "openjdk"
```
rashil2000, line 9:
```suggestion
            "java/oraclejdk",
            "java/openjdk"
```

rashil2000:
- .jar file can't be run directly, so create `bin` instead of `shortcut` (Scoop has special logic for handling .jar files)
- Is the app 32bit or 64bit?

zStruCat:
> * .jar file can't be run directly, so create `bin` instead of `shortcut` (Scoop has special logic for handling .jar bin files)
Are you sure about it? Cause it's a gui-program and can run well using shortcuts instead of bin with Java SE installed on my computer. There seems no need for shim to handle it.

>     * Is the app 32bit or 64bit?
jar file doesn't differentiate 32bits or 64 bits as they are byte code executed by JVM. Only the JVM and native libs have the differences of 32/64bits.

Also, am i understood correctly about manifest's architecture?
only 32-bit => no need for manifest.architecture.
only 64-bit => architecute.64bit
both => architecture.both

rashil2000:
> Are you sure about it? Cause it's a gui-program and can start and run well by just clicking its shortcut on my Java_SE-installed computer. There seems no need for shim to handle it.

How have you installed Java? If you used the installer, then it registers .jar files that can be run directly. But for users who install TimeSlotTracker through Scoop, it is very likely that they would have installed Java through Scoop too. And Java installed through Scoop does not register file associations (by design). So .jar files won't run on their own, in this case.

> jar file doesn't differentiate 32bits or 64 bits as they are byte code executed by JVM. Only the JVM and native libs have the differences of 32/64bits.

Ah, that makes sense. No change needed then.

> Also, am i understood correctly about manifest's architecture? only 32-bit => no need for manifest.architecture. only 64-bit => architecute.64bit both => architecture.both

Yes, correct.

zStruCat:
shortcuts already changed to bin

rashil2000:
Use this (slight modification with URL fragments):
```json
{
    "version": "1.3.23",
    "description": "TimeSlotTracker is an application to register your time.",
    "homepage": "https://github.com/TimeSlotTracker/timeslottracker-desktop",
    "suggest": {
        "JDK": [
            "java/oraclejdk",
            "java/openjdk"
        ]
    },
    "url": "https://github.com/TimeSlotTracker/timeslottracker-desktop/releases/download/v1.3.23/TimeSlotTracker-1.3.23-full.jar#/TimeSlotTracker.jar",
    "hash": "d4ab2676dec5ba56cabd032bd7ec37c2a989553bb7a50a6a41be92dda0464473",
    "bin": "TimeSlotTracker.jar",
    "checkver": "github",
    "autoupdate": {
        "url": "https://github.com/TimeSlotTracker/timeslottracker-desktop/releases/download/v$version/TimeSlotTracker-$version-full.jar#/TimeSlotTracker.jar"
    }
}
```

PR #7895, opened 2022-02-02
Fix extract_dir for FreeCad package.

thegala:
Wrong extract directory for FreeCad package.


rashil2000:
Change it in autoupdate too

PR #7947, opened 2022-02-10
eartrumpet: Add to startup via .reg file

### LINKED ISSUE #7929: EarTrumpet install UX does not match supported install channels
## Bug Report

**Package Name:** EarTrumpet

### Current Behaviour

Hi there, I'm one of the EarTrumpet owners 👋. Our users are not getting the default installation experience as our supported installation channels. In this case, the EarTrumpet app is not configured correctly to start with Windows, leading to confusion (and has resulted in an issue filed in our repository).

### Expected Behaviour

EarTrumpet to start up with Windows after Scoop installation.

### Additional context/output

https://github.com/File-New-Project/EarTrumpet/discussions/1053

### Possible Solution

See our [`ChocolateyInstall.ps1`](https://community.chocolatey.org/packages/eartrumpet#files):
```powershell
...
Write-Output "Adding shortcut to Startup"
New-ItemProperty -LiteralPath 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Run' -Name 'EarTrumpet' -Value $exePath -PropertyType String -Force -ea SilentlyContinue | Out-Null
```

issaclin32:
closes #7929

### DIFF
--- a/bucket/eartrumpet.json
+++ b/bucket/eartrumpet.json
@@ -14,10 +14,13 @@
     "post_install": [
         "$exepath = \"$dir\\EarTrumpet.exe\".Replace('\\', '\\\\')",
         "'add-startup', 'remove-startup' | ForEach-Object {",
-        "   $content = Get-Content \"$bucketsdir\\extras\\scripts\\eartrumpet\\$_.reg\"",
-        "   $content = $content.Replace('$exepath', $exepath)",
-        "   $content | Set-Content -Path \"$dir\\$_.reg\"",
-        "  }"
+        "    $content = Get-Content \"$bucketsdir\\extras\\scripts\\eartrumpet\\$_.reg\"",
+        "    $content = $content.Replace('$exepath', $exepath)",
+        "    if ($global) {",
+        "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "    }",
+        "    $content | Set-Content -Path \"$dir\\$_.reg\"",
+        "}"
     ],
     "uninstaller": {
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):
Personally, I feel that we shouldn't instruct Scoop to run the import command for uninstall/removal of registry entries. If the user manually imports them to install the .reg file, they can manually import them to uninstall it too.

rasa:
Also, the .reg file doesn't work if installed globally, as it's hardcoded for the current user. Is that intentional?

issaclin32:
> Also, the .reg file doesn't work if installed globally, as it's hardcoded for the current user. Is that intentional?

Nope. I forgot to add global option for this one.
Also we should probably check global options for other packages involving .reg (such as `grepwin`, `everything`, etc.)

issaclin32:
> Personally, I feel that we shouldn't instruct Scoop to run the import command for uninstall/removal of registry entries. If the user manually imports them to install the .reg file, they can manually import them to uninstall it too.

But sadly, I think many end-users will forget to remove the registry before uninstalling the app. Which leads to an awkward situation where the user sees **a context menu/startup that is not working**, and has no way to remove it via Scoop. (unless they reinstall the app)

riverar:
It doesn't make sense to have an option for HKLM for user-mode only app EarTrumpet.

issaclin32:
>Does Scoop install the app in a global location that is shared by all users or something?

Yes. 
If we install the app by `scoop install eartrumpet -g`, then it will be located at `C:\ProgramData\scoop\apps\eartrumpet`.

riverar:
Hm OK thanks.

PR #7963, opened 2022-02-14
Add persistent config to alacritty

### LINKED ISSUE #2430: Alacritty configuration
Alacritty on Windows automatically creates its configuration file in `%APPDATA%\alacritty\alacritty.yml`. To make it persistent and portable in a scoop fashion, this should probably be made into a persisting file.

Fortunately, alacritty has a parameter to configure the location of this file:

    --config-file <config-file>
        Specify alternative configuration file [default: $XDG_CONFIG_HOME/alacritty/alacritty.yml]

Gleydar:
Alacritty usually expects its configuration file in `%appdata%\alacritty\alacritty.yml`. Due to the portable nature of scoop, this is not an optimal solution.
With this change, we download an example configuration file and persist it, using the name `alacritty.yml.example`. This file can then be edited and changed to `alacritty.yml`. It will be picked up due to the additional program argument `--config-file <config-file>`. 

Closes #2430

- [x ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).


rashil2000:
Question: is it possible to use the `--config-file` twice in alacritty? 

If it accepts it only once, or if it ignores the value that's specified after the first instance of the flag, then users who want to use this flag externally will not be able to do so if we 'lock' it in the shim.

bucket/alacritty.json
```diff
@@ -6,18 +6,47 @@
...
+    "hash": [
+        "6253cc8b39c784f0369ab6e307a49c473eb81863df4ec7882ecfb6d7630a8811",
+        "c7c2f96113f3c2cdf0ef059cded558cc5746b3be60c6fef3b7040a3b8325c686"
+    ],
+    "bin": [
+        [
+            "alacritty.exe",
+            "alacritty.exe",
+            "--working-directory \"%USERPROFILE%\""
+        ],
```
rashil2000, line 22:
```suggestion
        "alacritty.exe",
```

This flag wasn't present before, so we should not add it.

bucket/alacritty.json
```diff
@@ -6,18 +6,47 @@
...
+        [
+            "alacritty.exe",
+            "alacritty-config.exe",
```
rashil2000, line 25:
```suggestion
            "alacritty-config",
```

Extension not needed

rashil2000 (changes requested):
You need to update autoupdate's URL field to account for yml file

PR #8015, opened 2022-02-21
task: Remove duplicated package. Use go-task in main bucket.

lyuha:
task packages is duplicate of go-task in main bucket.

Same manifest is here [bucket/main: go-task](https://github.com/ScoopInstaller/Main/blob/ca3674dd27cc7eceaff6cd30cb55f2c9ee936382/bucket/go-task.json).


rashil2000:
Could you also make a PR in Main bucket to rename the package to task.json?

lyuha:
Ok. I'm done.

ScoopInstaller/Main#3315

PR #8026, opened 2022-02-22
git-aliases: Add version 0.3.5

sitiom:

### DIFF
--- a/bucket/git-aliases.json
+++ b/bucket/git-aliases.json
@@ -15,7 +15,7 @@
     },
     "checkver": "github",
     "autoupdate": {
-        "url": "https://github.com/dahlbyk/posh-git/archive/v$version.zip",
+        "url": "https://github.com/gluons/powershell-git-aliases/archive/v$version.zip",
         "extract_dir": "powershell-git-aliases-$version\\src"
     }
 }
--- a/bucket/git-aliases.json
+++ b/bucket/git-aliases.json
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/gluons/powershell-git-aliases",
     "license": "MIT",
     "notes": [
-        "Use the module by running: 'Import-Module git-aliases'",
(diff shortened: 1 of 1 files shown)


bucket/git-aliases.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "url": "https://github.com/gluons/powershell-git-aliases/archive/v0.3.5.zip",
+    "hash": "6f51ffcb968de60ecd434bd5074726e6a8818713cf86821782c9d1ffd39b7ef5",
+    "extract_dir": "powershell-git-aliases-0.3.5\\src",
+    "psmodule": {
+        "name": "git-aliases"
+    },
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://github.com/dahlbyk/posh-git/archive/v$version.zip",
```
niheaven, line 18: Wrong URL

niheaven:
Why not use nupkg?

https://psg-prod-eastus.azureedge.net/packages/git-aliases.0.3.5.nupkg

sitiom:
> Why not use nupkg?
> 
> https://psg-prod-eastus.azureedge.net/packages/git-aliases.0.3.5.nupkg

Direct to source if possible. I also see [`posh-git`](https://github.com/ScoopInstaller/Extras/blob/master/bucket/posh-git.json) doing it.

niheaven:
`posh-git` is created before added to PSGallery, so source is used. Use a more stable and accessible host is preferred.

rashil2000:
A `suggest` for git should be added

sitiom:
> `posh-git` is created before added to PSGallery, so source is used. Use a more stable and accessible host is preferred.

I don't think so, posh-git has been in PSGallery since the beginning.

sitiom:
Switched to using PSGallery.

niheaven (approved):

Issue #8045, opened 2022-02-25
[Request] IMO (note- MSI installer only)


SonaliBendre:
### Information

**Name:** [name of new package]
IMO
**Description:** [clear and concise details of what it is]

**Homepage:** [a URL/link]
https://imo.im/
**Download link(s):** [URL(s)/link(s)]
imo free video calls and chat
**Some indication of popularity/repute:** [GitHub stars/software reviews etc.]
![image](https://user-images.githubusercontent.com/73757361/155697223-675d9fb4-5709-4514-a00c-f1c49d5e2f3b.png)
 More than 200M users

I have no idea about MSI Installers so can't make PR

rashil2000:
You can search in the repo, there are plenty of examples which use MSI

SonaliBendre:
https://github.com/Homebrew/formulae.brew.sh/blob/master/api/cask-source/imo.rb

SonaliBendre:
> You can search in the repo, there are plenty of examples which use MSI

I saw some manifests and also check scoop repo what it does in case of MSI apps but maybe my brain too small to understand

SonaliBendre:
https://github.com/ScoopInstaller/Extras/pull/8046

PR #8047, opened 2022-02-25
Add imo Messanger

SonaliBendre:
Closes https://github.com/ScoopInstaller/Extras/issues/8045

### DIFF
--- a/bucket/imo.json
+++ b/bucket/imo.json
@@ -1,0 +1,22 @@
+{
+    "version": "1.2.29.0",
+    "description": "Imo is a free, simple, and faster international video call & instant messaging app. (With tons of ads ofcourse)",
+    "homepage": "https://imo.im/",
+    "license": "Proprietary",
+    "url": "https://imo.im/winapp/current/ImoSetup.msi",
+    "hash": "sha512:abda3c2d5dfcae3421cbb9677945e8d5e2f648e44d93d7ab66f0fc6bad3640797c1909984efe94050bba602d28ce00ade8b465fc971d980ab907fd9524f6c343",
+    "extract_dir": "Imo Messenger",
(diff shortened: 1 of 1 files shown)


SonaliBendre:
There no version number in https://imo.im/winapp/current/ImoSetup.msi how to check version? If they release new version, hash willn't match
should we persist Logs?

bucket/imo.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "1.2.29.0",
+    "description": "Imo is a free, simple, and faster international video call & instant messaging app. (With tons of ads ofcourse)",
+    "homepage": "https://imo.im/",
+    "license": "Proprietary",
```
rasa, line 5: `Proprietary` is for when it costs money, and `Freeware` for when it's free to use. Which is appropriate here?

SonaliBendre: freeware

SonaliBendre: but it contains horrible ads

niheaven: Still Freeware, ads not means it is Proprietary

SonaliBendre: Sorry I thought the maintainers will change license

rashil2000:
How did you check version number for the manifest?

SonaliBendre:
I got the version number from app properties
image

rashil2000:
We need something like a website or link to check, we can't download it for checking version

SonaliBendre:
> We need something like a website or link to check, we can't download it for checking version

officially they don't provide or maybe its hidden. we have to create an unofficial api. what do you do in case of these apps?

rashil2000:
> what do you do in case of these apps?

We don't accept these apps

PR #8052, opened 2022-02-26
goldendict@1.5.0-RC2-372-gc3ff15f: make goldendict portable

zStruCat:
According to goldendict>help>03_Portable_Mode, goldendict can be made portable if a folder named portable is under its dirertory, in which case the dictionaries must be stored within the program.

### DIFF
--- a/bucket/goldendict.json
+++ b/bucket/goldendict.json
@@ -13,6 +13,10 @@
             "GoldenDict"
         ]
     ],
+    "persist": [
+        "portable",
+        "content"
+    ],
     "checkver": {
         "url": "https://sourceforge.net/p/goldendict/activity/feed",
         "regex": "GoldenDict-([\\w.-]+)_\\(QT_486\\)\\.7z"


rashil2000:
If we are making this portable, we should do a data migration for people who had saved user data in the non-portable location. See https://github.com/ScoopInstaller/Extras/blob/e766b6346213105dadf2e1e44f506fd3bc30cf80/bucket/vscode.json#L44-L51 for example.

zStruCat:
> If we are making this portable, we should do a data migration for people who had saved user data in the non-portable location. See
> 
> https://github.com/ScoopInstaller/Extras/blob/e766b6346213105dadf2e1e44f506fd3bc30cf80/bucket/vscode.json#L44-L51
> for example.
Non-portable GoldenDict user have to specify the location(s) of the directory for dict files, and the GoldenDict **doesn't** copy the dict file to some default location such as %APPDATA% or %USERPROFILE%. Technically it's possible to read and edit the config (an xml file) and copy accordingly, but i guess it's would be doing too much.
Shall i start a new manifest named goldendict-portable?

rashil2000:
Oh, I thought it would be storing the files in a standard location like AppData or Documents. If that is not the case, then this PR is good to go.

PR #8058, opened 2022-02-27
Fixed Webex checkver regex to match updated website

### LINKED ISSUE #8057: Webex checkver is no longer reporting the newest version
## Bug Report

**Package Name:** Webex

### Current Behaviour

There was a recent hash check failure reported as #7994, the hash changed because a new version was released, however the checkver for Webex is returning an old version.

### Expected Behaviour

Checkver for Webex should find the latest version.

### Additional context/output

Checkver is matching some old text on the Checkver URL, the regex needs to be updated to handle the recent format change.

### Possible Solution

Update the regex to match the new format.
I have have a PR incoming for this.

MathewBlock:
Closes #8057

The Webex URL that checkver uses has been updated to change the format of how they present the version from:
`Version: xx.x.x.xxxxx` to `Windows—xx.x.x.xxxxx`.

Note, there is a complication here due to how Scopp downloads the URL and processes the regex, pattern matching on the emdash character is not possible. This is because the UTF-8 website is being interpreted as Windows-1252. So the emdash actually reports as `â€”`. For reference this is a common encoding issue which you can read about [here](https://www.i18nqa.com/debug/utf8-debug.html).

I have not updated the version or hash at this time, I'll leave that for the automated process which should work as normal from now on.

### DIFF
--- a/bucket/webex.json
+++ b/bucket/webex.json
@@ -1,5 +1,5 @@
 {
-    "version": "42.1.0.21190",
+    "version": "42.2.0.21486",
     "description": "Video and audio conferencing",
     "homepage": "https://www.webex.com/",
     "license": {
--- a/bucket/webex.json
+++ b/bucket/webex.json
@@ -25,7 +25,7 @@
     ],
     "checkver": {
         "url": "https://help.webex.com/article/mqkve8",
-        "regex": "Version:\\s([\\d.]+)"
+        "regex": "Windowsâ€”([\\d.]+)"
     },
     "autoupdate": {
         "architecture": {
--- a/bucket/webex.json
+++ b/bucket/webex.json
@@ -1,5 +1,5 @@
 {
-    "version": "42.2.0.21486",
+    "version": "42.3.0.21576",
     "description": "Video and audio conferencing",
     "homepage": "https://www.webex.com/",
     "license": {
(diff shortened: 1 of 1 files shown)


rashil2000:
I think we can update the version here, because the verifier would fail otherwise 😅

rashil2000:
See https://github.com/ScoopInstaller/Extras/runs/5348824770?check_suite_focus=true#step:3:262

MathewBlock:
I've run checkver locally and it passes, so I suspect the issue I encountered with UTF-8 being interpreted as Windows-1252 doesn't apply on GitHub actions? or perhaps it uses a different code page? I guess I could try a different encoding to see if that helps, but then it wont work locally anymore.

I wonder if there are other examples of unicode in regex within the extras bucket?

rashil2000:
GitHub Actions might be using Windows-1200. Can you temporarily set it locally and modify the checkver accordingly?

MathewBlock:
Okay, I have a fix, but it involves fixing checkver.ps1 by explicitly setting the encoding of the webclient to utf8.
I believe that without setting it explicitly it's defaulting to the local code page, which could be different on every machine, thus causing this issue. If this post on [stackoverflow ](https://stackoverflow.com/a/65269448) is to be believed then utf8 is the default for powershell 6+, so this change would be correcting that for powershell 5.

If this sounds reasonable I can pursue getting my small change into the Scoop base code, then updating this to be utf8.

I think that would be a more robust fix for this issue,

4ooc:
Replace the emdash character with [^\\d], than we could not care the character set.
Windows[^\\d]([\\d.]+)

MathewBlock:
@rashil2000 Thanks for the links, I've read through the history and while it might make good sense to switch to UTF-8 for checkver, there are certainly issues with updating other parts of the system. @4ooc raises a pragmatic point, why not just skip over the non-numbers. I tweaked the regex a little to accommodate the fact it comes out as 3 character in my codepage but the concept is the same. I also added some surrounding context to help refine the match.

Additionally, while deliberating this, Cisco released another patch, so I updated the version and hash again.

rashil2000:
No worries, thanks for the fix @MathewBlock and for the workaround @4ooc

PR #8076, opened 2022-03-03
isocreator: Add version 1.0

### LINKED ISSUE #5364: [Request] Add IsoCreator
IsoCreator C#.NET 2.0 application which creates ISO 9660 Joliet CD images from a given folder or volume. Can also create ISO with virtual files, given a tree representing a directory structure...

https://sourceforge.net/projects/iso-creator-cs/

issaclin32:
closes #5364

[ISO Creator](https://sourceforge.net/projects/iso-creator-cs/) is a simple app that creates ISO images from a given folder or volume.

**NOTES**:
* The app is built in 32-bit.
* Version number is not shown in the repo, but can be found in the application.
* *persist* is not needed because the app does not store any config by files or regsitry (tested on a fresh install VM)

### DIFF
--- a/bucket/isocreator.json
+++ b/bucket/isocreator.json
@@ -1,0 +1,14 @@
+{
+    "version": "1.0",
+    "description": "Creates ISO images from a given folder or volume.",
+    "homepage": "https://sourceforge.net/projects/iso-creator-cs/",
+    "license": "Public Domain",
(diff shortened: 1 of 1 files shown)


rashil2000:
- License is not SPDX


Unrelated, but:

> app does not store any config by registry (tested on a VM)

How does one check this?

rashil2000 (approved):
Oh okay, thanks for the info!

PR #2318, opened 2019-06-19
wsltty: move to dir before run install.bat, modify script (description, checkver)

amansheim-bm-net:
Closes #3146

PR #8006, opened 2022-02-19
Update WSLTTy

SonaliBendre:
Closes #XXXX

Relates to #XXXX

- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #8079, opened 2022-03-03
wsltty: Fix installer

issaclin32:
closes #2318

Related:
#8006
https://github.com/mintty/wsltty/issues/304


rashil2000:
Hmm, as noted in the linked issue, we should wait for the new release which includes the patched installer.

PR #8108, opened 2022-03-06
teamviewer@15.25.8: update version string

### LINKED ISSUE #8097: Update Teamviewer version
The issue is very simple. 

The version of TeamViewer that is distributed through scoop is downloaded from https://dl.teamviewer.com/download/version_15x/TeamViewerPortable.zip. As you can see, this zip file is not linked to a specific version: therefore if a new version of TeamViewer is released, the developer will update the zip file, but scoop will stay behind. 

This is happening nowadays: the version in scoop is `15.25.8.0` but the latest version reported on the official website is `15.27.3.0` (indeed, if you download the zip file and right click on the `TeamViewer.exe` file you will see this version).

The proposed change is simply to update the version in the json file corresponding to TeamViewer in the `extras `repository.

I can do the PR if you are ok with this.

maephisto666:
The version string was updated. The hash was right.
If the version string is old and you are already on that version, the update will not be triggered.

Closes #8097 

### DIFF
--- a/bucket/teamviewer.json
+++ b/bucket/teamviewer.json
@@ -1,5 +1,5 @@
 {
-    "version": "15.25.8",
+    "version": "15.27.3",
     "description": "Remote control, desktop sharing, online meetings, web conferencing and file transfer between computers",
     "homepage": "https://www.teamviewer.com",
     "license": {
--- a/bucket/teamviewer.json
+++ b/bucket/teamviewer.json
@@ -17,8 +17,8 @@
     ],
     "persist": "teamviewer.ini",
     "checkver": {
-        "url": "https://www.teamviewer.com/en/download/windows/",
-        "regex": "<p>Current version:\\s*([\\d.]+)\\s*<\\/p>"
+        "url": "https://community.teamviewer.com/English/categories/change-logs-en",
+        "regex": "\\[Windows\\] v([\\d.]+)"
     },
     "autoupdate": {
         "url": "https://download.teamviewer.com/download/version_$majorVersionx/TeamViewerPortable.zip"
--- a/bucket/teamviewer.json
+++ b/bucket/teamviewer.json
@@ -18,7 +18,7 @@
     "persist": "teamviewer.ini",
     "checkver": {
         "url": "https://www.teamviewer.com/en/download/windows/",
-        "regex": "Current version:\\s*([\\d.]+)\\s*</"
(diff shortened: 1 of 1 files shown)


rashil2000:
image

The version is 15.25.8

maephisto666:
> image
> 
> The version is 15.25.8

Thank you @rashil2000 for sharing the screenshot. Now things are a bit clearer.

This is what I see when I go to https://www.teamviewer.com/en/download/windows/. I am connected from Europe.

image

But I also did another test. I opened the link using a VPN connection from United States and guess what? The version was still the old one (like you highlighted). 

Funny thing though, as I mentioned in the first comment the hash was already right: I guess that means that maybe someone had my exact issue, they just updated the hash without changing the version (otherwise the `checkver `check would have failed like in my case).

Based on your experience @rashil2000, is there a way to manage the different versions of the websites available from different world regions?

rashil2000:
> Based on your experience @rashil2000, is there a way to manage the different versions of the websites available from different world regions?

Not currently. 

What is the app version shown inside the TeamViewer application? Is it different when you download using the USA VPN?

maephisto666:
> > Based on your experience @rashil2000, is there a way to manage the different versions of the websites available from different world regions?
> 
> Not currently.
> 
> What is the app version shown inside the TeamViewer application? Is it different when you download using the USA VPN?

So through my usual VPN (NordVPN) I get an error when I try to ultimately download the zip file (it is a Cloudflare error, so I believe they are allowing you to browse the website through a VPN but not to download the zip file through it). Then I tried a different free service (HideMyAss VPN) and in the picture you can see the package contains the updated version but the website still mentions the old version.

image


I can also try to re-install TeamViewer through scoop. What I do expect is that the folder created under `scoop/apps` will be called `15.25.8` but when I open the About information panel within the app I would see the newer version `15.27.3`. And I expect this because the hash is the one from the newer zip file.

rashil2000:
How about we use the changelog for getting the latest version? https://community.teamviewer.com/English/categories/change-logs-en

So it will become something like:
```json
"checkver": {
    "url": "https://community.teamviewer.com/English/categories/change-logs-en",
    "regex": "\\[Windows\\] v([\\d.]+)"
}
```

maephisto666:
Hey @rashil2000 that sounds like a good idea. I did not even know about that page. The regex will match the first occurrence and since everything is sorted from most recent to less recent it should always pick the latest version for Windows. 

Let me see my PR...

PR #8109, opened 2022-03-06
BatteryCare: Add version 0.9.36

fauzanelka:

Add BatteryCare v0.9.36 app manifest

### DIFF
--- a/bucket/battery-care.json
+++ b/bucket/battery-care.json
@@ -1,6 +1,6 @@
 {
     "version": "0.9.36",
-    "description": "BatteryCare is a software created to optimize the usage and performance of the modern laptop's battery. ",
+    "description": "Tool to optimize the usage and performance of the modern laptop's battery. ",
     "license": "Freeware",
     "homepage": "https://batterycare.net/",
     "url": "https://batterycare.net/files/BatteryCarePortable.zip",
--- a/bucket/battery-care.json
+++ b/bucket/battery-care.json
@@ -1,0 +1,18 @@
+{
+    "version": "0.9.36",
+    "description": "BatteryCare is a software created to optimize the usage and performance of the modern laptop's battery. ",
+    "license": "Freeware",
+    "homepage": "https://batterycare.net/",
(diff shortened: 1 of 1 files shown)


bucket/battery-care.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "0.9.36",
+    "description": "BatteryCare is a software created to optimize the usage and performance of the modern laptop's battery. ",
```
rashil2000, line 3:
```suggestion
    "description": "Tool to optimize the usage and performance of the modern laptop's battery. ",
```

rashil2000:
Add autoupdate field (it is required). 

Where is the user data/config stored?

fauzanelka:
> Add autoupdate field (it is required).
> 
> Where is the user data/config stored?

I forgot that, I see that BatteryCare loads user data in AppData directory. How do you set that?

bucket/battery-care.json
fauzanelka: Looks good, okay.

rashil2000:
If it is stored in AppData, then no change required. Although portable applications shouldn't really store in AppData.

fauzanelka:
What about the hash of the new version? The official URL only has one and it's the latest version.

rashil2000:
If hash is not available, we can skip putting it in autoupdate.

fauzanelka:
LGTM!

bucket/battery-care.json
```diff
@@ -0,0 +1,18 @@
...
+    "shortcuts": [
+        [
+            "BatteryCare.exe",
+            "BatteryCare"
+        ]
+    ],
+    "checkver": {
+        "url": "https://batterycare.net/en/download.html",
+        "regex": "Version ([\\d.]+)"
+    }
```
rashil2000, line 17:
```suggestion
    },
    "autoupdate": {
        "url": "https://batterycare.net/files/BatteryCarePortable.zip"
    }
```

PR #8124, opened 2022-03-09
(chore): Migrate CI to GitHub Actions

rashil2000:
Relates to https://github.com/ScoopInstaller/Main/pull/3361


rashil2000:
Pinging @r15ch13 to disable the AppVeyor pipeline

niheaven:
`#Requires -Modules @{ ModuleName = 'Pester'; MaximumVersion = '4.99' }` should be added to `test.ps1`

rashil2000:
It wasn't present before though. Why is it necessary?

niheaven:
As my previous comment, all scoop tests don't support Pester 5 which is preinstalled in GitHub's Windows image. If we don't add this line, GitHub Actions will try to use Pester 5 even if `init.ps1` in scoop core installed version 4. Appveyor doesn't has Pester installed so it will use version 4 that installed by `init.ps1`.

rashil2000:
Oh, didn't know that. Added it 👍

bin/test.ps1
```diff
@@ -1,2 +1,5 @@
+#Requires -Modules @{ ModuleName = 'Pester'; MaximumVersion = '4.99' }
+
 if(!$env:SCOOP_HOME) { $env:SCOOP_HOME = Resolve-Path (scoop prefix scoop) }
```
niheaven, line 3:
```suggestion
if(!$env:SCOOP_HOME) { $env:SCOOP_HOME = Convert-Path (scoop prefix scoop) }
```

niheaven: > Resolves the wildcard characters in a path, and displays the path contents.

See [Resolve-Path](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/resolve-path?view=powershell-5.1)

For most usage, we should use `Convert-Path` which just `Converts a path from a PowerShell path to a PowerShell provider path.` unless using `Resolve-Path -Relative`.

niheaven (changes requested):

bin/test.ps1
niheaven: Forgotten to change them in `Main` bucket...

niheaven (approved):

PR #8125, opened 2022-03-09
Fix DBeaver version check.

chrismeller:
With the release of version 22.0 on the 7th, the version check was failing. While the download page lists it as "22.0", the actual version used in file and checksum names was "22.0.0". Switching to use their Github releases page instead gets us the correct result.


chrismeller:
Thanks, @rashil2000!

PR #8130, opened 2022-03-09
amazon-workspaces@4.0.6.2415: update version and fix checkver logic

FilipDeVos:
The amazon-workspaces version check was validating the version by pinging the [winsparkle](https://github.com/vslavik/winsparkle) appcast config file for the x86 version. The 32-bit version is no longer updated by Amazon, so to make scoop work properly, the url had to be updated to point to the x64 version.

The [documentation](https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-ag.pdf#%5B%7B%22num%22%3A183%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C72%2C152.414%2Cnull%5D) references the x86 url only so I had to decompile the `workspaces.dll` file to find the correct filename. 

``` csharp
public string GetAppcastFileName()
{
	if (Environment.Is64BitOperatingSystem)
	{
		return "WorkSpacesAppCastx64.xml";
	}
	return "WorkSpacesAppCast.xml";
}
```
The download url is composed out of:
`https://s3.amazonaws.com/workspaces-client-updates/<environment>/<region>/<platform>/WorkSpacesAppCastx64.xml`

Combining this information makes the new url for 64-bit only: 
https://s3.amazonaws.com/workspaces-client-updates/prod/global/windows/WorkSpacesAppCastx64.xml

For reference, the download url for the outdated 32-bit version of amazon workspaces client is:
https://d2td7dqidlhjx7.cloudfront.net/prod/pdt/windows/x86/Amazon+WorkSpaces.msi

### DIFF
--- a/bucket/amazon-workspaces.json
+++ b/bucket/amazon-workspaces.json
@@ -30,6 +30,10 @@
         "regex": "title>Version ([\\d.]+)</"
     },
     "autoupdate": {
-        "url": "https://d2td7dqidlhjx7.cloudfront.net/prod/global/windows/Amazon+WorkSpaces.msi"
+        "architecture": {
+            "64bit": {
+                "url": "https://d2td7dqidlhjx7.cloudfront.net/prod/global/windows/Amazon+WorkSpaces.msi"
+            }
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


rashil2000:
Move the url and hash inside architecture block too

PR #8145, opened 2022-03-12
godot: Update shortcut

sitiom:
Better naming

### DIFF
--- a/bucket/godot.json
+++ b/bucket/godot.json
@@ -18,7 +18,7 @@
     "shortcuts": [
         [
             "godot.exe",
-            "Godot"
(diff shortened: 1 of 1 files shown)


rashil2000:
Why is this needed? 

What is the name of the shortcut when Godot is installed using their installer (non-scoop) way?

sitiom:
> Why is this needed?
> 
> What is the name of the shortcut when Godot is installed using their installer (non-scoop) way?

From their [unofficial installer](https://hugo.pro/projects/godot-builds/) and from Linux, the name is `Godot Engine`, so the name is matching.

PR #8150, opened 2022-03-13
appium: update to version 1.22.2, add autoupdate

tech189:
- Updated to latest version
- Added autoupdate
- Update shortcut to new name
- No hash available


rashil2000:
Why not use the zip file?

tech189:
The zip file only contains the 64bit version, while the exe contains both 32bit and 64bit versions!

PR #8154, opened 2022-03-14
discord: Update to version 1.0.9004

### LINKED ISSUE #8051: Update discord version
The discord version we are using from scoop is very very old.
discord now don't update nupkg so either we will have have extract DiscordSetup.exe or use https://github.com/portapps/discord-portable . I think we should use https://github.com/portapps/discord-portable because its portable and stores data in scoop\apps\discord\data folder instead %appdata%\discord (the electron shit) and also its [configurable](https://portapps.io/doc/configuration/)

MathewBlock:
Closes #8051

I used Fiddler to capture network traffic to see what URL the updater was pinging to get the new version. The old checkver URL still exists, but has become out of date. From downloading the installer from the website manually I was able to find the new download URL scheme, then substitute in the correct version.

Checkver now uses jsonpath to grab the updater's URL then regex to extract the version from there. There is a version field array within the JSON I wanted to use but I couldn't figure out whether jsonpath lets me concat 3 elements to build up the version, so instead I went with regex parsing the URL. Note that URLs uses in this json are for the updater, we don't use them.

The install now requires two unzip operations, the first unpacks the DiscordSetup.exe file to reveal the inner .nupkg. From there I use an installer script to unpack that package and move the payload we need into the correct location (I used blisk.json as a reference, please correct me if there is a better way).

I also re-ordered post_install to match the master order specified in the contribution guide.

### DIFF
--- a/bucket/discord.json
+++ b/bucket/discord.json
@@ -1,5 +1,5 @@
 {
-    "version": "0.0.311",
+    "version": "1.0.9004",
     "description": "Free Voice and Text Chat",
     "homepage": "https://discordapp.com/",
     "license": {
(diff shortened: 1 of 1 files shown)


MathewBlock:
Apologies, I've found some issues, this will need some more work before it is ready.

rashil2000:
Kindly mark it as draft for the time being

MathewBlock:
This is currently blocked because if you run `disable-auto-update.ps1` as instructed, Discord won't load. I could remove it, but then Discord will update itself whenever it likes. Also, anyone who has previously run that script will be in the broken state, they would need to reverse the change that script does to get a working application again.

There are some potential solutions in the linked issue #8051 but they require more investigation.

MathewBlock:
I'm going to abandon this effort in favour of just using the portable app version from GitHub (separate PR incoming).

PR #8171, opened 2022-03-17
quassel: Add version 0.14.0

SonaliBendre:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/quassel.json
+++ b/bucket/quassel.json
@@ -43,6 +43,7 @@
             "-c $dir\\config"
         ]
     ],
+    "persist": "config",
     "checkver": {
         "github": "https://github.com/quassel/quassel"
     },
--- a/bucket/quassel.json
+++ b/bucket/quassel.json
@@ -1,0 +1,49 @@
+{
+    "description": "Modern, cross-platform, distributed IRC client based on the Qt framework.",
+    "homepage": "https://quassel-irc.org",
+    "license": "GPL-2.0|GPL-3.0",
+    "version": "0.14.0",
+    "url": "https://quassel-irc.org/pub/quassel-windows-0.14.0.7z",
+    "hash": "sha256:5dbe003cd92e09aa243206ab09cc9193e49995708e275365091372b18299b2fb",
+    "autoupdate": {
+        "url": "https://quassel-irc.org/pub/quassel-windows-$version.7z"
+    },
+    "checkver": {
+        "github": "https://github.com/quassel/quassel"
+    },
+    "bin": [
+        [
+            "quassel.exe",
+            "quassel.exe",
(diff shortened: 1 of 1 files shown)


rashil2000:
- Order of fields not followed
- Outdated SPDX License identifier
- Is this a 32bit or 64bit app?
- Do all the binaries support commandline arguments?

bucket/quassel.json
```diff
@@ -0,0 +1,49 @@
+{
+    "description": "Modern, cross-platform, distributed IRC client based on the Qt framework.",
+    "homepage": "https://quassel-irc.org",
+    "license": "GPL-2.0|GPL-3.0",
```
issaclin32, line 4:
```suggestion
    "license": "GPL-2.0-or-later|GPL-3.0-or-later",
```
Check **"Identifier"** at https://spdx.org/licenses/

https://github.com/quassel/quassel/blob/master/gpl-3.0.txt has `or any later versions` in it, therefore it is `GPL-3.0-or-later`.

bucket/quassel.json
```diff
@@ -0,0 +1,49 @@
...
+    "version": "0.14.0",
+    "url": "https://quassel-irc.org/pub/quassel-windows-0.14.0.7z",
```
issaclin32, line 6: You can check the app is 32 or 64-bit using pelook (`scoop install pelook`)
If the app is 32-bit, then keep as it is (both 32-bit and 64-bit machine can run it)
If the app is 64-bit. then use
```
"architecture" {
    "64bit":  {
        "url": ...
        "hash": ...
    }
```

issaclin32:
`bin` is only needed if  (1) The app is supposed to run under command-line, or (2) The app supports command-line arguments.

You can either check [readme](https://github.com/quassel/quassel/blob/master/README.md) or read the source code on main program to find it out. In the case of *quassel*, there are some command-line arguments.
```
Once you've set up Quassel, you may:

Add more users: quasselcore --add-user
Change the password of an existing user: quasselcore --change-userpass=username
See all available options: quasselcore --help
```

issaclin32:
I updated the manifest because the author has not responded for a while.

**NOTES**:
* The app is built in 64-bit.
* **quassel** supports command-line args. See [ReadMe](https://github.com/quassel/quassel/blob/master/README.md) for example.
* *persist* is not needed because config is at `$Env:AppData\quassel-irc.org`. I tried tweaking some settings and creating profiles/connections, but `$dir\config` does not appear.

rashil2000 (changes requested):
We should wrap the `$dir\\config` part in double quotes to account for spaces in path

rashil2000 (approved):

PR #8195, opened 2022-03-23
ani-cli@1.9: Suggest mpv instead of depending

### LINKED ISSUE #8194: ani-cli installs mpv even if mpv-git is installed
## Bug Report

**Package Name:** ani-cli

### Current Behaviour

installing ani-cli installs mpv but i have mpv-git which is already configured and is available as mpv in PATH

### Expected Behaviour

mpv-git gets accepted as a dependency as it is almost identical to mpv

### Additional context/output

[line 10 in ani-cli.json having mpv as a dependency](https://github.com/ScoopInstaller/Extras/blob/master/bucket/ani-cli.json#L10)

### Possible Solution

i'm not knowledgeable on scoop packages so i'm not entirely sure if that is possible but basically checking if either mpv or mpv-git and depending on which ever is already available

if that is not possible maybe it can not have mpv as a dependency at all and add a simple note to install mpv (or mpv-git)

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.2

**Additional software:**

#### Scoop Configuration

```json
{
  "lastupdate": "2022-03-23T06:52:51.5882149+02:00",
  "SCOOP_REPO": "https://github.com/lukesampson/scoop",
  "SCOOP_BRANCH": "master",
  "aria2-warning-enabled": false
}
```

x6r:
Closes #8194

### DIFF
--- a/bucket/ani-cli.json
+++ b/bucket/ani-cli.json
@@ -7,7 +7,10 @@
     "hash": "31cf0d80e2277e5da9f5e04147bc51f48e393855b0ba1159b96b4bd0dbd3308f",
     "depends": "aria2",
     "suggest": {
-        "mpv": "mpv"
+        "mpv": [
+            "mpv",
+            "mpv-git"
+        ]
     },
     "bin": "ani-cli",
     "checkver": "github",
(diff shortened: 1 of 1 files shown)


bucket/ani-cli.json
```diff
@@ -5,10 +5,10 @@
     "license": "GPL-3.0-or-later",
     "url": "https://github.com/pystardust/ani-cli/releases/download/v1.9/ani-cli",
     "hash": "31cf0d80e2277e5da9f5e04147bc51f48e393855b0ba1159b96b4bd0dbd3308f",
-    "depends": [
-        "aria2",
-        "mpv"
-    ],
+    "depends": "aria2",
+    "suggest": {
+        "mpv": "mpv"
```
rashil2000, line 10:
```suggestion
        "mpv": [
            "mpv",
            "mpv-git"
        ]
```

PR #8210, opened 2022-03-26
coretemp: Fix persist, Add language packs

### LINKED ISSUE #5083: Coretemp won't boot
All that happens when I start Coretemp is that a black square shows up in the tray.

When I hover over it it disappears since the app just apparently crashes on startup.

I check back from time to time and install the software through Scoop, but the issue sadly remains to this day.

Thanks in advance.

issaclin32:
fixes #5083

* **CoreTemp** will crash on startup if `coretemp.ini` is empty. Therefore we need to manually persist it.
* Includes **language packs** in the package. ~I think it won't hurt to download the language packs with the main program, as they are relatively small in size.~ Use an additional script to download the language pack.


rashil2000:
To avoid making the manifest large, maybe we can provide a helper script for users to download the language pack themselves (and persist it, if required)

bucket/coretemp.json
```diff
@@ -6,24 +6,38 @@
...
+    "notes": "Download language packs by running: $dir\\download-language-packs",
     "architecture": {
         "64bit": {
             "url": "https://www.alcpu.com/CoreTemp/CoreTemp64.zip",
             "hash": "1287dca98adecd6e8e27272f96f206ad9d60fb84b61fe0a353865ee81bd1a088"
         },
         "32bit": {
             "url": "https://www.alcpu.com/CoreTemp/CoreTemp32.zip",
-            "hash": "5b71c1109d8cacc3763ba4dd470d6ec5f426c69a4b675121d1557e9b54e2c0db"
+            "hash":"5b71c1109d8cacc3763ba4dd470d6ec5f426c69a4b675121d1557e9b54e2c0db"
```
rashil2000, line 17:
```suggestion
            "hash": "5b71c1109d8cacc3763ba4dd470d6ec5f426c69a4b675121d1557e9b54e2c0db"
```

scripts/coretemp/download-language-packs.cmd
```diff
@@ -0,0 +1 @@
+powershell %~dp0download-language-packs.ps1
```
rashil2000, line 1: I think just the ps1 file should be sufficient, no? We don't usually have Cmd scripts in the repo

scripts/coretemp/download-language-packs.ps1
```diff
@@ -0,0 +1,42 @@
+$language_pack_urls = @(
```
rashil2000, line 1: Slight chance of optimization here: making an array of just the `zh-CN` part, the rest of the URL is identical for all entries.

scripts/coretemp/download-language-packs.cmd
rashil2000: We could update the note to `Download language packs by running: $dir\\download-language-packs in PowerShell`

issaclin32: How about this:
`Run $dir\download-language-packs.ps1 (powershell $dir\download-language-packs.ps1 under cmd)`

rashil2000: Yup, seems good

rashil2000 (approved):

PR #8243, opened 2022-03-31
flat-assembler: Update to version 1.73.30, add license URL

tech189:
Relates to Excavator issue 'URL is not valid', checkver works okay locally... https://github.com/ScoopInstaller/Extras/runs/5777899107?check_suite_focus=true#step:3:230

### DIFF
--- a/bucket/flat-assembler.json
+++ b/bucket/flat-assembler.json
@@ -1,10 +1,13 @@
 {
-    "version": "1.73.29",
+    "version": "1.73.30",
     "description": "Open source assembly language compiler",
     "homepage": "https://flatassembler.net/",
-    "license": "Unknown",
-    "url": "https://flatassembler.net/fasmw17329.zip",
-    "hash": "6995d164069dd45d45a8228440935ef38e444a35d700af7a339c67b3d5c67d23",
+    "license": {
+        "identifier": "Unknown",
+        "url": "https://github.com/tgrysztar/fasm/blob/master/LICENSE.TXT"
+    },
+    "url": "https://flatassembler.net/fasmw17330.zip",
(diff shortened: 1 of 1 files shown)


bucket/flat-assembler.json
```diff
@@ -1,10 +1,13 @@
 {
-    "version": "1.73.29",
+    "version": "1.73.30",
     "description": "Open source assembly language compiler",
     "homepage": "https://flatassembler.net/",
-    "license": "Unknown",
-    "url": "https://flatassembler.net/fasmw17329.zip",
-    "hash": "6995d164069dd45d45a8228440935ef38e444a35d700af7a339c67b3d5c67d23",
+    "license": {
+        "identifier": "Unknown",
```
issaclin32, line 6:
```suggestion
        "identifier": "Freeware",
```
I think the license is marked as **unknown** because "the author of the manifest" could not find the license document. It is actually **freeware**.

```
This program is free for commercial and non-commercial use as long as
the following conditions are adhered to.
```

tech189:
The workflow says there is a problem with the site, an SSL issue, but the logs don't say what the 'inner exception' is.

```
INFO: Hashes

OperationStopped: C:\Users\runneradmin\SCOOP\apps\scoop\current\lib\install.ps1:142
Line |
 142 |          throw $e
     |          ~~~~~~~~
     | The SSL connection could not be established, see inner exception.
```

tech189:
Tracked it down to this line - is there a way to get it to output the full exception/traceback?

https://github.com/ScoopInstaller/Scoop/blob/04b30de9dde613acec8bb0014b42f381f83e2ec9/lib/install.ps1#L142

rashil2000:
We could just check the entire error object ($_), but for that we need to be able to reproduce the error locally.

rashil2000:
Verified locally, I think we should just merge it

PR #8245, opened 2022-04-01
Update Windows Terminal to v1.12.10732

hmalphettes:
This update to Windows Terminal has a fix for the issue "Element not found" affecting a number of scoop users:
https://github.com/microsoft/terminal/issues/12305

The download URL scheme has changed.
Note that the release notes specify that the Win10 flavor of the URL will work just as well for Win11:

https://github.com/microsoft/terminal/releases/tag/v1.12.10732.0:
> If you install the Windows 10 verison on Windows 11, it will probably automatically upgrade itself to the Windows 11
version. It turns out that it is impossible to have two bundles with the same version number, so it has to be this
way.

### DIFF
--- a/bucket/windows-terminal.json
+++ b/bucket/windows-terminal.json
@@ -1,13 +1,13 @@
 {
-    "version": "1.12.10393.0",
+    "version": "1.12.10732.0",
     "description": "The new Windows Terminal, and the original Windows console host - all in the same place!",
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

hmalphettes:
Thanks a lot to the maintainers for all your care.

I double confirm that this works on Windows 10 and on Windows 11.

PR #8255, opened 2022-04-02
psreadline: add version 2.2.2

astelmachonak-nydig:
Adds PSReadLine PowerShell module

### DIFF
--- a/bucket/psreadline.json
+++ b/bucket/psreadline.json
@@ -1,0 +1,19 @@
+{
+    "version": "2.2.2",
+    "description": "A bash inspired readline implementation for PowerShell",
+    "homepage": "https://github.com/PowerShell/PSReadLine",
+    "license": "BSD-2-Clause",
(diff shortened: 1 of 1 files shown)


bucket/psreadline.json
```diff
@@ -0,0 +1,23 @@
+{
+    "version": "2.2.2",
+    "description": "A bash inspired readline implementation for PowerShell",
+    "homepage": "https://github.com/PowerShell/PSReadLine",
+    "license": "BSD-2-Clause",
+    "notes": [
+        "Use the module by running: 'Import-Module PSReadLine'",
```
rashil2000, line 7: AFAIK, the highest version of PSReadline is imported automatically, if found in PSModulePath. So this might not be needed.

astelmachonak-nydig: You are right - just validated and it works without explicit import. Fixed.

PR #8263, opened 2022-04-04
sk-ssh-agent: Add version 1.4

astelmachonak-nydig:
SK SSH Agent is pretty new, but very useful application for Windows. It works as SSH Agent for both putty and openssh and supports  FIDO/U2F security keys for authentication (along with regular keys). No other application on windows can do this (ssh-agent from openssh on Windows 10 works, but require to be started as service, so not usable without admin permissions).

### DIFF
--- a/bucket/sk-ssh-agent.json
+++ b/bucket/sk-ssh-agent.json
@@ -1,0 +1,15 @@
+{
+    "version": "1.4",
+    "description": "An SSH agent that supports using FIDO/U2F security keys with PuTTY and OpenSSH for Windows.",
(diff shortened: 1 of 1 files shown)


rashil2000:
Hey @astelmachonak-nydig, thanks for the PR!

Unfortunately this project is very new, and as part of our rudimentary scrutiny we wait for projects to become somewhat popular before adding them to official buckets. The reason for this is that these projects distribute binaries, and a certain level of trust is essential for typical users to run them.

astelmachonak-nydig:
@rashil2000 np, I see your point. I will host this internally meanwhile.

PR #8276, opened 2022-04-07
gephi: Update to version 0.9.3, fix autoupdate

tech189:
Relates to Excavator unable to autoupdate: https://github.com/ScoopInstaller/Extras/runs/5864968342?check_suite_focus=true#step:3:220


rashil2000:
The `-x64` might mean that it needs an architecture 64bit block now

tech189:
No it looks like it still ships 32-bit and 64-bit binaries. But I need to fix the post_install script as they now bundle the JRE, so no need to edit the conf to point to `JAVA_HOME`

tech189:
Fixed now, I have tested updating from 0.9.2 to 0.9.3 and installing from scratch.

PR #8285, opened 2022-04-09
(chore): Fix git conflicts with line endings

### LINKED ISSUE #7546: Some files cause occasionally Git conflicts on updating Scoop
## Bug Report

**Package Name:** issue does not apply to specific packages, the repository itself is affected

### Current Behaviour

Occasionally Git conflicts happen on `.reg` files and manifest `.json` files. `scoop update` throws an error due to that.
I can't understand when exactly this happens. It can be that it happens when new `.reg` files are added to the Git index.
Last time it happened to me with these two `.reg` files which were added two days ago: https://github.com/ScoopInstaller/Extras/tree/master/scripts/everything

### Expected Behaviour

No Git conflicts should happen.

### Additional context/output

Executing a `git reset --hard` didn't help to fix this issue. I had to do this:
```bash
git rm .gitattributes
git add -A
git reset --hard
```
Source: https://stackoverflow.com/a/19588910/7593710

### Possible Solution

Specify how to deal with `.reg` and other text files in `.gitattributes`.

I'm not 100% sure but this could be a solution:
```
# Set default behavior to automatically normalize line endings.
# If Git decides that the content is text, its line endings will be normalized to CRLF in the working tree on checkout.
# In the Git index/repository the files will always be stored with LF.
* text=auto eol=crlf
```

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:**
```
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      18362  1801
```

**Additional software:** not relevant

#### Scoop Configuration

```json
{
    "lastupdate":  "2021-12-16T08:50:52.5175566+01:00",
    "SCOOP_REPO":  "https://github.com/lukesampson/scoop",
    "SCOOP_BRANCH":  "master",
    "aria2-enabled":  true,
    "aria2-warning-enabled":  false
}
```

maksymgendin:
Closes #7546


tech189:
(I think you accidentally added `hwp-converter.json` to this PR branch)

.editorconfig
```diff
@@ -15,6 +15,6 @@ trim_trailing_whitespace = true
 [*.{yml,yaml}]
 indent_size = 2
 
-[*.{[Bb][Aa][Tt],[Cc][Mm][Dd]}]
-# DOS/Win *requires* BAT/CMD files to have CRLF newlines
+[*.{[Bb][Aa][Tt],[Cc][Mm][Dd],reg,ps1}]
```
rashil2000, line 18:
```suggestion
[*.{[Bb][Aa][Tt],[Cc][Mm][Dd],[Rr][Ee][Gg],[Pp][Ss]1}]
```

maksymgendin:
> (I think you accidentally added `hwp-converter.json` to this PR branch)

Yes, it was half-consciously. I was checking and fixing the index for line endings and came across this file  which is checked in with 755 modifiers instead of 644 like other files. But I will revert this to keep the PR cleaner.

maksymgendin:
> If you need to specify these filetypes explicitly, shouldn't `.bat` and `.cmd` be listed as well? (The command processor needs these endings to interpret the files I believe.)

I'm almost sure that we don't need to specify them if newer Git versions are used. But not sure how older GIt versions behave - will the be able to auto-detect all text files? Not sure how to deal with this, need your support.

1️⃣ Remove **all** text file concretizations from `.gitattributes` and just keep `* text=auto eol=crlf`
2️⃣ Add concretizations for `.bat` and `.cmd` files to `.gitattributes`

.editorconfig
```diff
@@ -15,6 +15,6 @@ trim_trailing_whitespace = true
 [*.{yml,yaml}]
 indent_size = 2
 
-[*.{[Bb][Aa][Tt],[Cc][Mm][Dd]}]
-# DOS/Win *requires* BAT/CMD files to have CRLF newlines
+[*.{[Bb][Aa][Tt],[Cc][Mm][Dd],[Rr][Ee][Gg],[Pp][Ss]1}]
```
maksymgendin, line 18: Then a fair question would also be here - do we need to specify handling of these file endings? CRLF is already the default, see above.

rasa: In my experience, `.reg` and `.ps1` files can have LF line endings. Are we supporting older versions of Windows here? If so, how old? XP?

maksymgendin: I think Win 7 is the oldest version which is supported, right @rashil2000?

But does it matter here? We already have set CRLF for all files as the **default**. Does it make sense to additionally mention `.bat`, `.cmd`, `.reg` and `.ps1`? Seems to me like duplication...

rashil2000: The oldest version we support is 7.

If we've already put a default for all text files as CRLF, then we shouldn't be special-casing extensions after that.

maksymgendin: Alright, removed the duplication 👍

maksymgendin:
I cleaned up a bit. Hopefully `text=auto` instead of just `text` will make it work like expected 🙂

.editorconfig
```diff
@@ -14,7 +14,3 @@ trim_trailing_whitespace = true
 
 [*.{yml,yaml}]
 indent_size = 2
-
-[*.{[Bb][Aa][Tt],[Cc][Mm][Dd]}]
-# DOS/Win *requires* BAT/CMD files to have CRLF newlines
```
rashil2000, line 19: This is still required though, no? Batch scripts won't work on LF. 
(I'm aware this setting doesn't affect git checkouts).

maksymgendin: We have set CRLF as the default above, so all text files should be CRLF. As far as I understand 🙂 If we would have set LF as the default, then we would need this block.

rashil2000 (approved):

issaclin32 (approved):

PR #8299, opened 2022-04-11
Neovide@0.7.0: current version is outdated. wsl mode does not work in 0.5.0

Lite5h4dow:
Update neovide to beta 0.7.0 (WSL no longer works in version 0.5.0, 0.7.0 does work)

- [✔️] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/neovide.json
+++ b/bucket/neovide.json
@@ -1,5 +1,5 @@
 {
-    "version": "0.5.0",
+    "version": "0.7.0",
     "description": "A simple GUI for Neovim",
     "homepage": "https://github.com/Kethku/neovide",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


issaclin32:
Neovide is not updated by our bot because we only check stable releases by default. (ver 0.7.0 is pre-release). However, in this case, we should definitely use ver 0.7.0 first.

Thank you for pointing out!

issaclin32 (approved):

rashil2000:
@issaclin32 pre-release version of Neovide is already present in Versions bucket - https://github.com/ScoopInstaller/Versions/blob/master/bucket/neovide-dev.json.

tech189:
GitHub actions bot just reversed the manifest to 0.5.0 - https://github.com/ScoopInstaller/Extras/commit/de30fc4462fff5023ff9abe21b19e79f67afe545

issaclin32:
> GitHub actions bot just reversed the manifest to 0.5.0 - [de30fc4](https://github.com/ScoopInstaller/Extras/commit/de30fc4462fff5023ff9abe21b19e79f67afe545)

Thanks for reporting.

I just [disabled checkver](https://github.com/ScoopInstaller/Extras/commit/3862605c0807c1dc1f880c3ceccb5e13a368cd12) so that it stays in ver 0.7.0.

There is probably a bug in the *checkver* machanism. Will investigate later.

rashil2000:
> There is probably a bug in the checkver machanism. 

The checkver is working as intended. checkver changes the manifest to what it sees as latest, regardless of whether the change is an upgrade or a downgrade. This is to keep parity with upstream developers who sometimes recall latest releases due to bugs or security vulnerabilities.

issaclin32:
> This is to keep parity with upstream developers who sometimes recall latest releases due to bugs or security vulnerabilities.

Oh. I see.

PR #8304, opened 2022-04-11
ccleaner: Fix bin and shortcuts

DoveBoy:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/ccleaner.json
+++ b/bucket/ccleaner.json
@@ -21,22 +21,22 @@
         "64bit": {
             "bin": [
                 [
-                    "CCleaner64.exe",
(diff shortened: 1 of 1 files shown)


instinctualjealousy:
The packaging change that caused this issue also broke the persists from what I can tell- the program still expects ccleaner.ini to be alongside the executable, so persisting it to the root directory no longer works. This probably applies to all the other persisted items, but I am uncertain. I know it won't read my winapp2.ini from anywhere other than alongside executable though. Likely two hardlinks will need to be made from persist, one to x86, one to x64, if that's possible in the syntax (which I'm unfamiliar with). Might try to crack at scoop packages here at some point.

PR #8309, opened 2022-04-12
folding-at-home: Add checkver script

issaclin32:
[Folding@Home](https://foldingathome.org/?lng=en) now uses JS generated page, which makes us unable to match regex by fetching the page.

This fixes the problem by adding a checkver script.

### DIFF
--- a/bucket/folding-at-home.json
+++ b/bucket/folding-at-home.json
@@ -39,7 +39,36 @@
         "GPUs.txt",
         "config.xml"
     ],
-    "checkver": "fah-installer_([\\d.]+)_x86\\.exe",
+    "checkver": {
+        "script": [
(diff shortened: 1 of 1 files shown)


issaclin32:
The script works fine locally. Not sure why PR checks failed.
```
INFO: Checkver

Log of Output:

    GET with 0-byte payload 
    received -byte response of content type text/html 
    GET with 0-byte payload 
    received -byte response of content type text/html 
     
    Invoke-WebRequest:  
    Line | 
      18 |  $links = $(Invoke-WebRequest($url2)).Links 
         |             ~~~~~~~~~~~~~~~~~~~~~~~~ 
         |  
    404 Not Found 
     
    404 Not Found 
    nginx/1.14.2 
```

rashil2000 (approved):

PR #8311, opened 2022-04-13
dopamine: Persist settings file (Solve #4327)

### LINKED ISSUE #4327: Dopamine settings not persisted
Next time opening Dopamine after updating it always starts with the configuration screen.  
Looks like some settings aren't getting persisted

Rafeqm:
Closes #4327

### DIFF
--- a/bucket/dopamine.json
+++ b/bucket/dopamine.json
@@ -5,6 +5,7 @@
     "license": "GPL-3.0-only",
     "url": "https://www.digimezzo.com/content/software/dopamine/Dopamine%202.0.9%20(Release)%20-%20Portable.zip",
     "hash": "fd5ec9aa1c6f9a2a21244a65f9d77ef20614422c9ce6df75614ba8edb09c8b08",
+    "post_install": "if (!(Test-Path \"$persist_dir\\Dopamine\\Settings.xml\")) { New-Item \"$dir\\Dopamine\\Settings.xml\" | Out-Null }",
     "shortcuts": [
         [
             "Dopamine.exe",
--- a/bucket/dopamine.json
+++ b/bucket/dopamine.json
@@ -11,6 +11,7 @@
             "Dopamine"
         ]
     ],
+    "persist": "Dopamine\\Settings.xml",
     "checkver": {
         "url": "https://www.digimezzo.com/content/software/dopamine",
         "regex": "/Dopamine\\s+([\\d.]+)\\s+\\(Release\\)",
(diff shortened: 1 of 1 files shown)


bucket/dopamine.json
```diff
@@ -11,6 +11,7 @@
             "Dopamine"
         ]
     ],
+    "persist": "Dopamine\\Settings.xml",
```
rashil2000, line 14: On a fresh installation, does this file already exist? Or is it created later?

Rafeqm: Could be both. I don't know cause I haven't intalled it yet. If the file is created later, can scoop create that file? Of course, the file content will be edited by the app itself, or in worse scenario the file will be tampered with the new settings applied.
The first time I know scoop can persist _file_ (not just folder) is after installing Jetbrains IDEs via scoop. It also the first time I installed those IDEs.

rashil2000: If the file doesn't exist on a fresh installation, Scoop will assume the item is a _folder_ and create a folder hard link. To create the file, you can do something like this: https://github.com/ScoopInstaller/Extras/blob/943b50ec4541a630533d695a25edf1f54e4d65f3/bucket/q-dir.json#L39

Rafeqm: What is that `start.qdr`?

bucket/dopamine.json
```diff
@@ -5,6 +5,7 @@
     "license": "GPL-3.0-only",
     "url": "https://www.digimezzo.com/content/software/dopamine/Dopamine%202.0.9%20(Release)%20-%20Portable.zip",
     "hash": "fd5ec9aa1c6f9a2a21244a65f9d77ef20614422c9ce6df75614ba8edb09c8b08",
+    "post_install": "if (!(Test-Path \"$persist_dir\\Dopamine\\Settings.xml\")) { New-Item \"$dir\\Dopamine\\Settings.xml\" | Out-Null }",
```
rashil2000, line 8:
```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\Dopamine\\Settings.xml\")) { New-Item \"$dir\\Dopamine\\Settings.xml\" | Out-Null }",
```

Rafeqm: Umm isn't it got checked after installation?

rashil2000: Hard link is created before post_install. So we need to create the file before the hard link creation.

PR #8424, opened 2022-05-04
uivonim: Add version 0.29.0

issaclin32:
[Uivonim](https://github.com/smolck/uivonim) is a maintained fork of **Veonim** (`extras/veonim`, deprecated), a GUI frontend for Neovim.

**NOTES**:
* *persist* is not needed because config is at `$Env:AppData\uivonim`.


rashil2000:
We should add a bin field (it supports args)

rashil2000 (approved):

PR #8442, opened 2022-05-09
ivpn-client: Add version 3.8.7

### LINKED ISSUE #7912: [Request] ivpn
Package Request Information
Name: ivpn

Description: Cheap, open source and reliable VPN from gibraltar

Homepage: https://www.ivpn.net/apps-windows/

Download link(s): https://repo.ivpn.net/windows/bin/IVPN-Client-v3.5.2.exe

Some indication of popularity/repute: Github  83 stars

issaclin32:
closes #7912

[IVPN Client](https://www.ivpn.net/apps-windows/) is an open-source VPN software.

**NOTES**:
* The installation of *IVPN* involves (1) installing TAP driver, (2) modifyng registry and (3) installing service. It is possible to do that via *pre_install*, but it would be hard to maintain. A similar case can be found in *OpenVPN* (`extras/openvpn`).
* See [NSIS script](https://github.com/ivpn/desktop-app/blob/master/ui/References/Windows/Installer/IVPN%20Client.nsi) for details about the installation.
* `IVPN Service.exe` will only work if it is under `$Env:ProgramFiles`. Therefore we should not use *INSTALLDIR* switch here.

### DIFF
--- a/bucket/ivpn-client.json
+++ b/bucket/ivpn-client.json
@@ -1,0 +1,38 @@
+{
+    "version": "3.8.7",
+    "description": "Open-source VPN software",
+    "homepage": "https://www.ivpn.net/apps-windows/",
+    "license": "GPL-3.0-only",
+    "architecture": {
+        "64bit": {
+            "url": "https://repo.ivpn.net/windows/bin/IVPN-Client-v3.8.7.exe#/setup.exe",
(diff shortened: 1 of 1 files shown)


issaclin32:
> Since this does not use the Scoop directory for installation, I'd recommend not adding it here. non-portable bucket would be a better fit.

Okay. I will try to add it to non-portable bucket later.

PR #8446, opened 2022-05-09
cdex: Add version 2.24

### LINKED ISSUE #5481: [Request] Add CDex (portable)
Hello!

Can you please consider adding CDex? https://cdex.mu/download

Standard version: http://mirror.cdex.mu/CDex-2.24.exe
The standard version is [told to contain adware](https://www.chip.de/downloads/CDex_12997142.html), but there is also a portable version (without adware) available.

The portable version is several steps behind the standard version, though:
http://mirror.cdex.mu/CDex%201.77%20Portable/CDex-1.77-portable-unicode.zip

Thank you so much for your consideration, and all the best for 2021! 👍

issaclin32:
closes #5481

[CDex](https://cdex.mu/) is an open-source digital audio CD extractor.

**NOTES**:
* The app is built in 32-bit.
* *persist* is not needed because config is at `$Env:AppData\CDex`

### DIFF
--- a/bucket/cdex.json
+++ b/bucket/cdex.json
@@ -1,0 +1,30 @@
+{
+    "version": "2.24",
+    "description": "Open-source Digital Audio CD Extractor",
+    "homepage": "https://cdex.mu/",
+    "license": "GPL-3.0-or-later",
+    "url": "http://mirror.cdex.mu/CDex-2.24.exe#/dl.7z_",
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):
Is it likely to receive any updates? (only 2 new ones since 2015 are there)

issaclin32:
> Is it likely to receive any updates? (only 2 new ones since 2015 are there)

I think it is safe to remove *checkver* in this case. Thanks for the info!

PR #8489, opened 2022-05-17
inkscape@1.2_2022-05-15: Fix checkver script

### LINKED ISSUE #8488: inkscape@1.2_2022-05-15 non-deterministic hash
Each download of Inkscape produces a different hash: #8480 #8481 #8482 #8484 #8486 #8487

magneticflux-:
Closes #8488 

Reran `checkver.ps1` and it works now. The file name changed and the old file doesn't return 404 and downloads HTML instead.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/inkscape.json
+++ b/bucket/inkscape.json
@@ -5,13 +5,13 @@
     "license": "GPL-3.0-or-later",
     "architecture": {
         "64bit": {
-            "url": "https://inkscape.org/en/gallery/item/33465/inkscape-1.2_2022-05-15_dc2aedaf03-x64.7z",
-            "hash": "e7dedf1fd0288bea2196a3bbdab06e364ee13a239c7ac14329a2e1f8f547b248",
+            "url": "https://media.inkscape.org/dl/resources/file/inkscape-1.2_2022-05-15_dc2aedaf03-x64.7z",
+            "hash": "38b5b7687aeca2762e4026fbf5f4586582b5c832cc6a30f8bca608c509e8ac31",
             "extract_dir": "inkscape-1.2_2022-05-15_dc2aedaf03-x64"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


bucket/inkscape.json
```diff
@@ -32,25 +32,30 @@
...
+        "regex": "(?<version>[\\d.]+_[\\d-]+)\\s(?<win32biturl>.+)\\s(?<win64biturl>.+)\\s(?<inkscapedir>.+)\\s(?<win32bithashurl>.+)\\s(?<win64bithashurl>.+)"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
                 "url": "https://inkscape.org/en$matchWin64biturl",
-                "extract_dir": "$matchInkscapedir-x64"
+                "extract_dir": "$matchInkscapedir-x64",
+                "hash": {
+                    "url": "$matchWin32bithashurl"
```
issaclin32, line 49:
```suggestion
                    "url": "$matchWin32bithashurl",
                    "regex": "$sha256"
```

issaclin32: I think there is something wrong *Scoop*'s hash matching machanism.
Therefore we need to **manually** match `$sha256` here. Will try to fix it later.
([related code](https://github.com/ScoopInstaller/Scoop/blob/master/lib/autoupdate.ps1#L70))

issaclin32 (approved):

tech189:
Nice work! Your checkver & autoupdate is way simpler. I got a fix working just now but my manifest is 7 lines longer so no need! 🚀

rashil2000:
Aah, the variable $version should not be used in the checkver script. It is causing problems as seen before, https://github.com/ScoopInstaller/Extras/issues/8494

magneticflux-:
Wasn't `$version` already being set before this PR? I just changed what came after it.

PR #8535, opened 2022-05-24
spotify: Run uninstaller only on uninstall

sitiom:
Fixes https://github.com/ScoopInstaller/Extras/discussions/8534

### DIFF
--- a/bucket/spotify.json
+++ b/bucket/spotify.json
@@ -25,7 +25,10 @@
         ]
     ],
     "uninstaller": {
-        "script": "Start-Process -Wait \"$dir\\Spotify.exe\" -ArgumentList '/Uninstall', '/Silent'"
+        "script": [
+            "if ($cmd -ne 'uninstall') { return }",
+            "Start-Process -Wait \"$dir\\Spotify.exe\" -ArgumentList '/Uninstall', '/Silent'"
+        ]
     },
     "checkver": {
         "script": [


bucket/spotify.json
```diff
@@ -25,7 +25,12 @@
         ]
     ],
     "uninstaller": {
-        "script": "Start-Process -Wait \"$dir\\Spotify.exe\" -ArgumentList '/Uninstall', '/Silent'"
+        "script": [
+            "# Run only on uninstall",
+            "if ($purge -ne $null) {",
```
issaclin32, line 30:
```suggestion
            "if ($purge) {",
```

rashil2000:
Purge is a special case of uninstall. We should actually check for uninstallation and run the command thereafter. See https://github.com/ScoopInstaller/Extras/blob/edfc2d23b082ce74c5529eb1a017e373a69ba6cd/bucket/archwsl.json#L33 for example.

sitiom:
> Purge is a special case of uninstall. We should actually check for uninstallation and run the command thereafter. See
> 
> https://github.com/ScoopInstaller/Extras/blob/edfc2d23b082ce74c5529eb1a017e373a69ba6cd/bucket/archwsl.json#L33
> 
> for example.

The Spotify uninstaller actually "purges" the config data. Just removing the spotify directory is already good enough.

rashil2000:
When I manually run the Spotify uninstaller, it deletes data in both LocalAppData and AppData. This behaviour should be the same when I run `scoop uninstall spotify`, which the current script in the PR is not doing (it only does that when you pass the purge flag).

sitiom:
So it has to match the installer's uninstall behavior 🤔

PR #8542, opened 2022-05-25
Speedtest GUI: Add version 1.10.163

### LINKED ISSUE #7136: (Request) add Speedtest-gui
PackageIdentifier: Speedtest
PackageVersion: 1.0
PackageLocale: en-US
Publisher: ookla
PackageName: speedtest-gui
License: proprietary
ShortDescription: GUI app for speedtest
Installers:

Architecture: x64
InstallerType: exe
InstallerUrl: https://install.speedtest.net/app/cli/ookla-speedtest-1.0.0-win64.zip
ManifestType: singleton
ManifestVersion: 1.0.0

Vedabahu:
Added Speedtest GUI.
Homepage is https://speedtest.net/
This can be done online only but some people prefer the app.

I have not added the autoupdate part as the url remains same even after new version is released.
The checkver contains html elements as the regex part was not detecting things properly.

Closes #7136

### DIFF
--- a/bucket/speedtest.json
+++ b/bucket/speedtest.json
@@ -1,5 +1,4 @@
 {
-    "##": "Autoupdate has not been added as the url will remain same.",
     "version": "1.10.163",
     "description": "Internet Speedtest for Windows GUI",
     "homepage": "https://www.speedtest.net/apps/windows",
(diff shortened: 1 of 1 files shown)


rashil2000:
URL should be added to autoupdate even if it stays the same - this is so that hashes can be updated

Vedabahu:
> URL should be added to autoupdate even if it stays the same - this is so that hashes can be updated

Ok I will do it.

This app is built using Unity so it makes a folder in `appdata\LocalLow`. Is it necessary to add a script that deletes that folder when uninstalling?

image

rashil2000:
I think we can leave it. What's there in that folder btw?

Vedabahu:
> I think we can leave it. What's there in that folder btw?

image

The `speedtest.db` stores records of the speedtest made and its results, `output_log.txt` has some normal loading and unloading logs. `config` & `values` has things about analytics like sending user usage data and other things to the app devs.

Vedabahu:
The uninstaller does not remove that file so I think it can be left alone.

bucket/speedtest.json
```diff
@@ -0,0 +1,38 @@
+{
+    "version": "1.10.163",
+    "description": "Internet Speedtest for Windows GUI",
```
rashil2000, line 3:
```suggestion
    "description": "Internet Speedtest GUI for Windows",
```
😅

Issue #8538, opened 2022-05-24
[Request] Icecast


hsn10:
### Information

**Name:** icecast

**Description:** [clear and concise details of what it is]

**Homepage:** [ICE cast](https://icecast.org)

**Download link(s):** [win32 dl](https://downloads.xiph.org/releases/icecast/icecast_win32_2.4.4.exe)

**Some indication of popularity/repute:** Only free streaming server I know. Can do radio and TV.

PR #8545, opened 2022-05-25
icecast: Add version 2.4.3

issaclin32:
closes https://github.com/ScoopInstaller/Extras/issues/8538

[IceCast](https://icecast.org/) is a streaming media (audio/video) server which supports Ogg (Vorbis and Theora), Opus, WebM and MP3 streams.

**NOTES**:
* This package is in *Extras* (rather than *Main*) because of non-standard install methods.

* *IceCast* used to have Windows GUI and Windows service installation, but was removed in 2.3.3.
> https://icecast.org/faq/
What happened to the Windows GUI? I only see a black window with some text.
Icecast used to have a graphical user interface on Windows, but that was deprecated in 2.3.3 as it wasn’t frequently used and hard to maintain. Considering that most of the things the GUI could do now are possible using the Icecast web interface, it’s not needed anymore.
The black window with text you see is the command line running the Icecast server, just keep that window open. When you want to quit Icecast, hit Ctrl+C and confirm that you want to quit.
......
How can I run Icecast as a Service on Windows?
We do not have service capabilities in Icecast anymore, currently. Please see this Wiki page for instruction how to set up Icecast as a service using a third party service manager.

* This packages uses *nssm* instead of `sc.exe create ...` because we are installing a batch file as Windows service. Also this is the recommended method in [the documentation](https://wiki.xiph.org/Icecast_Server/Setup_Windows_Service).

### DIFF
--- a/bucket/icecast.json
+++ b/bucket/icecast.json
@@ -38,8 +38,8 @@
         "script": [
             "# if user uninstalls (but not updates) the app, then uninstall IceCast service",
             "if ($cmd -eq 'uninstall') {",
-            "    sc.exe query IceCast | Out-Null",
-            "    if ($LASTEXITCODE -eq 0) {",
+            "    $service = Get-Service 'IceCast' -ErrorAction SilentlyContinue",
+            "    if ($service) {",
             "        if(!(is_admin)) { error 'Admin right is required to uninstall IceCast service'; break }",
             "        sc.exe delete IceCast",
             "    }",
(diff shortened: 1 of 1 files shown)


rashil2000:
Since we're not directly running nssm during installation, I recommend making it a `suggest` instead of `depends`

rashil2000:
Outside of .bat files, in PowerShell statements, it would be better to use `Get-Service` and `Remove-Service` instead of `sc.exe query` and `sc.exe delete`

issaclin32:
I use .bat file because it is the most compatible way. If the user runs `install-service.ps1` **under cmd**, the script may not be run by Powershell, but opened by *Powershell ISE*, *VSCode*, ...., depending on what the file association is.

I think it would be better to enable users to run scripts whether they are using *cmd* or *Powershell*, unless there is a clear advantage of using .ps1 scripts.

rashil2000:
I was talking about the `uninstaller.script` 😅

rashil2000 (approved):

Issue #8490, opened 2022-05-17
[Request] FreeRapid


NatoBoram:
### Information

**Name:** FreeRapid

**Description:**
> FreeRapid is an easy to use Java downloader that supports downloading from Rapidshare, Youtube, Facebook, Picasa and other file-sharing services, include the new ones like mega.co.nz.
> Simply copy and paste your links from a browser to this application. FreeRapid Downloader will handle the rest itself. No more clicking or uncomfortable waiting. 

**Homepage:** https://wordrider.net/freerapid/index.html

**Download link(s):** https://wordrider.net/freerapid/download.htm

Downloads are hidden behind ad-walls. Specifically, the kind of ad-wall that FreeRapid is made to bypass.

Source code can be downloaded with `git svn clone http://svn.wordrider.net/svn svn.wordrider.net`. Username and passwords are `wordrider`.

**Some indication of popularity/repute:** About 3K topics posted in its forum: https://wordrider.net/forum/index.php?9

Vedabahu:
This app was last updated on 18th May 2014. It also uses Java 6 or 7. So can you please move this issue to the [ScoopInstaller/Versions](https://github.com/ScoopInstaller/Versions) repo.

NatoBoram:
I just finished cloning it with `git svn` and its latest changes look like this:

> * 98ec202fa - (HEAD -> master, git-svn) CeskaTelevize.cz - fixed for more link types (5 months ago) <ntoskrnl>
> * 854ea0e6a - git-svn-id: http://svn.wordrider.net/svn@5826 40407085-c429-0410-a8d0-eb1a82597f6e (5 months ago) <ntoskrnl>
> * 484a7aa16 - NovaPlus - updated (5 months ago) <ntoskrnl>
> * f0b108018 - CeskaTelevize.cz - fixed (5 months ago) <ntoskrnl>
> * 89c8c2d90 - YouTube.com - fixed (11 months ago) <ntoskrnl>
> * 4377db178 - Mega.nz - support new link format (1 year, 3 months ago) <ntoskrnl>
> * 7811dc7d2 - YouTube.com - support non-DASH formats (1 year, 3 months ago) <ntoskrnl>
> * 3cd414621 - YouTube.com - fixed and added AV1 formats (1 year, 3 months ago) <ntoskrnl>
> * 4fdab37ad - ulozto update (2 years, 2 months ago) <Vity>

I think it's updated solely through its plugins.

[...]

PR #8575, opened 2022-05-29
freerapid: Add version 0.9u4

issaclin32:
* closes https://github.com/ScoopInstaller/Versions/issues/532
* related: #8490

[FreeRapid](http://wordrider.net/freerapid/) is a Java downloader that supports downloading from Rapidshare, Youtube, Facebook, Picasa and other file-sharing services.

**NOTES**:
* **file source**: *FreeRapid* now puts their download behind a **pay-wall**.
In https://github.com/ScoopInstaller/Versions/issues/532, @NatoBoram states that there is a way to fetch files directly from their SVN.

However I don't think we should download from SVN because (1) they may close the access of SVN because of unusual number of connections, etc.  (2) This is some sort of "hacking", which should not be encouraged, especially in an open-source software.

Instead, we can just use the file fetched by `archive.org` before they put the pay-wall, and let users update "legally" through the app itself.

* **bin/shortcuts**: There is `frd.exe` in the file, but it basically calls *javaw* to start `frd.jar` (The app does not have a bundled JRE). I think it is more straightforward to just use `frd.jar` for bin ~and shortcut~.

Also see **$dir\doc\readme.txt**
```
additional parameters for launching are:

java -jar frd.jar [-h -v -d -r -D<property>=<value> -m -p]

options
  -h (--help,-?)      print this message
  -v (--version)      print the version information and exit
  -d (--debug)        print debugging information
  -r (--reset)        reset user properties to default values  
  -m (--minim)        minimize main window on start  
  -Dproperty=value    Passes properties and values to the application (mostly for debug or testing purposes)
  -p (--portable)     configuration files will be stored in the 'config'
                      folder, all file paths will be saved relatively to FRD
                      folder (if possible) - useful for USB FLASH drives

If value of option -D is set 'default' (without ') default value will be used.
```

### DIFF
--- a/bucket/freerapid.json
+++ b/bucket/freerapid.json
@@ -16,7 +16,7 @@
     "shortcuts": [
         [
             "frd.exe",
-            "FreeRapid Downloader",
+            "FreeRapid Downloader"
         ]
     ],
     "persist": [
--- a/bucket/freerapid.json
+++ b/bucket/freerapid.json
@@ -15,10 +15,8 @@
     "bin": "frd.jar",
     "shortcuts": [
         [
-            "frd.jar",
+            "frd.exe",
             "FreeRapid Downloader",
-            "",
-            "frd.ico"
         ]
     ],
     "persist": [
--- a/bucket/freerapid.json
+++ b/bucket/freerapid.json
@@ -1,0 +1,28 @@
+{
+    "version": "0.9u4",
+    "description": "A Java downloader that supports downloading from Rapidshare, Youtube, Facebook, Picasa and other file-sharing services.",
+    "homepage": "http://wordrider.net/freerapid/",
+    "license": {
+        "identifier": "Freeware",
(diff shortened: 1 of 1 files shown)


bucket/freerapid.json
```diff
@@ -0,0 +1,28 @@
...
+    "url": "https://raw.githubusercontent.com/ScoopInstaller/Binary/master/freerapid/FreeRapid-0.9u4.zip",
+    "hash": "f26d570b63e9591e438a625f87eee80480b160f077a4d7aae433c8d3e0d20ab9",
+    "suggest": {
+        "JAVA Runtime Environment": "java/openjdk"
+    },
+    "extract_dir": "FreeRapid-0.9u4",
+    "bin": "frd.jar",
+    "shortcuts": [
+        [
+            "frd.jar",
```
rashil2000, line 18: This will require .jar extension to be registered to be handled by the user's JRE

rashil2000 (approved):

PR #8640, opened 2022-06-08
dotnet-generic-unpacker: Add version 1.0.0.1

issaclin32:
**[.NET Generic Unpacker](https://ntcore.com/?page_id=353)** is a program to dump .NET packed applications.

**NOTES**:
* Only **32-bit version** of .NET Generic Unpacker can dump **32-bit executables**. The same rule applies to 64-bit version.
> Download the x64 version of the .NET Generic Unpacker only if the process is not an x86 process. In all other cases download the x86 version.
* *autoupdate* is not needed because last update was in 2008.

### DIFF
--- a/bucket/dotnet-generic-unpacker.json
+++ b/bucket/dotnet-generic-unpacker.json
@@ -1,9 +1,9 @@
 {
-    "##": "Only 32-bit version of .NET Generic Unpacker can dump 32-bit executables. (The same rule applies to 64-bit version)",
     "version": "1.0.0.1",
     "description": "A program to dump .NET packed applications",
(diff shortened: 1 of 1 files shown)


bucket/dotnet-generic-unpacker.json
```diff
@@ -0,0 +1,44 @@
+{
+    "##": "Only 32-bit version of .NET Generic Unpacker can dump 32-bit executables. (The same rule applies to 64-bit version)",
```
rashil2000, line 2: This comment can be converted into notes so that users can also see

rashil2000 (approved):

PR #8672, opened 2022-06-12
phoenix-protector: Add version 1.8.0.1

issaclin32:
**[Phoenix Protector](https://ntcore.com/?page_id=384)** is a code obfuscator for .NET assemblies.

**NOTES**:
* This app requires *ilmerge* (`extras/ilmerge`) to work properly. However the app does not detect *ilmerge* from PATH, but rather prompts the user to configure it upon first run. Therefore I added the **.reg file** to simplify the task.
* *autoupdate* is not needed because last update was in 2012.
* Config location: `HCKU:\SOFTWARE\NTCore\Phoenix Protector`.

**SCREENSHOTS**:


rashil2000:
The reg files are showing up as "binary" in GitHub's web interface. Would it cause text encoding problems like before?

issaclin32:
> The reg files are showing up as "binary" in GitHub's web interface. Would it cause text encoding problems like before?

Thanks for reminding.
I changed the encoding from UTF-16LE to UTF-8, so that it is treated as a **text** file.

I think the problem described in #7546 (which **blocks bucket updates**) is caused by inconsistant line endings, rather than UTF-16LE encoding.

niheaven (approved):

PR #8681, opened 2022-06-13
imo-messenger: Add version 1.4.1.6

### LINKED ISSUE #8677: [Request] IMO Messenger
### Information

**Name:** [name of new package]
imo
**Description:** [clear and concise details of what it is]
With Imo for Windows you can message and video chat with your friends and family for free, no matter what device they are on.

**Homepage:** [a URL/link]
https://imo.im/
**Download link(s):** [URL(s)/link(s)]
https://static-web.imoim.net/as/indigo-static/imo-app/imo-Windows/ImoSetup_1.4.1.6_Release.exe
**Some indication of popularity/repute:** [GitHub stars/software reviews etc.]
image

issaclin32:
* Closes #8677

* **[IMO Messenger](https://imo.im/)** is a free video calls and chat software.

**NOTES**:
* **package naming**: I named the package `imo-meessenger` (instead of `imo`) because both the shortcut *(created by the installer)* and window title says `imo Messenger`.

* `pre_uninstall` is used here because imo Messenger **stays in tray** after closing the window.

* **checkver**: using **third-party site** for *checkver* because imo's homepage **requires JS to load**.
  I think it won't hurt because *imo* keeps their old releases on thier site. (which means it will not cause the installation failures even if the third-party site need a bit of time to catch up with the update). If there is a better source for *checkver*, please tell me. Thanks!

### DIFF
--- a/bucket/imo-messenger.json
+++ b/bucket/imo-messenger.json
@@ -20,7 +20,7 @@
     "shortcuts": [
         [
             "ImoDesktopApp.exe",
-            "Imo Messenger"
+            "imo Messenger"
         ]
     ],
     "persist": [
--- a/bucket/imo-messenger.json
+++ b/bucket/imo-messenger.json
@@ -1,0 +1,37 @@
+{
+    "##": "The homepage requires JS to load (not suitable for checkver)",
+    "version": "1.4.1.6",
+    "description": "Video calls and chat software",
+    "homepage": "https://imo.im/",
+    "depends": "extras/uniextract2",
+    "license": {
+        "identifier": "Freeware",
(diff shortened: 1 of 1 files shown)


tech189 (commented):
It installs and works okay for me. Only issue is that the Universal Extractor pops up a 'getting started' window during installation. This only happens the first time you use Universal Extractor as I couldn't get it to appear when reinstalling. Can we make it run silently?

tech189 (approved):
The install is almost completely silent now, the windows that pop up don't require any user interaction and disappear almost immediately, so I'd say that's okay!

PR #8690, opened 2022-06-14
windows-terminal: fix context menu error in drive root folder

### LINKED ISSUE #8689: windows terminal: context menu not work in drive root folder
## Bug Report

**Package Name:** winsows-terminal

### Current Behaviour

Context menu launches windows terminal with `could not access starting directory` error in disk root folder

### Expected Behaviour

launch windows terminal in current location

### Additional context/output

it is a known issue for paths on windows

### Possible Solution

replace `"wt.exe" -d "%V"` with `"wt.exe" -d "%V/"` or `"wt.exe" -d "."`

### System details

**Windows version:** windows 10

**OS architecture:** 64bit

**PowerShell version:** 5.1.19041.1682

#### Scoop Configuration

```json
//# Your configuration here
```

AkariiinMKII:
Closes #8689


rashil2000:
How has this been tested?

chadcatlett:
Could the powershell gallery being down be causing this?

AkariiinMKII:
> How has this been tested?

well, here is an example:
While in disk volume C, the `%V` returns `C:\`, and the command being `"wt.exe" -d "C:\"`, the backslash is treated as a escape character in this situation

rashil2000:
> > How has this been tested?
> 
> Well, here is an example: While in disk volume C, the `%V` returns `C:\`, and the command being `"wt.exe" -d "C:\"`, somehow the backslash is treated as a escape character in this situation, then wt returns error `could not access starting directory "C:""` Fortunately `C:/` is a valid path for wt

Since we just added a `/` in this case wouldn't the commandline now expand to `C:\/`?

AkariiinMKII:
> > > How has this been tested?
> > 
> > Well, here is an example: While in disk volume C, the `%V` returns `C:\`, and the command being `"wt.exe" -d "C:\"`, somehow the backslash is treated as a escape character in this situation, then wt returns error `could not access starting directory "C:""` Fortunately `C:/` is a valid path for wt
> 
> Since we just added a `/` in this case wouldn't the commandline now expand to `C:\/`?

That's possible... But according to the error report above, I think that may finally be `C:/`. I guess that by using the variable `%V`, only path for disk root folder ends with a backslash, which is misprocessed and causes this error. Anyway, both `wt -d C:/` and `wt -d C:\/` are workable in this case

AkariiinMKII:
And I found `"wt.exe" -d "."` also works in drive root folder, maybe we should turn to use this?

rashil2000:
I think it will be better to see how Windows Terminal installed from the Store does it

jasongodev:
Is this related to #9333 ?

AkariiinMKII:
> Is this related to #9333 ?

yes, that's another solution.

HUMORCE:
#9333

PR #8696, opened 2022-06-15
soundswitch: Add version 6.2.4

ClassicDarkChocolate:
[SoundSwitch](https://soundswitch.aaflalo.me/): Switch your default playback devices and/or recording devices using simple hotkeys

- No persist needed:

  

### DIFF
--- a/bucket/soundswitch.json
+++ b/bucket/soundswitch.json
@@ -1,0 +1,21 @@
+{
+    "version": "6.2.4",
+    "description": "Switch your default playback devices and/or recording devices using simple hotkeys",
+    "homepage": "https://soundswitch.aaflalo.me/",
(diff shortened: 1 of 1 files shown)


tech189 (commented):
Installs fine but I'm not sure we should hardcode the `.0` in the autoupdate url as previous releases have had versions ending in other numbers.

Also it would be good to move the both the url and hash and the url in autoupdate to an architecture block as the app is 64bit only.

ClassicDarkChocolate:
> Also it would be good to move the both the url and hash and the url in autoupdate to an architecture block as the app is 64bit only.

Thanks. I did not find that it dropped support for 32bit.

tech189 (approved):

PR #8698, opened 2022-06-15
Updated strokesplus to v0.5.6.6

Xelbayria:
updated link to the new website in JSON
updated the auto-update link in JSON
updated the old version to the new version

Closes #3939

Relates to #XXXX

- [X] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #8703, opened 2022-06-16
strokesplus.net: Add version 0.5.6.7

### LINKED ISSUE #3939: [Request] Add StrokesPlus.net
### 1. Checklist

1. [x] The app I am requesting [**does not already exist**](https://scoop-docs.netlify.app/apps/?query=StrokesPlus) on **<https://scoop-docs.netlify.app/apps>**
1. [x] [**There is no open issue for this app**](https://github.com/lukesampson/scoop-extras/issues?q=is%3Aissue+is%3Aopen+StrokesPlus)
1. [x] The issue title starts “[Request] Add”
1. [x] The download URL is public and not locked behind a paywall/login

### 2. Package details

1. Software project URL: **<https://www.strokesplus.net/>**
1. Downloads URL: **<https://www.strokesplus.net/Downloads>**
1. What versions does this software have? Installer and Portable
1. What dependencies required for this software? See “[**Resources**](https://www.strokesplus.net/Downloads)” section on official Downloads page
1. Have this software command-line interface (CLI)? No
1. Software summary/short description: Free software, by which user can call hotkeys, macros and other by mouse gestures. See [**official site**](https://www.strokesplus.net/) for more details.

### 3. Note

Please, do not confuse StrokesPlus.net and StrokesPlus, that [**already exist in Scoop**](https://scoop-docs.netlify.app/apps/?query=StrokesPlus). StrokesPlus is no longer maintained software, its latest version [**was released in March 2016**](https://www.strokesplus.com/). StrokesPlus.net is [**actively maintained on April, 2020**](https://www.strokesplus.net/ChangeLog.txt).

Thanks.

Xelbayria:
Closes #3939

Relates to #8698


Xelbayria:
Alright, guys, please make sure everything is good.

issaclin32 (approved):

Xelbayria:
ah. I forgot about this: 

~IMPORTANT: Extract with 7-Zip or another program that does not set Zone.Identifier or you will get an error during "Initializing Script Engine..." stating "Operation is not supported". Alternatively, you can go through each DLL file, right-click, select Properties and click the Unblock button or check box. Windows is essentially marking these DLLs as untrusted, so while StrokesPlus.net will start, the DLLs are not able to be loaded.~

NOTE: This will not be able to interact with elevated windows (e.g. Task Manager) unless you run as administrator. You can use the settingsfile="C:\Path\SettingsFile.ext" command line parameter to specify a custom location for your settings (make sure to move your file there first, or you will get the default settings)

Due to "NOTE", I would recommend using %appdata% to store the settings so it wouldn't get resettled to the default settings. Besides that, I already installed The software and there are no problems with DLL which is why I stroked the information ~ ~

bucket/strokesplus.net.json
```diff
@@ -0,0 +1,39 @@
...
+    "architecture": {
+        "64bit": {
+            "url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_0.5.6.6.zip",
+            "hash": "b1c82604910676bc7bb3cd43b02a278bc01f51b212a6c180aab59f0a6ec86e9f"
+        },
+        "32bit": {
+            "url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_0.5.6.6.zip",
+            "hash": "b1c82604910676bc7bb3cd43b02a278bc01f51b212a6c180aab59f0a6ec86e9f"
+        }
+    },
```
tech189, line 18: Please remove the architecture block - there is no difference between the 64bit and 32bit url and hash. The app can run on 32bit computers too.

bucket/strokesplus.net.json
```diff
@@ -0,0 +1,39 @@
...
+    "shortcuts": [
+        [
+            "StrokesPlus.net.exe",
+            "StrokesPlus.net"
+        ]
+    ],
```
tech189, line 24:
```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\strokesplussettings.json\")) { Copy-Item \"$dir\\Default_StrokesPlus.net.json\" \"$dir\\strokesplussettings.json\" | Out-Null }",
    "shortcuts": [
        [
            "StrokesPlus.net.exe",
            "StrokesPlus.net",
            "settingsfile=\"$dir\\strokesplussettings.json\""
        ]
    ],
    "persist": "strokesplussettings.json",
    "pre_uninstall": [
        "# The following must be done as StrokesPlus deletes the original file which was linked to the persist folder when it saves settings, rather than editing the file.",
        "Copy-Item \"$dir\\strokesplussettings.json\" \"$persist_dir\\strokesplussettings.json\" | Out-Null"
    ],
```

Allows settings to be persisted. If it's a first time install, the default settings are copied, so no error on startup.

bucket/strokesplus.net.json
```diff
@@ -0,0 +1,39 @@
...
+        "regex": "([\\d.]+)"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_$version.zip"
+            },
+            "32bit": {
+                "url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_$version.zip"
+            }
```
tech189, line 36: Please also remove the architecture here.

Xelbayria: Ok.. you are suggesting that it should look like this: 
```
"64bit": {
             "url": "the_url_line" 
}
```
both "architecture" and "32bit" are removed. is that correct?

bucket/strokesplus.net.json
Xelbayria: ok. I'm confused about `strokesplussettings.json` I have checked and am sure the correct file is `StrokePlus.net.json` and can confirm that `Default_StrokePlus.net.json` exists
When I change the setting of my strokeplus to a custom setting, `Default_StrokePlus.net.json` disappeared and `StrokePlus.net.json` appeared.

bucket/strokesplus.net.json
issaclin32: > both "architecture" and "32bit" are removed. is that correct?

No. It should be something like:

```
"url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_0.5.6.6.zip",
"hash": "b1c82604910676bc7bb3cd43b02a278bc01f51b212a6c180aab59f0a6ec86e9f"

....

"autoupdate": {
    "url": "https://cdn.strokesplus.net/files/StrokesPlus.net_Portable_$version.zip"
}
```

bucket/strokesplus.net.json
Xelbayria: It looks like this doesn't account for the other file `StrokesPlus.net.json` when it only copies `Default_Strokesplus.net.json`.  
Is there a way to add `Strokesplus.net.json`??

bucket/strokesplus.net.json
Xelbayria: ah got it.

bucket/strokesplus.net.json
tech189: No need to account for `StrokesPlus.net.json` as it is only created when no custom setting file is provided. Scoop's shortcut makes the app use the custom one `strokesplussettings.json`.

Xelbayria: alright. I understand now. 
👍

tech189:
```suggestion
    "shortcuts": [
        [
            "StrokesPlus.net.exe",
            "StrokesPlus.net",
            "settingsfile=\"$dir\\strokesplussettings.json\""
        ]
    ],
```
While updating to 0.5.6.7, you missed out this line and so it stopped loading the custom settings file, please add it back and then I can approve this PR.

Xelbayria:
@tech189 
I've made another change to the JSON. It should be all good, so please check to make sure it is.

tech189 (approved):

Xelbayria:
Alright. Now everything is all good. The last step is adding `strokesplus.json` and `strokesplus.net.json` to the master branch of Extras.  How does this work? I am curious if you don't mind my question

tech189:
> How does this work? I am curious if you don't mind my question

One of the maintainers can now squash the commits into one and add it to Extras, shall I do it @issaclin32?

issaclin32 (approved):

issaclin32:
> Just a quick question, how does `auto-update` works?

The [Excavator action](https://github.com/ScoopInstaller/Extras/actions/runs/2536860491) will run **every 4 hours** to browse through all `checkver` of **all the packages in the bucket**. If the package has an update, then use the information in `autoupdate` to update the file `xxx.json`.

You can check [.github/workflows](https://github.com/ScoopInstaller/Extras/tree/master/.github/workflows) and [ScoopInstaller/GithubActions](https://github.com/ScoopInstaller/GithubActions) for more details.

These Github actions are [available for your own bucket](https://github.com/ScoopInstaller/BucketTemplate) too.

issaclin32:
> > Just a quick question, how does `auto-update` works?
> 
> The [Excavator action](https://github.com/ScoopInstaller/Extras/actions/runs/2536860491) will run **every 4 hours** to browse through all `checkver` of **all the packages in the bucket**. If the package has an update, then use the information in `autoupdate` to update the file `xxx.json`.
> 
> You can check [.github/workflows](https://github.com/ScoopInstaller/Extras/tree/master/.github/workflows) and [ScoopInstaller/GithubActions](https://github.com/ScoopInstaller/GithubActions) for more details.
> 
> These Github actions are [available for your own bucket](https://github.com/ScoopInstaller/BucketTemplate) too.

Xelbayria:
alright, It's strange that the version of Strokesplus.net is not updated to the latest version. I ran a quick test and the script regarding `checkver` is not working...  

what's the regex for `x.x.x.x - MM/DD/YYYY`? where x.x.x.x are the version.

issaclin32:
> alright, It's strange that the version of Strokesplus.net is not updated to the latest version. I ran a quick test and the script regarding checkver is not working...

Thanks for reporting.
The problem is actually caused by access error of the website. The returned message was `0x5: Access Denied`, causing the regex to filter `0` as the version.

Xelbayria:
> > alright, It's strange that the version of Strokesplus.net is not updated to the latest version. I ran a quick test and the script regarding checkver is not working...
> 
> Thanks for reporting. The problem is actually caused by access error of the website. The returned message was `0x5: Access Denied`, causing the regex to capture `0` as the version.

I see. and saw the solution. I tried 
```
Latest Release: ([\\d.]+)
```
it doesn't work. I don't realize that `:` required \ and <b>  I've to look up why it's required. Well Your solution taught me that

Xelbayria:
> There is a mistake in the `pre_install` section, and I didn't find it when reviewing the code. Sorry about that.😓

No worries. Glad I run some tests to make sure it's working as intended. I saw the changes you made. I have my guess on why. 
 

> No. I am just matching `<b>` src

Ah, I understand now. 

> **Powershell** uses the .NET variant of RegEx, so some rules may be different from the RegEx libraries in JavaScript, Python, etc. You can select ".NET" on [Regex101.com](https://regex101.com/) to check it. regex

so it was .NET variant of RegEx. I was trying to find out which one. and yeah I know about Regex101.com and it's very useful. Thank you for answering my questions.

PR #8725, opened 2022-06-21
smplayer: Add version 22.2.0.0

### LINKED ISSUE #8701: [Request] SMPlayer
Package Request Information
Name: SMPlayer

Description: a video player forked from MPV

Homepage: https://www.smplayer.info/en/info

Download link(s): https://www.fosshub.com/SMPlayer.html/smplayer-22.2.0-x64.exe

PS: I know there is a similar package https://github.com/ScoopInstaller/Extras/pull/429. But, the old request is portable version

issaclin32:
* closes #8701

* **[SMPlayer](https://www.smplayer.info/)** is a video player forked from MPV. It can also play Youtube videos, search and download subtitles, and includes other features like a thumbnail generator and audio and video filters.

**NOTES**:
* not using `#/dl.7z` here because it will be treated as part of the input `www.fosshub.com/SMPlayer.html?dwl=...`, causing URL error.
* *persist*: **.m3u8** are playlists encoded in UTF-8. ([Wikipedia](https://en.wikipedia.org/wiki/M3U))

### DIFF
--- a/bucket/smplayer.json
+++ b/bucket/smplayer.json
@@ -17,7 +17,7 @@
         "Rename-Item \"$dir\\SMPlayer.html\" 'dl.7z'",
         "Expand-7zipArchive \"$dir\\dl.7z\" \"$dir\" -ExtractDir 'smplayer-portable' -Removal | Out-Null",
         "'hdpi.ini', 'playlist.ini', 'favorites.m3u8', 'radio.m3u8', 'tv.m3u8' | ForEach-Object {",
-        "    if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" }",
+        "    if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" | Out-Null }",
         "}",
         "if (Test-Path \"$persist_dir\\smplayer.ini\") { Copy-Item \"$persist_dir\\smplayer.ini\" \"$dir\\smplayer.ini\" }"
     ],
--- a/bucket/smplayer.json
+++ b/bucket/smplayer.json
@@ -1,0 +1,52 @@
+{
+    "version": "22.2.0.0",
+    "description": "A video player forked from MPV. It can also play Youtube videos, search and download subtitles, and includes other features like a thumbnail generator and audio and video filters.",
+    "homepage": "https://www.smplayer.info/",
+    "license": "GPL-2.0-or-later",
+    "architecture": {
+        "64bit": {
+            "url": "https://www.fosshub.com/SMPlayer.html?dwl=smplayer-portable-22.2.0.0-x64.7z",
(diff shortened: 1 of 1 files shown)


bucket/smplayer.json
```diff
@@ -0,0 +1,52 @@
...
+        [
+            "smplayer.exe",
+            "SMPlayer"
+        ]
+    ],
+    "persist": [
+        "screenshots",
+        "hdpi.ini",
+        "playlist.ini",
+        "smplayer.ini",
```
tech189, line 33: Can't get settings to persist after uninstallation - settings save to smplayer.ini but it's not linked to the smplayer.ini in the persist folder. Will we have to do something similar to strokesplus.net? Such as copying the file to persist manually in pre_uninstall: https://github.com/ScoopInstaller/Extras/pull/8703/files#diff-ea8096c8380d8ac840470e604fd0ddbcb123178f5deabae8d768ffc3277d5148R20-R23

issaclin32: Yeah. We should definitely do that. Thank you for the details.

By the way, Copy-Item does not print output to screen by default.
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/copy-item?view=powershell-7.2
> -PassThru
Returns an object that represents the item with which you're working. By default, this cmdlet doesn't generate any output.

bucket/smplayer.json
```diff
@@ -0,0 +1,56 @@
...
+        "32bit": {
+            "url": "https://www.fosshub.com/SMPlayer.html?dwl=smplayer-portable-22.2.0.0-win32-qt5.6.7z",
+            "hash": "be10cdc34a9b6282e1e91e9b789825a11939576750f1d5be839e2a4892532402"
+        }
+    },
+    "pre_install": [
+        "Rename-Item \"$dir\\SMPlayer.html\" 'dl.7z'",
+        "Expand-7zipArchive \"$dir\\dl.7z\" \"$dir\" -ExtractDir 'smplayer-portable' -Removal | Out-Null",
+        "'hdpi.ini', 'playlist.ini', 'favorites.m3u8', 'radio.m3u8', 'tv.m3u8' | ForEach-Object {",
+        "    if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" }",
```
tech189, line 20:
```suggestion
        "    if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" | Out-Null }",
```
To make creating new ini files silent

bucket/smplayer.json
tech189: Great, it works now! And thanks for the info, I'll remove the Out-Null from strokesplus.net now

tech189: Oh I see you've already fixed it, thank you!

bucket/smplayer.json
issaclin32: Oh. I missed that. Thank you for reminding me.

PR #8734, opened 2022-06-21
smtube: Add version 21.10.0

SonaliBendre:
Closes #XXXX

Relates to #XXXX

~* Not ready until #8725 is merged.~

### DIFF
--- a/bucket/smtube.json
+++ b/bucket/smtube.json
@@ -1,0 +1,36 @@
+{
+    "version": "21.10.0",
+    "description": "YouTube browser for SMPlayer.",
+    "homepage": "https://www.smtube.org",
+    "license": "GPL-2.0-only",
+    "depends": "SMPlayer",
+    "architecture": {
+        "64bit": {
+            "url": "https://sourceforge.net/projects/smtube/files/SMTube/21.10.0/smtube-21.10.0-x64.exe#/dl.7z",
+            "hash": "sha1:776c7fc98370c22c1bbd640cb93e2a7d28950958"
+        },
+        "32bit": {
+            "url": "https://sourceforge.net/projects/smtube/files/SMTube/21.10.0/smtube-21.10.0-win32-qt5.6.exe#/dl.7z",
+            "hash": "sha1:b854f382c3440a1631dcfe9b3caa82cef688e123"
+        }
+    },
+    "installer": {
+        "script": [
+            "Remove-Item -LiteralPath \"$dir\\`$PLUGINSDIR\" -Force -Recurse",
+            "Copy-Item \"$dir\\*\" \"$(appdir SMPlayer $global)\\current\" -Force -Recurse"
+        ]
+    },
+    "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/smtube.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "21.10.0",
+    "description": "YouTube browser for SMPlayer.",
+    "homepage": "https://www.smtube.org",
+    "license": "GPL-2.0-only",
+    "depends": "SMPlayer",
```
issaclin32, line 6:
```suggestion
    "depends": "extras/smplayer",
```

bucket/smtube.json
```diff
@@ -0,0 +1,36 @@
+{
+    "version": "21.10.0",
+    "description": "YouTube browser for SMPlayer.",
```
issaclin32, line 3:
```suggestion
    "description": "YouTube browser extension for SMPlayer.",
```

bucket/smtube.json
```diff
@@ -0,0 +1,36 @@
...
+            "url": "https://sourceforge.net/projects/smtube/files/SMTube/21.10.0/smtube-21.10.0-win32-qt5.6.exe#/dl.7z",
+            "hash": "sha1:b854f382c3440a1631dcfe9b3caa82cef688e123"
+        }
+    },
+    "installer": {
+        "script": [
+            "Remove-Item -LiteralPath \"$dir\\`$PLUGINSDIR\" -Force -Recurse",
+            "Copy-Item \"$dir\\*\" \"$(appdir SMPlayer $global)\\current\" -Force -Recurse"
+        ]
+    },
```
issaclin32, line 22: This will cause 3 problems:
**(1)** `install.json` and `manifest.json` will also be copied to SMPlayer's directory. This will cause problems when user tries to update/uninstall SMPlayer.
Need to avoid this by something like:
```
Get-ChildItem \"$dir\" | ForEach-Object {
    if ($_.Name -match '(install|manifest)\.json') { return }
    Copy-Item ....
}
```
(2) The users will be able to install smtube properly, not cannot remove it unless they re-install SMPlayer. So you'll need to add an uninstall script that removes all files in smplayer's directory that exists in smtube's `$dir`.

(3) You will need to manually create shortcut for `SMPlayer.exe`, see [extras/openvpn-connect](https://github.com/ScoopInstaller/Extras/blob/master/bucket/openvpn-connect.json) for example.

I think a much simpler way to solve this is to just add a package named **smplayer-with-smtube**, which has a **checkver script** that checks updates from both *SMPlayer* and *SMTube*.

see [extras/inkscape](https://github.com/ScoopInstaller/Extras/blob/master/bucket/inkscape.json) for examples of checkver script.

issaclin32: Asking for you opinion about this: @tech189

tech189: I think an `smplayer-with-smtube` package is a good idea to avoid the problems you've listed above 👍

PR #8751, opened 2022-06-23
camstudio: Add version 2.7.4-r354

issaclin32:
**[CamStudio](https://camstudio.org/)** is a free desktop recorder and video streaming software.

**NOTES**:
* Aside from `controller.ini`, most of the config is stored at registry: `HKCU:\SOFTWARE\CamStudioOpenSource for Nick`.
* *autoupdate* is not needed because last update was in 2015.

### DIFF
--- a/bucket/camstudio.json
+++ b/bucket/camstudio.json
@@ -1,0 +1,30 @@
+{
+    "version": "2.7.4-r354",
+    "description": "Free desktop recorder and video streaming software",
+    "homepage": "https://camstudio.org/",
+    "license": "GPL-3.0-or-later",
+    "url": [
+        "https://eu2.contabostorage.com/a4054a9f5df24ecf87c4a33ac9b6477b:camstudio/CamStudioSetup.exe#/dl.exe",
+        "https://downloads.sourceforge.net/project/camstudio/legacy/CamStudioCodec_1.5_Setup.exe#/codec.7z_"
+    ],
+    "hash": [
+        "3fbe450e3267799f372a775e4c7e7c47c8bbffeebb0bd0c9690e63374eba8ba0",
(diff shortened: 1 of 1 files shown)


bucket/camstudio.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "2.7.4-r354",
+    "description": "Free desktop recorder and video streaming software",
+    "homepage": "https://camstudio.org/",
+    "license": "GPL-3.0-or-later",
```
tech189, line 5: "The version available from this website is developed independently under the terms of the GNU Public License V2."
Footer of the website says this, but then in the source code it says v3:
https://sourceforge.net/p/camstudio/mercurial/ci/default/tree/CamStudio2.7/license.txt
What should we put?

bucket/camstudio.json
```diff
@@ -0,0 +1,30 @@
...
+    "url": [
+        "https://eu2.contabostorage.com/a4054a9f5df24ecf87c4a33ac9b6477b:camstudio/CamStudioSetup.exe#/dl.exe",
```
tech189, line 7: Any difference between this link and the other one on the homepage: https://cs07112018.s3.us-east-2.amazonaws.com/CamStudioSetup.exe ?

bucket/camstudio.json
issaclin32: `gpl-3.0.txt` is also packed in the binaries.

I think they updated the license to GPL-3.0, but didn't update the info on the website.

bucket/camstudio.json
issaclin32: Not sure, but I guess they are just different download locations.
One of the server is in the US, and another in Europe.

tech189:
Yes, it might be a bit out of scope for Scoop to determine the correct/nearest mirror based on location! I think the alternative buckets work well for now.

PR #8753, opened 2022-06-24
vcredist2022: Update version info

b4imetu:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/vcredist2022.json
+++ b/bucket/vcredist2022.json
@@ -1,5 +1,5 @@
 {
-    "version": "14.32.31326",
+    "version": "14.32.31326.0",
     "description": "Microsoft Visual C++ Redistributable for Visual Studio 2022",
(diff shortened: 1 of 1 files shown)


tech189:
Is there a way we can automatically get the latest version number rather than manually updating it?

b4imetu:
Sorry, I don't know how to implement it, I'm a rookie in this Scoop project.

But I feel that the Scoop script is very difficult to implement for `vcredist2022`'s auto-update because the usual solution is not feasible.

Because `"autoupdate":` usually requires a static URL and $version information in the package name, but the `vcredist2022` package does not have either of these features.

I just modified it manually according to the previous update method.

b4imetu:
Just a hotfix for #8732 Please merge at your convenience ^^ @issaclin32 @tech189

issaclin32:
I am changing `version` to 14.32.31326 **.0** according to the file version.

vc

PR #8774, opened 2022-06-29
libreoffice: Switch to MSI installer, Add 64-bit support

### LINKED ISSUE #2824: No fonts in libreoffice-stable
It seems that the `libreoffice-stable` package comes with a limited number of fonts. The [Portable Apps](https://portableapps.com/apps/office/libreoffice_portable) site suggests to add them from a zip package. I wonder if you can create a full stable package. TIA

### LINKED ISSUE #8771: LibreOffice 64-bit
## Package Request
LibreOffice 64-bit. The LO that this bucket has now is 32-bit.

### Criteria

For a package to be acceptable in the Extras bucket, it should be:

- [x] a well-known and widely used tool (e.g. if it's a GitHub project, it should have at least 100 stars and 50 forks)
- [x] the latest stable version of the program

### Information

**Name:** LibreOffice64

**Description:** 64-bit build of LibreOffice

**Homepage:** https://www.libreoffice.org/

**Download link(s):** https://www.libreoffice.org/donate/dl/win-x86_64/7.3.4/en-US/LibreOffice_7.3.4_Win_x64.msi

**Some indication of popularity/repute:** A worse version of this app is already in the bucket.

issaclin32:
* closes #8771
* closes #2824

* Switch to **MSI installer** because the portable version does not have **64-bit** support.

* In terms of version updates, The "portable" version lags behind "stable" version. For example, the latest "portable" is `7.3.2`, and latest "stable" is `7.3.4`. However we can see that "stable" version has 64-bit support since version 7.2. ([download site](https://download.documentfoundation.org/libreoffice/stable/)).

* config will be located at `$Env:AppData\LibreOffice` instead of `$dir\Data`.

* The MSI installer contains the **multi-lingual** version of *LibreOffice* (same as the portable version `LibreOfficePortable_7.3.2_MultilingualStandard.paf.exe`)

* **Bundled fonts** are usable in *LibreOffice* apps after we moved it to the correct location.

libre_fonts


tech189:
Does this also include the fonts that were missing from the portable version? See: https://github.com/ScoopInstaller/Extras/issues/2824

issaclin32:
> Does this also include the fonts that were missing from the portable version? See: #2824

The fonts are **not** included because `Expand-InnoArchive` only extracts `APPDIR` by default.
I think it would be a good idea to include them as well, because some of the features (e.g. formula in *LibreOffice math*) requires the fonts to display properly. However I am not sure about whether we should do that in this package.

There are 2 solutions I can think of:
**(1)** Install it in this package
When user installs/updates `libreoffice`, checks if all the fonts are installed. If some of the fonts are not installed, Then install the missing fonts. **(admin rights required)**

By the way, there is a method to install without admin rights, see [nerd-fonts bucket](https://github.com/matthewjberger/scoop-nerd-fonts) for examples. But, this will **only work** for Win10 after `version 1809`, and Win11 before `build 22621`.

**(2)** Put a seperate package `libreoffice-fonts` in the [nerd-fonts bucket](https://github.com/matthewjberger/scoop-nerd-fonts), and add `notes` to recommend user to install it.

tech189:
I found out from the following website that fonts can be used if placed in `$dir\LibreOffice\share\fonts\truetype`: https://ask.libreoffice.org/t/use-local-fonts-for-soffice-exe-headless-on-windows/54944/3. They will now show up in each of the programs, so no need for a separate package or admin rights!

bucket/libreoffice.json
```diff
@@ -1,62 +1,69 @@
...
-    "hash": "ae81da2a2bfb27a0cbdeffa65f712f8483a4f20049cbd1e02de2a36e6c3ff9d5",
     "pre_install": [
-        "if (!(Test-Path \"$persist_dir\\Data\\settings\")) {",
-        "    New-Item \"$dir\\Data\\settings\\LibreOfficePortableSettings.ini\" -Value \"[LibreOfficePortableSettings]`nInvalidPackageWarningShown=$version.0\" -Force | Out-Null",
-        "    if (Test-Path \"$env:AppData\\LibreOffice\") {",
-        "        info \"Copying old '$env:AppData\\LibreOffice' to '$persist_dir\\Data\\settings'\"",
-        "        Get-Item \"$env:AppData\\LibreOffice\\*\\*\" | Copy-Item -Destination \"$dir\\Data\\settings\" -Recurse -Force",
-        "    } else {",
-        "        Copy-Item \"$dir\\App\\DefaultData\\*\" \"$dir\\Data\" -Recurse -Force",
-        "    }",
```
tech189, line 19: Should we migrate persisted data back to `$env:AppData\LibreOffice` and delete from persist?

bucket/libreoffice.json
```diff
@@ -1,62 +1,69 @@
...
+            "hash": "509c70c1c8136805480146b55e4bad5dc73b11ee47b4682b43cf07670109e176"
+        },
+        "32bit": {
+            "url": "https://download.documentfoundation.org/libreoffice/stable/7.3.4/win/x86/LibreOffice_7.3.4_Win_x86.msi",
+            "hash": "e73e2fd3d95f305a8517abe65c940e2b7de13d4271fb789f38f89c28df57933a"
+        }
     },
-    "url": "https://download.documentfoundation.org/libreoffice/portable/7.3.2/LibreOfficePortable_7.3.2_MultilingualStandard.paf.exe#/dl.7z",
-    "hash": "ae81da2a2bfb27a0cbdeffa65f712f8483a4f20049cbd1e02de2a36e6c3ff9d5",
     "pre_install": [
```
tech189, line 19: Is there anything we're missing from the pre_install of `libreoffice-fresh` (the old package that used the .msi installer too)? Or is that all covered already?
https://github.com/ScoopInstaller/Extras/blob/613239a4913ea2b16a7f4fa9eac2c480968759bf/deprecated/libreoffice-fresh.json#L19-L27

bucket/libreoffice.json
```diff
@@ -1,62 +1,69 @@
...
         [
-            "LibreOfficeWriterPortable.exe",
+            "LibreOffice\\program\\swriter.exe",
             "LibreOffice\\LibreOffice Writer"
         ]
     ],
-    "persist": "Data",
     "checkver": {
-        "url": "https://download.documentfoundation.org/libreoffice/portable/?C=M;O=D",
+        "url": "https://download.documentfoundation.org/libreoffice/stable/?C=M;O=D",
```
tech189, line 55: `libreoffice-fresh`'s checkver updates to 7.3.4.2, which was released 3rd June 2022. Should we copy that checkver too?

[Chocolatey](https://community.chocolatey.org/packages/libreoffice-fresh) and [Homebrew](https://formulae.brew.sh/cask/libreoffice) stick to 7.3.4 but [WinGet](https://github.com/microsoft/winget-pkgs/tree/master/manifests/t/TheDocumentFoundation/LibreOffice/7.3.4.2) uses 7.3.4.2.

Checking for updates within LibreOffice 7.3.4 reports that "LibreOffice 7.3 is up to date."

bucket/libreoffice.json
issaclin32: I prefer not.

If we are doing it, we should only copy from persist if `$env:AppData\LibreOffice` is **empty**.
Otherwise this will happen:

(1) User installs version `7.3.2`:  Config copied from `$persist_dir` -> `$Env:AppData\LibreOffice`.
(2) User updates to version `7.3.4`: Config copied again from `$persist_dir` -> `$Env:AppData\LibreOffice`. Overwriting existing config that are set by the user during the use of version `7.3.2`.

bucket/libreoffice.json
issaclin32: > libreoffice-fresh's checkver updates to 7.3.4.2, which was released 3rd June 2022. Should we copy that checkver too?
Of course. We should track any minor update if possible, so that hash error / installtion error will not happen.

bucket/libreoffice.json
issaclin32: > Is there anything we're missing from the pre_install of `libreoffice-fresh` (the old package that used the .msi installer too)? Or is that all covered already?

Thanks for the details.

I think we should add this.
I tried to extract the file with `lessmsi`, and the app **is** located under `LibreOffice` folder.
This can avoid problems when user sets **MSIEXTRACT_USE_LESSMSI** to **true**. (Although it should be a rare case)

bucket/libreoffice.json
tech189: I agree. We could also leave a note to the user but only if there is data in `$persist_dir` (i.e. upgrading from portable version)

bucket/libreoffice.json
tech189: If we add this and only extract the `LibreOffice` folder will we not get the fonts in the `Fonts` folder?

bucket/libreoffice.json
tech189: Done. I modified the checkver based on `libreoffice-fresh`'s checkver. It uses `.../latest/win/x86_64/...` to get the version number, hopefully that will be okay.

issaclin32:
**Bundled fonts** are usable in *LibreOffice* apps after we moved it to the correct location.
libre_fonts

issaclin32:
@tech189 I modified *pre_install* script to make it simpler, and tested installation with `MSIEXTRACT_USE_LESSMSI` both **on** and **off**. Does this look good to you?

tech189:
Almost! When you install with `MSIEXTRACT_USE_LESSMSI` set to `false`, the Fonts folder is duplicated - it is both in `$dir\LibreOffice\Fonts` and in `$dir\LibreOffice\share\fonts\truetype`. It only needs to be in the second folder. I think you just need to add the Fonts folder to the Remove-Item command on line 27.

tech189 (approved):

issaclin32:
I just tested again to make sure everything works fine.
Thank you for help reviewing the script!

PR #8798, opened 2022-07-04
zlib: Add version 1.2.12

issaclin32:
**[zlib](http://www.zlib.net/)** is a massively spiffy yet delicately unobtrusive compression library. Unencumbered by patents. It is being used in many well-known apps including *FFMpeg*, *MiKTex*, *caffe*, *PyTorch*, etc.

* This package is mainly for users who are building **projects that depends on zlib**.

* [The zlib license](https://spdx.org/licenses/Zlib.html) is an SPDX license.

* I did not add `$dir\bin` (which contains `zlib.dll`) to PATH because the file should be **bundled** with the app that is using it. Having `zlib.dll` in PATH may lead to some confusing situations.

* This package has to be **manually updated**, but I think it's fine because *zlib* does not receive updates very often. They have a recent bug fix in **2022.3**. The update before that one was in **2017**.

* `register.reg` is for **CMake** and other build tools to find *zlib*. The key needs to be in `HKLM`. See [CMake's FindZLIB](https://cmake.org/cmake/help/latest/module/FindZLIB.html) for details.

* I built the binaries with the following steps:
(1) Download zlib source code (http://www.zlib.net/zlib1212.zip) and extract it.
(2) Open the folder `zlib-1.2.12` with *cmake-gui*. Select VS2022 as the compiler.
(3) Set build location to `zlib-1.2.12/win32`, and set both `AMD64` and `ASM686` switches **off**.
(4) Press 'Configure' and 'Generate'
(5) Open VS2022 dev command prompt, and run `msbuild zlib-1.2.12\win32\zlib.sln /property:Configuration=Release`
(6) Orginize the product:
Put `zlib.dll` under `bin` folder.
Put `.lib` and `.exp` files under `lib` folder.
Copy `.h` files from source code, and put them under `include` folder.

### DIFF
--- a/bucket/zlib.json
+++ b/bucket/zlib.json
@@ -12,8 +12,8 @@
     "post_install": [
         "$dirpath = \"$dir\".Replace('\\', '\\\\')",
         "'register', 'unregister' | ForEach-Object {",
-        "  if (Test-Path \"$bucketsdir\\local\\scripts\\zlib\\$_.reg\") {",
-        "    $content = Get-Content \"$bucketsdir\\local\\scripts\\zlib\\$_.reg\"",
+        "  if (Test-Path \"$bucketsdir\\extras\\scripts\\zlib\\$_.reg\") {",
+        "    $content = Get-Content \"$bucketsdir\\extras\\scripts\\zlib\\$_.reg\"",
         "    $content = $content.Replace('$dir', $dirpath)",
         "    $content | Set-Content -Path \"$dir\\$_.reg\"",
         "  }",
--- a/bucket/zlib.json
+++ b/bucket/zlib.json
@@ -1,0 +1,32 @@
+{
+    "version": "1.2.12",
+    "description": "Massively spiffy yet delicately unobtrusive compression library. Unencumbered by patents.",
+    "homepage": "http://www.zlib.net/",
+    "license": "Zlib",
+    "url": "https://raw.githubusercontent.com/ScoopInstaller/Binary/master/zlib/zlib-1.2.12-windows-32bit-vc14.2.7z",
+    "hash": "0d7245a2e499320cb4803b3889df8a9051d2dfc0c53d2908a3d99b687943c68b",
+    "notes": [
+        "For CMake (and other build tools) to find this installation of zlib, run:",
+        "\"$dir\\register.reg\""
+    ],
+    "post_install": [
+        "$dirpath = \"$dir\".Replace('\\', '\\\\')",
+        "'register', 'unregister' | ForEach-Object {",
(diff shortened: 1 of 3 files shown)


tech189 (commented):
I'm not quite sure how to review this, I don't have VS installed so I can't easily try building the binary nor do I do any programming with C to test it - sorry!

bucket/zlib.json
```diff
@@ -0,0 +1,32 @@
...
+    "pre_uninstall": [
+        "if ($cmd -eq 'uninstall') {",
+        "    $key = Get-ItemProperty -Path 'HKLM:\\SOFTWARE\\GnuWin32\\Zlib' -Name 'InstallPath' -ErrorAction SilentlyContinue",
+        "    if ($key.InstallPath.Contains('apps\\zlib\\current')) {",
+        "        if (!(is_admin)) { error \"$app requies admin rights to unregister the installation path\"; break }",
+        "        Invoke-ExternalCommand reg -ArgumentList @('import', \"$dir\\unregister.reg\") -RunAs | Out-Null",
+        "    }",
+        "}"
+    ],
+    "checkver": "zlib ([\\d.]+)</B>"
```
rashil2000, line 31:
```suggestion
    ]
```

Checkver can be removed, otherwise it will consume an Excavator cycle

bucket/zlib.json
```diff
@@ -0,0 +1,31 @@
...
+        "    $content = Get-Content \"$bucketsdir\\extras\\scripts\\zlib\\$_.reg\"",
+        "    $content = $content.Replace('$dir', $dirpath)",
+        "    $content | Set-Content -Path \"$dir\\$_.reg\"",
+        "  }",
+        "}"
+    ],
+    "pre_uninstall": [
+        "if ($cmd -eq 'uninstall') {",
+        "    $key = Get-ItemProperty -Path 'HKLM:\\SOFTWARE\\GnuWin32\\Zlib' -Name 'InstallPath' -ErrorAction SilentlyContinue",
+        "    if ($key.InstallPath.Contains('apps\\zlib\\current')) {",
```
tech189, line 25:
```
Running pre_uninstall script...
InvalidOperation:
Line |
   3 |      if ($key.InstallPath.Contains('apps\zlib\current')) {
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | You cannot call a method on a null-valued expression.
Unlinking ~\scoop\apps\zlib\current
```
Getting this error on uninstallation.

issaclin32: Fixed. Thank you!

rashil2000 (approved):

PR #8808, opened 2022-07-06
selene-light: Add version 0.19.1

sitiom:
Selene without Roblox-tailored lints.

### DIFF
--- a/bucket/selene-light.json
+++ b/bucket/selene-light.json
@@ -1,0 +1,23 @@
+{
+    "version": "0.19.1",
+    "description": "A blazing-fast modern Lua linter written in Rust, without Roblox-tailored lints.",
+    "homepage": "https://kampfkarren.github.io/selene/",
+    "license": "MPL-2.0",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


rashil2000:
Is there any significant advantage of the light version? There isn't too much difference in size and I'd prefer to keep just one package (which has all features/lints).

sitiom:
> Is there any significant advantage of the light version? There isn't too much difference in size and I'd prefer to keep just one package (which has all features/lints).

Other than no Roblox lints, it's just the same. The light version is built with `--no-default-features`, and is almost half the full version's size. Maybe we can just provide `selene-light` as an option?

rashil2000:
> The light version is built with `--no-default-features`, and is almost half the full version's size. 

Yeah, but they're both less than 3 MB so size is not a factor here.

> Maybe we can just provide `selene-light` as an option?

I didn't get you?

sitiom:
Makes sense, they are almost identical and both are small in size. The developer plans to expand selene with more features only available in the full version, but that's for a future time. We could add it here at a later time when it makes sense.

PR #8858, opened 2022-07-12
ffqueue: persist ffqueue.cfg

candrapersada:
Closes #XXXX

Relates to #XXXX

- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/ffqueue.json
+++ b/bucket/ffqueue.json
@@ -23,6 +23,12 @@
             "FFQueue Documentation"
         ]
     ],
+    "pre_install": [
+    "$manifest.persist | ForEach-Object {",
+        "    if (-not (Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" -ItemType file | Out-Null }",
+        "}"
+    ],
+    "persist": "ffqueue.cfg",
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

issaclin32:
Thanks!

candrapersada:
image

issaclin32:
> why persist ffqueue in delete?
and how to use persist ffqueue new?

Because `ffqueue.cfg` has its **unique header**. If the header is not found in the file, then *FFQueue* will report error and crash. So we can only **copy it before uninstalling** to preserve the file.

We use `New-Item` in *pre_install* because `ffqueue.cfg` does not exist in the archive, so we need to create it first so that we can **persist** it.
`New-Item` creates an **empty file**. This would work fine for most apps, but not for *FFQueue*.

Another solution is to use `Set-Content` to set the initial content of the config file. This method can be useful for apps requires something like "a proper JSON" or "a proper INI", but I would **not recommend it** for the case of *FFQueue*, because the header may change in the future.

PR #8870, opened 2022-07-14
SpotX: Add version 1.2

NeoFoxxo:
SpotX blocks ads and updates for the desktop version of Spotify, disabling podcasts and more.


issaclin32:
I think this app should go to the [Nonportable bucket](https://github.com/ScoopInstaller/Nonportable).

The script downloads and installs Spotify **via Spotify's installer**, and we cannot simply uninstall the app by `scoop uninstall spotx`.

PR #8911, opened 2022-07-23
AppFlowy: Fix extract_dir

b4imetu:
Fix `Creating shortcut for AppFlowy (app_flowy.exe) failed: Couldn't find D:\Scoop\apps\appflowy\current\app_flowy.exe`


tech189 (commented):
How about instead of a redundant subfolder "AppFlowy" we could move all of the files into the root `$dir` with `"extract_dir": "AppFlowy",`? Then we don't need to change the shortcut.

tech189 (approved):

PR #8948, opened 2022-07-29
scoop-completion: add auto-import scripts

AkariiinMKII:
- Add a function module to modify PS profile
- Auto-import scoop-completion module to PS profile 
- Remove residual folder

Referenced from dorado bucket. I'm not familiar with function modules for PowerShell, edits from old hands would be appreciated. : )

### DIFF
--- a/bucket/scoop-completion.json
+++ b/bucket/scoop-completion.json
@@ -3,6 +3,10 @@
     "description": "A Scoop tab completion module for PowerShell",
     "homepage": "https://github.com/Moeologist/scoop-completion",
     "license": "MIT",
+    "notes": [
+        "Run following command to import scoop-completion automatically in Powershell:",
+        "$dir\\append-to-profile.ps1"
+    ],
     "url": "https://github.com/Moeologist/scoop-completion/archive/v0.2.3.zip",
     "hash": "59fcda6e723b2c0bf7430668ccfb57086d82e9472733a01f9f7f9315768b5e5f",
     "extract_dir": "scoop-completion-0.2.3\\src",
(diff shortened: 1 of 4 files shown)


issaclin32:
In official buckets, we tend to leave this **as an option**. Some users may not want `scoop-completion` to be automatically loaded, because they have other modules that may conflict with it.

I would suggest adding `$dir\add-to-profile.ps1`, and `$dir\remove-from-profile.ps1`,
and use something like:
```
"notes": "Run \"$dir\\add-to-profile.ps1\" to import scoop-completion automatically in Powershell",
...
...
...
"pre_uninstall": "&\"$dir\\remove-from-profile.ps1\"",
```

AkariiinMKII:
> In official buckets, we tend to leave this **as an option**. Some users may not want `scoop-completion` to be automatically loaded, because they have other modules that may conflict with it.
> 
> I would suggest adding `scripts\scoop-completion\add-to-profile.ps1` and `scripts\scoop-completion\remove-from-profile.ps1`,
> and use something like:
> ```
> "notes": "Run \"$dir\\add-to-profile.ps1\" to import scoop-completion automatically in Powershell",
> ...
> ...
> "pre_install": [
>     "'add-to-profile.ps1', 'remove-from-profile.ps1' | ForEach-Object {",
>     "  if (Test-Path \"$bucketsdir\\extras\\scripts\\scoop-completion\\$_\") {",
>     "      Copy-Item \"$bucketsdir\\extras\\scripts\\scoop-completion\\$_\" \"$dir\\\"",
>     "  }",
>     "}",
> ]
> ...
> ...
> "pre_uninstall": "&\"$dir\\remove-from-profile.ps1\"",
> ```

In this situation, I recommend running both add and remove script manually, otherwise the add-to-profile script should run each time the module is updated.
As for the function module, it should be compatible with the whole bucket, how about keeping it for common usage?

issaclin32:
Oh. I forgot to add this in `pre_uninstall`.
```
"pre_uninstall": "if ($cmd -eq 'uninstall') { &\"$dir\\remove-from-profile.ps1\" }"
```
I this way, the user will not need to run *add-to-profile script* during updates.

AkariiinMKII:
> Oh. I forgot to add this in `pre_uninstall`.
> ```
> "pre_uninstall": "if ($cmd -eq 'uninstall') { &\"$dir\\remove-from-profile.ps1\" }"
> ```
> I this way, the user will not need to run *add-to-profile script* during updates.

Great 👍 I'll try this way

issaclin32:
> As for the function module, it should be compatible with the whole bucket, how about keeping it for common usage?

I am fine with keeping it for common usage.

issaclin32:
I think we should copy `ModifyPSProfile.psm1` to **$dir** as well.
If there is some update in `ModifyPSProfile.psm1`, this can avoid potential problems in old installations.

AkariiinMKII:
> I think we should copy `ModifyPSProfile.psm1` to **$dir** as well.
> If there is some update in `ModifyPSProfile.psm1`, this can avoid potential problems in old installations.

Placing the module file in scripts directory makes it able to receive hot fix by updating bucket, I think it might be better.

issaclin32:
> Placing the module file in scripts directory makes it able to receive hot fix by updating bucket, I think it might be better.

Okay. Let's leave it this way.

issaclin32 (approved):

PR #9015, opened 2022-08-08
librewolf: added extra config files to persist, and minor hash fix

brian6932:
also fixed minor mistake in hash

Those config files/dirs are required by userchrome configurations, so it makes sense to persist them

### DIFF
--- a/bucket/librewolf.json
+++ b/bucket/librewolf.json
@@ -13,9 +13,9 @@
         "$lw_name = (Get-Item \"$dir\\librewolf-*\").FullName",
         "Move-Item \"$lw_name\\*\" \"$dir\"",
         "Remove-Item \"$lw_name\"",
-        "if (Test-Path \"$dir\\Librewolf\\*.js\") { Copy-Item \"$dir\\Librewolf\\*.js\" \"$persist_dir\\\" }"
+        "if (Test-Path \"$persist_dir\\Librewolf\\*.js\") { Copy-Item \"$persist_dir\\Librewolf\\*.js\" \"$dir\\Librewolf\\\" }"
     ],
-    "pre_uninstall": "if (Test-Path \"$persist_dir\\Librewolf\\*.js\") { Copy-Item \"$persist_dir\\Librewolf\\*.js\" \"$dir\\\" }",
+    "pre_uninstall": "if (Test-Path \"$dir\\Librewolf\\*.js\") { Copy-Item \"$dir\\Librewolf\\*.js\" \"$persist_dir\\Librewolf\\\" }",
     "bin": [
         [
             "librewolf-portable.exe",
--- a/bucket/librewolf.json
+++ b/bucket/librewolf.json
@@ -1,12 +1,12 @@
 {
-    "version": "103.0.1-1",
+    "version": "103.0.1-2",
     "description": "A fork of Firefox, focused on privacy, security and freedom.",
     "homepage": "https://librewolf.net/",
     "license": "MPL-2.0",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

issaclin32:
Thanks for the fix.
Could you provide some information/documentation about `default` and `config.js`?

*Scoop* uses the portable version of *LibreWolf*, and there is `(INSTALL DIRECTORY)\Librewolf\defaults`.
I am not sure whether you are trying to persist
**(1)** `(INSTALL DIRECTORY)\Librewolf\defaults` and `(INSTALL DIRECTORY)\Librewolf\config.js`
, or just **(2)** `(INSTALL DIRECTORY)\defaults` and `(INSTALL DIRECTORY)\config.js`

Also, since `config.js` does not exist in the archive, we need to create it first by:
```
"pre_install": "if (!(Test-Path \"$persist_dir\\config.js\")) { New-Item \"$dir\\config.js\" | Out-Null }"
```
Otherwise *Scoop* will create `config.js` as a folder by default.

brian6932:
> Could you provide some information/documentation about `default` and `config.js`?

https://support.mozilla.org/en-US/kb/customizing-firefox-using-autoconfig

> I am not sure whether you are trying to persist
> **(1)** `(INSTALL DIRECTORY)\Librewolf\defaults` and `(INSTALL DIRECTORY)\Librewolf\config.js`
> , or just **(2)** `(INSTALL DIRECTORY)\defaults` and `(INSTALL DIRECTORY)\config.js`

Fixed, also I should've specified `defaults/prefs` as a nested dir

> Also, since `config.js` does not exist in the archive, we need to create it first by:
> 
> ```
> "pre_install": "if (!(Test-Path \"$persist_dir\\config.js\")) { New-Item \"$dir\\config.js\" | Out-Null }"
> ```
> 
> Otherwise _Scoop_ will create `config.js` as a folder by default.

Added, minor correction, it does not have to be `config.js`, any `*.js` file should work as long as the name is set as `general.config.filename` in a `*.js` file in `defaults/prefs`, is it possible to use a wildcard with persist?

bucket/librewolf.json
```diff
@@ -26,7 +27,11 @@
             "LibreWolf"
         ]
     ],
-    "persist": "Profiles",
+    "persist": [
+        "Profiles",
+        "LibreWolf\\defaults\\prefs",
+        "LibreWolf\\*.js"
```
issaclin32, line 33:
```suggestion
        "LibreWolf\\defaults\\prefs"
```

**persist** does not support wildcards.
If order to save `*.js`, we need to copy it in `pre_install` and `pre_uninstall` like:

```
"pre_install": [
    "......",
    "if (Test-Path \"$dir\\Librewolf\\*.js\") { Copy-Item \"$dir\\Librewolf\\*.js\" \"$persist_dir\\\" }"
]
"pre_uninstall": "if (Test-Path \"$persist_dir\\Librewolf\\*.js\") { Copy-Item \"$persist_dir\\Librewolf\\*.js\" \"$dir\\\" }"
```

brian6932:
Uhh, it seems like https://gitlab.com/librewolf-community/browser/windows/uploads/c5782ae771d4ded03bdfc2f4a3850264/sha256sums.txt literally does not exist in https://gitlab.com/api/v4/projects/13852981/releases, should hash verification be removed all together?

brian6932:
ok nvm, I fixed it

brian6932:
why not?

brian6932:
oh the id is different

brian6932:
so apparently the order of the id and sha changes in the array, your solution is best 👍

issaclin32:
Looks fine now. Thank you for the PR.

issaclin32 (approved):

PR #9030, opened 2022-08-11
vcredist*: Fixed ArgumentList in post_install scripts.

ErtugrulSener:
Hi guys, this is my first MR in this repo, please advice me on anything I am doing wrong.
I had the issue today, that the ArgumentList that was passed to the vcredist2013 package was ignored (/quiet),
so scoop was always opening the prompt to accept the license informations to install.

That halts my automation script and I realized this behaviour is not expected, it's a small bug in the post_install script.
So.. there is the fix!

### DIFF
--- a/bucket/vcredist2008.json
+++ b/bucket/vcredist2008.json
@@ -21,7 +21,7 @@
         "    1638 = 'This product is already installed';",
         "    3010 = 'A restart is required to complete the installation';",
         "}",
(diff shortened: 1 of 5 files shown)


niheaven:
Does https://github.com/ScoopInstaller/Scoop/pull/5065 break this? @lewis-yeung

niheaven (approved):

ErtugrulSener:
<3

lewis-yeung:
@niheaven Yes but no, the current behaviors are exactly what `ArgumentList` should have.

PR #9033, opened 2022-08-11
Update the hash for the new version of anydesk

nothingbutlucas:
Anydesk update to version 7.0.14
This is the new hash of the exe file


issaclin32:
Thanks for the fix.
However the hash was fixed by the bot [during version update](https://github.com/ScoopInstaller/Extras/commit/4410a5891468f82570ecffdf8700cbec9a85c1b0), so this PR can be closed.

PR #9084, opened 2022-08-23
vscode, vscodium: fix reg import cmds in notes

lunarwhite:

Users cannot import reg using previous commands in `"notes"`

### DIFF
--- a/bucket/vscode.json
+++ b/bucket/vscode.json
@@ -7,8 +7,8 @@
         "url": "https://code.visualstudio.com/License/"
     },
     "notes": [
-        "Add Visual Studio Code as a context menu option by running: '$dir\\install-context.reg'",
-        "For file associations, run '$dir\\install-associations.reg'"
+        "Add Visual Studio Code as a context menu option by running: 'reg import \"$dir\\install-context.reg\"'",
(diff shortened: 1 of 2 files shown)


tech189 (commented):
I think this is okay since other apps already have this but note that it bypasses the warning that would come up if you ran/double-clicked the reg file:
```
Adding information can unintentionally change or delete values and cause components
to stop working correctly. If you do not trust the source of this information in
~\scoop\apps\vscodium\current\install-context.reg, do not add it to the registry.

Are you sure you want to continue?

[Yes] [No]
```

lunarwhite:
> I think this is okay since other apps already have this but note that it bypasses the warning that would come up if you ran/double-clicked the reg file:
> 
> ```
> Adding information can unintentionally change or delete values and cause components
> to stop working correctly. If you do not trust the source of this information in
> ~\scoop\apps\vscodium\current\install-context.reg, do not add it to the registry.
> 
> Are you sure you want to continue?
> 
> [Yes] [No]
> ```

Agree. But as a user with little usage or dev experience, it‘s quite confusing.

So I suppose it's essential to have enough information about `notes` in every mainfest file.

rasa (approved):

PR #9110, opened 2022-08-26
Generate PowerShell profile modifier without templete files

AkariiinMKII:
This PR is related to #8948 , which added a PowerShell module. 
The module is aimed to provide a tool for each psmodule-like app to edit PowerShell profile, but all of them needs templet files created in advance, so I added a function to generate files without template.

### DIFF
--- a/bucket/scoop-completion.json
+++ b/bucket/scoop-completion.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "notes": [
         "Run following command to import scoop-completion automatically in Powershell:",
-        "$dir\\append-to-profile.ps1"
+        "$dir\\add-profile-content.ps1"
     ],
     "url": "https://github.com/Moeologist/scoop-completion/archive/v0.2.3.zip",
     "hash": "59fcda6e723b2c0bf7430668ccfb57086d82e9472733a01f9f7f9315768b5e5f",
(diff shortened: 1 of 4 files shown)


bucket/scoop-completion.json
```diff
@@ -16,21 +16,21 @@
...
-        "'append-to-profile.ps1', 'remove-from-profile.ps1' | ForEach-Object {",
-        "    if ((Test-Path \"$ScriptsDir\\$_\") -and (Test-Path \"$ModifierPath\")) {",
-        "        Copy-Item \"$ScriptsDir\\$_\" \"$dir\\\"",
-        "        $Content = (Get-Content \"$dir\\$_\").Replace('$ModifyPSProfile', \"$ModifierPath\")",
-        "        $content | Set-Content -Path \"$dir\\$_\"",
-        "    } else {",
-        "        Write-Host 'Missing files, please update scoop buckets and reinstall this module again' -ForegroundColor Red",
-        "    }",
+        "$BucketDir = Find-BucketDirectory -Root -Name extras",
+        "$UtilsPath = $BucketDir | Join-Path -ChildPath \"scripts\\ModifyPSProfile.psm1\"",
```
issaclin32, line 20:
```suggestion
        "$UtilsPath = $BucketDir | Join-Path -ChildPath \"scripts\\ModifyPSProfile.psm1\"",
        "Unblock-File $UtilsPath",
```
This can avoid some permission issues

scripts/ModifyPSProfile.psm1
```diff
@@ -35,9 +98,15 @@ function RemovefromProfile {
         [string] $Content
     )
 
-    ((Get-Content -Path $profile -raw) -replace "[\r\n]*$content",'').trim() | Set-Content $profile -NoNewLine
+    ((Get-Content -Path $PROFILE -raw) -replace "[\r\n]*$Content",'').trim() | Set-Content $PROFILE -NoNewLine
```
issaclin32, line 101:
```suggestion
    if ($null -eq (Get-Content -Path $PROFILE -raw)) { return }
    ((Get-Content -Path $PROFILE -raw) -replace "[\r\n]*$Content",'').Trim() | Set-Content $PROFILE -NoNewLine
```
When testing this script, I noticed that when the content in `$PROFILE` is empty or only contains spaces and line seperators, the return value of `(Get-Content...)` will be **$null** rather than an empty string.
This adds an extra check to avoid errors.

By the way, [this is why I put $null in the left](https://evotec.xyz/the-curious-case-of-null-should-be-on-the-left-side-of-equality-comparisons-psscriptanalyzer/).

AkariiinMKII:
@issaclin32 Thanks, the script might be OK now.

issaclin32 (approved):

PR #9114, opened 2022-08-27
vnote: Update to version 3.15.0

Trim21:
in previous PR https://github.com/ScoopInstaller/Extras/pull/9080 I update the file url for 3.14.0 only, but didn't find that the file hash also changed.

🤔 if i recall correctly there is a CI to check file hash?


issaclin32:
> if i recall correctly there is a CI to check file hash?

It can be invoked with `/verify`. I forgot to do that.

PR #9118, opened 2022-08-28
sunshine: Fix persist

sitiom:
The `config` folder contains template files that are copied to the root folder when running Sunshine for the first time.


issaclin32:
Since these files does not exist at the beginning (but rather created upon first run)
We need to create the files in `pre_install`, otherwise *Scoop* will persist them as folders. (by default)

```
"'apps.json', 'sunshine.conf' | ForEach-Object {"
"}",
```

sitiom:
It would be better to copy the default configs from the `config` folder:
```json
"'apps.json', 'sunshine.conf' | ForEach-Object {",
"    if (!(Test-Path \"$persist_dir\\$_\")) { Copy-Item \"$dir\\config\\$_\" \"$dir\\$_\" | Out-Null }",
"}",
```

issaclin32:
> It would be better to copy the default configs from the `config` folder (which is what Sunshine does):

Oh. I see.

PR #9165, opened 2022-09-05
quarto@v1.1.189: Use new `quarto.exe` binary

cderv:
This PR modies the `bin` property to use the new `quarto.exe` binary now included in the bundle. This new exe file solves some issues in some system and is the recommended way to call quarto now.


niheaven:
Why not move it to 'main'?

cderv:
I don't know about the rules between Extras and Main bucket. `quarto` is CLI only tool if it matters. I would be glad to see it in main

PR #9228, opened 2022-09-13
simple-dnscrypt: add version 0.7.1

KrisSchneider:
Created simple-dnscrypt.json


issaclin32:
I think we can just name the package as `simple-dnscrypt`. Same as the window title.
title

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,68 @@
...
+    "suggest": {
+        "vcredist2017": "extras/vcredist2017"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/0.7.1/SimpleDNSCrypt64.msi",
+            "hash": "5BA0C94774ACD7ADE18446B4351E7D28A948C490AEFC50DC6C08C13551F19F7F",
+            "post_install": [
+                "if (Test-Path \"$dir\\SimpleDNSCrypt64.msi\") {Remove-Item \"$dir\\SimpleDNSCrypt64.msi\" | Out-Null}"
+            ]
```
issaclin32, line 15: I don't understand why this `post_install` section is needed.
Is the MSI file not being removed after extraction?

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,68 @@
...
+                ]
+            },
+            "32bit": {
+                "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/$version/SimpleDNSCrypt.msi",
+                "post_install": [
+                    "if (Test-Path \"$dir\\SimpleDNSCrypt.msi\") {Remove-Item \"$dir\\SimpleDNSCrypt.msi\" | Out-Null}"
+                ]
+            }
+        }
+    }
```
issaclin32, line 66:
```suggestion
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/$version/SimpleDNSCrypt64.msi"
            },
            "32bit": {
                "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/$version/SimpleDNSCrypt.msi"
            }
        }
    }
```

`autoupdate` -> `post_install` is only needed if there is something in *post_install* that has to be modified when version changes. (For example, if the program is named something like `someapp-2.1.3.exe`)

If the script has `64bit` -> `post_install`, but has no `autoupdate` -> `64bit` -> `post_install`, then, the existing `64bit` -> `post_install` will **remain unchanged between updates**.

KrisSchneider: Do you mean it is unnecessary to have that same 2 scripts twice? Does that mean I don't need the `post_install` scripts in the `auto_update` section?

bucket/simple-dnscrypt.json
KrisSchneider: Yes the **.msi** is not being removed after installation. Which is why I have that `post_install` script to remove it.

KrisSchneider:
> I think we can just name the package as `simple-dnscrypt`. Same as the window title.
> 
> title
> 
> 

That works for me. I think that name will be easier to remember.

bucket/simple-dnscrypt.json
issaclin32: > Does that mean I don't need the `post_install` scripts in the `autoupdate` section?

Yes. There is no need to have `post_install` scripts in the `autoupdate` section, unless `post_install` contains something has to be changed when "version" changes.

bucket/simple-dnscrypt.json
issaclin32: Weird...This should not be happening.

To reassure, I tried installing the manifest (without `post_install`) on my own PC and a VM, and both of them didn't have the problem.

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,62 @@
+{
+    "version": "0.7.1",
+    "description": "An easy to use GUI for dnscrypt-proxy",
+    "homepage": "https://simplednscrypt.org/",
+    "license": "MIT",
+    "suggest": {
+        "vcredist2017": "extras/vcredist2017"
```
issaclin32, line 7:
```suggestion
        "vcredist2015-2022": "extras/vcredist2022"
```
[VCRedist2022 now covers 2015-2022](https://docs.microsoft.com/en-US/cpp/windows/latest-supported-vc-redist?view=msvc-170)

bucket/simple-dnscrypt.json
KrisSchneider: Oh maybe I was just being paranoid again. You can remove it from the manifest if you want to.

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,62 @@
...
+            "post_install": [
+                "if (Test-Path \"$dir\\SimpleDNSCrypt.msi\") {Remove-Item \"$dir\\SimpleDNSCrypt.msi\" | Out-Null}"
+            ]
+        }
+    },
+    "pre_install": [
+        "if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
+        "",
+        "if (!(Test-Path \"$env:LOCALAPPDATA\\SimpleDnsCrypt\")) {New-Item -Path \"$env:LOCALAPPDATA\" -Name \"SimpleDnsCrypt\" -ItemType \"directory\" -Force | Out-Null}"
+    ],
```
issaclin32, line 29:
```suggestion
```

I tried installing version 0.7.0, then update to version 0.7.1, and the config  is being tranferred without a problem, so I think we can just let app handle it.

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,62 @@
...
+    "persist": [
+        "dnscrypt-proxy\\"
+    ],
+    "pre_uninstall": [
+        "if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
+        "",
+        "if (Get-Process -Name \"SimpleDnsCrypt\") {Stop-Process -Name \"SimpleDnsCrypt\" | Out-Null}",
+        "if (Get-Service -Name \"dnscrypt-proxy\") {Stop-Service -Name \"dnscrypt-proxy\"| Out-Null; Remove-Service -Name \"dnscrypt-proxy\" | Out-Null}",
+        "if (Test-Path \"$dir\") {Remove-Item \"$dir\" -Force -Recurse | Out-Null}",
+        "if (Test-Path \"$persist_dir\") {Remove-Item \"$persist_dir\" -Force -Recurse | Out-Null}"
```
issaclin32, line 46:
```suggestion
```

I don't think this is needed. These folders should be automatically removed or persisted.

bucket/simple-dnscrypt.json
KrisSchneider: The reason I have it there is because on my end, the app fails to start without this folder in appdata/local.

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,62 @@
...
+    "shortcuts": [
+        [
+            "SimpleDnsCrypt.exe",
+            "Simple DnsCrypt"
+        ]
+    ],
+    "bin": "SimpleDnsCrypt.exe",
+    "persist": [
+        "dnscrypt-proxy\\"
+    ],
```
issaclin32, line 39:
```suggestion
    "persist": "dnscrypt-proxy\\dnscrypt-proxy.toml",
```
If we *persist* the whole folder, it will block the update of `dnscrypt-proxy` when user updates the app.

I tried tweaking the config, and found the only file that is related to user config is `dnscrypt-proxy.toml`. Other files such as `public-resolvers.md` is updated according to the data on the app's server.

bucket/simple-dnscrypt.json
KrisSchneider: > ```suggestion
> 
> ```
> 
> 
> 
> I don't think this is needed. These folders should be automatically removed or persisted.

When I tried uninstalling the app without that part of the script, I sometimes get an error that says `ERROR couldn't remove '~\scoop\apps\simple-dnscrypt\current' it may be in use`

bucket/simple-dnscrypt.json
issaclin32: > If your not having any problems, then maybe I was just being paranoid again.

Okay. I will remove it from the manifest.

Probably something like antivirus or indexing services (such as [Everything](https://github.com/ScoopInstaller/Extras/blob/master/bucket/everything.json)) is blocking the file.

bucket/simple-dnscrypt.json
issaclin32: Okay, Let's leave this command, but I don't think we need **admin rights** to create new folders in `$env:LOCALAPPDATA`

bucket/simple-dnscrypt.json
KrisSchneider: Hey Isaac, I going to be heading to bed now. I will tell you when my response to the issue later.

issaclin32: > Hey Isaac, I going to be heading to bed now. I will tell you when my response to the issue later.

No worries. We are not in a rush.

bucket/simple-dnscrypt.json
KrisSchneider: Ok I just tested it, it turns out you're right. I will get rid of the `(!(is_admin))` command.

bucket/simple-dnscrypt.json
KrisSchneider: I fixed it, check the latest commit.

bucket/simple-dnscrypt.json
KrisSchneider: Probably could be the **Everything.exe** service I run on my computer.

rasa (approved):
Replaces https://github.com/rasa/scoops/blob/master/bucket/simple-dnscrypt.json

KrisSchneider:
Thanks @rasa

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,63 @@
...
+        "dnscrypt-proxy\\blacklist.txt",
+        "dnscrypt-proxy\\blocked.log",
+        "dnscrypt-proxy\\query.log",
+        "dnscrypt-proxy\\domain-blacklist.txt",
+        "dnscrypt-proxy\\domain-whitelist.txt"
+    ],
+    "pre_uninstall": [
+        "if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
+        "",
+        "Start-Process \"$dir\\Uninstall.exe\" | Out-Null",
```
issaclin32, line 45:
```suggestion
        "Start-Process -Wait \"$dir\\Uninstall.exe\" | Out-Null",
```

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,63 @@
...
+            "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/0.7.1/SimpleDNSCrypt64.msi",
+            "hash": "5BA0C94774ACD7ADE18446B4351E7D28A948C490AEFC50DC6C08C13551F19F7F"
+        },
+        "32bit": {
+            "url": "https://github.com/bitbeans/SimpleDnsCrypt/releases/download/0.7.1/SimpleDNSCrypt.msi",
+            "hash": "CAC6B8119AF3DA70A5977A2E73EE5214303990FEDD7ABBC5F247A1E2CF19A663"
+        }
+    },
+    "pre_install": [
+        "if (!(Test-Path \"$env:LOCALAPPDATA\\SimpleDnsCrypt\")) {New-Item -Path \"$env:LOCALAPPDATA\" -Name \"SimpleDnsCrypt\" -ItemType \"directory\" -Force | Out-Null}",
```
issaclin32, line 20:
```suggestion
        "if (!(Test-Path \"$env:LOCALAPPDATA\\SimpleDnsCrypt\")) { New-Item \"$env:LOCALAPPDATA\\SimpleDnsCrypt\" -ItemType directory -Force | Out-Null }",
```

This looks cleaner

issaclin32, line 20:
```suggestion
        "if (!(Test-Path \"$env:LOCALAPPDATA\\SimpleDnsCrypt\")) { New-Item \"$env:LOCALAPPDATA\\SimpleDnsCrypt\" -ItemType Directory -Force | Out-Null }",
```
This looks cleaner

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,63 @@
...
+        "New-Item \"$dir\\dnscrypt-proxy\\blacklist.txt\" -ItemType \"file\" | Out-Null",
+        "New-Item \"$dir\\dnscrypt-proxy\\blocked.log\" -ItemType \"file\" | Out-Null",
+        "New-Item \"$dir\\dnscrypt-proxy\\query.log\" -ItemType \"file\" | Out-Null",
+        "New-Item \"$dir\\dnscrypt-proxy\\domain-blacklist.txt\" -ItemType \"file\" | Out-Null",
+        "New-Item \"$dir\\dnscrypt-proxy\\domain-whitelist.txt\" -ItemType \"file\" | Out-Null"
```
issaclin32, line 25:
```suggestion
        "'blacklist.txt', 'blocked.log', 'query.log', 'domain-blacklist.txt', 'domain-whitelist.txt' | ForEach-Object {",
        "    if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" | Out-Null }",
        "}"
```
For `New-Item`, `ItemType` is `File` by default.

bucket/simple-dnscrypt.json
```diff
@@ -0,0 +1,63 @@
...
+        "dnscrypt-proxy\\query.log",
+        "dnscrypt-proxy\\domain-blacklist.txt",
+        "dnscrypt-proxy\\domain-whitelist.txt"
+    ],
+    "pre_uninstall": [
+        "if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
+        "",
+        "Start-Process \"$dir\\Uninstall.exe\" | Out-Null",
+        "if (Get-Process -Name \"SimpleDnsCrypt\" | Out-Null) {Stop-Process -Name \"SimpleDnsCrypt\" | Out-Null}",
+        "if (Get-Service -Name \"dnscrypt-proxy\" | Out-Null) {Stop-Service -Name \"dnscrypt-proxy\"| Out-Null; Remove-Service -Name \"dnscrypt-proxy\" | Out-Null}"
```
issaclin32, line 47:
```suggestion
        "if (Get-Process -Name 'SimpleDnsCrypt' -ErrorAction SilentlyContinue) { Stop-Process -Name 'SimpleDnsCrypt' | Out-Null }",
        "$service = Get-Service -Name 'dnscrypt-proxy' -ErrorAction SilentlyContinue",
        "if ($service.Status -ne 'Stopped') {",
        "    if (!(is_admin)) { error 'Admin rights are required to stop dnscrypt-proxy service'; break }",
        "    Stop-Service -Name 'dnscrypt-proxy' | Out-Null",
        "}",
        "if (($service) -and ($cmd -eq 'uninstall')) {",
        "    if (!(is_admin)) { error 'Admin rights are required to remove dnscrypt-proxy service'; break }",
        "    Remove-Service -Name 'dnscrypt-proxy' | Out-Null",
        "}"
```

bucket/simple-dnscrypt.json
KrisSchneider: Looks good

bucket/simple-dnscrypt.json
KrisSchneider: Looks better than mine, I'll go with it.

rasa:
I went to delete my version, and found
https://github.com/rasa/scoops/blob/master/scripts/simple-dnscrypt/simple-dnscrypt.ps1
which this version lacks. My apologies for not pointing this out sooner,

KrisSchneider:
@rasa I believe you can easily disable the built-in auto update through the settings. 
(SimpleDnsCrypt)g2a254_09-18-2022

rasa:
I know, but I think generally scoop disables auto-update logic. If we are going to leave it in, we may want to alert the user that it's on, and how to disable it.

KrisSchneider:
Should I add a `notes` section to the manifest that tells the user how to disable built-in auto update?
Or could you please add the simple-dnscrypt script to the **script** folder of this bucket?

PR #9282, opened 2022-09-21
elasticsearch: Use `bin` path to avoid missing shims for older versions

marcolongol:
if you try to install a older version like '7.1.1' with scoop install <app>@<version> the shim fails to be created due to changes in the app structure,

actually the scoop installer itself should ignore missing shims in the cases were you are downgrading to a given version, I think that would be more appropriate.

in the meanwhile we can fix this by just putting the entire Elasticsearch 'bin' folder into PATH


rasa (approved):

marcolongol:
Still getting the error... how long till we merge?
image

chawyehsu (approved):

PR #9297, opened 2022-09-23
Update KDE Kate to version 22.08.1

### LINKED ISSUE #9289: KDE Kate out of date
## Bug Report

scoop.kate 

### Current Behaviour

scoop.kate version is [22.04.01](https://scoop.sh/#/apps?q=kate)

### Expected Behaviour

latest version is [22.08.1](https://github.com/KDE/kate/tags)

KrisSchneider:
Closes #9289 

### DIFF
--- a/bucket/kate.json
+++ b/bucket/kate.json
@@ -18,7 +18,7 @@
     ],
     "checkver": {
         "url": "https://binary-factory.kde.org/view/Windows%2064-bit/job/Kate_Release_win64/",
-        "regex": "kate-(\\d+)-"
+        "regex": "kate-(\\d+)-(?<build>\\d+))-msvc"
     },
     "autoupdate": {
         "architecture": {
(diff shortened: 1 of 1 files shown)


tech189:
I'm not sure we should do this in Extras, at the moment this app is only updated when a stable version is released. If this PR is merged, then the app will be updated daily regardless of whether any changes have been made - it's a daily build. There have been [no changes](https://binary-factory.kde.org/job/Kate_Release_win64/changes) in the last 10 builds. Maybe instead we could add this PR's manifest as a `stable-nightly` version to the Versions bucket. What do you think @rashil2000?

rashil2000:
> Maybe instead we could add this PR's manifest as a `stable-nightly` version to the Versions bucket. What do you think @rashil2000?

We used to do this before I switched them to update only on stable releases. There's no point in updating when there's been no commits. 


We should just update the download URL and regex to match the website.

rashil2000:
**Edit:**

Merged in #9298 

We also need to do this for `okular`, `kdeconnect`, and `filelight`.

rashil2000:
Merged in #9298

PR #9353, opened 2022-09-28
kscript: Add version 4.1.1

Dany-coder778:
Added kscript.json to the bucket 

### DIFF
--- a/bucket/kscript.json
+++ b/bucket/kscript.json
@@ -3,8 +3,9 @@
     "description": "Enhanced scripting support for Kotlin.",
     "homepage": "https://github.com/kscripting/kscript",
     "license": "MIT",
-    "notes": "This program will also install kotlin via Scoop.",
-    "depends": "kotlin",
+    "suggest": {
+        "Kotlin": "main/kotlin"
+    },
     "url": "https://github.com/kscripting/kscript/releases/download/v4.1.1/kscript-4.1.1-bin.zip",
     "hash": "127d169f5dc0f67800b99552c0db0e8b662de4dbe1ff8f2078a7c1cb9a5ed5c1",
     "extract_dir": "kscript-4.1.1", 
--- a/bucket/kscript.json
+++ b/bucket/kscript.json
@@ -15,3 +15,4 @@
         "extract_dir": "kscript-$version"
     }
 }
+
--- a/bucket/kscript.json
+++ b/bucket/kscript.json
@@ -3,9 +3,8 @@
     "description": "Enhanced scripting support for Kotlin.",
     "homepage": "https://github.com/kscripting/kscript",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


goto1134:
There should be a dependency on kotlin, as the [installation guide](https://github.com/kscripting/kscript#installation) says

goto1134:
shouldn't it be `depends` instead of `suggest`?

Dany-coder778:
Maybe the user already has a kotlin compiler in their PATH and doesn't want another one.  It would still be fine for me to put it as a dependency, but maybe some people might not like having kotlin installed as well.  I also noticed that many jsons only suggest installing the visual c redistributables and don't enforce it as a dependency.

goto1134:
It is an interesting question: how should it work in scoop? Looking and macOS brew, it force installs the dependencies.

But brew is de facto standard macOS tool. We have so many alternative package managers on windows... It might be reasonable not to install the dependencies.

Maybe we need maintainers to shed some light on the desired behavior?

goto1134:
I looked up the documentation, and it says:

> ... when you install an app, Scoop will assume you want to install anything required that you don't already have and install these automatically too.

So that means kotlin should be in `depends`, I guess.

Dany-coder778:
I replaced suggest with depends and added a note informing the user of installing kotlin.

bucket/kscript.json
```diff
@@ -0,0 +1,17 @@
...
+    "url": "https://github.com/kscripting/kscript/releases/download/v4.1.1/kscript-4.1.1-bin.zip",
+    "hash": "127d169f5dc0f67800b99552c0db0e8b662de4dbe1ff8f2078a7c1cb9a5ed5c1",
+    "extract_dir": "kscript-4.1.1", 
+    "bin": "bin\\kscript.bat",
+    "checkver": "github",
+    "autoupdate": {
+        "url": "https://github.com/kscripting/kscript/releases/download/v$version/kscript-$version-bin.zip",
+        "extract_dir": "kscript-$version"
+    }
+}
```
goto1134, line 17:
```suggestion
}

```

goto1134:
@Dany-coder778, there is an error
@HUMORCE, can you please review this PR? It has been stale for a lot of time but contains a pretty useful piece of software

bucket/kscript.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "4.1.1",
+    "description": "Enhanced scripting support for Kotlin.",
+    "homepage": "https://github.com/kscripting/kscript",
+    "license": "MIT",
+    "notes": "This program will also install kotlin via Scoop.",
+    "depends": "kotlin",
```
HUMORCE, line 7:
```suggestion
    "suggest": {
        "Kotlin": "main/kotlin"
    },
```

PR #9355, opened 2022-09-28
arduino-old: Add version 1.8.19

Dany-coder778:
Added arduino-old.json to the bucket


Zliced13:
I think this should go to the **versions** bucket instead of extras, since it's an alternative version of the same app.

Dany-coder778:
https://github.com/ScoopInstaller/Versions/pull/715

PR #9363, opened 2022-09-29
usb-tree-view: Add version 3.8.1

RavenMacDaddy:
App description: View USB devices in a tree-like structure.

Credit goes to @404NetworkError for his manifest [here](https://github.com/404NetworkError/scoop-bucket/blob/main/bucket/usbtreeview.json) which this is based on.

### DIFF
--- a/bucket/usb-tree-view.json
+++ b/bucket/usb-tree-view.json
@@ -13,6 +13,11 @@
             "hash": "405e6d9ffb145fa76f35a2ff9a4b4ff520d16f70cfa559852c21872c73a18f85"
         }
     },
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\UsbTreeView.ini\")) {",
+        "   New-Item \"$dir\\UsbTreeView.ini\" | Out-Null",
+        "}"
+    ],
     "bin": [
         [
             "UsbTreeView.exe",
@@ -25,6 +30,7 @@
             "USB Device Tree Viewer"
         ]
     ],
+    "persist": "UsbTreeView.ini",
     "checkver": {
         "url": "https://www.uwe-sieber.de/usbtreeview_e.html",
         "regex": "USB Device Tree Viewer V([\\d.]+)"
--- a/bucket/usb-tree-view.json
+++ b/bucket/usb-tree-view.json
@@ -1,0 +1,32 @@
+{
+    "version": "3.8.1",
+    "description": "View USB devices in a tree-like structure",
+    "homepage": "https://www.uwe-sieber.de/usbtreeview_e.html",
+    "license": "Freeware",
+    "architecture": {
+        "64bit": {
+            "url": "https://www.uwe-sieber.de/files/UsbTreeView_x64.zip",
+            "hash": "43f33cf1bf9b6ea41e36840923a85791b06803cb9f588a4c3b288e2539d0c63b"
+        },
+        "32bit": {
+            "url": "https://www.uwe-sieber.de/files/UsbTreeView_Win32.zip",
+            "hash": "405e6d9ffb145fa76f35a2ff9a4b4ff520d16f70cfa559852c21872c73a18f85"
+        }
+    },
+    "bin": [
+        [
+            "UsbTreeView.exe",
+            "usbtreeview"
+        ]
+    ],
+    "shortcuts": [
+        [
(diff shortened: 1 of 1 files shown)


rashil2000:
My only concern is that if a bucket isn't official, it's not discoverable for the average Scoop user.

RavenMacDaddy:
Scoop in terms of CLIs and package managers is one of the more organized alternatives.

One might argue that having everything available always, ala Chocolatey or Homebrew, is a non-issue as we're talking about manifests describing what needs to be done to manage software. 

However, I think Scoop strikes a good balance and agree with @rashil2000, which is why I continue to work with the known buckets accordingly.

rasa:
> My only concern is that if a bucket isn't official, it's not discoverable for the average Scoop user.

I'm not sure I follow. The user needs to manually add the extras bucket to search, or install apps from it, yes? 

So if we throw everything in extras, how do we distribute the maintenance workload?

Perhaps we need to expand the definition of an "official" bucket. We currently have [three](https://github.com/ScoopInstaller/Scoop/blob/master/buckets.json#L5) "official" buckets outside our organization. If well known/maintained buckets were added to the official bucket list, would that alleviate the need to move everything into extras?

I know I'm sounding like a broken record, but I think it's important for Scoop's long term health and viability to have a plan on how best to build a distributed infrastructure. But if I'm a lone voice, perhaps I need to stfu.

rashil2000:
> If well known/maintained buckets were added to the official bucket list, would that alleviate the need to move everything into extras?

I'm okay with merging rasa/scoops into ScoopInstaller (or making it an official, but non-org bucket). But since it's a (seemingly) random collection of applications, what would we categorize it as?

RavenMacDaddy:
Maintaining any bucket, whether it being additions or corrections, is always a collective contribution process.

I certainly don't care if someone else has created a manifest in the first place; if I see something that needs correction because I've interest in that particular manifest and therefore stumble upon it, I'll correct it either way.

On the subject of official buckets, one way to do it is not to necessarily categorize a particular one, though obviously if there's a clear theme, by all means.

However, simply having a sub-category of buckets named **Community** which contains well maintained and extensive repositories (buckets) would mean easier additions of buckets for the user, as they can simply add them as easily as one of the originals.

It would also mean a list of buckets that can be expanded over time as beneficial repositories are discovered by the community.

rasa (approved):

rasa:
I am not asking that rasa/scoops be made official. I'm suggesting we consider how best to distribute the workload of app maintenance. One way would be to create official buckets for a particular category. It could be a publisher, say Sordum, or it could be by licensing, say open, free, non-free, or it could be by category, games, editors, dev tools, etc.

Or we could throw everything in main (cli) and extras (gui), but then partition the apps inside the bucket folder.

RavenMacDaddy:
Something like what's first mentioned sounds reasonable to me; subdirs or specific buckets for specific publishers, depending on the size of their catalog.

It's also akin to the Nirsoft bucket that already exists in the list.

I wouldn't mind posting these ideas as a discussion (maybe a poll?) to keep things OT for the PR.

However, it does seem more fitting you do it as both the initiator of the discussion and with ownership of the ideas presented.

rasa:
@RavenMacDaddy Yes, I'm sorry to hijack your PR with a meta discussion of bucket organization. I've approved this PR to not let our aside sideline your effort. Once it's accepted I will delete my version in scoops.

chawyehsu:
> I know I'm sounding like a broken record, but I think it's important for Scoop's long term health and viability to have a plan on how best to build a distributed infrastructure. But if I'm a lone voice, perhaps I need to stfu.

@rasa Believe me, I stand with you for the voice of *building a distributed infrastructure*.

HUMORCE:
> My only concern is that if a bucket isn't official, it's not discoverable for the average Scoop user.

Only official buckets are displayed in results by default, after some migrations/updates of Scoop Search. where the manifests of official bucket is clearly marked, all manifests should be displayed.

jasongodev (changes requested):
@RavenMacDaddy please fix the manifest to pass the test.

RavenMacDaddy:
@jasongodev Considering the discussion above not going anywhere in regards to the PR I'm gonna put my resources elsewhere - closing.

PR #9411, opened 2022-10-07
archisteamfarm: Add bin

redactedscribe:
Although ArchiSteamFarm comes with its own interactive console, you may want to run it directly from a console window you already have open instead.


HUMORCE:
#9476

PR #9534, opened 2022-10-22
ipfilter-updater: Fix homepage

brian6932:

### DIFF
--- a/bucket/ipfilter-updater.json
+++ b/bucket/ipfilter-updater.json
@@ -14,8 +14,9 @@
         ]
     ],
     "checkver": {
-        "url": "https://github.com/DavidMoore/ipfilter/releases",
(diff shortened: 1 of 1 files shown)


brian6932:
While I'm at it, I may as well improve the checkver

brian6932:
@rasa Can you merge this when you get the chance? pretty small PR

brian6932:
@niheaven The reason \S was used was so the -beta flags would count, as they did previously when the checkver was \w

niheaven:
The tag itself contains `beta`, and the regex extracts everything.

niheaven:
CleanShot 2022-11-13 at 03 02 10@2x

brian6932:
Oh I wasn't aware, ok np

PR #9541, opened 2022-10-22
spotube: Fix `autoupdate.url` and `autoupdate.hash.regex`

### LINKED ISSUE #9539: The app manifest of Spotube doesn't implement auto-update correctly
## Bug Report

**Package Name:** Spotube

### Current Behaviour

The current app manifest of the package [https://github.com/ScoopInstaller/Extras/blob/master/bucket/spotube.json] doesn't implement auto-update correctly because of the following reasons:
1) The URL in autoupdate is referencing a static version: v2.3.0, which makes auto-update malfunction and always set the download version in the URL to v2.3.0.
2) The Hash Regex in autoupdate references a static hash (Which I believe is the hash of v2.3.0) which makes the regex not work in newer versions. I tried to make a regex that would work for all versions but I'm not too good at regex so I couldn't do it.

### Expected Behaviour

Auto-update should work.

### Additional context/output

None

### Possible Solution

1) The URL in autoupdate should be replaced with the following: ``https://github.com/KRTirtho/spotube/releases/download/v$version/Spotube-windows-x86_64-setup.exe``.
2) The Hash Regex in autoupdate shouldn't reference a static hash since the hash changes in every update.

### System details

System details are not relevant to the bug.

Zliced13:
Closes #9539

### DIFF
--- a/bucket/spotube.json
+++ b/bucket/spotube.json
@@ -27,7 +27,7 @@
         },
         "hash": {
             "url": "$baseurl/RELEASE.sha256sum",
-            "regex": "49b637ccac5f091a6c5f4825a37727a9387c152a947fd7c351cb0520ddb34201\\s+\\./$basename"
+            "regex": "$sha256\\s+\\./$basename"
         }
     }
 }
--- a/bucket/spotube.json
+++ b/bucket/spotube.json
@@ -22,7 +22,7 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/KRTirtho/spotube/releases/download/v2.3.0/Spotube-windows-x86_64-setup.exe"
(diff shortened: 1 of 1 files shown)


M6HqVBcddw2qaN4s:
@Zliced13 Correct me if I'm wrong, but the hash regex still references a static hash which wouldn't work with newer versions?

Zliced13:
No it will work just, fine. Did you look at the feedback the bot gave above?

M6HqVBcddw2qaN4s:
@Zliced13 Just tested if it would be able to get the hash from v2.5.0's SHA256sum and it doesn't.
https://user-images.githubusercontent.com/61710729/197376113-45349f94-90e1-452d-85b8-338a5656944a.png

Zliced13:
@M6HqVBcddw2qaN4s you were right, I looked at the manifest again and fixed the `autoupdate.hash.regex`

HUMORCE (approved):

PR #9592, opened 2022-10-28
sfss: Add version 1.0.0

jewlexx:

### DIFF
--- a/bucket/sfss.json
+++ b/bucket/sfss.json
@@ -3,13 +3,28 @@
     "description": "Stupid Fast Scoop Search. Sub-second scoop searching, with a drop-in replacement",
     "homepage": "https://github.com/jewlexx/sfss",
     "license": "Apache-2.0",
-    "url": "https://github.com/jewlexx/sfss/releases/download/v1.0.0/sfss.exe",
-    "hash": "70856E0BB3341D7DB5E505B1C1D4C6428003EC55568D9AF2D33A8413BF566161",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/jewlexx/sfss/releases/download/v1.0.0/sfss-x86_64.exe#sfss.exe",
+            "hash": "23CB72BB87A2ACC57E59E7C67702805EC55DC54455050AF0F21A9197CC7AF9FC"
+        },
+        "32bit": {
+            "url": "https://github.com/jewlexx/sfss/releases/download/v1.0.0/sfss-i686.exe#sfss.exe",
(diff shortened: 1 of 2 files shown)


HUMORCE:
`autoupdate.hash` is missing.

jewlexx:
> autoupdate.hash is missing.

Is that just the same as the current version's hash?

HUMORCE:
oh, you are the owner of the repo. I mention this because the sha256 value is including in the release notes, you created a file for hash checking, it also okay.

screenshot_20221028132001

...but the app does not seem to work on a 32bit OS.

> architecture: If the app has 32- and 64-bit versions, architecture can be used to wrap the differences ([example](https://github.com/ScoopInstaller/Main/blob/master/bucket/7zip.json)).
> 32bit|64bit: contains architecture-specific instructions (bin, checkver, extract_dir, hash, installer, pre_install, post_install, shortcuts, uninstaller, url, and msi [msi is deprecated]).
> 
> https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties

niheaven:
Yes, a 64bit app.

image

niheaven:
image

HUMORCE:
@niheaven you have forgotten to edit `autoupdate` part. while it works, but it does not meet the Standards.

> If the application provides only a 32bit download, the architecture field is not required. In all other cases, architecture field is mandatory.

niheaven:
For autoupdate, non-arch one will update both, and this mechanism is designed. So if there's only 64bit arch, it is no need to use arch in autoupdate.

HUMORCE:
I understand about that, but it does not mean that this meets the standards. this is part of all the other cases in the guide.

jewlexx:
@HUMORCE I have added a 32bit version, and tested the downloads on my machine, and they seem to work fine

niheaven:
> All the properties except autoupdate.note can be set globally for all architectures or for each architecture separately (under architecture.64bit or architecture.32bit). Global properties can be used to update each architectural properties, i.e., if only setted globally, autoupdate.url is used to update either architecture.64bit.url or architecture.32bit.url.

https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#properties-of-autoupdate

@HUMORCE I mean this is *designed*, at east for `autoupdate` field. I preferred to use uniform url and hash if there's only one architecture to be updated, and of course, others could use nested from. There're no standards here.

HUMORCE:
emmm, you are right. and I used the designed feature in this manifest when last edit. I simply thought a clear/explicit statement would be easier for more collaborators to read and edit.

PR #9640, opened 2022-11-03
freetube: Update to version 0.18.0-beta, and fix `architecture` property

Zliced13:
Closes #9634

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

Issue #9634, opened 2022-11-02
[CHANGE] FreeTube not being updated to the latest version 


Xelbayria:
## Bug Report

**Package Name:** Freetube

### Current Behaviour

The version of Freetubeshould have been updated to the late version but it hasn't since the new version is out at (NOV 1, 2022 @ 9:25pm). I took a look at the JSON file. the REGEX for the version of Freetube is incorrect. 

### Expected Behaviour

The Freetube app is to be updated to the latest version via Github's "ACTIONS". 

### Possible Solution

The REGEX for JSON to check for Freetube's version needs to be changed. 

### System details

**Windows version:** 10

**OS architecture:** 64

**PowerShell version:** 7.2.7

Zliced13:
Could you show how you would fix the `checkver`?

Zliced13:
@Xelbayria  do you urgently need a fixed version of this app?

Zliced13:
Just incase you need it I have a fixed version in my personal bucket which you can use until the **freetube** manifest in this bucket is fixed.

```
scoop bucket add MyScoop 'https://github.com/Zliced13/MyScoop'
```
```
scoop install freetube_fixed
```

Xelbayria:
> Just incase you need it, I have a fixed version in my personal bucket which you can use until the **freetube** manifest in this bucket is fixed.
> 
> ```
> scoop bucket add MyScoop 'https://github.com/Zliced13/MyScoop'
> ```
> 
> ```
> scoop install freetube_fixed
> ```

Thanks, I'll use it for now.

PR #9648, opened 2022-11-03
Update freetube.json

webac:
New Beta version: 0.18.0-beta


Xelbayria:
There is already PR for this: #9640 and an issue report from #9634

PR #9703, opened 2022-11-09
vcredist & vcredist2022: add arm64 archictecture

dahlia:

----

This adds specializations to vcredist & vcredist2022 for x86 (32-bit) and ARM64 architectures.  They installs a different set of redistributables for each architecture:

- x86 (64-bit): both *VC_redist.x64.exe* and *VC_redist.x86.exe* as it's done so far.
- x86 (32-bit): only *VC_redist.x86.exe*.
- ARM64: all of *VC_redist.arm64.exe*, *VC_redist.x64.exe*, and *VC_redist.x86.exe* for compatibility with non-arm64 executables.

I am not sure if this change is appropriate.  Please tell me if I did something wrong!

### DIFF
--- a/bucket/vcredist.json
+++ b/bucket/vcredist.json
@@ -14,21 +14,59 @@
         "vcredist2012",
         "vcredist2013"
     ],
-    "url": [
-        "https://download.visualstudio.microsoft.com/download/pr/bcb0cef1-f8cb-4311-8a5c-650a5b694eab/2257B3FBE3C7559DE8B31170155A433FAF5B83829E67C589D5674FF086B868B9/VC_redist.x64.exe",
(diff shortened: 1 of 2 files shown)


niheaven:
I've not a x86 machine, and could x64 installer be installed in x86?

dahlia:
> I've not a x86 machine, and could x64 installer be installed in x86?

Although I have no x86 machine either, I believe x64 installer does not work on x86.

niheaven:
So the question is, why there's nobody complaint it?

dahlia:
Hmm… fair enough.  Then I'm going to close this pull request.  Thanks!

PR #9772, opened 2022-11-18
godot & godot-mono: Enhance `pre_install` script

Zliced13:

### DIFF
--- a/bucket/godot-mono.json
+++ b/bucket/godot-mono.json
@@ -15,8 +15,8 @@
             "extract_dir": "Godot_v3.5.1-stable_mono_win32"
         }
     },
-    "pre_install": "Rename-Item \"$dir\\Godot_*_console.cmd\" 'godot-mono-cli.cmd'; Rename-Item \"$dir\\Godot_*.exe\" 'godot-mono.exe'",
-    "bin": "godot-mono-cli.cmd",
+    "pre_install": "Remove-Item \"$dir\\Godot_*_console.cmd\"; Rename-Item \"$dir\\Godot_*.exe\" 'godot-mono.exe'",
+    "bin": "godot-mono.exe",
     "shortcuts": [
         [
             "godot-mono.exe",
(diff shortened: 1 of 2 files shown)


niheaven:
I'm not sure why whole file is changed, so just change `pre_install` in a direct commit.

drilus:
This commit breaks the install for me. I think you need:

``"Get-ChildItem \"$dir\\Godot_*.exe\"``

in order to properly rename a wildcard

Zliced13:
You don't have to worry about this commit merging @drilus. @niheaven closed this issue already.

ikenichiro:
I think @drilus is saying that the installation is still broken after e7843a9 .
I also get the following error when trying to install godot with the current configuration.

```
+ Rename-Item "$dir\Godot_*.exe" 'godot.exe'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Rename-Item]、PSArgumentException
    + FullyQualifiedErrorId : Argument,Microsoft.PowerShell.Commands.RenameItemCommand
```
Could you guys take a look? @niheaven @Zliced13

niheaven:
It works for me in PS5 and PS7, how about yours? @Zliced13 

image

Zliced13:
I was also able to install **godot** just fine.
(WindowsTerminal)c1K2k4_11-22-2022

ikenichiro:
I was able to reproduce the error and also find a working around on multiple environments, but I'm not exactly sure why this error occurs.
1. Execute ``scoop install godot`` having PowerShell's working directory anywhere inside drive C: (e.g. C: or C;\Users\someuser\scoop\)

Executing the install command when having the working directory at any other drive (ex. D: ) works as expected.

godot-not-installing

This screenshot is from PSVersion 5.1.19041.1645, but I have also confirmed the same errors on 5.1.19041.1682.
Both environments run having the WIndows environment in Japanese, however, the username does not include any multi-byte chars (only composed of non-special ASCII, all small-case a-z ).

ikenichiro:
confirmed fixed by #9825 .
Many thanks for the fast work ;)

PR #9786, opened 2022-11-19
firefox: use scoop profile by default

alealexpro100:
Use scoop profile by default. Previous profile will not be lost.
Also, this behavior makes less changes in appdata directory.
File prefs.js took from [here](https://portableapps.com/apps/internet/firefox_portable)


HUMORCE:
the preferred solution for profile of mozilla series software: implement https://github.com/ScoopInstaller/Scoop/issues/5234

PR #9840, opened 2022-11-25
mp3tag: Enhance `pre_install` script, add `pre_uninstall` script, and change `persist`

Zliced13:
This allows for mp3tag to be added as a shell extension when the user installs this app.

### DIFF
--- a/bucket/mp3tag.json
+++ b/bucket/mp3tag.json
@@ -47,11 +47,12 @@
         "mp3tag.cfg"
     ],
     "pre_uninstall": [
-        "if (!(is_admin)) {error \"$app requires admin rights to $cmd\"; break}",
         "Copy-Item \"$dir\\export\\*\" \"$persist_dir\\export\" -Recurse -ErrorAction 'SilentlyContinue'",
-        "if (![System.Environment]::Is64BitOperatingSystem) { Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell32.dll\", '/s') }",
-        "elseif ([System.Environment]::Is64BitOperatingSystem) { Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell.dll\", '/s') }",
-        "Stop-Process -Name 'explorer' -Force; Start-Sleep -Seconds 4"
+        "if (Test-Path 'HKLM:\\SOFTWARE\\Classes\\Directory\\shellex\\ContextMenuHandlers\\Mp3tagShell\\') {",
+        "   if (!(is_admin)) {error \"$app requires admin rights to $cmd\"; break}",
+        "   Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell32.dll\", '/s')",
+        "   Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell.dll\", '/s')",
+        "}"
     ],
     "checkver": "Mp3tag\\s+v([\\w.]+)",
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


HUMORCE (changes requested):
exists restrict of installation

Zliced13:
@HUMORCE can you read my comment on the manifest when you have the time?

HUMORCE:
where is it?

Zliced13:
Ok since GitHub will not show comments for whatever stupid reason I'll just paste it here. This is for lines 16 to 17.

@HUMORCE This part of the script will not work without administrator privileges. Should I make this optional instead?

HUMORCE:
> since GitHub will not show comments for whatever stupid reason.

you need to `submit review` after `start a review`.

> make this optional

yes

HUMORCE:
> exists restrict of installation

applies to uninstallation also.

for all operations that require admin privileges: use `notes` or `warn() | info()` instead, let users execute them manully.

e.g. some items in `scoop-checkup` need to be executed manually by the user rather than requesting admin privileges to execute when installing Scoop.

> Scoop installs programs from the command line with a minimal amount of friction.
> 
> src: https://github.com/ScoopInstaller/Scoop#what-does-scoop-do

HUMORCE (approved):

redactedscribe:
@Zliced13 Thanks for the enhancements, but for what reason was persist limited to only a select few files? The manifest was correct to back up all of `/data/` and `/export/`, as well as `/mp3tag.cfg`. These two directories and `/mp3tag.cfg` is what Mp3tag exports as a backup when you use the File menu "Save configuration" feature. Limiting persist to only a few files within `/data/` misses some user-customised files and the same goes for omitting all of `/export/`.

There is also `/Mp3tagSettings.zip` which is automatically created after each termination of the application. If there's a good reason behind limiting persist as you have, at minimum `/Mp3tagSettings.zip` should also be persisted because this contains all of the user's settings (the three items I've mentioned).

Would it be possible that you fix this?

Zliced13:
Someone is already fixing this in https://github.com/ScoopInstaller/Extras/pull/10167

Zliced13:
> @Zliced13 Thanks for the enhancements, but for what reason was persist limited to only a select few files? The manifest was correct to back up all of `/data/` and `/export/`, as well as `/mp3tag.cfg`. These two directories and `/mp3tag.cfg` is what Mp3tag exports as a backup when you use the File menu "Save configuration" feature. Limiting persist to only a few files within `/data/` misses some user-customised files and the same goes for omitting all of `/export/`.
> 
> 
> 
> There is also `/Mp3tagSettings.zip` which is automatically created after each termination of the application. If there's a good reason behind limiting persist as you have, at minimum `/Mp3tagSettings.zip` should also be persisted because this contains all of the user's settings (the three items I've mentioned).
> 
> 
> 
> Would it be possible that you fix this?

If you read my previous comments, you would know why I had to use `pre_(un)install` scripts to manually copy the files instead of the using the built in `persist` property.

redactedscribe:
I see now that you are actually copying the `/export/` directory in the pre inst/uninstall lines, but I'm still confused about why you're only persisting a select number of files from within `/data/` instead of all of them.

Zliced13:
Because not all the files in the `/data/` directly need to be copied to the persist folder.

redactedscribe:
> Because not all the files in the `/data/` directory need to be copied to the persist folder.

This is the contents of `/data/` for me:
mp3tag-data

- If `actions` is not persisted, all user created/modified tagging actions will be lost on uninstall.
- If `sources` is not persisted, all user created/modified source scripts for fetching metadata will be lost on uninstall.
- If `genres.ini` is not persisted, all user-defined genres tags will be lost on uninstall.
- If `tools.ini` is not persisted, all user-defined external program paths and parameters will be lost on uninstall.
- `empty.mte` appears to be a template which I haven't modified, however others may have.
- `freedb.src` may also be of importance to anyone who may have edited it.

My `/data/` folder may not contain all the potential files/folders that Mp3tag can create as user data, I don't know. What I do know is that I'd not want to lose the files which you have chosen to mark as expendable via their omission from the manifest. Backing up `/data/` as a whole would avoid losing any of this user data, much of which is important.

Zliced13:
Next PR, I will try to add that feature to mp3tag.

PR #9845, opened 2022-11-25
Zim: Add Version 0.75.1

xmeadow:
I was surprised to be the first to create this pull request. But it would be very nice if zim wiki could be included in the extra bucket.

Json worked as expected.


xmeadow:
For people who want to learn about Zim: https://zim-wiki.org/

bucket/zim.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version":  "0.75.1",
+    "license":  "GPL-2.0-only",
+    "description":"Zim is a graphical text editor used to maintain a collection of wiki pages. Each page can contain links to other pages, simple formatting and images. Pages are stored in a folder structure, like in an outliner, and can have attachments. Creating a new page is as easy as linking to a nonexistent page. All data is stored in plain text files with wiki formatting. Various plugins provide additional functionality, like a task list manager, an equation editor, a tray icon, and support for version control.",
```
rasa, line 4: This description seems longer than usual. Our [docs](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#required-properties) say it should be "one line." I would interpret this to mean one sentence, or two small sentences, at most. Would you mind shortening this? Perhaps:
```json
    "description":"A graphical text editor to maintain wiki pages.",
    "##": "Each page can contain links to other pages, simple formatting and images. Pages are stored in a folder structure, like in an outliner, and can have attachments. Creating a new page is as easy as linking to a nonexistent page. All data is stored in plain text files with wiki formatting. Various plugins provide additional functionality, like a task list manager, an equation editor, a tray icon, and support for version control.",
```

rasa (changes requested):

bucket/zim.json
xmeadow: You are right. I commited the shorter description. Thanks.

xmeadow:
@rasa please help. I really can't figure out why the hash is wrong. Also checkver should work according to my brain. I don't know why scoop uses \\ instead of a single \ in regex but I tesed with regex and compared with several other manifests. I just need a helping hand. ;)

xmeadow:
ah. i learned the hard way that scoop does not automatically detect the highest version when determining the version. It was a pain to figure out but after a bunch of dumb commits all tests passed ;)

tech189:
Thanks for your contribution, I have merged the most recent pull request instead which is based on yours with a few tweaks to get the persisted folders working!

PR #8887, opened 2022-07-17
houdoku: add version 2.9.2

SonaliBendre:
Closes #XXXX

Relates to #XXXX

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #9868, opened 2022-11-29
houdoku: Fix arch

sitiom:
Fixes https://github.com/ScoopInstaller/Extras/pull/8887#discussion_r1034279768

### DIFF
--- a/bucket/houdoku.json
+++ b/bucket/houdoku.json
@@ -3,8 +3,12 @@
     "description": "Manga reader and library manager for the desktop",
     "homepage": "https://houdoku.org/",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


niheaven:
Where's the config files for `zip` edition? If it is still in $APPDATA, using exe and given hash is better.

sitiom:
> Where's the config files for `zip` edition? If it is still in $APPDATA, using exe and given hash is better.

It's still in `%appdata%`, and with the actual zip version, we do not need the `pre_install` field and the manifest is much simpler.

niheaven:
Given hash and smaller download size are preferred than simpler manifest. So just nest the arch in this PR.

PR #9897, opened 2022-12-03
xampp@8.1.12-0: Fix config and Add setup

alealexpro100:
File `xampp-control.ini` is not in archive by default, so we create it.
Scoop links it as directory, not a file, if it does not exist, which leads to incorrect start of application.
Also add automatic PHP setup (does not change system files).

### DIFF
--- a/bucket/xampp.json
+++ b/bucket/xampp.json
@@ -3,7 +3,6 @@
     "description": "Apache distribution containing MariaDB, PHP, and Perl",
     "homepage": "https://www.apachefriends.org/index.html",
     "license": "GPL-2.0-only",
-    "notes": "Follow the instructions on '$dir\\readme_en.txt' to set up XAMPP.",
     "suggest": {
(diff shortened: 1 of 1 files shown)


bucket/xampp.json
```diff
@@ -14,6 +13,12 @@
         }
     },
     "extract_dir": "xampp",
+    "pre_install": [
+        "if (-Not (Test-Path \"$dir\\xampp-control.ini\")) {",
+        "   New-Item \"$dir\\xampp-control.ini\"",
+        "}"
+    ],
+    "post_install": "Start-Process -FilePath \"$dir\\setup_xampp.bat\" -ArgumentList \"< nul\" -WorkingDirectory \"$dir\"",
```
HUMORCE, line 21:
```suggestion
    "post_install": "Start-Process -FilePath \"$dir\\setup_xampp.bat\" -ArgumentList \"< nul\" -WorkingDirectory \"$dir\" -Wait",
```

PR #9899, opened 2022-12-03
megasync: updated checkver regex AND add post_install script for shell extension

jasongodev:
**This update adds a post_install script for MEGAsync's context menu and sync indicator icons.**

The MEGAsync installers have a unique case regarding how it register the shell extension responsible for the context menu and sync indicator icons.

In the original installation process, MEGAsync installer will dynamically create the ShellExtX64.dll using the $R0 file of the installer.

To make it more complicated, the 32-bit installer has 2 "$R0" file inside the installer. The original install process will dynamically extract this as ShellExtX32.dll and ShellExtX64.dll accordingly.

However, the Scoop installation process extracts these files in the $dir, then deletes the "dollar files" using the pre_install script
```
"Remove-Item \"$dir\\`$*\", \"$dir\\Megaupdater.exe\", \"$dir\\uninst*\" -Recurse"
```

While this deletes all the dollar-files and dollar-folders that are not necessary, it also deletes the $R0 file that should have been renamed to ShellExtX64.dll.

Another complication is for 32-bit installer where 2 "$R0" files are extracted in $dir causing it to overwrite the other. The order of the overwrite is unpredictable which can cause the $R0 that is 32bit to be overwritten by the 64-bit version and thus will not register correctly in a 32-bit machine.

**So the proposed changes does the following to solve these issues:**

1. Let the pre_install script to run as is, deleting all the dollar-files and dollar-folders. Why? Because there's no use of retaining $R0 in pre_install given that in 32-bit systems it might be the wrong overwritten file out of the two.
2. Determine the bit using the $architecture variable. $bit will be used later in the script.
```
"$bit = \"$architecture\".Substring(0,2)",
```
3. A post_install script was added that will extract the $R0 files from the cached installer.
4. The 7z command uses the -aou flag to auto-rename duplicate files. This will create $R0 and $R0_1 for 32-bit installer. For the 64-bit installer only $R0 exists.
```
"7z e -bso0 -bse0 -bsp0 -aou -o\"$dir\" \"$cachedir\\megasync#$version#https_mega.nz_MEGAsyncSetup$bit.exe_dl.7z\" `$R0",
```
5. Once extracted, we let 7z determine the file by listing its details. Since the dll files are compressed and can be read by 7z, it will output the string "CPU = x64" for 64-bit dll, while none for 32-bit dll. This allows us to rename $R0 and $R0_1 appropriately depending on their bit. Remember that we can't know which is 32 or 64-bit among both $R0 and $R0_1 due to the installer compilation, hence this step is necessary.
```
"if (Test-Path \"$dir\\`$R0\") { if ((7z l -bse0 \"$dir\\`$R0\") -like \"*CPU = x64*\") { Rename-Item \"$dir\\`$R0\" \"$dir\\ShellExtX64.dll\" } else { Rename-Item \"$dir\\`$R0\" \"$dir\\ShellExtX32.dll\" } }",
"if (Test-Path \"$dir\\`$R0_1\") { if ((7z l -bse0 \"$dir\\`$R0_1\") -like \"*CPU = x64*\") { Rename-Item \"$dir\\`$R0_1\" \"$dir\\ShellExtX64.dll\" } else { Rename-Item \"$dir\\`$R0_1\" \"$dir\\ShellExtX32.dll\" } }",
```
6. An instruction on how to enable the shell extension is printed in the screen.

I have tested this for both 32-bit and 64-bit systems and it works.

- I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/megasync.json
+++ b/bucket/megasync.json
@@ -41,7 +41,8 @@
         ]
     ],
     "checkver": {
-        "github": "https://github.com/meganz/MEGAsync"
+        "url": "https://api.github.com/repos/meganz/MEGAsync/releases",
+        "regex": "v(\d+.\d+.\d+).\d+_Win"
     },
     "autoupdate": {
         "architecture": {
--- a/bucket/megasync.json
+++ b/bucket/megasync.json
@@ -4,7 +4,7 @@
         "https://github.com/meganz/MEGAsync/blob/58d3ed7f4a10d08b0fa908639a65deac6a6bde1b/src/MEGAUpdater/Preferences.h#L9",
         "https://github.com/meganz/MEGAsync/blob/58d3ed7f4a10d08b0fa908639a65deac6a6bde1b/src/MEGASync/control/UpdateTask.cpp#L93-L111"
     ],
-    "version": "4.9.0",
+    "version": "4.9.0.0",
     "description": "Client for automated synchronization between local folder and MEGA cloud",
     "homepage": "https://mega.nz",
     "license": {
@@ -42,7 +42,7 @@
     ],
     "checkver": {
         "url": "https://api.github.com/repos/meganz/MEGAsync/releases",
-        "regex": "v(\\d+.\\d+.\\d+).\\d+_Win"
+        "regex": "v(\\d+.\\d+.\\d+.\\d+)_Win"
     },
     "autoupdate": {
         "architecture": {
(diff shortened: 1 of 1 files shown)


Zliced13:
@jasongodev you should edit the `pre_install` script to not remove the _`$PERSISTDIR`_ directory.

jasongodev:
@Zliced13 There's no use of retaining $R0 in pre_install given that in 32-bit systems the installer has 2 copies of this in the same root folder which means it will overwrite itself. We don't have a way to know which is the 32 or 64 bit dll from this overwritten file. **The surest way** is to re-extract them from the installer, which 7zip safely does by not overwriting it. Instead it creates $R0 and $R0_1 which we can scan whether they are 32 or 64 bit.

Zliced13:
> @Zliced13 There's no use of retaining $R0 in pre_install given that in 32-bit systems the installer has 2 copies of this in the same root folder which means it will overwrite itself. We don't have a way to know which is the 32 or 64 bit dll from this overwritten file. **The surest way** is to re-extract them from the installer, which 7zip safely does by not overwriting it. Instead it creates $R0 and $R0_1 which we can scan whether they are 32 or 64 bit.

Oh, so that is why. Sorry for the misunderstanding of your script.

Zliced13:
You just have to wait for a maintainer to your PR.

PR #9950, opened 2022-12-08
mp3tag: Fix `pre_(un)install` script

Zliced13:
**mp3tag** has a problem where the program will not uninstall if the context menu is installed. This edit should fix that issue.
(WindowsTerminal)Z9J2w1_12-08-2022

### DIFF
--- a/bucket/mp3tag.json
+++ b/bucket/mp3tag.json
@@ -30,7 +30,9 @@
         "   if (!(Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" | Out-Null }",
         "}",
         "Remove-Item \"$dir\\`$*\", \"$dir\\Mp3tagUninst*\" -Recurse",
-        "Start-Process \"$dir\\mp3tag.exe\" -Verb 'Open' -WindowStyle 'Minimized'; Start-Sleep -Seconds 3; Stop-Process -Name 'mp3tag'",
+        "if (!(Test-Path \"$persist_dir\\data\\usrfields.ini\")) {",
+        "   Start-Process \"$dir\\mp3tag.exe\" -Verb 'Open' -WindowStyle 'Minimized'; Start-Sleep -Seconds 3; Stop-Process -Name 'mp3tag'",
+        "}",
         "Copy-Item \"$persist_dir\\export\\*\" \"$dir\\export\" -Recurse -ErrorAction 'SilentlyContinue'"
     ],
     "bin": "mp3tag.exe",
--- a/bucket/mp3tag.json
+++ b/bucket/mp3tag.json
@@ -52,7 +52,7 @@
     "pre_uninstall": [
         "Copy-Item \"$dir\\export\\*\" \"$persist_dir\\export\" -Recurse -ErrorAction 'SilentlyContinue'",
         "if (Test-Path 'HKLM:\\SOFTWARE\\Classes\\Directory\\shellex\\ContextMenuHandlers\\Mp3tagShell\\') {",
-        "   if (!(is_admin)) {error \"$app requires admin rights to $cmd\"; break}",
+        "   if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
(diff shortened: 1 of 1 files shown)


Zliced13:
@HUMORCE could you please approve this PR when you have the time?

Zliced13:
**EDIT:** It also has this issue where the tag panel is completely blank, this PR will also address that issue.
(Mp3tag)a2K6j9_12-08-2022

bucket/mp3tag.json
```diff
@@ -52,6 +55,7 @@
         "   if (!(is_admin)) {error \"$app requires admin rights to $cmd\"; break}",
         "   Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell32.dll\", '/s')",
         "   Start-Process 'regsvr32' -Wait -Verb 'RunAs' -ArgumentList @('/u', \"$dir\\Mp3tagShell.dll\", '/s')",
+        "   Stop-Process -Name 'explorer' -Force; Start-Sleep -Seconds 2",
```
HUMORCE, line 58: Unacceptable

Zliced13: How is this _Unacceptable_ to you?

Zliced13: If I don't use **Stop-Process** to restart explorer, then unregistering the context wouldn't happen.

HUMORCE: what if I have some explorer windows open that I am working on? that's why I always say to let the user execute certain commands/scripts manually

Zliced13: Should I edit the script to break the **uninstallation** process if they have the context menu registered? Then tell them how to uninstall the context menu?

HUMORCE: I think all of your changes can be revert except for the persistence part. if you think the context menu is important, you should create a copy of it in the nonportable bucket.

the app works just as fine even without a context menu, and this is the expected behaviour for most portable apps. I don't see why you would want to change that.

i'm not saying that apps installed via Scoop cannot get the full experience. but these are extras setup steps in fact. I don't want it in my context menu, even when I use the app.(AND this is the reason most people use Scoop instead of winget/choco/official installer of app)

Zliced13: Fixed it.

Zliced13 (commented):
@HUMORCE This is not going to work, a blank `usrfields.ini` file will cause the fields in mp3tag to become completely blank.

Zliced13:
Proof:
> **EDIT:** It also has this issue where the tag panel is completely blank, this PR will also address that issue. (Mp3tag)a2K6j9_12-08-2022

Zliced13:
Additionally, Why do you remove the `pre_uninstall` section? It was there was there to remove the context menu option if the user activated it.

HUMORCE:
> This is not going to work, a blank usrfields.ini file will cause the fields in the Tag Panel in mp3tag to become completely blank.

I dont use this app so I dont know about this, but `data\usrfields.ini` was created as a dir, before.

> Additionally, why did you remove the pre_uninstall section? It was there to remove the context menu option if the user activated it.

the context menu cannot be uninstalled when we do not kill explorer, is it?

`"##": "An external script is needs to (un)install the context menu.",`

Zliced13:
No I have tested it. In the previous commit, I have found a way to remove the context menu without having to restart **explorer.exe**. You can try it out for yourself, in my bucket.
```
scoop bucket add MyScoop 'https://github.com/Zliced13/MyScoop'
```
```
scoop install mp3tag_fixed
```

Zliced13:
Make sure to run the commands to add **mp3tag** as a context menu option, then run the scoop uninstall command. It should uninstall the context menu in the uninstallation process.

HUMORCE:
ok, why use script to copying/moving `export\*` instead of `persist` here?

Zliced13:
Because the `export` folder is part of the mp3tag archive. If I were to persist the whole folder, it would end up creating endless copies of `export.orginal` folders in the `scoop/apps/mp3tag/current` folder, which no one probably wants.

PR #10082, opened 2022-12-22
(fix) freeplane@1.10.6  java version

H-Tung:
Freeplane [requires Java 11 to 17 to be installed](https://docs.freeplane.org/getting-started/Distributions.html), while openjdk/oraclejdk is 19.

dsBuffer

### DIFF
--- a/bucket/freeplane.json
+++ b/bucket/freeplane.json
@@ -5,7 +5,7 @@
     "license": "GPL-2.0-or-later",
     "suggest": {
         "JDK": [
-            "java/oraclejdk17",
+            "java/openjdk11",
             "java/openjdk17"
         ]
     },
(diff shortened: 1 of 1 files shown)


bucket/freeplane.json
```diff
@@ -5,8 +5,8 @@
     "license": "GPL-2.0-or-later",
     "suggest": {
         "JDK": [
-            "java/oraclejdk",
-            "java/openjdk"
+            "java/oraclejdk17",
```
HUMORCE, line 8: `java/oraclejdk17` does not exist, remove this item or replace it with `java/oraclejdk-lts`.

HUMORCE (changes requested):

bucket/freeplane.json
HUMORCE: OR replace them with `java/openjdk11` and `java/openjdk17`, then user will know that it can run in Java 11 to 17 without view the documentation.

PR #10110, opened 2022-12-25
sfsu: Fix `notes` and `bin`

Zliced13:
**sfsu** has recently been changed to become a single binary instead of three. This fix was made to accommodate that.


jewlexx:
This is a duplicate of #10107

PR #10111, opened 2022-12-25
Add chatgpt 0.6.10

Slach:
ChatGPT Desktop Application use rust tauri instead of Electron, use custom prompts


rasa (approved):

sitiom:
Note that this is an unofficial application, it seems that there isn't any such indication in the description.

- This will be replaced with the official application in #18684

PR #10127, opened 2022-12-28
threema: Fix `pre_install` for PowerShell 5

Zliced13:

### DIFF
--- a/bucket/threema.json
+++ b/bucket/threema.json
@@ -10,7 +10,7 @@
         }
     },
     "pre_install": [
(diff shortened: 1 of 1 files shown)


tech189:
Could you explain why you propose this change? I am able to install the manifest okay as-is.

Zliced13:
I changed it because `Rename-Item` does not work with wildcards in **PowerShell 5**, I made this change to make it backwards compatible.

tech189 (approved):
Thanks!

Issue #9643, opened 2022-11-03
PowerToys 0.64.0 doesn't work properly


raycheung:
## Bug Report

**Package Name:** powertoys

### Current Behaviour

A couple modules won't run properly.

### Expected Behaviour

All modules should run.

### Additional context/output

DLL files are now hardlinked into subfolders after this commit: https://github.com/microsoft/PowerToys/commit/db191b8b75ad456d5eb94eaa0aa0c7b59ecec900 .

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.7

**Additional software:** [(optional) e.g. ConEmu, Git]

#### Scoop Configuration

```json
{
  "last_update": "2022-11-03T12:56:10.2105809+08:00",
  "scoop_repo": "https://github.com/ScoopInstaller/Scoop",
  "scoop_branch": "master"
}
```

Issue #9419, opened 2022-10-08
PowerToys 0.63.0 doesn't work properly


raycheung:
## Bug Report

**Package Name:** powertoys

### Current Behaviour

The Settings page doesn't not show. A couple modules won't run properly.

### Expected Behaviour

The Settings page and all modules should run.

### Additional context/output

DLL files are now hardlinked into subfolders after this commit: https://github.com/microsoft/PowerToys/commit/665851300dd0adbbbc338788314dcfbb4a10e50b .

### Possible Solution

Replicate the new hardlink steps in the installer script with simple powershell post-install script.

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.6

**Additional software:** [(optional) e.g. ConEmu, Git]

#### Scoop Configuration

```json
{
  "lastupdate": "2022-10-08T11:27:45.6662818+08:00",
  "SCOOP_REPO": "https://github.com/ScoopInstaller/Scoop",
  "SCOOP_BRANCH": "master"
}
```

PR #10177, opened 2023-01-04
PowerToys@0.66+: Remove auto-update

jaimecbernardo:
PowerToys adopted a strategy of creating links in the file system to share dependencies between utilities. The fact that the PowerToys Scoop package doesn't run the installer means that these dependencies don't get created except by doing extra steps.
This means that auto-updating the PowerToys manifest leads to installations that do not work properly at runtime. This has happened with 0.63 and 0.64. ( #9419 and #9643 , respectively).

This is bound to happen again in 0.66, set to release this week, since the inclusion of self-contained .NET binaries: https://github.com/microsoft/PowerToys/commit/6ac508fb933d0a30e1176baae34981f91e0984cf

This commit removes the auto-update feature, as the scoop manifest needs extra additions to have a working 0.66 installation.

Relates to #9419 and #9643

### DIFF
--- a/bucket/powertoys.json
+++ b/bucket/powertoys.json
@@ -37,17 +37,5 @@
             "PowerToys.exe",
             "PowerToys"
         ]
-    ],
-    "checkver": "github",
-    "autoupdate": {
-        "architecture": {
-            "64bit": {
-                "url": "https://github.com/microsoft/PowerToys/releases/download/v$version/PowerToysSetup-$version-x64.exe",
-                "hash": {
-                    "url": "https://github.com/microsoft/PowerToys/releases/tag/v$version",
-                    "regex": ">$sha256<"
-                }
-            }
-        }
-    }
+    ]
 }


jaimecbernardo:
@raycheung , what do you think? This makes sense right?

jaimecbernardo:
Hi @niheaven,
Mentioning you since you were the one who merged the fixes for the mentioned issues.
There will be broken installations of PowerToys if the auto-updater runs after PowerToys 0.66 gets released.

jaimecbernardo:
Disclaimer: I'm a part of the PowerToys core team, by the way: https://github.com/microsoft/PowerToys/blob/9a77c3a87a98ff66a6ade26f36ffe3869481cdd6/COMMUNITY.md#powertoys-core-team

HUMORCE:
unnecessary in fact, because we usually fix/rollback manifests when they are not working properly.

If your team is thinking about the UX for Scoop users, please consider portable installation.

jaimecbernardo:
We plan on removing the need for UAC in coming releases, so perhaps the "non-extra" PowerToys manifest would be enough to serve the users who use this one as well.

jaimecbernardo:
Regarding the need for this change: Users who install a wrong update (without the hardlinks) would need to manually reinstall PowerToys to get it to work properly after the manifest is fixed, correct?

HUMORCE:
> Users who install a wrong update (without the hardlinks) would need to manually reinstall PowerToys to get it to work properly after the manifest is fixed, correct?

Yes

Issue #10066, opened 2022-12-20
autohotkey@2.0.0: decompress error


dertuxmalwieder:
```
Downloading new version
Starting download with aria2 ...
Download: Download Results:
Download: gid   |stat|avg speed  |path/URI
Download: ======+====+===========+=======================================================
Download: d404b5|OK  |   2.3MiB/s|C:/Users/stop_/scoop/cache/autohotkey#2.0.0#https_github.com_Lexikos_AutoHotkey_L_releases_download_v2.0.0_AutoHotkey_2.0.0_setup.exe_dl.7z
Download: Status Legend:
Download: (OK):download completed.
Checking hash of AutoHotkey_2.0.0_setup.exe ... ok.
Uninstalling 'autohotkey' (1.1.36.02)
Installing 'autohotkey' (2.0.0) [64bit] from extras bucket
Loading AutoHotkey_2.0.0_setup.exe from cache.
Extracting dl.7z ... ERROR Exit code was 2!
Failed to extract files from C:\Users\stop_\scoop\apps\autohotkey\2.0.0\dl.7z.
Log file:
  ~\scoop\apps\autohotkey\2.0.0\7zip.log

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/ScoopInstaller/Extras/issues/new?title=autohotkey%402.0.0%3a+decompress+error
```

Contents of said log:

```
7-Zip 22.01 (x64) : Copyright (c) 1999-2022 Igor Pavlov : 2022-07-15

Scanning the drive for archives:
1 file, 2910208 bytes (2842 KiB)

Extracting archive: C:\Users\stop_\scoop\apps\autohotkey\2.0.0\dl.7z
WARNING:
C:\Users\stop_\scoop\apps\autohotkey\2.0.0\dl.7z
Cannot open the file as [7z] archive
The file is open as [PE] archive

--
Path = C:\Users\stop_\scoop\apps\autohotkey\2.0.0\dl.7z
Open WARNING: Cannot open the file as [7z] archive
Type = PE
Physical Size = 2910208
CPU = x86
Characteristics = Executable 32-bit NoRelocs
Created = 2022-12-20 08:14:22
Headers Size = 4096
Checksum = 0
Image Size = 5414912
Section Alignment = 4096
File Alignment = 512
Code Size = 2875392
Initialized Data Size = 40960
Uninitialized Data Size = 2494464
Linker Version = 14.33
OS Version = 6.0
Image Version = 0.0
[...]

PR #10185, opened 2023-01-05
autohotkey@2.0.2: Path escape 

HUMORCE:
Relates to https://github.com/ScoopInstaller/Extras/commit/3bdc740916fbb64121df51b2bcc051501e9bd6f6#commitcomment-95121357

### DIFF
--- a/bucket/autohotkey.json
+++ b/bucket/autohotkey.json
@@ -17,7 +17,7 @@
         "script": [
             "$ahk = @('/script', \"`\"$dir\\installer\\UX\\install.ahk`\"\", '/to', \"`\"$dir`\"\", '/user', '/silent')",
             "if ($global) { $ahk = @('/script', \"`\"$dir\\installer\\UX\\install.ahk`\"\", '/to', \"`\"$dir`\"\", '/silent') }",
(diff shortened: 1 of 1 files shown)


rashil2000:
When installing, this popped up and I chose "Yes".

image

Installation proceeded successfully. But then I ran my usual script, and this popped up: 
image

What happens if I select "Yes" here? If it downloads v1.1.36.02, will it persist across updates made to Autohotkey by Scoop?

rashil2000:
> Have you installed AHK with official installer before? I have not tested this case

No, I have only installed it using Scoop.

> All sub-versions will be uninstalled.

Then we need a way to persist it, right? Otherwise users will have to download specific versions again and again.

----

Another question: do autohotkey developers actually recommend updating packages to v2? There seems to be no backward compatibility at the moment.

HUMORCE:
> No, I have only installed it using Scoop.

unknown factors..

> Then we need a way to persist it, right? Otherwise users will have to download specific versions again and again.

I will do this in another PR.

> Another question: do autohotkey developers actually recommend updating packages to v2? There seems to be no backward compatibility at the moment.

I have not used AHK for quite time, so I am not sure about that. but the launcher of v2 contains script version detection.

refer. #10066

rashil2000:
> I will do this in another PR.

Yes, please do! It is very essential.

PR #10244, opened 2023-01-15
lazy-posh-git: Add version 0.2.0

goto1134:
Adds lazy-posh-git version 0.1.0

### DIFF
--- a/bucket/lazy-posh-git.json
+++ b/bucket/lazy-posh-git.json
@@ -1,7 +1,7 @@
 {
     "version": "0.2.0",
     "description": "PowerShell proxy command around Set-Location to defer import of posh-git module until one changes working directory to the root of a git directory.",
-    "homepage": "https://github.com/bergmeister/lazy-posh-git/",
+    "homepage": "https://github.com/bergmeister/lazy-posh-git",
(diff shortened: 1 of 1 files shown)


HUMORCE:
Maybe we should switch to gh releases? both of checkver and download url.

goto1134:
I looked at posh-git package and used the same approach. If there is a reason for switching the source, surely I could do it!

HUMORCE:
> If there is a reason for switching the source

- Avoid downloading from a fixed-location CDN
- High level version check

goto1134:
@HUMORCE, that sounds reasonable to me! Thank you for the suggestions. Should we fix posh-git too?

HUMORCE (approved):

HUMORCE:
> Should we fix posh-git too?

of course, if you have time.

```
❯ scoop search posh-git
Results from local buckets...

Name             Version    Source   Binaries
----             -------    ------   --------
posh-git         1.1.0      extras
posh-git-beta    1.1.0      versions
posh-git-nightly 4b746d6e48 versions
```

goto1134:
Unfortunately, posh-git releases are not published on github

PR #10306, opened 2023-01-23
ddu@18.0.6.0: remove extract_dir

hu3rror:
Closes #10295

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #10315, opened 2023-01-24
ddu: Fix `pre_install`

brian6932:
Relates to https://github.com/ScoopInstaller/Extras/pull/10306#issuecomment-1401780945

### DIFF
--- a/bucket/ddu.json
+++ b/bucket/ddu.json
@@ -6,6 +6,7 @@
     "url": "https://www.wagnardsoft.com/DDU/download/DDU%20v18.0.6.0.exe#/dl.7z",
     "hash": "sha1:9ef1fba66e055c1b370fa6fd8ed867e418fbb0a7",
     "extract_dir": "DDU v18.0.6.0",
+    "pre_install": "Get-ChildItem \"$dir\" -Exclude *.exe, Settings | Remove-Item -Force -Recurse",
     "shortcuts": [
         [
             "Display Driver Uninstaller.exe",


HUMORCE:
The extras part contains useless part of files, right?

PDB are not in it.

Scoop remove actually extras files for packages only.

HUMORCE:
There is no manifest to actively delete pdb-like files.

HUMORCE:
> 7Zip dump

What I mean is that the pdb file is not a redundant file

This change will also cause other files to be deleted: `Issues and solutions.txt`, `Licence.txt`, `Readme.txt` and files that may be added in the future.

brian6932:
>What I mean is that the pdb file is not a redundant file

How exactly is it useful to someone that just wants to use ddu?
It doesn't seem worth it to keep a pdb file for the 99% of users that won't interact with it. If a user wanted it, they could always grab it themselves was kinda my point, if this is the main point of contention, then it can just be added to the exclude rule.

>This change will also cause other files to be deleted: `Issues and solutions.txt`, `Licence.txt`, `Readme.txt` and files that may be added in the future.

Well this isn't really useful to someone that just wants the program, it's on their site. Bundling READMEs and licenses with programs that people don't use doesn't really make sense to me, these are publicly available after all, in general it just makes package size larger for no real gain.

This is really a minor thing in the first place, so if it is that hard to agree with, I'm perfectly fine with closing this, or altering it to not be an exclude rule, and/or to loop some deletions.

HUMORCE:
Most packages retain the pdb file, and ddu is no exception. even if these files are used by only 1 person, it is included in the distribution.

For minimal installation or other specific requirements: consider creating copies in your own bucket.

PR #10330, opened 2023-01-26
tailscale: Add note and scripts for adding to startup programs

### LINKED ISSUE #10329: tailscale: No option to add to startup folder
We should display a note to users with instructions on how to add Tailscale to their startup folder. Also, we should be able to provide a script for adding and removing the shortcut from the startup folder.

shayne:
Make it easier for users to add Tailscale to the startup programs folder by providing a note with scripts to add/remove from the startup folder.

Changes:
 - add `"notes"` with instructions for adding to to startup programs 
 - `pre_install` move is_admin check to fail early
 - `pre_install` generates `tailscale-{add,remove}-startup` scripts
 - `uninstaller` also removes the startup shortcut

Closes #10329


shayne:
@tech189, any objections to merging this in?

tech189:
Not sure if it's necessary to create a script to add the program to start up within Scoop's installation, it would probably be better to do this within the application itself. For example, QuickLook's menu:
image

However, getting Scoop to manage this is a [popular feature request](https://github.com/ScoopInstaller/scoop/issues/2276) and since that's not currently possible, we could add this if no one else objects. We would just need to put the script into the `scripts` folder rather than into the program's folder.

shayne:
Thanks! I would prefer to include the script (and am happy to move it to a `scripts` folder) if that's OK. Short of having it automatically added to start-up programs, showing users the note with an actionable command is ideal since, for Tailscale, it's almost always what's wanted.

shayne:
Updated to use `scripts`

HUMORCE:
Use `notes` instead of script.

```
Add to startup:
New-ItemProperty -Path HKCU:\Software\Microsoft\Windows\CurrentVersion\Run -Name Tailscale -PropertyType String -Value "$dir\\tailscale-ipn.exe"

Remove from startup:
Remove-ItemProperty -Path HKCU:\Software\Microsoft\Windows\CurrentVersion\Run -Name Tailscale
```

StarsbySea (approved):
LGTM. Thanks!

shayne:
LMK if we want to use notes instead of what is currently posted

StarsbySea:
> LMK if we want to use notes instead of what is currently posted

Maybe you should take HUMORCE's advice, after all he is a member of ScoopInstaller and has a lot more experience than me 😂

jasongodev:
What's the progress of this PR? Is this ready for review?

shayne:
@jasongodev, it's ready if folks are OK with the script approach. The notes approach is OK, but the line length causes line breaks in PowerShell, making it fail to copy and paste, which is a real bummer.

image

jasongodev (changes requested):
@shayne Please run a checkver update of this manifest to update the version and hash. This will pass the verifier bot test.

shayne:
Thanks, done!

shayne:
Whoops! Didn't amend the commit properly. 😅

StarsbySea (approved):

jasongodev (approved):

shayne:
It looks like we're approved. Any chance we could land this PR? 🙏

tech189 (approved):
Tested startup scripts and checked Tailscale started at login 👍

PR #10426, opened 2023-02-03
sophiapp: Remove "SophiApp.exe.config" from `persist`

Zliced13:
Reason: This file is already included inside the zip file, and only exists to help the **SophiApp.exe** file. There is no need to persist it.


Zliced13:
This also fixes the issue that prevents SophiApp from launching at all.

PR #10529, opened 2023-02-17
Duplicate manifests. Remove foxit-reader.json

cvladan:
There is exactly the same package in another manifest `foxit-pdf-reader.json` That one is newer, and the manifest naming is more appropriate, as the name of the tool is "Foxit PDF Reader" and not "Foxit Reader". In any case, this manifest should be removed.

- [+] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).


niheaven:
Will be fixed when we have some new feature.

PR #10697, opened 2023-03-08
(chore): Upgrade http to https

StarsbySea:
With the help of ChatGPT, I used the following code to check all the json files and replace the url that could be upgraded to https.

```python
import os
import json
import requests

# Function to convert http url to https url
def convert_url(url):
    if url.startswith('http://'):
        url = 'https://' + url[len('http://'):]
    return url

# Function to process individual JSON file
def process_file(file_path):
    proxies = {
        'http': '127.0.0.1:5551',
        'https': '127.0.0.1:5551'
    }

    # Open and read JSON file
    with open(file_path, 'r', encoding='utf-8') as f:
        data = json.load(f)
    file_modify = False

    # Iterate over JSON key-value pairs
    for key, value in data.items():
        # Check if value is a string and starts with 'http://'
        if isinstance(value, str) and value.startswith('http://'):
            # Convert 'http://' to 'https://'
            https_url = convert_url(value)
            try:
                # Send a HEAD request to the https URL using proxies
                response = requests.head(https_url, proxies=proxies, allow_redirects=True)
                # If the request was redirected and the final response is successful, replace the http URL with the final URL in JSON data
                if response.status_code == 200:
                    if key == 'url':
                        # Some download addresses are redirected according to the user IP and region, we do not save the redirected address, we just upgrade the protocol to https
                        data[key] = https_url
                    else:
                        data[key] = response.url
                    file_modify = True
            except requests.exceptions.RequestException:
                # If an error occurs while sending the request, move on to the next key-value pair
                pass

    # If any changes were made to the JSON data, write it back to the file
    if file_modify == True:
        with open(file_path, 'w', encoding='utf-8') as f:
            json.dump(data, f, indent=4)
            f.write('\n')
        print(f"Modified file: {file_path}")

# Function to process all JSON files in a directory
def process_directory(directory):
    for dirpath, dirnames, filenames in os.walk(directory):
        for filename in filenames:
            if filename.endswith('.json'):
                file_path = os.path.join(dirpath, filename)
                process_file(file_path)

if __name__ == '__main__':
    # Modify directory variable to point to the target directory
    directory = r'C:\Users\Kevin\EXtras\bucket'
    process_directory(directory)

```

### DIFF
--- a/bucket/3d-fix-manager.json
+++ b/bucket/3d-fix-manager.json
@@ -1,7 +1,7 @@
 {
     "version": "1.85",
     "description": "Tool for Installing Nvidia 3D Vision Fixes and changing 3D related settings such as 3D depth and convergence while gaming.",
-    "homepage": "http://helixmod.blogspot.com/2017/05/3d-fix-manager.html",
+    "homepage": "https://helixmod.blogspot.com/2017/05/3d-fix-manager.html",
     "license": "Freeware",
     "architecture": {
         "64bit": {
--- a/bucket/aida64extreme.json
+++ b/bucket/aida64extreme.json
@@ -6,7 +6,7 @@
         "identifier": "Shareware",
         "url": "https://www.aida64.com/licensing"
     },
-    "url": "http://download.aida64.com/aida64extreme685.zip",
+    "url": "https://download.aida64.com/aida64extreme685.zip",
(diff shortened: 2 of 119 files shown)


HUMORCE:
We have spent some time on #10981.

PR #10703, opened 2023-03-09
USBImager: fixed wrong hash

dsm:
fixed wrong hash 


jasongodev (changes requested):
This is not the correct way to update the hash.

The autoupdate hash routine is also incorrect.

Please create a new hash routine that detects the regex pattern of the current download file. The hash file mentioned contains a lot of hashes. You need to use regex to check for it.

dsm:
thanks for review

niheaven (approved):

PR #10819, opened 2023-03-21
camunda-modeler: fix user data dir

jfastnacht:
`camunda-modeler` saves files to AppData

### DIFF
--- a/bucket/camunda-modeler.json
+++ b/bucket/camunda-modeler.json
@@ -14,15 +14,25 @@
         }
     },
     "bin": "Camunda Modeler.exe",
+    "pre_install": [
+        "if (Test-Path \"$env:AppData\\camunda-modeler\") {",
+        "    info '[Persistent AppData]: Moving user data...'",
+        "    ensure \"$persist_dir\\UserData\" | Out-Null",
+        "    Copy-Item \"$env:AppData\\camunda-modeler\\*\" \"$persist_dir\\UserData\\\" -Recurse -Force -ErrorAction 'SilentlyContinue'",
+        "    Remove-Item \"$env:AppData\\camunda-modeler\" -Recurse -Force -ErrorAction 'SilentlyContinue'",
+        "}"
+    ],
     "shortcuts": [
         [
             "Camunda Modeler.exe",
-            "Camunda Modeler"
+            "Camunda Modeler",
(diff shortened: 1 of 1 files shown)


jfastnacht:
`camunda-modeler` is an Electron app saving some of its data in a user data directory in AppData, which is (AFAIK) not desired in the scoop-main or scoop-extras buckets.

I stumbled over this, because I wanted to add element templates, which you can either save in the app directory or in the user data directory. There is no persisted folder right now, so if I do so and I want to move my installation of camunda-modeler by copying the persist folder, I will lose my configuration and element templates in `camunda-modeler`.

Regarding the `pre_install` - I copied that from `ariang-native` in this bucket, so I assumed it's the way to go. In this case a scoop installation and a non-scoop installation won't work together anymore.

There is also the `resources` folder within the `camunda-modeler` folder itself, but this does not solely belong to the user, since it contains the `app.asar` which is part of the downloaded package, so if it should be persisted as well, the manifest needs a custom installation script.

References:
* [ariang-native pre_install](https://github.com/ScoopInstaller/Extras/blob/5ac9804cd82adf244c08db342148b4a2ba3072a1/bucket/ariang-native.json#L16-L23)
* [camunda-modeler Flags](https://docs.camunda.io/docs/components/modeler/desktop-modeler/flags/)
* [camunda-modeler Search Paths](https://docs.camunda.io/docs/components/modeler/desktop-modeler/search-paths/)

bucket/camunda-modeler.json
```diff
@@ -14,15 +14,25 @@
         }
     },
     "bin": "Camunda Modeler.exe",
+    "pre_install": [
+        "if (Test-Path \"$env:AppData\\camunda-modeler\") {",
+        "    info '[Persistent AppData]: Moving user data...'",
+        "    ensure \"$persist_dir\\UserData\" | Out-Null",
+        "    Copy-Item \"$env:AppData\\camunda-modeler\\*\" \"$persist_dir\\UserData\\\" -Recurse -Force -ErrorAction 'SilentlyContinue'",
+        "    Remove-Item \"$env:AppData\\camunda-modeler\" -Recurse -Force -ErrorAction 'SilentlyContinue'",
```
HUMORCE, line 22: prompts the users where the data is being moved from and to.

PR #10823, opened 2023-03-21
imhex: Fixed `pre_uninstall`, `checkver`, and `autoupdate.url`.

Zliced13:
Fixed it so that it only removes the context menu when the user passes the **uninstall** parameter.


niheaven:
Have you test it locally by `checkver` before creating the PR? Too many `/verify`s here.

BTW, why `reg delete` instead of `Remove-Item`?

Zliced13:
`reg delete` is much faster than `Remove-Item`. And yes, I have tested the manifest locally on my machine.

niheaven (approved):

Issue #10912, opened 2023-04-06
powertoys@0.69.0: hash check failed


jack-mil:
The Github release page has changed which breaks the regex hash lookup. 
Power Toys 0.69 now supports a User Only installation, which Scoop should use by default. This does not require admin privileges to install and should avoid a UAC popup when installing by Scoop. 

I will work on a PR for this, updating the powertoys.json manifest.

PR #10915, opened 2023-04-06
Update Powertoys 0.69.0 to use User Installer

jack-mil:
Power Toys 0.69 now supports a User Only installation, which Scoop should use by default. This does not require admin privileges to install and should avoid a UAC popup when installing by Scoop.

This requires updating the hash check regex and the downloaded filename.

I will work on a PR for this, updating the powertoys.json manifest.
Relates to #10912 

### DIFF
--- a/bucket/powertoys.json
+++ b/bucket/powertoys.json
@@ -16,13 +16,13 @@
     "installer": {
         "script": [
             "Expand-DarkArchive \"$dir\\$fname\" \"$dir\\.tmp\"",
-            "Get-ChildItem \"$dir\\.tmp\\AttachedContainer\\PowerToysSetup*.msi\" | Rename-Item -NewName 'PowerToysSetup.msi' -Force",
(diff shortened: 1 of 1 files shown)


jack-mil:
I have fixed the hash regex, but the installation fails due to extracting the MSI. I will look into this, but am not familiar with the MSI installer.

jack-mil:
I have fixed the install script 
and updated the post install script to skip hardlinking the WinAppSDK dlls for the Settings Module, as those seem to exist in the Settings module by default now. 
Without this fix, the post-install script would produce errors about symlinks already existing. However, these errors can be ignored and the install works regardless. 
This fix just cleans up the command line output

StarsbySea:
local test pass
屏幕截图 2023-04-07 230342

StarsbySea (approved):
local test pass
屏幕截图 2023-04-07 230342

PR #9268, opened 2022-09-20
Add QGIS

marcolongol:
- [X] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #11074, opened 2023-04-17
Add: osgeo4w

marcolongol:


HUMORCE:
A network installer? I prefer #9268 but I don't know how to test if QGIS is deployed correctly.

marcolongol:
osgeo4w is even recommended by QGIS itself -> https://qgis.org/en/site/forusers/download.html

image

it also installs other essential packages for geospatial data processing such as gdal. I made the manifest in a way that the environment variables for these are also being set during install. -- I personally prefer having osgeo4w right away than have to deal with distinct QGIS versions on my system.

HUMORCE:
> osgeo4w is even recommended by QGIS itself

I know about this, its documentation recommends the network installer.

> it also installs other essential packages for geospatial data processing such as gdal.

Scoop(or other package managers) can also do this. manifests such as `main/poetry` is prone to the problem of the network installer/script not installing correctly.

PR #11097, opened 2023-04-22
p4v@2023.1-2431464: Fix hash

Sep102:
Apparently this installer changed upstream:
```
Checking hash of p4vinst64.msi ... ERROR Hash check failed!
App:         extras/p4v
URL:         https://cdist2.perforce.com/perforce/r23.1/bin.ntx64/p4vinst64.msi
First bytes: D0 CF 11 E0 A1 B1 1A E1
Expected:    f4cd6cbd5ee622fc543222ed19dbdd82520e9ca8d84160a3753660fd2b3ceea1
Actual:      77dc21705c1e6817eed02b6e44243f97bdfc6386c41fdb02f5af9c8c3bc9fc6c
```

### DIFF
--- a/bucket/p4v.json
+++ b/bucket/p4v.json
@@ -9,7 +9,7 @@
     "architecture": {
         "64bit": {
             "url": "https://cdist2.perforce.com/perforce/r23.1/bin.ntx64/p4vinst64.msi",
-            "hash": "f4cd6cbd5ee622fc543222ed19dbdd82520e9ca8d84160a3753660fd2b3ceea1"
+            "hash": "77dc21705c1e6817eed02b6e44243f97bdfc6386c41fdb02f5af9c8c3bc9fc6c"
         }
     },
     "extract_dir": "Perforce",


jasongodev (commented):
I believe the hashing function works fine. Usually the hash is updated automatically by Scoop so we don't need to manually update the hash like this.

HUMORCE (approved):
[The bot](https://github.com/ScoopInstaller/GithubActions) will automatically fix the hash for relevant manifest, when you create a new issue as instructed by the error output.

Sep102:
> [The bot](https://github.com/ScoopInstaller/GithubActions) will automatically fix the hash for relevant manifest, when you create a new issue as instructed by the error output.

Ah, good to know, thank you. I thought perhaps that was reporting the broken hash for a maintainer to update and submitting a PR would save the effort of that but knowing it is auto-maintained by the bot I'll be sure to use that mechanism in the future.

PR #11111, opened 2023-04-23
vidiot: Update to version 0.3.38 and fix autoupdate

StarsbySea:
vidiot: 0.3.38 (scoop version is 0.3.37) autoupdate available
Autoupdating vidiot
Could not find hash in https://sourceforge.net/projects/vidiot/files
Downloading Vidiot-0.3.38-win32.zip to compute hashes!


HUMORCE:
For this case, temporarily point 32bit download url to `Vidiot-0.3.38-win64.exe#/dl.7z`.

aliesbelik:
As per answer from the author in [discussion](https://sourceforge.net/p/vidiot/discussion/general/thread/e6a0aed717/) support of the 32bit version has been dropped (starting with 0.3.38).

Should we remove 32bit option from manifest?

PR #11132, opened 2023-04-24
office-tool-plus: Update to version 10.1.7.0 and add more architecture

StarsbySea:

### DIFF
--- a/bucket/office-tool-plus.json
+++ b/bucket/office-tool-plus.json
@@ -1,23 +1,23 @@
 {
-    "version": "10.1.7.0",
+    "version": "10.1.7.1",
     "description": "A powerful and useful tool for Office deployments",
     "homepage": "https://otp.landian.vip/",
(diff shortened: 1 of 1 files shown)


HUMORCE:
Use the zips that without runtime and suggests `windowsdesktop-runtime-lts`

StarsbySea:
I think it would be better to provide a version that includes the runtime. Firstly, users may install the wrong runtime and the software will not work, and secondly, it is not a good idea to use administrator rights to install a runtime on the system in order to use a (non-administrative) tool.

HUMORCE:
You are right, all program distributors want to integrate runtimes.

HUMORCE:
The arm64 package does not exist in gh releases, currently.

HUMORCE (changes requested):
10.1.7.1

HUMORCE (approved):

PR #11312, opened 2023-05-21
bulk-crap-uninstaller: Changed url, added hash extraction

Zliced13:

### DIFF
--- a/bucket/bulk-crap-uninstaller.json
+++ b/bucket/bulk-crap-uninstaller.json
@@ -3,7 +3,7 @@
     "description": "Bulk program uninstaller with advanced automation",
     "homepage": "https://www.bcuninstaller.com/",
     "license": "Apache-2.0",
-    "url": "https://github.com/Klocman/Bulk-Crap-Uninstaller/releases/download/v5.6/BCUninstaller_5.6_portable.zip",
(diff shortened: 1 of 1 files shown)


bucket/bulk-crap-uninstaller.json
```diff
@@ -3,7 +3,7 @@
     "description": "Bulk program uninstaller with advanced automation",
     "homepage": "https://www.bcuninstaller.com/",
     "license": "Apache-2.0",
-    "url": "https://github.com/Klocman/Bulk-Crap-Uninstaller/releases/download/v5.6/BCUninstaller_5.6_portable.zip",
+    "url": "https://ftp.halifax.rwth-aachen.de/osdn/bulk-crap-uninstaller/78782/BCUninstaller_5.6_portable.zip",
```
HUMORCE, line 6: `https://dotsrc.dl.osdn.net/osdn/bulk-crap-uninstaller/78782/BCUninstaller_5.6_portable.zip`

HUMORCE (approved):

bitoolean:
Auto-update URL times out for me (after several seconds, it won't load, or it only loads the text if you're lucky), so this change to "OSDN" maybe isn't all that great? There's a new version out, which would've easily been detected through either github or sourceforge. Does this change just need to be reverted?

PR #11335, opened 2023-05-24
windows-terminal: Switch to portable distribution

kazzarin:
The latest release added an official portable mode
https://github.com/microsoft/terminal/releases/tag/v1.17.11391.0
https://learn.microsoft.com/en-ca/windows/terminal/distributions#windows-terminal-portable

### DIFF
--- a/bucket/windows-terminal.json
+++ b/bucket/windows-terminal.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.16.10261.0",
+    "version": "1.17.11391.0",
     "description": "The new Windows Terminal, and the original Windows console host - all in the same place!",
     "homepage": "https://github.com/microsoft/terminal",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


HUMORCE (approved):

igitur:
I think it would've been prudent to display a warning that the settings file location has changed. For a while I thought all my settings have reverted to their default state. Luckily I could find my original settings file in my user profile and copy it over the new file. But honestly, I think this change could've been handled a bit better.

DavideCanton:
+1, I did not even know that 1.17 introduced portable mode. A warning (or a copy of the current settings, but it may have been hard to perform in a script) would have been better.

HUMORCE:
@igitur @DavideCanton 

Sorry, hint about portable mode has been added to manfiest now. https://github.com/ScoopInstaller/Extras/commit/e1c1203e71357317377823a7d6dfdc2ef97f220f

kazzarin:
It seems I overlooked that when testing, since I typically only check using a new install within the sandbox.

@HUMORCE Would it be preferable to have the migration be part of the installer script? Something like: 
```
"$settings_path = \"$env:LOCALAPPDATA\\Packages\\Microsoft.WindowsTerminal_8wekyb3d8bbwe\\LocalState""
"if (!(Test-Path \"$persist_dir\\settings\") -and (Test-Path $settings_path)) {"
"    info \"Copying old settings to '$persist_dir\\settings'\""
"    Copy-Item \"$settings_path\\*\" \"$dir\\settings\" -Recurse -Force"
"}"
```

HUMORCE:
Sure, but set a deadline for it.

Automatic migration side effects `scoop uninstall <app> --purge`.

igitur:
Not sold on it, but I think a migration script is overkill. A simple console warning in bright orange or something should suffice.

HUMORCE:
I've moved the hint to pre_install stage with warn message. https://github.com/ScoopInstaller/Extras/commit/dfb4c5e1e3c727d69e5c01d4447844a0ee844cbe

PR #11357, opened 2023-05-27
Update vcredist2022 and add autoupdate

r15ch13:
- Updated the main package and vcredist to version `14.36.32532.0`.
- Also added an autoupdate script. It uses the permalink to download 1 MB from the `vc_redist.x64.exe` executable and extracts the version from it. 😄
- Removed 2015, 2017 and 2019 because they are all included in the current version.
- Removed suggestions for 2015-2019
- Removed suggestions for the main packages because the suggestions should only target one version


HUMORCE (approved):

bucket/lively.json
```diff
@@ -4,8 +4,7 @@
     "description": "A free and open-source software that allows users to set animated desktop wallpapers and screensavers. ",
     "license": "GPL-3.0",
     "suggest": {
-        ".NET Desktop Runtime": "extras/windowsdesktop-runtime",
-        "VCRedist": "extras/vcredist"
+        ".NET Desktop Runtime": "extras/windowsdesktop-runtime"
```
HUMORCE, line 7:
```suggestion
        ".NET Desktop Runtime": "extras/windowsdesktop-runtime",
        "vcredist": "extras/vcredist2022"
```

https://github.com/rocksdanister/lively/blob/cdcfc3009e988022f440886a856eb24d2f3e4014/src/installer/Script.iss#L97

bucket/droidcam.json
```diff
@@ -6,9 +6,6 @@
         "identifier": "Freeware",
         "url": "https://www.dev47apps.com/droidcam-terms-privacy.html"
     },
-    "suggest": {
-        "vcredist": "extras/vcredist"
```
HUMORCE, line 10: Its installer contains vcredist2015, so suggests `extras/vcredist2022`

r15ch13: Tested it in Windows Sandbox (Windows 11) and it worked without installing vcredist, same goes for `lively`

HUMORCE: Required by specific functions/features, e.g. https://github.com/rocksdanister/lively/wiki/Common-Problems#web-wallpapers-not-working

I believe there are a lot of manifests that missing vcredist suggestion, Scoop doesn't handle them as a dependency, so not matter.

PR #11668, opened 2023-07-31
emacs: Add cmdproxy.exe to shims

nitincodery:
Emacs on windows needs `shell-file-name` to be set to path of `cmdproxy.exe` manually, by adding this into bin paths, `cmdproxy.exe` is available directly without changing paths explicitly everytime on update.

### DIFF
--- a/bucket/emacs.json
+++ b/bucket/emacs.json
@@ -16,7 +16,8 @@
         "bin\\emacs.exe",
         "bin\\emacsclientw.exe",
         "bin\\etags.exe",
-        "bin\\ctags.exe"
+        "bin\\ctags.exe",
+        "libexec\\emacs\\$version\\x86_64-w64-mingw32\\cmdproxy.exe"
     ],
     "shortcuts": [
         [


HUMORCE:
Hey, I am not an emacs user, where is the doc about `shell-file-name`?

nitincodery:
Open Emacs on windows system, `Ctrl+h`, then press `v`, then type `shell-file-name`, it will show its default value to `D:/emacs-build/emacs-28.2/nt/cmdproxy.exe` which is set during build process of emacs by archive provider, whereas when installed via scoop it must change its value to path `%USERPROFILE%/scoop/apps/emacs/current/libexec/emacs/28.2/x86_64-w64-mingw32/cmdproxy.exe`.

So, my point is, the variable `shell-file-name` is used to open shell inside Emacs, which is `cmd.exe` under windows, but, `cmdproxy.exe` is a [proxy shell](https://github.com/emacs-mirror/emacs/blob/master/nt/cmdproxy.c) designed for use with Emacs on Windows. It is a binary part of Emacs distribution (i.e. made by Emacs developers) in order to call Windows subprocesses.

So, it is needed to be manually updated after installation or update via scoop, because of version number is part of `cmdproxy.exe` path, so adding it to shims will make it available directly inside Emacs and will get updated when Emacs gets an update.

nitincodery:
Also, why? when I added emacs.json to my [private scoop bucket](github.com/nitincodery/exe), it doesn't installs emacs, and gives error in movedir function inside core.ps1, saying cannot find emacs-29.1?

HUMORCE:
> Also, why? when I added emacs.json to my [private scoop bucket](github.com/nitincodery/exe), it doesn't installs emacs, and gives error in movedir function inside core.ps1, saying cannot find emacs-29.1?

I think you are talking about https://github.com/ScoopInstaller/Extras/commit/2dbd05bcd80fd0ef37d9b6e4db470f725a219079

for main issues, i will test them tomorrow.

nitincodery:
When updating via extras/emacs it is updating as well as installing successfully, but when I use emacs.json via my private bucket it shows same error as mentioned in issue.

HUMORCE:
> When updating via extras/emacs it is updating as well as installing successfully, but when I use emacs.json via my private bucket it shows same error as mentioned in issue.

See https://github.com/ScoopInstaller/Extras/commit/2dbd05bcd80fd0ef37d9b6e4db470f725a219079

---

The value of `shell-file-name` is as your expect, even if `cmdproxy.exe` has not added to shims.

Screenshot_230801044535

HUMORCE:
> Also, I was trying to create custom shim myself (for cmdproxy.exe before), how can I make shim myself? And can it be made for .cmd/.ps1/.ahk files too?

`scoop help shim`

PR #11695, opened 2023-08-04
Remove Browser Tamer from scoop

aloneguid:
I'm an author of Browser Tamer (bt) utility and I would like it to be permanently removed from Scoop. Let me know if dropping `bt.json` is enough? Thanks.


HUMORCE:
Due to the Apache-2.0 license are selected, you cannot prohibit 3rd-parties from distributing it.

aloneguid:
Scoop has a very ancient version which is not supported/in use any more and has bugs. I'm don't have time to support Scoop, so I'd like it to be removed. I don't mind if anyone adds it in future though.

HUMORCE:
After this PR created, I have updated it to the latest version and adapted it to portable.

PR #11751, opened 2023-08-17
wxFormBuilder: Add version 4.0.0

igitur:
This PR adds wxFormBuilder. Do I really have to open a separate issue for this? Will do so if required.


aliesbelik:
@igitur, could you please

- update to the latest version (4.0.0 at the moment);
- switch to portable versions instead of installers (available in 4.0.0):
```
 wxFormBuilder-4.0.0-ucrt-x86_64.zip
 wxFormBuilder-4.0.0-i686.zip
```

Otherwise looks good.

PR #11171, opened 2023-04-28
android-sdk: Deprecate manifest

HUMORCE:
Obsolete tool, use `main/android-clt` instead:
```
> scoop install main/android-clt
> sdkmanager --install "platform-tools"
```

Relates to #9081

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

Issue #10413, opened 2023-02-02
[Bug]: flutter depends on temurin8-jdk, but it can't run on that version of JDK


dodumosu:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

flutter

### Expected/Current Behaviour

flutter depends on temurin8-jdk, but actually running the flutter command to accept the Android SDK licenses fails on that JDK version.

### Steps to Reproduce

```console
install temurin8-jdk (without any other JDK/JRE), install flutter, then try to run `flutter doctor --android-licenses` (to accept the Android SDK licenses)

PS> scoop install temurin8-jdk
PS> scoop install flutter
PS> flutter doctor --android-licenses
```

### Possible Solution

Installing a newer JDK (i have temurin17-jdk installed personally) allows 

### Scoop and Buckets Version

```console
Current Scoop version:
v0.3.1 - Released at 2022-11-15

'extras' bucket:
31469bc0a (HEAD -> master, origin/master, origin/HEAD) postman@10.9.0: Fix hash (Closes #10410)

'java' bucket:
53ffa7f8 (HEAD -> master, origin/master, origin/HEAD) temurin19-nightly-jdk: Update to version 19.0.2-7.0.202302020347

'main' bucket:
4694db89f (HEAD -> master, origin/master, origin/HEAD) keptn: Add version 1.1.0 (#4412)

'nerd-fonts' bucket:
ac2c7cb (HEAD -> master, origin/master, origin/HEAD) andika: Update to version 6.200

'versions' bucket:
14e96329b (HEAD -> master, origin/master, origin/HEAD) sharex-dev: Fix url (#974)
```

### Scoop Config

```console
last_update         aria2-enabled scoop_repo                              scoop_branch
-----------         ------------- ----------                              ------------
2/2/2023 8:56:03 PM          True https://github.com/ScoopInstaller/Scoop master
```

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.2.8
PSEdition                      Core
GitCommitId                    7.2.8
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
[...]

Issue #10335, opened 2023-01-27
[Bug]: Android-Sdk cannot shim emulator.exe


jewlexx:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

android-sdk

### Expected/Current Behaviour

When installing `android-sdk`, it attempts to shim 'emulator\\emulator.exe' and 'emualtor\\emulator-check.exe', both of which are no longer in the 'emulator' folder and are now in the 'tools' folder with everything else.

### Steps to Reproduce

```console
PS> scoop install android-sdk
Installing 'android-sdk' (4333796) [64bit] from extras bucket
Loading sdk-tools-windows-4333796.zip from cache.
Checking hash of sdk-tools-windows-4333796.zip ... ok.
Extracting sdk-tools-windows-4333796.zip ... done.
Linking F:\scoop\apps\android-sdk\current => F:\scoop\apps\android-sdk\4333796
Creating shim for 'proguard'.
Creating shim for 'proguardgui'.
Creating shim for 'retrace'.
Creating shim for 'emulator'.
Can't shim 'emulator\emulator.exe': File doesn't exist.
```

### Possible Solution

The shim paths should be changed from 'emulator//' to 'tools//' for 'emulator.exe' and 'emulator-check.exe' 

### Scoop and Buckets Version

```console
PS> scoop --version
Current Scoop version:
v0.3.1 - Released at 2022-11-15

'dorado' bucket:
36fc912a (HEAD -> master, origin/master, origin/HEAD) wasmer: Update to version v3.1.1

'emulators' bucket:
854ef78f (HEAD -> master, origin/master, origin/HEAD) yuzu-dev: Update to version EA-3336

'extras' bucket:
b6bed650f (HEAD -> master, origin/master, origin/HEAD) thisiswin11@1.4.0: Fix hash (Closes #10333)

'games' bucket:
d6f30760 (HEAD -> master, origin/master, origin/HEAD) yuzu-pineapple: Update to version EA-3334

'java' bucket:
d7da401c (HEAD -> master, origin/master, origin/HEAD) openjdk21: Update to version 21-7-ea

'lemon' bucket:
e4f0bbb4 (HEAD -> master, origin/master, origin/HEAD) ytt: Update to version 0.44.3

'main' bucket:
[...]

Issue #9081, opened 2022-08-22
android-sdk: post_install errors


redactedscribe:
## Bug Report

**Package Name:** android-sdk

### Current Behaviour
Current manifest `post_install`:

```powershell
    "post_install": [
        "if (Test-Path \"$(appdir adb $global)\") {",
        "    cmd /c mklink /j \"$dir\\platform-tools\" \"$(appdir adb $global)\\current\\platform-tools\" | out-null",
        "    attrib \"$(appdir adb $global)\\current\\platform-tools\" +R /L",
        "}",
        "&\"$dir\\tools\\bin\\sdkmanager.bat\" '--install' '\"cmdline-tools;latest\"'"
    ],
```

The errors on install that I'm getting:

```
<...>
Running post_install script...

The system cannot find the path specified.
The system cannot find the path specified.
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlSchema
        at com.android.repository.api.SchemaModule$SchemaModuleVersion.<init>(SchemaModule.java:156)
        at com.android.repository.api.SchemaModule.<init>(SchemaModule.java:75)
        at com.android.sdklib.repository.AndroidSdkHandler.<clinit>(AndroidSdkHandler.java:81)
        at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:73)
        at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:48)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlSchema
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
        ... 5 more
```

Trying to run `sdkmanager.bat` also outputs the above error, so maybe that's directly related?

### Expected Behaviour

No errors on install.

### System details

**Windows version:** 11
**OS architecture:** 64bit

**PowerShell version:** 7.2.6

#### Scoop Configuration

```json
{
  "lastupdate": "2022-08-22T23:27:09.7381354+03:00",
  "SCOOP_REPO": "https://github.com/ScoopInstaller/Scoop",
  "SCOOP_BRANCH": "master",
[...]

PR #11786, opened 2023-08-25
android-emulator: Add version 10330179

silverkorn:
The official emulator for Android development, based on QEMU.

Few notices:
- `checkver`/`hash` use a `script` to make subsequent request to a dynamically generated URL from the iframe containing the version list.
  This iframe is never resolved by any Powershell's Web request objects/methods, so an additional request has been added.
- The created `checkver`->`regex` ignores **Beta** and **Canary** releases
- Only x64 binaries are now distributed and supported

### DIFF
--- a/bucket/android-emulator.json
+++ b/bucket/android-emulator.json
@@ -1,0 +1,59 @@
+{
+    "version": "10330179",
+    "description": "The official emulator for Android development, based on QEMU.",
+    "homepage": "https://developer.android.com/studio/releases/emulator",
+    "license": {
+        "identifier": "GPL-2.0-only",
+        "url": "https://cs.android.com/android/platform/superproject/+/main:prebuilts/android-emulator/NOTICE"
+    },
+    "notes": [
+        "You can manually install assistant drivers within a Powershell terminal by running the following commands:",
+        "",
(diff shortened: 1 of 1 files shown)


silverkorn:
Hmm... I don't see how we can have a `hash` for the `autoupdate` if we can't use a `script` technique.
Unless you have a solution, should I just remove that part?

HUMORCE:
Why not just install it through SDK manager(main/andoird-clt)?

silverkorn:
Oh... Apparently I was using the `sdkmanager` from the deprecated `android-sdk` and it didn't seem to be part of the available list (same for the `platform-tools`, which was using `adb` as a scoop dependency and symlinking to it's directory)

But indeed, I see it now as well as the `platform-tools`.

Feel free to cancel this PR.
Thanks!

HUMORCE:
Since no one maintains android related manifests, I think the best practice is install `main/android-clt` then take what you need.

Related:
https://github.com/ScoopInstaller/Extras/pull/11171, https://github.com/ScoopInstaller/Extras/issues/10335, https://github.com/ScoopInstaller/Extras/issues/10413, https://github.com/ScoopInstaller/Extras/issues/9081

PR #11808, opened 2023-08-29
sublime-text: Add ability to select sublime-text in context menu

silver886:
Add ability to select sublime-text in context menu, both win10 and win11, `Open with` > `Choose another app`.

image
image


silver886:
@HUMORCE this has been a while. Is there anything I can do to move this forward?

PR #11835, opened 2023-09-04
archi@5.1.0: fix download URL

bugrasan:
Archi download url fix, based on the following error:
```
> scoop install archi
WARN  Purging previous failed installation of archi.
ERROR 'archi' isn't installed correctly.
Removing older version (5.1.0).
'archi' was uninstalled.
Installing 'archi' (5.1.0) [64bit] from extras bucket
URL https://www.archimatetool.com/downloads/archi.php?/5.1.0/Archi-Win64-5.1.0.zip is not valid
```

### DIFF
--- a/bucket/archi.json
+++ b/bucket/archi.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://www.archimatetool.com/downloads/archi.php?/5.1.0/Archi-Win64-5.1.0.zip",
+            "url": "https://www.archimatetool.com/downloads/archi5.php?/5.1.0/Archi-Win64-5.1.0.zip",
             "hash": "sha1:0048230b72fb1774a64b2b6ed3c2c14b123db7a0"
         }
     },
@@ -24,7 +24,7 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://www.archimatetool.com/downloads/archi.php?/$version/Archi-Win64-$version.zip"
+                "url": "https://www.archimatetool.com/downloads/archi5.php?/$version/Archi-Win64-$version.zip"
             }
         },
         "hash": {
--- a/bucket/archi.json
+++ b/bucket/archi.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/archi.json
```diff
@@ -5,7 +5,7 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://www.archimatetool.com/downloads/archi.php?/5.1.0/Archi-Win64-5.1.0.zip",
+            "url": "https://www.archimatetool.com/downloads/archi5.php?/5.1.0/Archi-Win64-5.1.0.zip",
```
Thomah, line 8: Hello !

A little hyphen is missin here. The URL should be : 

`https://www.archimatetool.com/downloads/archi-5.php?/5.1.0/Archi-Win64-5.1.0.zip`

bucket/archi.json
```diff
@@ -24,7 +24,7 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://www.archimatetool.com/downloads/archi.php?/$version/Archi-Win64-$version.zip"
+                "url": "https://www.archimatetool.com/downloads/archi5.php?/$version/Archi-Win64-$version.zip"
```
Thomah, line 27: Like my other comment, the URL should be : 

`https://www.archimatetool.com/downloads/archi-5.php?/$version/Archi-Win64-$version.zip`

Thomah (changes requested):
Hi !

I was looking to install Archi with Scoop but I met the same problem as you. I did some tests on your PR but it seems like there is a little mistake on the URLs.

Also, as mentioned in [CONTRIBUTING.md](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md), can you rename the PR with  : `archi@5.1.0: fix download URL` please ?

Can you make the change please ?

bucket/archi.json
bugrasan: thanks, the URL seems to changed again.

bucket/archi.json
bugrasan: - fixed second URL
- also renamed the PR

Thomah (approved):

aliesbelik:
Fixed in #12155.

PR #11873, opened 2023-09-11
ilspy: Update to version 8.1

whut:
ilspy: Update to version 8.1

Also updated checkver regex and autoupdate url, as now there is -x64 suffix in the ILSpy zip.

There is also ARM build of ILSpy, I assume it is not relevant for scoop?


HUMORCE:
#12093

PR #11997, opened 2023-10-07
cnping: Add version 1.0.0

ivanik7:
Minimal Graphical IPV4 Ping/HTTP Ping Tool.

https://github.com/cntools/cnping

### DIFF
--- a/bucket/cnping.json
+++ b/bucket/cnping.json
@@ -14,8 +14,7 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/cntools/cnping/releases/download/$version/cnping.exe",
-                "hash": "ed61944a9d720fd0ad429fa7ec97c2357fa805301532dd43049ffc5666bfdc8e"
+                "url": "https://github.com/cntools/cnping/releases/download/$version/cnping.exe"
             }
         }
     }
--- a/bucket/cnping.json
+++ b/bucket/cnping.json
@@ -1,0 +1,22 @@
+{
+    "version": "1.0.0",
+    "description": "Minimal Graphical IPV4 Ping/HTTP Ping Tool.",
+    "homepage": "https://github.com/cntools/cnping",
+    "license": "MIT",
+    "architecture": {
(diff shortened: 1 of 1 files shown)


bucket/cnping.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    "bin": "cnping.exe",
+    "checkver": "github",
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/cntools/cnping/releases/download/$version/cnping.exe",
+                "hash": "ed61944a9d720fd0ad429fa7ec97c2357fa805301532dd43049ffc5666bfdc8e"
+            }
+        }
```
aliesbelik, line 20:
```suggestion
        "architecture": {
            "64bit": {
                "url": "https://github.com/cntools/cnping/releases/download/$version/cnping.exe"
            }
        },
        "hash": {
            "url": "https://github.com/cntools/cnping/releases/tag/$version",
            "find": "SHA256:\\s+$sha256\\s+$basename"
        }
```

aliesbelik (changes requested):
Otherwise looks good.

bucket/cnping.json
ivanik7: This hash regex rely on human-written release description. For example, for beta versions there is no hash in description. I don't think this is more reliable than just specifing hash in manifest.

aliesbelik: > This hash regex rely on human-written release description. For example, for beta versions there is no hash in description. I don't think this is more reliable than just specifing hash in manifest.

Hi @ivanik7.

The idea is that [autoupdate](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate) section is used to update manifest automatically (when `checkver` detects new stable version). So we shouldn't use any hard-coded values here: either we can detect hash automatically (from release notes, file, etc), or we exclude autoupdate.hash from manifest completely and hash will be calculated for downloaded binary.

aliesbelik (approved):

PR #12099, opened 2023-10-25
teraterm@5.0: Switch to github release

topstone:
OSDN is currently unstable, so [current redistribution is via GitHub](https://github.com/TeraTermProject/osdn-download/releases).

### DIFF
--- a/bucket/musicbee.json
+++ b/bucket/musicbee.json
@@ -1,5 +1,5 @@
 {
-    "version": "3.5.8447",
+    "version": "3.5.8698",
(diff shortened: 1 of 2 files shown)


topstone:
Which is the better way?

1.  To modify autoupdate-related fields and remake PR.
2.  To leave them as-is.

CryoRenegade:
> Which is the better way?
> 
>     1. To modify autoupdate-related fields and remake PR.
> 
>     2. To leave them as-is.

Changing them over to github will be the better option, you can keep adding to this PR via your branch. If you need help let us know!

HUMORCE (approved):

PR #12336, opened 2023-11-26
QuickInstaller: Add version 0.2.8748.37063

ath0mas:
[QuickInstaller](https://quickinstaller.net/) v0.2


ath0mas:
> Does it need to be unpacked?

It already has `"innosetup": true`, but are you thinking of anything more?

I have been using it as-is from my own bucket since I created this PR, so I suppose it's ok.

Issue #12533, opened 2023-12-26
[Request]: otvdm


filip2cz:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

otvdm

### Description

16-bit Windows (Windows 1.x, 2.x, 3.0, 3.1, etc.) emulator for 64-bit Windows.

### Homepage

https://github.com/otya128/winevdm/

### Download Link(s)

https://github.com/otya128/winevdm/releases/download/v0.9.0/otvdm-v0.9.0.zip

### Some Indication of Popularity/Repute

Stars: 2.3k
Forks: 137

StudentWeis:
#12534

PR #12534, opened 2023-12-26
otvdm: Add version 0.9.0

filip2cz:
Relates to [#12533](https://github.com/ScoopInstaller/Extras/issues/12533)


StudentWeis:
Do you need to consider 'Run "install" shortcut or right-click on install.inf and select "Install"' to install this software?

filip2cz:
> Do you need to consider 'Run "install" shortcut or right-click on install.inf and select "Install"' to install this software?

You can do it, but it is not necesary to run this software. Should I include it?

PR #12580, opened 2024-01-08
feat: add nushell shell. A new type of shell (www.nushell.sh)

MrStrik3:
Add nushell to the Extras scoop bucket. nushell is a new type of shell which is increasing in popularity


HUMORCE:
nushell already exists in the Main bucket: https://github.com/ScoopInstaller/Main/blob/master/bucket/nu.json

PR #12606, opened 2024-01-13
scribus: small correction checkver regex

LaPingvino:
The regex for checkver was temporarily changed to something inferior that did work in the meantime. This should work better and follows what it used to be before the change to the new stable branch.

### DIFF
--- a/bucket/scribus.json
+++ b/bucket/scribus.json
@@ -27,7 +27,7 @@
     ],
     "checkver": {
         "url": "https://www.scribus.net/downloads/",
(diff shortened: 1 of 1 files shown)


LaPingvino:
The autoupdate was wrong, corrected the 1.6.1 link and autoupdate link as well. Sorry!

LaPingvino:
/verify again because I double checked the link and it should be right...

LaPingvino:
/verify again because I double checked the link and it should be right...

LaPingvino:
Ok I give up, someone please give this a look, take over and correct. I also already replaced my Windows with Linux again so it's not for my use anymore xD

PR #12675, opened 2024-01-21
koboldcpp@1.55.1: Fix hash

Slach:


aliesbelik:
@Slach Looks like the original hash value is right, please recheck.

aliesbelik:
Closing since no changes required.

PR #12758, opened 2024-02-06
thorium: Update to version 121.0.6167.204, fix autoupdate & checkver

Icey-Glitch:
Updates the manifest to support the new versioning in the thorium repo. Now looks for SSE3 version in the release to maintain support for older cpus as before, also adds architecture properties.

### DIFF
--- a/bucket/thorium.json
+++ b/bucket/thorium.json
@@ -6,8 +6,12 @@
         "identifier": "BSD 3-Clause",
         "url": "https://github.com/Alex313031/Thorium-Win/blob/main/LICENSE"
     },
-    "url": "https://github.com/Alex313031/Thorium-Win/releases/download/M119.0.6045.214/Thorium_119.0.6045.214.zip",
-    "hash": "0af9e4391dd67948732ffc1d8b8e194003163b757d7c84ae7191d814548dddac",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


Icey-Glitch:
what is the status on this?

aliesbelik:
Fixes #12832.

aliesbelik:
I'm not sure if we should switch from AVX version (with which manifest was originally merged for version M117.0.5938.157) to SSE3.

On the other hand, IMO, the "default" manifest should operate either the most capable version (which is AVX2) or the most compatible one (which is SSE3), the other modifications could be hosted in `Versions` bucket.
But there is already PR for AVX2 in `Versions`: ScoopInstaller/Versions#1438.

So I'm a bit stuck how to handle this situation properly.
@HUMORCE, @niheaven, @rashil2000, could you please advise?

The same refers to Mercury browser (#12132, #12811) from the same author that has 4 release modifications (SSE3/SSE4/AVX/AVX2).

Icey-Glitch:
I think the most reasonable way is to have the most compatible version be in extras.
Versions should have everything else; If AVX2 was the used here i feel like the naming would have to signify that so it would kinda defeat the purpose of it being in extras.

aliesbelik:
Ok, since no other reactions, switched to SSE3 option and updated to the latest version (121.0.6167.204).
For AVX2 please track ScoopInstaller/Versions#1438 in `Versions`.

PR #12785, opened 2024-02-11
LunarVim@1.3.0: Update install.ps1 hash

toby7002:
LunarVim install.ps1 hash is not correct so I downloaded the `install.ps1` script and hash it manually to get a correct hash  

### DIFF
--- a/bucket/lunarvim.json
+++ b/bucket/lunarvim.json
@@ -9,7 +9,7 @@
         "https://raw.githubusercontent.com/LunarVim/LunarVim/master/utils/bin/lvim.ps1"
     ],
     "hash": [
-        "d4c8fc99b9e92ee5c92d267341040075618d7c096161bafa4f385d7b64c474f2",
+        "fc6f0ebc6104cd085c58e35ce737a367afcda2e10080a506212762eae5294d52",
         "90799fcdc9dd2ca8959281fabeed2683562a9df59943fbec63405266ebde8f0e",
         "1a04e5c7981598105d26f6340d9c6f80c2486515842a797e3918151861e713a1"
     ],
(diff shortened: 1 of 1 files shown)


aliesbelik:
Looks like hash is correct but for the version 1.3.0, and urls point to the files from master.
I guess this will cause the issue each time any of files will be updated.

aliesbelik:
In addition to the previous ones.

Added `checkver`, `autoupdate` and versioned urls to the manifest.
Looks like the powershell scripts used in manifest are quite rarely changed but using their master versions along with calculated hashes will cause "hash check failed" issues like this one after each change in the files.

I'm not an active (neo|lunar)vim user so if you see better ways to resolve this let's discuss.

toby7002:
All goods but should we added [git](https://cli.github.com/), [make](https://www.gnu.org/software/make/), [pip](https://pypi.org/project/pip/), [python](https://www.python.org/), [npm](https://npmjs.com/), [node](https://nodejs.org/), [cargo](https://www.rust-lang.org/tools/install) and [ripgrep](https://github.com/BurntSushi/ripgrep)? Because LunarVim requires them 🤔

aliesbelik:
> All good but should we added [git](https://cli.github.com/), [make](https://www.gnu.org/software/make/), [pip](https://pypi.org/project/pip/), [python](https://www.python.org/), [npm](https://npmjs.com/), [node](https://nodejs.org/), [cargo](https://www.rust-lang.org/tools/install) and [ripgrep](https://github.com/BurntSushi/ripgrep)? Because LunarVim requires them 🤔

Oh, there are a lot of them.
I've added `neovim` only as a primer, suppose it could even go to [depends section](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties):

> depends:
>   Runtime dependencies for the app which will be installed automatically.

If all the other prerequisites aren't absolutely necessary for runtime they could be added to `suggest`:

> suggest:
>   Display a message suggesting optional apps that provide complementary features.

PR #12886, opened 2024-03-01
Paisa: Add version 0.6.6 (CLI only)

### LINKED ISSUE #12885: [Request]: Paisa
### Prerequisites

### Criteria

- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

Paisa

### Description

Personal Finance Manager

### Homepage

https://paisa.fyi/

### Download Link(s)

https://github.com/ananthakumaran/paisa/releases/latest

2K stars, 91 forks on GitHub.

shivjm:
Closes #12885 

## Complications

In order to make it portable, I set the `PAISA_CONFIG` environment variable to `$persist_dir/paisa.yaml`. Without it, Paisa requires the config file to be specified at runtime or uses its non-portable defaults. However, when that variable is set, Paisa *requires* a configuration file, so without any intervention, it’ll fail every time. My solution is to test whether paisa.yaml and main.ledger exist after installation, and run `paisa init` if not (creating both). I would appreciate any better ideas.

I don’t know whether setting `PAISA_CONFIG` here will override any custom value the user has. That would be undesirable. Again, suggestions are welcome.

Note that this package only includes the CLI. I don’t personally have a need for the GUI, but it could be added in future; whether that would be a separate package or an additional binary here is a question I’m not currently able to answer due to my limited knowledge.


shivjm:
I’d left out `-Force` when creating the directory, so I just added that.

Issue #12856, opened 2024-02-26
[Bug]: czkawka-gui is not updated to 7.0.0 due to archive name change


soredake:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

czkawka-gui

### Expected/Current Behaviour

czkawka-gui is not updated to 7.0.0 due to archive name change, `windows_czkawka_gui.zip` ->  `windows_czkawka_gui_gtk_410.zip`

### Steps to Reproduce

```console
❯ scoop install czkawka-gui
Updating Scoop...
Scoop was updated successfully!
WARN  'czkawka-gui' (6.1.0) is already installed.
Use 'scoop update czkawka-gui' to install a new version.
```

### Possible Solution

Update archive name in package manifest https://github.com/ScoopInstaller/Extras/blob/500ba541c0d32839ed46c03aecf6a0a5efebd08b/bucket/czkawka-gui.json#L25

### Scoop and Buckets Version

```console
v0.3.1 - Released at 2022-11-15

'extras' bucket:
500ba541c (HEAD -> master, origin/master, origin/HEAD) metasploit-framework: Update to version 6.3.58-20240226113509

'games' bucket:
4bbc1fed6 (HEAD -> master, origin/master, origin/HEAD) yuzu: Update to version 0-1727

'java' bucket:
0b128307 (HEAD -> master, origin/master, origin/HEAD) temurin8-nightly-jre: Update to version 8.0.412-4.0.202402231854

'main' bucket:
c9a5c65e2 (HEAD -> master, origin/master, origin/HEAD) poetry: Update to version 1.8.1

'naderi' bucket:
a7380487 (HEAD -> master, origin/master, origin/HEAD) chord: Deprecated

'nonportable' bucket:
0694d98 (HEAD -> master, origin/master, origin/HEAD) 10x-np: Update to version 1.0.129

'sysinternals' bucket:
e669cc0 (HEAD -> main, origin/main, origin/HEAD) chore(ci): switch to `pull_request_target` trigger

'versions' bucket:
4d4344c4c (HEAD -> master, origin/master, origin/HEAD) vscode-insiders: Update to version 1.87.0-insider+1708955303482
```

### Scoop Config

```console
last_update         use_external_7zip scoop_repo                              scoop_branch
[...]

PR #12915, opened 2024-03-06
czkawka-gui: Update to 7.0.0

pinjeff:
Updates czkawka-gui to new version while still using the gtk gui as the slint ui is still experimental


aliesbelik:
Fixes #12856.

Any argumentation which build - gtk_46, gtk_410 - to use is welcome.

pinjeff:
both versions work on my windows 11 machine
but my windows 10 machine can only launch the gtk_46 version, so that's what I went with

PR #12933, opened 2024-03-10
spoon: add version v0.0.1

Bios-Marcel:
This adds https://github.com/Bios-Marcel/spoon

It's a couple of milliseconds slower than https://github.com/shilangyu/scoop-search and offers completion, similar to https://github.com/Moeologist/scoop-completion.

The idea is to add more to this, as time goes on and improve it where possible. It currently doesn't offer any destructive capabilities, such as cleanup / uninstall and so on. Available commands are mostly wrappers around scoop.

### DIFF
--- a/bucket/spoon.json
+++ b/bucket/spoon.json
@@ -1,0 +1,39 @@
+{
+    "version": "0.0.1",
+    "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


Bios-Marcel:
I've executed shasum -a 256 on a freshly downloaded binary and it's the same as specified in the manifest. So I don't really know what the bot wants from me :)

PR #13088, opened 2024-04-10
doublecmd: Use portable edition

### LINKED ISSUE #2181: Doublecmd .msi instead portable edition?
This is not an issue itself but an observation. Since doublecmd has a portable zipped version would not be better to use that rather using the regular msi, unpacking, etc.?

Benefit here would be that portable edition stores config files in the portable app directorty instead than in the roaming profile (as the regular .msi does this by default).

Changing installation from .msi to the .zip edition would require to preserve some files in the portable app when updating but I think this is something scoop does in the manifests.

Thanks,
Ricard

### LINKED ISSUE #10849: DoubleCMD not portable
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

doublecmd

### Expected/Current Behaviour

double commander package uses common installation, but this app supports a configurable settings folder (user home / app folder). is it possible to preconfigure (postinstall) to use scoop `persist` folder?

### Steps to Reproduce

```console
* install doublecmd
* check `persist` folder - no settings for app
```

### Possible Solution

some post-install script should edit settings.ini

### Scoop and Buckets Version

```console
latest
```

### Scoop Config

```console
common
```

### PowerShell Version

```console
latest
```

### Additional Softwares

_No response_

niheaven:
- Closes #2181
- Closes #10849

### DIFF
--- a/bucket/doublecmd.json
+++ b/bucket/doublecmd.json
@@ -5,15 +5,15 @@
     "license": "GPL-2.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://downloads.sourceforge.net/project/doublecmd/DC%20for%20Windows%2064%20bit/Double%20Commander%201.1.12/doublecmd-1.1.12.x86_64-win64.msi",
-            "hash": "sha1:ec72ac1c583983a1ba3ff5ed5dc34faf4b5830ba"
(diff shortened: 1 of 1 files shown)


goyalyashpal:
awesome. i did one more thing in my local manifests... use the github releases for link. but that shouldn't be any difference

PR #13155, opened 2024-04-21
RemoveDrive: Add version 3.4.3

svraka:
RemoveDrive prepares drives for "Safe Removal" by commandline. The app has been included in a popular but not "known" bucket by @rasa ([here](https://github.com/rasa/scoops/blob/master/bucket/removedrive.json)), and the manifest was taken from there. I think this app deserves to be in one of the "known" buckets as well for its usefulness. The app fulfils all but one [criteria for the Main bucket](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket), hence the PR in this bucket.

> * a reasonably well-known and widely used developer tool (e.g. if it's a GitHub project, it should have at least 500 stars and 150 forks)

As a freeware, not open source program it is hard to gauge its popularity. It is recommended in a [top answer](https://superuser.com/a/443163) in a [popular Stackoverflow post](https://superuser.com/questions/443162/remove-usb-device-from-command-line) on how to remove USB device from command line and the program seems to be maintained with regular updates.

> * the latest stable version of the program

It is the latest version.

> * the full version i.e. not a trial version

It is a full version of a freeware program.

> * a fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

No need for pre/post install scripts, although there is no version-specific URL, and as far is I can tell, old versions are not available from the developer's site.

> * a non-GUI tool

It is a non-GUI tool.


svraka:
Updated to 3.4.4.

PR #13231, opened 2024-05-04
plugdata: Add version 0.8.3

make-42:


make-42:
Resolves https://github.com/ScoopInstaller/Extras/issues/13313

aliesbelik:
1. Please fix manifest ([log](https://github.com/ScoopInstaller/Extras/actions/runs/8949548959/job/25141807478?pr=13231#step:4:26)).
2. Add 32bit version as well.

amreus:
The `bin` shim entry uses spaces which would be unexpected to a command line user. The user would need to quote the command as in:

```
> "plug data"
```

So maybe use `plugdata`.

In general I am OK with a `bin` shim for gui programs, but if there are no command line options then there is probably no need for one. 

And thank you for contributing!

make-42:
Sure thing! I modified the manifest to conform with your requirements!

PR #12997, opened 2024-03-24
alacritty@0.13.2: Add Open with Alacritty context menu

jenbroek:
Adds a bundled registry file to optionally add an "Open Alacritty here" context menu entry just as Alacritty's own installer[^1] does.

Closes #12983

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

[^1]: https://github.com/alacritty/alacritty/blob/master/alacritty/windows/wix/alacritty.wxs#L41-L49

PR #13358, opened 2024-05-26
add context menu for alacritty and fix sublime bug

rafaeloledo:
Inspired on the #12997, add more registry entries to work with right click on folder directly. Alter the pop-up text to be `Open with Alacritty` instead of `Open Alacritty here` to be cohesive with `sublime-text`  and `vscode` examples.

Closes #12997

Sublime:
Fix wrong path in background directory. If i open in the root of any drive, it opens a file.
Correct behavior: open the folder instead.

Wrong:
image

Correct:
image

Accordingly to my tests, surrounding the %v sequence with escaped `"` in alacritty causes it to reproduce the same behavior of sublime's bug being corrected. But this behavior is unpredictable, it internaly transforms the path and `cd` in, despite of accusing the error.

image

### DIFF
--- a/scripts/alacritty/install-context.reg
+++ b/scripts/alacritty/install-context.reg
@@ -1,0 +1,13 @@
+Windows Registry Editor Version 5.00
+
+[HKEY_CURRENT_USER\Software\Classes\Directory\shell\Open with &Alacritty]
+@="Open with &Alacritty"
+"Icon"="$alacritty"
+[HKEY_CURRENT_USER\Software\Classes\Directory\shell\Open with &Alacritty\command]
+@="\"$alacritty\" --working-directory %v"
+
+[HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\Open with &Alacritty]
+@="Open with &Alacritty"
+"Icon"="$alacritty"
+[HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\Open with &Alacritty\command]
+@="\"$alacritty\" --working-directory %v"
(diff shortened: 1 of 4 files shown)


scripts/alacritty/install-context.reg
```diff
@@ -0,0 +1,13 @@
...
+@="Open with &Alacritty"
+"Icon"="$alacritty"
+[HKEY_CURRENT_USER\Software\Classes\Directory\shell\Open with &Alacritty\command]
+@="\"$alacritty\" --working-directory %v"
+
+[HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\Open with &Alacritty]
+@="Open with &Alacritty"
+"Icon"="$alacritty"
+[HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\Open with &Alacritty\command]
+@="\"$alacritty\" --working-directory %v"
```
jenbroek, line 13: You need to use `\"%v\"` here (see: https://github.com/alacritty/alacritty/pull/7875#issuecomment-2024441897). To be sure, I also tested this by creating a directory named '力 九' (just copied some random Hanzi). With just `%v`, Alacritty won't launch, but it does with `\"%v\"`. I imagine the same would hold true for the Sublime Text registry script.

jenbroek (changes requested):

scripts/alacritty/install-context.reg
rafaeloledo: Thank you for clarifying this. I think that's no solution for root dir, it depends on microsoft supporting Unicode sequences in the registry. Closing. Feel free to merge the diffs in your PR.

PR #13565, opened 2024-07-05
wingetui: rename manifest to unigetui.json and update

redactedscribe:

WingetUI has recently completed its rename to [UniGetUI](https://github.com/marticliment/UniGetUI).

**Note:** Since UniGetUI still uses WingetUI as the executable's name, the manifest will have to be updated in the future if the process is renamed to UniGetUI.exe.


marticliment:
fix marticliment/UniGetUI#2434

aliesbelik:
Fixed by #14311.

marticliment:
I am afraid the issue does not get fixed by the renaming, I attempted to move the package to nonportable since the extras package with innoextract is broken (some installation steps are required for UniGetUI to work proprtly)

PR #13666, opened 2024-07-20
dbeaver6: persist configuration

raisercostin:
Persist configs by creating a symlink at $env:APPDATA\\DBeaverData to <scoop>/persist/dbeaver/DBeaverData - https://dbeaver.com/docs/dbeaver/Workspace-Location/

### DIFF
--- a/bucket/ab-av1.json
+++ b/bucket/ab-av1.json
@@ -1,13 +1,13 @@
 {
-    "version": "0.11.5",
+    "version": "0.11.6",
     "description": "AV1 re-encoding using ffmpeg, svt-av1 & vmaf.",
     "homepage": "https://github.com/alexheretic/ab-av1",
     "license": "MIT",
     "depends": "main/ffmpeg",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/alexheretic/ab-av1/releases/download/v0.11.5/ab-av1.exe",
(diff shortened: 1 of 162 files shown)


raisercostin:
Any reason for not accepting this? Should I do something more?

rasa (commented):
@raisercostin Perhaps we need to implement this a different way, as symbolic links can only be created by:

1. Administrators: By default, creating symbolic links requires administrative privileges.

2. Developer Mode: If Developer Mode is enabled, non-administrative users can create symbolic links without requiring elevated privileges. Developer Mode grants the SeCreateSymbolicLinkPrivilege to regular users.

3. Manually Assigning Privileges: Non-administrative users can be granted the SeCreateSymbolicLinkPrivilege manually through Local Group Policy Editor or other privilege assignment tools. This allows them to create symbolic links without enabling Developer Mode or having admin rights.

jonathanMelly:
Hello,

To have the workspace persisted in standard "$persist" location, I would like to suggest to use the "-data" argument.
Here is the patch if it can help.

[dbeaver-workspace.patch](https://github.com/user-attachments/files/18658853/dbeaver-workspace.patch)

What do you think, could it make everybody happy (no symlink and workspace in $persist folder) ?

PR #13684, opened 2024-07-24
kcc@6.1.0: Remove innosetup, add CLI tools

StormPooper:
KCC wasn't installing for me, turns out it is a portable .exe file and not an InnoSetup installer, so I've updated it to just use it directly. I've also added the shortcut in `autoupdate` as it uses the version in the filename. Tested locally and installs fine.


StormPooper:
Fixed lint error from build triggered by @niheaven.

PR #13751, opened 2024-08-03
binance: Add version 1.53.0

3xp0rt:
binance: Add version 1.53.0


aliesbelik:
Please add `checkver` and `autoupdate.hash` sections.
As well using versioned download link seems to be a better option.
[dorado/binance](https://github.com/chawyehsu/dorado/blob/master/bucket/binance.json) could be used for reference.

PR #13813, opened 2024-08-13
ultraiso@9.7.6.3860: Append a URL fragment(#/)

Bo26fhmC5M:
Please append a URL fragment(#/) to change extension of downloaded file from php to exe.
Otherwise, a file with php extension will be downloaded and installation will fail.

### DIFF
--- a/bucket/ultraiso.json
+++ b/bucket/ultraiso.json
@@ -6,7 +6,7 @@
         "identifier": "Proprietary",
         "url": "https://www.ezbsystems.com/pad/ultraiso.xml"
     },
-    "url": "http://www.ezbsystems.com/dl1.php?file=uiso9_pe.exe",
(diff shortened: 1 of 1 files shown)


aliesbelik:
Looks like manifest for `ultraiso` is up to date and doesn't require additional changes.
Please double check your setup.

Bo26fhmC5M:
> Looks like manifest for `ultraiso` is up to date and doesn't require additional changes. Please double check your setup.

When I download the URL from the browser, it is downloaded with the ".exe" extension, but scoop downloaded it with the ".php" extension. Is this a problem only occurring in my environment?

rasa (approved):
This change, while possibly unneeded, can't hurt, and a few other innosetup's have this hack.

PR #13853, opened 2024-08-21
googlechrome: Remove persist User Data

b4imetu:
Closes #XXXX

Relates to https://github.com/ScoopInstaller/Extras/pull/7573#issuecomment-1000041441

In my humble opinion, it is unnecessary to save User Data in persist.
> Same for Google Chrome - the non portable manifest crashes the browser (linked issue).
- There is no such problem for Google Chrome.
- It will cause two Chrome icons with different profiles to appear in the taskbar.
- The default browser opening method will also open a new profile.
> troublesome for ordinary users.
097b1a046725dfab7ad2c330684e221


b4imetu:
for [versions/googlechrome-beta](https://github.com/ScoopInstaller/Versions/blob/master/bucket/googlechrome-beta.json)
[versions/googlechrome-canary](https://github.com/ScoopInstaller/Versions/blob/master/bucket/googlechrome-canary.json)
[versions/googlechrome-dev](https://github.com/ScoopInstaller/Versions/blob/master/bucket/googlechrome-dev.json)
they do not persist User Data and they work well.

b4imetu:
please review this ty @aliesbelik @sitiom @rashil2000

chawyehsu:
There wasn't any message left for this breaking change. The missing of my chrome profile confused me after the update.

At least the old persisted profile data should be copied to AppData and there should be a message notifying the change.

chawyehsu:
Also, this made the persisted profile data being isolated and requires a manual deletion.

rasa:
@chawyehsu Sorry 'bout that. I didn't realize this would break existing installs, or I would have never accepted it as is. I'll be more careful next time.

In my feeble defense, the PR was several months ago, and no one voiced a concern. Perhaps in these kinds of PRs, we should require at least 2, or even 3, approvals before merging.

rasa:
@przemoc Thank you for the detailed writeup, and my sincere apologies for merging this. I didn't think about how this might effect existing users, I was thinking more about it being consistent with the other chrome packages.

In hindsight, it was wrong to merge it in without at least some discussion as to the consequences.

I think you are correct that most Scoop users want their user data dir under Scoop. If they want it under `%AppData%`, they would most likely use an `-np` package.

So, should we undo this change, and if so, change the other chrome packages too, to match? It seems being consistent is important, both for users and maintainers.

@ScoopInstaller/maintainers thoughts?

uiolee:
> As I looked into Google Chrome shortcut in `%APPDATA%\Microsoft\Windows\Start Menu\Programs\Scoop Apps\` vs PR I noticed that I still have `--user-data-dir` even though PR was meant to remove it, IIUC. Maybe shortcuts are not recreated or something like that, have not investigated it.

I also encounter such problems. the `--user-data-dir` in shortcuts still refer to '$dir\User Data\'

uiolee:
Such changes make me feel sad. Personally, I think that the `persist` of Scoop is a good design. I don't know why someone doesn't want it. 

Chrome itself supports specified `--user-data-dir`. I don't think there is a reason not to use this useful feat.

EFLKumo:
I don't think this PR is responsible for scoop users and it violates the design principles of scoop. I uploaded the manifest that still has the data persistent storage [here](https://github.com/EFLKumo/jam).

PR #13949, opened 2024-09-04
7tt: Add startup entry

brian6932:
This mirrors the installer's startup handling, `post_install` is preferred as `$dir` won't set version numbers.


rasa (approved):
What a terrible name for an .exe. Spaces *and* a `+` symbol.

PR #14075, opened 2024-09-24
v2rayn: Change behaviors of preinstall

ProjektMing:

For previous pre_uinstall cause dependence on app's specific version, I consider using SymbolicLink instead of HardLink.
I've checked on my PC and got no problems so far.


rasa:
@ProjektMing @stevenlele has a good point. SymbolicLinks may be problematic, as I think Windows still requires administrator rights, or "Developer mode" to enable them. Not all users will have this setup.

ProjektMing:
Understood. I just saw hardlink make it hard to upgrade independently, but ignored other factors.

PR #14080, opened 2024-09-24
add teams-for-linux

ArjixWasTaken:
Contrary to the app's name, it provides a windows build as well :)


ArjixWasTaken:
The app doesn't appear to have cli usage, other than handling the url protocol `msteams`.
Should I remove the `bin` property?

ArjixWasTaken:
Converted to draft because it appears I can make the app persistently store data in the persist folder.

https://github.com/IsmaelMartinez/teams-for-linux/blob/develop/app/index.js#L85-L89

PR #14121, opened 2024-10-02
sucrose: Add version 24.10.1.0

### LINKED ISSUE #13793: [Request]: Add Sucrose Wallpaper Engine
### Prerequisites

### Criteria

- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

Sucrose Wallpaper Engine

### Description

A versatile wallpaper engine that brings life to your desktop with a wide range of interactive themes.

### Homepage

https://github.com/Taiizor/Sucrose/

### Download Link(s)

https://github.com/Taiizor/Sucrose/releases/download/v$version/Sucrose_Bundle_.NET_Framework_4.8_x64_$version.exe

305 stars

Taiizor:
Closes #13793

Relates to #13793


StudentWeis:
You'll need to add a newline at the end of the file.
> Error: [-] files end with a newline 65ms (64ms|1ms)

Taiizor:
> You'll need to add a newline at the end of the file.
> 
> > Error: [-] files end with a newline 65ms (64ms|1ms)

I think it worked.

Jogai:
A bit late to the party, but out of curiosity, why does this use the older .NET packages, instead of the newer net9.0 based ones? The latter is more performant afaik.

Taiizor:
> A bit late to the party, but out of curiosity, why does this use the older .NET packages, instead of the newer net9.0 based ones? The latter is more performant afaik.

Yes, you’re right, but since most Windows devices already come with .NET Framework 4.8 preinstalled, Sucrose doesn’t require any extra installation to work. However, this isn’t the case with .NET 9 — for Sucrose to run, it needs to be installed separately on the device.

Jogai:
I think all the builds are selfcontained, so no matter which framework, it will run if the architecture is correct.

Taiizor:
> I think all the builds are selfcontained, so no matter which framework, it will run if the architecture is correct.

The default will very soon switch to .NET 9. However, for backward compatibility and to ensure the Bundle file works on nearly every device, it will continue to be named and compiled as .NET Framework 4.8.

PR #14230, opened 2024-10-16
syncthingtray: fix config persist

qianlongzt:
fix config file syncthingtray.ini persist

### DIFF
--- a/bucket/syncthingtray.json
+++ b/bucket/syncthingtray.json
@@ -18,9 +18,18 @@
     },
     "pre_install": [
         "Move-Item \"$dir\\syncthingtray-*-mingw32.exe\" \"$dir\\syncthingtray.exe\"",
-        "Move-Item \"$dir\\syncthingtray-*-mingw32-cli.exe\" \"$dir\\syncthingtray-cli.exe\"",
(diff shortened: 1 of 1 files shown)


Higramx:
I think what's preferable is making the symlink:
```
"if (!(Test-Path \"$persist_dir\\$config\")) { New-Item \"$persist_dir\\$config\" -ItemType File | Out-Null }",
"New-Item -Type SymbolicLink -Value $persist_dir\\$config -Path $dir\\$config | Out-Null"
```
and removing the "move-item" parts.
I also assume the "ensure" things aren't necessary (in that case).

qianlongzt:
Yes, I tested the symlink, and it works correctly. It does not delete the symlink; it deletes the target file and recreates it in the target location. `ensure` is still necessary, or an error might occur.

But symlink need administrator privilege

image

Higramx:
Does persist not require admin privilege? I have not tested without.

qianlongzt:
hard link and junction: normal user
symlink: admin or enable developer mode

https://github.com/ScoopInstaller/Scoop/blob/859d1db51bcc840903d5280567846ae2f7207ca2/lib/diagnostic.ps1#L59

Higramx:
ok thanks

PR #14243, opened 2024-10-18
llrt: Add version 0.2.2-beta

trdthg:
close ScoopInstaller/Versions#2822 

Added an initial manifest to install LLRT [awslabs/llrt](https://github.com/awslabs/llrt).

### DIFF
--- a/bucket/llrt.json
+++ b/bucket/llrt.json
@@ -1,0 +1,23 @@
+{
+    "version": "0.2.2-beta",
+    "description": "LLRT (Low Latency Runtime) is an experimental, lightweight JavaScript runtime designed to address the growing demand for fast and efficient Serverless applications.",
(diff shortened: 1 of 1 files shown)


z-Fng:
Extras bucket does not accept unstable packages. Consider renaming it to `llrt-beta` and adding it to [Versions bucket](https://github.com/ScoopInstaller/Versions) instead.

> ### For a package to be acceptable in this bucket, it should be:
> 1. **Reasonably well-known and widely used. e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks** *
> 2. **English interface (or at least English documentation)** *
> 3. **Latest stable version** *
> 4. Full version (i.e. not a trial version)
> 5. Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

Issue #13248, opened 2024-05-07
[Bug]: Godot Installation running in the background


sylbeth:
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

godot

### Expected/Current Behaviour

When executing godot, it is expected to retrieve the standard output from the terminal. I am using godot-rust, and I need it for debugging purposes of my code. The problem is, every time I run the command, nothing gets printed, and I am able to execute other commands. I talked about it in the godot-rust server and it seems likely that the scoop intallation might be running godot on the background, but it should be in the foreground, since I believe it is more useful to be able to see the output than, just not.

### Steps to Reproduce

```console
PS> scoop install godot
PS> godot
```

### Possible Solution

If it is running in the background, running it in the foreground might solve it.

### Scoop and Buckets Version

```console
PS> scoop --version
Current Scoop version:
d285bb08 (HEAD -> master, tag: v0.4.1, origin/master, origin/HEAD) chore(release): Bump to version 0.4.1 (#5924)

'main' bucket:
4288604a9 (HEAD -> master, origin/master, origin/HEAD) vector: Update to version 0.38.0

'extras' bucket:
b0ba16044 (HEAD -> master, origin/master, origin/HEAD) ucheck: Update to version 5.4.0
```

### Scoop Config

```console
last_update         scoop_repo                              scoop_branch
-----------         ----------                              ------------
07/05/2024 21:48:59 https://github.com/ScoopInstaller/Scoop master
```

### PowerShell Version

```console
PS> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.19041.4355
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.19041.4355
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

sylbeth:
As an addendum, I believe this is the intended behaviour from godot since in their [documentation](https://docs.godotengine.org/en/stable/tutorials/editor/command_line_tutorial.html) they expect those arguments to work, but, for example, ```-h```, which doesn't even output anything.

sylbeth:
[...]

PR #13484, opened 2024-06-20
Godot: Self-contained mode

Bobrokus:
I added a line to the preinstall script to create a `_sc_` file. This makes the installation portable.
I also kept the console exe available, as it can be useful for debugging.
Lastly, I made the `editor_data` persistent.

Relates to #13483

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #14268, opened 2024-10-21
godot: default to console

gsuuon:
This defaults godot to the console binary. I found it strange that `godot --help` did nothing - since scoop is a terminal installer I think being console-first makes sense. If this is too breaking (not sure if anyone's depending on that behavior) we could also just add `godot-cli` as the console alias.

Closes: #13248 
Related: #13484 (draft) -- I don't address portability in this PR

### DIFF
--- a/bucket/godot.json
+++ b/bucket/godot.json
@@ -14,10 +14,10 @@
         }
     },
     "pre_install": [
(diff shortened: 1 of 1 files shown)


rasa (approved):
Good call: all bin commands should be CLI, GUIs live in shortcuts.

Issue #13476, opened 2024-06-19
[Request]: keyguard (bitwarden alternative gui)


nitrocode:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

keyguard

### Description

 Keyguard is an alternative client for the Bitwarden® platform, created to provide the best user experience possible. 

### Homepage

https://github.com/AChep/keyguard-app

### Download Link(s)

https://github.com/AChep/keyguard-app/releases/tag/r20240612

### Some Indication of Popularity/Repute

800+ stars

magisk317:
any  update？

magisk317:
I'm glad to see that this problem has been solved, but I still can't see the return result in the command line. What might be the reason?

**Search directly for "Keyguard"：**
```
scoop search keyguard
WARN  No matches found.
```

**Neither does fuzzy search：**
```
scoop search key
Results from local buckets...

Name                        Version          Source   Binaries
----                        -------          ------   --------
bind                        9.16.50          main     dnssec-dsfromkey.exe | dnssec-importkey.exe | dnssec-keyfromlabe…
git-with-openssh            2.47.0.windows.2 main     ssh-keygen.exe | ssh-keyscan.exe
gnupg1                      1.4.23           main     gpgkeys_curl.exe | gpgkeys_finger.exe | gpgkeys_hkp.exe | gpgkey…
gow                         0.8.0            main     lesskey.exe
less                        668              main     lesskey.exe
mls-software-openssh        9.8p1-1          main     ssh-keygen.exe | ssh-keyscan.exe
openssh                     9.8.1.0p1        main     ssh-keygen.exe | ssh-keyscan.exe
rage                        0.11.0           main     rage-keygen.exe
rktools2k3                  1.0              main     remapkey.exe
unxutils                    2007.03.01       main     lesskey.exe
yubikey-personalization     1.19.0           main
age                         1.2.0            extras   age-keygen.exe
autohotkey                  2.0.18           extras
cfssl                       1.6.5            extras   cfssl-newkey.exe
[...]

PR #14348, opened 2024-11-02
keyguard: Add 1.6.4-20241102

AChep:
Closes https://github.com/ScoopInstaller/Extras/issues/13476

- [+] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/keyguard.json
+++ b/bucket/keyguard.json
@@ -1,0 +1,34 @@
+{
+    "version": "1.6.4-20241102",
+    "description": "Keyguard is an alternative client for the Bitwarden® platform, created to provide the best user experience possible",
+    "homepage": "https://github.com/AChep/keyguard-app",
+    "license": {
+        "identifier": "Proprietary",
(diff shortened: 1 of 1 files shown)


AChep:
Please let me know how the update process goes. Should I create a PR each time a new version of the app is released?

aliesbelik:
> Please let me know how the update process goes.

Normally manifest will be [regularly checked](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#testing-and-running-autoupdate) (using `checkver` condition) and automatically updated (using `autoupdate` template) in case of need by running github action, unless any changes in release or binaries naming schema appear so that manual fix will be required.

> Should I create a PR each time a new version of the app is released?

No.
In case of significant changes in the upstream project (release/binaries naming changes, binaries format change, moving to another hosting, additional/deprecated binaries) it would be nice to check the manifest as well if you are interested in its relevance.

PR #14381, opened 2024-11-08
openai-translator@0.4.33: Fix shortcut

TiancongLx:
The shortcut is "OpenAI Translator.exe" now, other than "app.exe".


aliesbelik:
Reverts #13743.

PR #14478, opened 2024-11-29
neohtop: Add version 1.1.2

Slach:
NeoHTOP - cross platform htop on steroids

package request fix https://github.com/ScoopInstaller/Extras/issues/14812


aliesbelik:
Relates to Abdenasser/neohtop#71.

bucket/neohtop.json
```diff
@@ -0,0 +1,22 @@
...
+    "hash": "75f6ea91670bb4888bcb82572967c394570c5cfa7000fc30ea08a16981dfa77a",
+    "bin": "NeoHtop.exe",
+    "shortcuts": [
+        [
+            "NeoHtop.exe",
+            "NeoHtop"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/Abdenasser/neohtop",
```
pratikpc, line 16: In this case, it lookslike the checkver can be simplified to 

```json
    "checkver": "github",
```

This is because:-
1. Regex is not needed as it matches the default regex
2. The URL provided to the GitHub parameter matches the project's url

Slach: fixed in https://github.com/ScoopInstaller/Extras/pull/14478/commits/a1d26090b740ca62b226795d38b7f35b7428769e

pratikpc:
Verify would fail as neohtop upstream is now on v1.1.2

You can take advantage of Scoop Check Version app to update the manifest.

Also remember to update the title.

I do not have merge access but other than that, LGTM

pratikpc (approved):

Issue #14427, opened 2024-11-19
vscode@1.96.0-insider: download via aria2 failed


s0racat:

Miqueas:
The latest stable is still 1.95.3, idk why the manifest was changed, the version clearly says insider, that's a whole another channel 🤔

mattcargile:
Yes, I would expect _insiders_ version to be installed from the `versions` bucket.

aliesbelik:
Fixed by 5e481d29e.

TheLastZombie:
It broke again: 6d1418c522d05ab36804934bf49b77db3eaab7d1

s0racat:
shoud i reopen this issue?

TheLastZombie:
I guess so. Not sure if maintainers get notifications regarding closed issues.

s0racat:
I reopened this issue

Skylark13:
Looks like it got re-updated (downgraded?) to 1.95.3 (1.96.0-insider at 7:34 AM EST, 1.95.3 at 9:28 AM EST)
https://github.com/ScoopInstaller/Extras/commit/a1885dd67e321c5e4709afd47410bafda25b4068
Is it just ping-ponging between versions at this point?

as000fm:
When back to not working version 1.96.0-insider _just an hour ago..._
Why this version is in the Extra bucket instead of being in the Version bucket **only**?

OPVL:
+1

PR #14484, opened 2024-11-30
vscode: revert 1.96.0-insider -> 1.95.3

OPVL:
Reverted manifest to previous version `1.95.3` from erroneous insiders version.

Validated file checksums are still correct. A more permanent fix is required and an investigation as to why this keeps happening.

Relates to #14427


aliesbelik:
Fixed by bot in 7e0a95b.

Guess could CI issue in upstream.

PR #14528, opened 2024-12-07
Added Cromite browser to the bucket

Tomurisk:
I added Cromite browser to the bucket. Cromite is a fork of Chromium, that has cross-platform binaries, including Android. It is a great addition along Thorium, ungoogled-chromium, Woolyss, and so.


Tomurisk:
I would like to inform that the package was thoroughly manually checked, and installation proceeds successfully

Tomurisk:
@niheaven @aliesbelik Could I get any feedback? Thank you

PR #14540, opened 2024-12-09
wsl2-distro-manager@1.8.15: fix link to windows patched version

### LINKED ISSUE #14539: wsl2-distro-manager@1.8.15: hash check failed
There is currently an issue with the package `wsl2-distro-manager` because they did a patch release that is not accounted for in the manifest. 

See https://github.com/bostrot/wsl2-distro-manager/releases/tag/v1.8.15

The package for windows is now named [wsl2-distro-manager-v1.8.15+1.zip](https://github.com/bostrot/wsl2-distro-manager/releases/download/v1.8.15/wsl2-distro-manager-v1.8.15+1.zip) and published after the offical release, but they did not updated the version number as it only concerns windows (not mac)

It is possible to only patch the manifest, as it seems this will not happen often. 

````patch
diff --git a/bucket/wsl2-distro-manager.json b/bucket/wsl2-distro-manager.json
index af34d4ca6..d7d3403e5 100644
--- a/bucket/wsl2-distro-manager.json
+++ b/bucket/wsl2-distro-manager.json
@@ -5,8 +5,8 @@
     "architecture": {
         "64bit": {
-            "url": "https://github.com/bostrot/wsl2-distro-manager/releases/download/v1.8.15/wsl2-distro-manager-v1.8.15.zip",
-            "hash": "f7e6e96a9a06f9f4f7300f13cb0594acb3b2f073e51925ce0cbe5d12624c459a"
+            "url": "https://github.com/bostrot/wsl2-distro-manager/releases/download/v1.8.15/wsl2-distro-manager-v1.8.15+1.zip",
+            "hash": "1d6cd20e8f26dc69eee60f8a894c3584b1740178ce39e1a029f1b0b1d8257f40"
         }
     },
     "shortcuts": [
````

Or the manifest needs to be adapted to this special case in case it is happening again. 

Not sure what you prefer

cderv:
Closes #14539 

It just patches the manifest with the right url. And next auto update should be ok for 1.8.16

However, if the project do another time a patch release like this, then the problem from #14539 will happen again. 

Though it seems enough to do that for now to no complexify the manifest. 

### DIFF
--- a/bucket/wsl2-distro-manager.json
+++ b/bucket/wsl2-distro-manager.json
@@ -5,8 +5,8 @@
     "license": "GPL-3.0-or-later",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/bostrot/wsl2-distro-manager/releases/download/v1.8.15/wsl2-distro-manager-v1.8.15.zip",
-            "hash": "f7e6e96a9a06f9f4f7300f13cb0594acb3b2f073e51925ce0cbe5d12624c459a"
+            "url": "https://github.com/bostrot/wsl2-distro-manager/releases/download/v1.8.15/wsl2-distro-manager-v1.8.15+1.zip",
+            "hash": "1d6cd20e8f26dc69eee60f8a894c3584b1740178ce39e1a029f1b0b1d8257f40"
         }
     },
     "shortcuts": [


cderv:
> Though it seems enough to do that for now to no complexify the manifest.

In fact it is not... It seems checks expect autoupdate to be working and so to be fixed for this manifest. 

Should there be an exception here or do you want me to come up with a valid autoupdate here using regex probably ?

Issue #5318, opened 2020-12-16
Looks cannot update from sourceforge.net via aria2


Mosney:
I noticed recent update failed resources are all stored at Sourceforge and I cannot download it properly. There is anyone else as same as me?

> PS C:\Users\fst> scoop info aria2
Name: aria2
Description: Lightweight multi-protocol & multi-source command-line download utility
Version: 1.35.0-1
Website: https://aria2.github.io/
License: GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later.html)
Manifest:
  D:\Scoop\buckets\main\bucket\aria2.json
Installed:
  D:\Scoop\apps\aria2\1.35.0-1
Binaries:
  aria2c.exe

```
PS C:\Users\fst> scoop update qbittorrent-portable
qbittorrent-portable: 4.3.0.1 -> 4.3.1
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
Updating one outdated app:
Updating 'qbittorrent-portable' (4.3.0.1 -> 4.3.1)
Downloading new version
Starting download with aria2 ...
Download: 12/17 02:28:09 [ERROR] CUID#8 - Download aborted. URI=https://downloads.sourceforge.net/project/qbittorrent/qbDownload: Exception: [AbstractCommand.cc:351] errorCode=1 URI=https://downloads.sourceforge.net/project/qbittorrent/qbitDownload:   -> [SocketCore.cc:1021] errorCode=1 SSL/TLS handshake failure: Error: 鐢变簬鍚婇攢鏈嶅姟鍣ㄥ凡鑴辨満锛屽悐閿Download: Download Results:
Download: gid   |stat|avg speed  |path/URI
Download: ======+====+===========+=======================================================
Download: e2bd00|ERR |       0B/s|D:/Scoop/cache/qbittorrent-portable#4.3.1#https_downloads.sourceforge.net_project_qbittorrent_qbittorrent-win32_qbittorrent-4.3.1_qbittorrent_4.3.1_x64_setup.exe_dl.7z
Download: Status Legend:
Download: (ERR):error occurred.
Download: aria2 will resume download if the transfer is restarted.
Download: If there are any errors, then see the log file. See '-l' option in help/man page for details.

ERROR Download failed! (Error 1) An unknown error occurred
[...]

Issue #4418, opened 2020-07-15
winscp@5.17.6: download via aria2 failed


todoubaba:
```
ERROR Download failed! (Error 1) An unknown error occurred
ERROR https://downloads.sourceforge.net/project/winscp/WinSCP/5.17.6/WinSCP-5.17.6-Portable.zip
    dir=C:\Users\xxx\scoop\cache
    out=winscp#5.17.6#https_downloads.sourceforge.net_project_winscp_WinSCP_5.17.6_WinSCP-5.17.6-Portable.zip

ERROR & 'C:\Users\xxx\scoop\apps\aria2\current\aria2c.exe' --input-file='C:\Users\xxx\scoop\cache\winscp.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/7.0 (Windows NT 10.0; Win64; x64; Core)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn --enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=8064 --continue
```

Issue #827, opened 2018-03-04
winscp link broken


marknelissen:
When installing winscp 5.13, I get an error 500 on the used url.
`URL https://sourceforge.net/projects/winscp/files/WinSCP/5.13/WinSCP-5.13-Portable.zip is not valid`

r15ch13:
This seems to be a SourceForge problem.
Even https://sourceforge.net/projects/winscp/files/ doesn't work.

marknelissen:
Is the version on sourceforge a repackaged version? Why not use the one directly from winscp.net?

I can download it from there, but don't know how to manually inject it into scoop, as a temporary workaround.

marknelissen:
Situation is back to normal.

PR #14546, opened 2024-12-10
winscp: fix download url & autoupdate url

vsimko:
I replaced the Sourceforge URL with the one provided by WinSCP directly.

Relates to #5318, #4418, #827

### DIFF
--- a/bucket/winscp.json
+++ b/bucket/winscp.json
@@ -3,7 +3,7 @@
     "description": "Copy files between a local computer and remote servers using FTP, FTPS, SCP, SFTP, WebDAV or S3 file transfer protocols.",
     "homepage": "https://winscp.net",
     "license": "GPL-3.0-or-later",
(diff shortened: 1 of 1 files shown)


CryoTheRenegade:
Looks like the page you are referencing is a download placeholder for a cdn, the actual link from their cdn is "https://cdn.winscp.net/files/WinSCP-6.3.6-Portable.zip?secure=9FweEKieEPd1v28jDRhStA==,1734027149" which may have rotating keys to prevent automatic downloads. Sourceforge is still a viable source for downloading WinSCP and does not seem to be going anywhere.

vsimko:
you're unfortunately right.
the hash code should be the same even when downloading through a cdn. I'm
still trying to find out why it works for me locally but not in the CI
Pipeline.

vsimko:
What about the following link, would it make sense?
https://files03.tchspt.com/down/WinSCP-6.3.6-Portable.zip

PR #14564, opened 2024-12-12
sd-card-formatter: Extract instead of execute

brian6932:


brian6932:
Seems like they have gone back to the old dl path naming scheme on their server.

brian6932:
@rasa I guess I should've kept #14568 linked, it can be closed.

PR #14652, opened 2024-12-29
catalyst: Rename to catalyst-browser

jdev082:
Renames the `catalyst` package to `catalyst-browser` to be more consistent with upstream changes and other software repositories.


jdev082:
Recreated PR because the old one was broken.

rasa (commented):
@jdev082 Renames concern me, as users who have this installed, will not be informed of the rename. The package will simply disappear.

Is there some way we can alert the user to the rename? For example, we create a new catalyst-browser, and change catalyst.json to require it as a dependency? I haven't tested that, so I'm not sure that's gonna work for them.

Or at least add a note in the old version's description and notes, that says it's been renamed, and to uninstall the old, and reinstall the new?

PR #15045, opened 2025-03-10
windbg: Add version 1.2603.20001.0

### LINKED ISSUE #4289: [Request] Add WinDbg
Hello, please add WinDbg.

WinDbg is the *official Windows debugger, made by Microsoft**.
It is used to debug native Windows programs in both user mode and kernel mode.
https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/
It is a free program but it doesn't come with Windows.

As you can see [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/debugger-download-tools), to install WinDbg you need to download the [Windows SDK installer](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/) and choose to install "Debugging Tools for Windows" only.

The Windows SDK installer is a WiX installer. It contains manifests which contain the version, and links (and hashes) to other installers, among them is the link to the actual MSI installer belonging to WinDbg (`X64 Debuggers And Tools-x64_en-us.msi`).

Thanks!

ohhithere:
Closes #4289

Couple notes:
* Minimum requirements based on winget's [windbg manifest](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/WinDbg/1.2502.25002.0/Microsoft.WinDbg.installer.yaml)
* Manifest based on Scoop's [winget manifest](https://github.com/ScoopInstaller/Main/blob/master/bucket/winget.json)

### DIFF
--- a/bucket/windbg.json
+++ b/bucket/windbg.json
@@ -1,0 +1,46 @@
+{
+    "version": "1.2603.20001.0",
+    "description": "WinDbg is a debugger that can be used to analyze crash dumps, debug live user-mode and kernel-mode code, and examine CPU registers and memory.",
+    "homepage": "https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://learn.microsoft.com/en-us/legal/windows-sdk/windbg-preview---software-license-terms"
+    },
+    "notes": "At least Windows 10 build 17763 is required to use WinDbg.",
+    "url": "https://windbg.download.prss.microsoft.com/dbazure/prod/1-2603-20001-0/windbg.msixbundle#/dl.7z",
+    "hash": "d98e7027fc84962dcd93c270195e84b27bacfcf942935043fa457c87aea83532",
+    "architecture": {
+        "64bit": {
+            "pre_install": "Get-ChildItem \"$dir\" -Exclude '*x64.msix' | Remove-Item -Force -Recurse"
+        },
+        "32bit": {
+            "pre_install": "Get-ChildItem \"$dir\" -Exclude '*x86.msix' | Remove-Item -Force -Recurse"
+        },
+        "arm64": {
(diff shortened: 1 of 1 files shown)


haussmann:
Hello @ohhithere ! Thank you very much for developing this manifest. I committed some recommendations:

- Updated checkver to allow for more flexibility and simplicity in detecting the newer versions
- Added a link to the license url
- Added a note regarding stored settings being out of scoops control (similar to other manifests)
- Remove the minimum windows 10 build requirements note, since Win10 seems to be out of support anyway
- Simplifies and hardens the installation script

Feel free to revert / overwrite, but in case you do, please submit an explanation.

ohhithere:
Hi!
Thanks for taking a look at this. Good change to also link to the license! I also didn't know of `$dashVersion` - neat. At this stage and for what I want, I have reverted the rest and here's why:

* I do think the note should contain compatibility information for Windows 10. It remains supported by Microsoft for many use-cases and even if it weren't, we need not follow Microsoft's policy.
* I don't think the note on scoop not persisting the data itself is needed - that is the default and I would argue expected. Applications, especially in the `extras` bucket, have all kinds of state and side effects. Scoop "persisting" data is mostly to help applications which just want to use their install directory. An absolute minority of manifests notify the user of this.
* I am not sure you tested the change - but it likely breaks the install at least on arm. The separation of `pre_install` and `install` has semantic purpose and your change would just do the install in `pre_install` but more importantly as per the [documentation](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests), the `$architecture` variable only has `64bit` and `32bit` as possible values, neither of which map neatly (not to mention arm) so that creates the need for string manipulation in the routine which I think ought to be avoided. As noted in the original message of the PR, this pattern I am using is also used by the `winget` manifest in `Main`.
* The regex I wrote extracts the version from a changelog which is inherently brittle so strong matching is required. I noticed and exploited the pattern. The modified regex is much riskier and doesn't enforce the version format they stick to.
* In the autoupdate section, you removed the `#/dl.7z` which, without having tested it, very likely breaks auto-extraction and thus the entire install flow.

PR #10652, opened 2023-03-02
bitwarden: Changed `pre_install`, added `architecture`, and added `autoupdate` hash extraction

FlawlessCasual17:
This was done, because the Bitwarden electron app was NOT made to be portable. This change was made so the portable config folder's contents will be copied over to the `Appdata\Roaming` directory.

See #10651 if you want to know more of what I am talking about.

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #15053, opened 2025-03-11
bitwarden@2025.2.1: Make portable

o-l-a-v:
BREAKING CHANGE:

* Will not use `$env:APPDATA\Bitwarden` anymore.

---

Builds on closed PRs by @FlawlessCasual17:

* <https://github.com/ScoopInstaller/Extras/pull/10652>
* <https://github.com/ScoopInstaller/Extras/pull/13077>

Changes from PRs above:

* Download `nsis.7z` given the OS architecture, cuts download size to 1/3
* Add hash extraction on autoupdate.
* Add more reliable checkver logic

What I added:

* Use `--user-data-dir=\"$dir\\data\"` to have Bitwarden installed by Scoop keep its data inside Scoop.
  * Like other manifests do too: <https://github.com/search?q=repo%3AScoopInstaller%2FExtras%20%22--user-data-dir%22&type=code>
* Persist `data`.

### DIFF
--- a/bucket/bitwarden.json
+++ b/bucket/bitwarden.json
@@ -3,29 +3,53 @@
     "description": "Password management solutions for individuals, teams, and business organizations",
     "homepage": "https://bitwarden.com",
     "license": "GPL-3.0-or-later",
-    "url": "https://github.com/bitwarden/clients/releases/download/desktop-v2025.2.1/Bitwarden-Portable-2025.2.1.exe",
-    "hash": "6b705acd9c3ee391d72d993c649f309e652cf111e2b1fa24d03ad1880f0fb0aa",
+    "architecture": {
+        "32bit": {
+            "url": "https://github.com/bitwarden/clients/releases/download/desktop-v2025.2.1/bitwarden-2025.2.1-ia32.nsis.7z",
+            "hash": "2f9959351c62ba020704f03d600593b8e59fbc161f2db3d3170e10b4a46f8589"
+        },
+        "64bit": {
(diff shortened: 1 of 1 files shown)


jasongodev:
I was tagged in the previously closed PR that uses portable version of the Bitwarden. Looking at this new PR, it looks okay to me.

pratikpc:
Is it possible in Scoop to add default arguments to the bin exe generated?

I do not use Bitwarden so am just curious if without it, this would not end up being a breaking change?

pratikpc:
Also if possible can't the persist directory name be retained?

Otherwise, this would end up as a bigger breaking change

o-l-a-v:
The persist dir has no function as of now, so changing it is not a breaking change.

pratikpc:
@o-l-a-v but in portable mode is that not the user data directory?

Sorry I am not a user of Bitwarden Client but I read the code here

https://github.com/bitwarden/clients/blob/fdcb6ff7a2fbefeccbaf5b5bcd4243f8fb84c09a/apps/desktop/src/main.ts#L94

With Electron, the environment variable is set for portable builds. 

https://www.electron.build/nsis.html#portable

Electron has a portable target

https://github.com/bitwarden/clients/blob/fdcb6ff7a2fbefeccbaf5b5bcd4243f8fb84c09a/apps/desktop/electron-builder.json#L86

o-l-a-v:
Damn, you're probably right @pratikpc. I read these previous discussions, but got it wrong.

* <https://github.com/ScoopInstaller/Extras/issues/4843>
* <https://github.com/ScoopInstaller/Extras/pull/8241>

Probably best to keep the persist directory name, I agree.

Issue #15097, opened 2025-03-18
[Bug]: install script fails with commands specific to version 2.0


AchimGaedkeLynker:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

gimp

### Expected/Current Behaviour

Installation fails (I believe decompression was successful) due to potentially outdated file renaming.
It looks like the install script is too specific to version 2.x.

### Steps to Reproduce

```console
C:\Users\achimgaedke> scoop install gimp
Installing 'gimp' (3.0.0) [64bit] from 'extras' bucket
Loading gimp-3.0.0-setup.exe from cache
Checking hash of gimp-3.0.0-setup.exe ... ok.
Extracting gimp-3.0.0-setup.exe ... done.
Running installer script...Rename-Item : Cannot rename because item at 'lib\gimp\2.0\plug-ins\twain\twain,2.exe' does not exist.
At line:4 char:4
+    Rename-Item 'lib\gimp\2.0\plug-ins\twain\twain,2.exe' 'twain.exe'
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Rename-Item], PSInvalidOperationException
    + FullyQualifiedErrorId : InvalidOperation,Microsoft.PowerShell.Commands.RenameItemCommand

Remove-Item : Cannot find path 'C:\Users\achimgaedke\scoop\apps\gimp\3.0.0\lib\gimp\2.0\plug-ins\twain\twain,1.exe'
because it does not exist.
At line:5 char:4
+    Remove-Item 'lib\gimp\2.0\plug-ins\twain\twain,1.exe'
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\achimg...ain\twain,1.exe:String) [Remove-Item], ItemNotFoundEx
   ception
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.RemoveItemCommand

[...]

PR #15101, opened 2025-03-18
gimp: transition 2.x -> 3.x

AchimGaedkeLynker:
Relates to #15097

Changed the version number to the best of my knowledge and now it installs and runs for me (haven't worked much with it).

* I need some help with understanding the python path in `default.env`
* There are binaries with `,1` and `,2` but also `,3`. I don't know what `,3` is. It "works" for me/64 bit architecture... not sure for the other architectures.

- [X] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`

### DIFF
--- a/bucket/gimp.json
+++ b/bucket/gimp.json
@@ -11,45 +11,38 @@
             "Push-Location \"$dir\"",
             "Get-ChildItem -Filter '*.debug' -Recurse | Remove-Item -Recurse",
             "if ($architecture -eq '64bit') {",
-            "   Rename-Item 'lib\\gimp\\2.0\\plug-ins\\twain\\twain,2.exe' 'twain.exe'",
-            "   Remove-Item 'lib\\gimp\\2.0\\plug-ins\\twain\\twain,1.exe'",
             "   Get-ChildItem -Filter '*,1*' -Recurse | Rename-Item -NewName { $_.name -Replace ',\\d','' }",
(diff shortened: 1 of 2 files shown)


TayouVR:
I can't test this on other architectures, but this does work on my amd64 machine.

there was a 3.0 dev verion in Versions, maybe that can help? not sure about that though
https://github.com/ScoopInstaller/Versions/blob/262970ca723e424172bd30574bd6df51526ef72b/bucket/gimp-dev.json

niheaven:
Since GIMP v3 added arm64 support, Scoop could not handle its installer without https://github.com/ScoopInstaller/Scoop/pull/3502. Therefore, it would be better to hold the GIMP version to 2.x until the installer PR is merged.

sgtlaggy:
The `,1`, `,2`, and `,3` suffixes appear to refer to 32-bit, 64-bit, and ARM64 respectively from reading the `install_script.iss`. Not sure if this was the case before, but that would suggest the manifest here installs the 32-bit version regardless of whether the user is on a 32-bit or 64-bit OS.

niheaven:
It is consistent, with only `,1` and `,2` before. However, for 3.x, most libraries have three files: `,1` for 32-bit, `,2` for 64-bit, and `,3` for ARM64. Some libraries do not exist in 32-bit, so for those, `,1` is for 64-bit and `,2` is for ARM64, making the situation more complex.

PR #15124, opened 2025-03-20
nomacs: Re-add checkver and update to 3.21.0

tmewett:
See commits for description.

### DIFF
--- a/bucket/nomacs.json
+++ b/bucket/nomacs.json
@@ -1,15 +1,15 @@
 {
-    "version": "3.16.1709",
+    "version": "3.21.0",
     "description": "Image viewer to view all common image formats including RAW and psd images",
     "homepage": "https://nomacs.org",
     "license": "GPL-3.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/nomacs/nomacs/releases/download/3.16.224/nomacs-portable-win.zip",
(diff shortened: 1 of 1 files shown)


aliesbelik:
Looks like `3.16.1709` is still the latest _stable_ release, referred in official documentation as well ([Installation](https://nomacs.org/docs/getting-started/installation/)).
Beta, RC or any other kinds of preview versions belong to the [Versions](https://github.com/ScoopInstaller/Versions) bucket.

tmewett:
Oh interesting, you're right. I don't exactly understand their versioning; I've opened nomacs/nomacs#1289 to hopefully clarify whether 3.16 is still the latest stable

tmewett:
The answer seems to be:

> We are not maintaining a stable branch.

I.e. it seems that the 3.16 line is not maintained.

Since nomacs is only maintaining one line of releases, then this line should be included in 'extras' bucket and hence nomacs should be updated to 3.21. Thoughts? On the other hand, they don't seem to have fully come to terms with this fact upstream.

arvdk:
I believe this PR can be closed because the issues it addresses have already been resolved.

Currently:
`"checkver": {
        "github": "https://github.com/nomacs/nomacs"
    },`

aliesbelik:
Fixed by #16924.

PR #15212, opened 2025-04-04
NetworkMiner 3.0 Released

erik4711:
https://netresec.com/?b=254caa9

There is also a static URL for the 3.0 download available (if that helps):
https://download.netresec.com/networkminer/NetworkMiner_3-0.zip


aliesbelik:
Updated automatically in abd8991.

PR #15229, opened 2025-04-05
focus-editor: Update shim

rawhuul:


HUMORCE:
I change the PR for creating both `focus` and `focus-editor` shim. In case people who wants a short command, shim conflicts can be manage via `scoop shim -h`

PR #15368, opened 2025-04-29
windhawk@1.6: Fix update

m417z:
I'm not familiar with Scoop, but this is my attempt at fixing https://github.com/ramensoftware/windhawk/issues/490.
@redactedscribe please review, thanks.

### DIFF
--- a/bucket/windhawk.json
+++ b/bucket/windhawk.json
@@ -23,16 +23,8 @@
             "/D=$dir"
         ]
     },
-    "post_install": [
-        "if((test-path \"$dir\\AppData.original\")) {",
-        "    Copy-Item \"$dir\\AppData.original\\*\"  \"$dir\\AppData\" -Exclude \"*.ini\" -Recurse -Force",
-        "    Remove-Item \"$dir\\AppData.original\" -Recurse -Force | Out-Null",
-        "    Remove-Item \"$dir\\windhawk.ini.original\" -Recurse -Force | Out-Null",
-        "}"
-    ],
     "persist": [
-        "AppData",
-        "windhawk.ini"
+        "AppData"
     ],
     "shortcuts": [
         [


m417z:
> persisting `windhawk.ini` is probably desirable

Why? It's read-only, and for the non-portable installation, it resides in the Program Files folder. It seems to me that it can be treated as the other read-only, non-persistant files.

> I would ask commentors on the issue to test this.

Will do, although I'm not sure any of them is proficient enough with Scoop packages. If you could test it, that would be great, as the author of this package you know it best. Thanks.

proJM-Coding:
The issue which is trying to be fixed is that when `windhawk.dll` is called by windhawk it will call with the 1.5.1 version of the engine since scoop will persist the `windhawk.ini` file. To fix the issue manualy you need to edit the `windhawk.ini` file to have the EnginePath=Engine\1.6.1. 

I don't see why persisting windhawk.ini is probably desirable as it is the reason for this issue.

SquirrelKiev:
The contents of windhawk.ini isn't anything a user would need to change anyway, as it all has to do with where app data is located.
```
[Storage]
Portable=1
CompilerPath=Compiler
EnginePath=Engine\1.7.1
UIPath=UI
AppDataPath=AppData
```

SorYoshino:
Based on my testing, the updated manifest installs without any issues.  

It only needs to have its field order adjusted to comply with the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

bucket/windhawk.json
```diff
@@ -23,16 +23,8 @@
             "/D=$dir"
         ]
     },
-    "post_install": [
-        "if((test-path \"$dir\\AppData.original\")) {",
-        "    Copy-Item \"$dir\\AppData.original\\*\"  \"$dir\\AppData\" -Exclude \"*.ini\" -Recurse -Force",
-        "    Remove-Item \"$dir\\AppData.original\" -Recurse -Force | Out-Null",
-        "    Remove-Item \"$dir\\windhawk.ini.original\" -Recurse -Force | Out-Null",
-        "}"
-    ],
```
z-Fng, line 32: Simply removing the logic to persist `windhawk.ini` would work. Is it still necessary to remove the logic for persisting AppData?

Note: `AppData.original` is actually the `AppData` folder inside the package. Without this logic, the `AppData\Engine` directory wouldn’t get updated because it's persisted.

SorYoshino: Following my testing, the scripts within `post_install` are redundant.

```powershell
┏[ D:\Software\Scoop\Local\cache]
└─> Start-Process -FilePath .\windhawk#1.7.3#0fff666.exe -ArgumentList @('/S', '/portable', '/D=.\windhawk') -Wait

┏[ D:\Software\Scoop\Local\cache]
└─> Get-ChildItem -Path '.\Windhawk\AppData' -Recurse

        Directory: D:\Software\Scoop\Local\cache\Windhawk\AppData


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d----        2025/12/11      2:10                  Engine
-a---        2025/12/11      2:10            360   settings.ini

        Directory: D:\Software\Scoop\Local\cache\Windhawk\AppData\Engine


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d----        2025/12/11      2:10                  ModsWritable
d----        2025/12/11      2:10                  Symbols
-a---        2025/12/11      2:10            320   settings.ini
-a---        2025/12/11      2:10             46 󰈙  warning.txt

        Directory: D:\Software\Scoop\Local\cache\Windhawk\AppData\Engine\ModsWritable


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a---        2025/12/11      2:10             46 󰈙  warning.txt

        Directory: D:\Software\Scoop\Local\cache\Windhawk\AppData\Engine\Symbols


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a---        2025/12/11      2:10             46 󰈙  warning.txt

```

SorYoshino: > `AppData\Engine` directory

This folder is automatically generated when the programme is running. As shown in the results above, the `AppData` folder contains only `settings.ini` and some `warning.txt` files.

m417z: I thought that the "persist" section takes care of this. I didn't find documentation about the `.original` suffix. I also didn't understand why it copies everything but .ini files.

z-Fng: > I thought that the "persist" section takes care of this. I didn't find documentation about the .original suffix.

After the installer completes, Scoop will create a symbolic link for the persist folder to the specified directory. If a folder with the same name already exists after installation, it will first rename that folder to `folder.original` before creating the symbolic link.

> I also didn't understand why it copies everything but .ini files.

This post_install script assumes that new files exist inside the `AppData.original` folder after installation and need to be copied out. It also assumes that the `.ini` files are the default configuration file and are no longer needed, since custom configuration files already exist.

Removing the `post_install` script means that after an update, any files or any changes to the directory structure inside `AppData` that are introduced in a new version may not be used/applied.

z-Fng: > Removing the `post_install` script means that after an update, any files or any changes to the directory structure inside `AppData` that are introduced in a new version may not be used/applied.

If these changes look good to you, we can merge this PR.

m417z: > After the installer completes, Scoop will create a symbolic link for the persist folder to the specified directory. If a folder with the same name already exists after installation, it will first rename that folder to `folder.original` before creating the symbolic link.

In this case, it makes sense to copy all files, including .ini files. That's what Windhawk expects.

Also, the `warning.txt` files are there because the installer applies permissive access attributes to these folders:
* `AppData\Engine`: GenericRead + GenericExecute
* `AppData\Engine\ModsWritable`: FullAccess
* `AppData\Engine\Symbols`: FullAccess

It'd be simpler to have Scoop create the symbolic link first, then run the installer. But if that's not possible, this can also be done post-install as following:
```cmd
icacls AppData\Engine /grant everyone:RX
icacls AppData\Engine /grant *S-1-15-2-1:RX
icacls AppData\Engine /grant *S-1-15-2-2:RX

icacls AppData\Engine\ModsWritable /grant everyone:F
icacls AppData\Engine\ModsWritable /grant *S-1-15-2-1:F
icacls AppData\Engine\ModsWritable /grant *S-1-15-2-2:F

icacls AppData\Engine\Symbols /grant everyone:F
icacls AppData\Engine\Symbols /grant *S-1-15-2-1:F
icacls AppData\Engine\Symbols /grant *S-1-15-2-2:F
```
Without this, mods that apply to sandboxed processes won't work.

> It also assumes that the `.ini` files are the default configuration file and are no longer needed, since custom configuration files already exist.

Is it during an installation or an update?

On first run, Windhawk expects to find the files that the installer placed there.

During an update, the installer might make changes to the files, and Windhawk expects to see these changes.

> Removing the `post_install` script means that after an update, any files or any changes to the directory structure inside AppData that are introduced in a new version may not be used/applied.

But isn't the symbolic link already there on an update?

z-Fng: > In this case, it makes sense to copy all files, including .ini files. That's what Windhawk expects.

Shouldn't files like `AppData\settings.ini` be protected from being overwritten by default configuration file?

> Is it during an installation or an update?

Based on the post_install logic in the manifest, this applies to both cases, as long as it's not a fresh installation.

> But isn't the symbolic link already there on an update?

Not exactly. Each version is actually installed in a separate folder, such as `~/Scoop/apps/windhawk/1.7.3`. The symbolic link will be created only after installation is complete. Symbolic links are created after the installer runs.

m417z: > Shouldn't files like `AppData\settings.ini` be protected from being overwritten by default configuration file?

The installer doesn't override ini files with default configuration files.
* When installing, the `AppData` folder, including the default configuration files, is created.
* When updating, the installer does the necessary changes to the existing ini files, if needed.

I'm not sure how the update flow works in Scoop and how it differs from installation.

> Not exactly. Each version is actually installed in a separate folder, such as `~/Scoop/apps/windhawk/1.7.3`. The symbolic link will be created only after installation is complete. Symbolic links are created after the installer runs.

So if I'm updating from 1.7.2 to 1.7.3, what is the flow?
1. `~/Scoop/apps/windhawk/1.7.3` is created
2. The installer is being run
3. The symbolic link is created
4. `post_install` is executed

When among these steps is the previous data from the `1.7.2` folder copied to `1.7.3`?

Since the main issue here is that `1.7.3` ends up having the old `windhawk.ini` file, I assume that after 2 `AppData` is renamed to `AppData.original`, and 3 actually moves `1.7.2/AppData` to `1.7.3/AppData`?

It all complicates stuff. That means that any merging strategy that the Windhawk installer might decide to do has to be reimplemented in a `post_install` script.

z-Fng: > Since the main issue here is that `1.7.3` ends up having the old `windhawk.ini` file, I assume that after 2 `AppData` is renamed to `AppData.original`, and 3 actually moves `1.7.2/AppData` to `1.7.3/AppData`?

According to the original manifest, `windhawk.ini` will be replaced in step 3:
- Rename `~/Scoop/apps/windhawk/1.7.3/AppData` to `~/Scoop/apps/windhawk/1.7.3/AppData.original`.
- Link `~/Scoop/apps/windhawk/1.7.3/AppData` to `~/Scoop/persist/windhawk/AppData`.
- Rename `~/Scoop/apps/windhawk/1.7.3/windhawk.ini` to `~/Scoop/apps/windhawk/1.7.3/windhawk.ini.original`.
- Link `~/Scoop/apps/windhawk/1.7.3/windhawk.ini` to `~/Scoop/persist/windhawk/windhawk.ini`.

```json
    "persist": [
        "AppData",
        "windhawk.ini"
    ],
```

> Since the main issue here is that 1.7.3 ends up having the old windhawk.ini file,

This issue can be easily resolved by removing windhawk.ini from the persist property.

Can other issues be resolved by moving the execution of the installer to post_install, since the symbolic link has already been created at that point? Or adjust folder permissions when launching the program.

m417z: > moving the execution of the installer to post_install

That sounds like the best solution, if possible. Also, in this case, `.original` items will never be created if I understand correctly.

z-Fng: > Also, in this case, `.original` items will never be created if I understand correctly.

Yes, that's right. Typically, we treat all installers as portable installation packages, simply extracting the files without running the installer.

m417z: But in this case, since the args are specified, I assume it's running, right? How do I move it to post_install?

z-Fng: > Also, the `warning.txt` files are there because the installer applies permissive access attributes to these folders:
> 
> * `AppData\Engine`: GenericRead + GenericExecute
> * `AppData\Engine\ModsWritable`: FullAccess
> * `AppData\Engine\Symbols`: FullAccess

Can the operation of applying permissive access attributes to these folders be completed when running? Moving the execution of the installer to the post_install script may cause some other issues.

m417z: At the very least it will require releasing a new version, and that's not something I'm planning for now. What kind of issues might it cause?

If that's a problem, perhaps applying the permissive access attributes in post_install the best approach for now.

z-Fng: Shims and shortcuts are created before post_install. If the installer runs in post_install script, we need to create them manually.

m417z: Thanks. I also pushed a small change to create `settings.ini` in Unicode format. Currently it's ASCII-only, but it will allow setting Unicode content in the file in the future.

If everything else works, I think it's good to go.

z-Fng: I'm not sure why the language settings would be lost. Everything else is working fine.

m417z: See: https://devblogs.microsoft.com/oldnewthing/20240606-00/?p=109861

What I mean is that if I decide to store non-ASCII data in the file in the future (for example, a path and the user name is non-ASCII), then unless the file is created with Unicode encoding, the string might not be saved correctly.

z-Fng: Tested with the latest commit. No matter what language I set, it defaults to English after reinstallation. It's a bit weird, as other settings are presisted normally.

z-Fng:
/verify. Ready to merge.

z-Fng (approved):

bucket/windhawk.json
m417z: I misinterpreted your message earlier.

When reinstalling, I'd imagine that this should be false, right?
```powershell
if(!(test-path "$persist_dir\AppData\settings.ini")) {...}
```

Perhaps `$persist_dir` points to the wrong location for some reason?

z-Fng: You can reproduce it by running the following command twice. Between the two runs, change the language and some other settings -- you'll notice the language setting isn't preserved. This may need to be fixed upstream later.
```
Start-Process -FilePath .\windhawk#1.7.3.exe -ArgumentList @('/S', '/portable', "/D=D:\windhawk") -Wait
```

m417z: You're right, I can reproduce it. I looked at the code, and the language is always overridden. The reason is that in non-silent mode, the user is prompted for the language, so whatever is selected is applied.

I'll add proper handling for this case in the next update. For now, it's possible to specify the language in the command line with `/lang=1033`, but I don't see how it can help without some extra scripting.

Let's keep it as is for now. The issue will only occur once, until the next update with the fix.

z-Fng: Thanks for the heads up and the thorough explanation! If anything needs further adjustments, feel free to open an issue or submit a PR anytime.

PR #15598, opened 2025-06-08
beyondcompare: Persist additional user data

ltguillaume:
Several vital configuration files and folders are missing from the persistent data specification. This PR fixed that.

### DIFF
--- a/bucket/beyondcompare.json
+++ b/bucket/beyondcompare.json
@@ -54,23 +54,32 @@
         "   $CONT = @('<BCPreferences Version=\"2\" MinVersion=\"2\"><TBcPrefs><CheckForUpdatesDayAmt Value=\"0\"/></TBcPrefs></BCPreferences>')",
         "   Set-Content \"$dir\\$file\" $CONT -Encoding Ascii",
         "}",
-        "$file = 'BCColors.xml'",
-        "if (!(Test-Path \"$persist_dir\\$file\")) {",
-        "   New-Item \"$dir\\$file\" -ItemType File | Out-Null",
-        "}",
-        "if (Test-Path \"$persist_dir\\BC5Key.txt\") { Copy-Item \"$persist_dir\\BC5Key.txt\" \"$dir\" }"
+        "'BCColors.xml', 'BCCommands.xml', 'BCFileFormats.xml', 'BC5Key.txt' | ForEach-Object {",
+        "   if (!(Test-Path \"$persist_dir\\$_\")) {",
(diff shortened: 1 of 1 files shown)


bucket/beyondcompare.json
```diff
@@ -54,23 +54,32 @@
...
     "persist": [
+        "Helpers",
+        "Packers",
+        "BCColors.xml",
+        "BCCommands.xml",
+        "BCFileFormats.xml",
         "BCPreferences.xml",
         "BCState.xml",
-        "BCColors.xml"
+        "BC5Key.txt"
```
HUMORCE, line 71: It will generate a empty BC5Key.txt, will it effect opening the program? and XMLs

ltguillaume: There are no problems with having an empty `BC5Key.txt` file. It will just start in the trial mode.

bucket/beyondcompare.json
```diff
@@ -54,23 +54,32 @@
...
     ],
     "pre_uninstall": [
-        "'BCPreferences.xml', 'BCState.xml', 'BCColors.xml', 'BC5Key.txt' | ForEach-Object {",
+        "'Helpers', 'Packers' | ForEach-Object {",
+        "    if ((Test-Path \"$dir\\$_\") -And !((Get-Item -Path \"$dir\\$_\" -Force).LinkType -eq \"Junction\")) {",
+        "        New-Item \"$dir\\$_\" -ItemType Directory -ErrorAction Ignore | Out-Null",
```
z-Fng, line 76: Hi. What is the purpose of copying the folder? Can the built-in persist feature in scoop achieve this?

ltguillaume: Not necessary for new installations, but I believe it was for those updating from the old manifest. Otherwise they would lose the contents they added to those directories themselves when updating. That said, it might never have worked to begin with, because it seems that when updating the (pre)uninstall procedures of the `manifest.json` inside the app/beyondcompare/current directory, which is then still the old one, are used instead of the one in the updated bucket. Personally, I think this is a design mistake, because it won't allow anyone to fix mistakes with an updated manifest before updating.

So indeed, lines 74-79 can be removed.

ltguillaume: #16249

z-Fng: Got it, thanks a lot for your reply! ❤️ 

It was during my review of #16125 that I wasn’t sure about its purpose, so I came here to ask.

PR #15679, opened 2025-06-27
simplyserial: Add persistent settings.cfg

### LINKED ISSUE #15678: simplyserial: add settings.cfg as persistent
https://github.com/fasteddy516/SimplySerial?tab=readme-ov-file#global-settings
> When SimplySerial starts, it looks for a file called settings.cfg in its application folder (the same location as the ss.exe program file.) 

So I think it makes sense to make it a persistent file.

Xotchkass:
Closes #15678

https://github.com/fasteddy516/SimplySerial?tab=readme-ov-file#global-settings


HUMORCE:
The file should be created when it does not exists, otherwise persistence does not work. 

See also: https://github.com/ScoopInstaller/Scoop/issues/6369

HUMORCE (changes requested):
On a fresh installation, the file does not exists, create it on install script.

Xotchkass:
> The file should be created when it does not exists, otherwise persistence does not work.

Thanks. Didn't know that

PR #15693, opened 2025-07-01
logitech-omm: Update to version 2.5.358

JonBCraig:

This PR updates the `logitech-omm` manifest to version `2.5.358`.

**Note regarding `checkver`:**
The existing `checkver` logic in the manifest appears to be ineffective or broken, as it does not reliably detect new versions. Despite efforts, I was unable to identify a reliable new `checkver` pattern for this software at this time.

Therefore, the `checkver` section has been left unchanged in this pull request. This update primarily focuses on providing the latest working version for users. A dedicated issue might be beneficial to track and resolve the `checkver` problem in the future.

### DIFF
--- a/bucket/logitech-omm.json
+++ b/bucket/logitech-omm.json
@@ -18,9 +18,21 @@
         ]
     ],
     "checkver": {
-        "url": "https://support.logi.com/api/v2/help_center/en-us/articles.json?label_names=webcontent=productdownload,websoftware=d3e2d42a-2b53-11eb-bf13-a5ea3d8273f5&per_page=1",
-        "jsonpath": "$.articles[0].body",
-        "regex": ">([\\d.]+)<"
+        "script": [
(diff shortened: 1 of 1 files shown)


aliesbelik:
> Therefore, the `checkver` section has been left unchanged in this pull request. This update primarily focuses on providing the latest working version for users. A dedicated issue might be beneficial to track and resolve the `checkver` problem in the future.

Unfortunately, updating only download url without fixing `checkver` won't help: url/hash will be overwritten during the next update check.
I don't like the idea to switch to using script but looks like this is the only working method here.

JonBCraig:
Hi @aliesbelik,

Thank you for your message and for the new modifications to the PR.

I must admit I'm still relatively new to contributing to Scoop manifests, especially when it comes to the intricacies of checkver logic.

Your help and expertise are greatly appreciated!

brian6932:
I don't think this over-complicated checkver's required anymore, the last 7 articles have all had the version in the body.

PR #15738, opened 2025-07-08
Update godot.json updating hashes for virustotal compatibility

bfcns:
Updating the hash from 512 to 256 for virustotal compatiblity.

- [ ] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

### DIFF
--- a/bucket/godot.json
+++ b/bucket/godot.json
@@ -6,15 +6,15 @@
     "architecture": {
         "64bit": {
             "url": "https://github.com/godotengine/godot/releases/download/4.4.1-stable/Godot_v4.4.1-stable_win64.exe.zip",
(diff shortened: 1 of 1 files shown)


HUMORCE (changes requested):
`autoupdate.hash` should be updated, otherwise hash will refill with sha512 value when the manifest updating.

bfcns:
So where is this `autoupdate.hash`, why do we keep it sha512?

HUMORCE:
https://github.com/ScoopInstaller/Extras/blob/a5d669a1511b21f3a7404e9ab35e59706ce75060/bucket/godot.json#L52-L54

bfcns:
I do not know how to do that. There is no SHA256-SUMS.txt file there, but when you check https://github.com/godotengine/godot/releases assets each file has sha256 besides them.

PR #15777, opened 2025-07-13
googlechrome: Add arm64 and update source url

PIPIPIG233666:
use https://github.com/Bush2021/chrome_installer to track updates instead of relying on additional xml parsing


z-Fng:
We tend to avoid using unofficial sources.

PR #12699, opened 2024-01-25
cpu-z-arm64: add v1.01

AkariiinMKII:
### Help needed 

The package of arm64 version does not include config file, therefore a `pre_install` script was added to ensure the file before persist.
However, I have no WoA device to test the manifest, I'll appreciate if someone can come with help ^^

Closes #12698

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

PR #15840, opened 2025-07-19
cpu-z-arm64: Add version 1.04

artiga033:
Added as a seperated software because:
- this is what they did in their official site (<https://www.cpuid.com/softwares/cpu-z-arm64.html>). The name `cpu-z-arm64` is taken from the site url, too.
- It has different version number with x86/x64 version
- Theoretically, Windows on ARM users can run the x64 and arm64 version seperately for different purpose, the former for detecting the emulated chip info and the latter for the real CPU.

The shorcut name `CPU-Z ARM64` is take from the window title, in comparison with x64 version.

image

### DIFF
--- a/bucket/cpu-z-arm64.json
+++ b/bucket/cpu-z-arm64.json
@@ -1,0 +1,28 @@
+{
+    "version": "1.04",
+    "description": "System information software",
+    "homepage": "https://www.cpuid.com/softwares/cpu-z-arm64.html",
+    "license": "Freeware",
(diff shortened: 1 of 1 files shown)


aliesbelik:
Duplicates #12699.

artiga033:
is there any reason for that not being merged? For lack of test I can ensure that it works well at least with my surface pro 11 with snapdragon X Elite chips

z-Fng:
Closing this PR in favor of #12699 (merged).

PR #15971, opened 2025-08-11
eclipse-mat: Update to version 1.16.1

frankclr:
eclipse-mat@1.15.0 can't be updated. eclipse update address had changed, need to be correct,and update to 1.16.1

the orginal update address "https://www.eclipse.org/mat/downloads.php" can't be accessed now. so that no update more than two years.

### DIFF
--- a/bucket/eclipse-mat.json
+++ b/bucket/eclipse-mat.json
@@ -1,12 +1,12 @@
 {
-    "version": "1.15.0.20231206",
+    "version": "1.16.1.20250109",
     "description": "Eclipse Memory Analyzer",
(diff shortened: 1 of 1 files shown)


bucket/eclipse-mat.json
```diff
@@ -1,12 +1,12 @@
 {
-    "version": "1.15.0.20231206",
+    "version": "1.16.1.20250109",
     "description": "Eclipse Memory Analyzer",
     "homepage": "https://www.eclipse.org/mat/",
```
aliesbelik:
```suggestion
    "homepage": "https://eclipse.dev/mat/",
```

PR #16005, opened 2025-08-17
cherry-studio@1.5.6: Switch Cherry Studio manifest to portable version

GeorgeDong32:
- Switch Cherry Studio manifest to portable version
- Add notes and depends
- Optimize persist

- [ ] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

### DIFF
--- a/bucket/cherry-studio.json
+++ b/bucket/cherry-studio.json
@@ -3,35 +3,49 @@
     "description": "A desktop client that supports for multiple LLM providers.",
     "homepage": "https://cherry-ai.com",
     "license": "Apache-2.0",
+    "notes": "You'd better turn off update detection in Settings->About Us in Cherry Studio and use scoop to manage updates, otherwise it will cause some errors.",
+    "depends": "extras/vcredist2022",
+    "suggest": {
+        "vcredist": "extras/vcredist2022"
+    },
     "architecture": {
         "64bit": {
-            "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v1.5.6/Cherry-Studio-1.5.6-x64-setup.exe#/dl.7z",
-            "hash": "sha512:53ac4ce98e6a135ce23696c0c466976472bf3691bfc0ec6eb4f336bb7e43e2b83a79621459be0ef0ee3f435a4cc4af424ab7e445054bbd16cc130d25c46eaa9b",
-            "pre_install": "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-64.7z\" \"$dir\""
+            "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v1.5.6/Cherry-Studio-1.5.6-x64-portable.exe#/cherry-studio.exe",
+            "hash": "5b791570ee341d0cfd08247960707145f778f7755f0b855ef9186586bacdc629"
         },
         "arm64": {
-            "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v1.5.6/Cherry-Studio-1.5.6-arm64-setup.exe#/dl.7z",
-            "hash": "d86af4a3c5b4d22c41093514665eba5ecb85e1e3768f81f12328e5cf7fcdf84e",
-            "pre_install": "Expand-7zipArchive \"$dir\\`$PLUGINSDIR\\app-arm64.7z\" \"$dir\""
+            "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v1.5.6/Cherry-Studio-1.5.6-arm64-portable.exe#/cherry-studio.exe",
+            "hash": "549fc360e8da47f589df9c17566fe646d653477b261d8be04bc9556c81e2bf7a"
         }
     },
+    "post_install": [
+        "# Handle data directory migration",
+        "if (Test-Path \"$persist_dir\\data\") {",
+        "  $dataItems = Get-ChildItem \"$persist_dir\\data\" -ErrorAction SilentlyContinue",
+        "  if (-not $dataItems) {",
+        "    # persist_dir\\data exists but is empty, check if we need to copy from APPDATA",
+        "    if ((Test-Path \"$env:APPDATA\\CherryStudio\") -and (Get-ChildItem \"$env:APPDATA\\CherryStudio\" -ErrorAction SilentlyContinue)) {",
+        "      Copy-Item \"$env:APPDATA\\CherryStudio\\*\" \"$persist_dir\\data\" -Recurse -Force",
+        "      info \"Data copied from '$env:APPDATA\\CherryStudio' to persist directory.\"",
+        "    }",
+        "  }",
+        "} else {",
+        "  # persist_dir\\data does not exist",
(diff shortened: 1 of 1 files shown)


GeorgeDong32:
Hi here, I am a core developer of Cherry Studio. This PR aims to resolve previously encountered data and dependency issues during Scoop installation. Please let us know of any inaccuracies, and we will promptly address them.

bucket/cherry-studio.json
```diff
@@ -3,50 +3,60 @@
     "description": "A desktop client that supports for multiple LLM providers.",
     "homepage": "https://cherry-ai.com",
     "license": "Apache-2.0",
+    "notes": "You'd better turn off update detection in Settings->About Us in Cherry Studio and use scoop to manage updates, otherwise it will cause some errors.",
+    "depends": "extras/vcredist2022",
+    "suggest": {
+        "vcredist": "extras/vcredist2022"
+    },
```
Lutra-Fs: only keep `suggest` field is enough, as many times the users machine have external source `vcredist`. The other thing is many times you do not want to use `depends` and `suggest` at the same time, 

See https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests
depends: Runtime dependencies for the app which will be installed automatically. See also suggest (below) for an alternative to depends.

suggest: Display a message suggesting optional apps that provide complementary features. See [ant](https://github.com/ScoopInstaller/Main/blob/master/bucket/ant.json) for an example.
["Feature Name"] = [ "app1", "app2"... ]
e.g. "JDK": [ "extras/oraclejdk", "openjdk" ]
If any of the apps suggested for the feature are already installed, the feature will be treated as 'fulfilled' and the user won't see any suggestions.

bucket/cherry-studio.json
```diff
@@ -3,50 +3,60 @@
...
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v$version/Cherry-Studio-$version-x64-setup.exe#/dl.7z",
-                "hash": {
-                    "url": "$baseurl/latest.yml",
-                    "regex": "sha512: $base64"
-                }
+                "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v$version/Cherry-Studio-$version-x64-portable.exe#/cherry-studio.exe"
```
Lutra-Fs, line 60: Are you no longer provided pre-calculated checksum?

bucket/cherry-studio.json
```diff
@@ -3,50 +3,60 @@
...
-        "# Compatible with incorrect persistence method",
-        "if (Test-Path \"$persist_dir\\data\") {",
-        "  ensure \"$env:APPDATA\\CherryStudio\" | Out-Null",
-        "  Move-Item \"$persist_dir\\data\\*\" \"$env:APPDATA\\CherryStudio\" -Recurse",
-        "  Remove-Item \"$persist_dir\" -Recurse -Force",
-        "  info \"This manifest no longer provides persistence.\"",
-        "  info \"Your data has been moved to '$env:APPDATA\\CherryStudio'(default location).\"",
-        "}"
+    "persist": [
+        "data"
```
Lutra-Fs, line 52: I have not checked the reason that disables Cherry Studio's persistence, but Is Cherry Studio portable version fine with data precreated and is a symlink?

bucket/cherry-studio.json
```diff
@@ -3,50 +3,60 @@
...
+        "    }",
+        "  }",
+        "} else {",
+        "  # persist_dir\\data does not exist",
+        "  if ((Test-Path \"$env:APPDATA\\CherryStudio\") -and (Get-ChildItem \"$env:APPDATA\\CherryStudio\" -ErrorAction SilentlyContinue)) {",
+        "    ensure \"$persist_dir\\data\" | Out-Null",
+        "    Copy-Item \"$env:APPDATA\\CherryStudio\\*\" \"$persist_dir\\data\" -Recurse -Force",
+        "    info \"Data copied from '$env:APPDATA\\CherryStudio' to persist directory.\"",
+        "  }",
+        "}"
```
Lutra-Fs, line 43: See https://github.com/ScoopInstaller/Main/pull/6462#issuecomment-2812476661

bucket/cherry-studio.json
Lutra-Fs: I can see it is still available

GeorgeDong32: The latest.yml only calculates the SHA for the non-portable version; the portable version of the software has not been precomputed.

Lutra-Fs: Oh yeah, ok. That said, can you pre-calculate the SHA for the portable version as well? That would help us improve security. But totally fine if leave it as is.

bucket/cherry-studio.json
GeorgeDong32: The portable version uses the "data" folder located in the program directory, which is identical to the AppData/Cherry Studio folder used by the installed version. In our tests, the symbolic link for "persist" works correctly. Previously, "persist" was disabled because users of the installed version did not disable automatic updates, causing all data in the program directory to be deleted. This issue does not occur with the portable version.

bucket/cherry-studio.json
```diff
@@ -3,51 +3,65 @@
...
+                "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v$version/Cherry-Studio-$version-x64-portable.exe#/cherry-studio.exe"
             },
             "arm64": {
-                "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v$version/Cherry-Studio-$version-arm64-setup.exe#/dl.7z"
+                "url": "https://github.com/CherryHQ/cherry-studio/releases/download/v$version/Cherry-Studio-$version-arm64-portable.exe#/cherry-studio.exe"
             }
         }
     }
-}
+}
```
Lutra-Fs: Files should end with a newline. This is currently the reason that our CI failed.

bucket/cherry-studio.json
Lutra-Fs: Thanks for your help. Based on the recommendation in [ScoopInstaller/Main#6462](https://github.com/ScoopInstaller/Main/pull/6462), as long as this change does not result in data loss or issues for existing users, we should be fine. If user data is preserved and persisted correctly, a clear note in the manifest should be sufficient. So I would recommend you to remove the post_install script.

GeorgeDong32:
Unfortunately, I’ve sustained an injury that prevents me from sitting to code. I’m currently unable to work on this PR and will need a few days to recover before I can resume maintenance.

GeorgeDong32:
Thanks for the review. The `post_install` script remains necessary to handle **critical data migration when switching from non-portable to portable mode**.  

Here's why:  
1️⃣ **Previous install script behavior**:  
   The previous install script intentionally relocated old user data to `%APPDATA%` (as per legacy design) and does not persist any data in scoop.
2️⃣ **Portable mode requirement**:  
   Portable installations **must** store data in the `data` directory to maintain state across sessions, which will be persisted
3️⃣ **Migration gap**:  
   Without this script, users switching to portable mode would **lose all existing data** since the installer can't automatically reverse the `%APPDATA%` migration

PR #16018, opened 2025-08-20
lark: fix lark url as the download url has updated

cuichenli:


aliesbelik:
Guess it makes sense to add separate sections for `64bit` as well.

cuichenli:
> Guess it makes sense to add separate sections for `64bit` as well.

It seems that the 64-bit and 32-bit builds use different version tags. I’m not sure how to define this behavior in `checkver`. Could you share some examples or documentation that I can refer to? Thanks!

z-Fng:
/verify

> Guess it makes sense to add separate sections for `64bit` as well.

This will be resolved in #16533 instead.

Issue #13289, opened 2024-05-14
format-factory@5.8.1.0: download via aria2 failed


LinuxGratos:
Latest 5.17.0.0 can be found from there: http://pcfreetime.com/public/FFSetup5.17.0.0.exe
but download is very slow. Could/Should we mirror it somewhere?

aliesbelik:
Looks like manifest was [pinned to 5.8.1.0, with disabled autoupdate](https://github.com/ScoopInstaller/Extras/commit/001a67f10c3b89dc84148a726f1edc913993fa2d), as per https://github.com/ScoopInstaller/Extras/issues/7371#issuecomment-976839197.
It will be helpful if you could check the latest version with current manifest or update manifest properly to work with the latest version.

ciaranchen:
1. The download URL has been moved to http://down.pcgeshi.com/FormatFactory_5.12.0.0_setup.exe . Its server doesn't support the Referer header, so `#sourceforge.net.exe` has been added to the URL. `exe` ensures the downloaded file is an exe file.
2. The installer can't be decompressed using 7-zip because it switches to NSIS Installer. I modified the `installer` and other fields of manifest based on the [`anaconda3` manifest](https://github.com/ScoopInstaller/Extras/blob/master/bucket/anaconda3.json).

I'll submit a pull request with the changes then.

z-Fng:
Probably it's time to deprecate it and move it to the [Nonportable](https://github.com/ScoopInstaller/Nonportable) bucket.

Issue #7371, opened 2021-11-20
format-factory@5.9.0.0: decompress error


CALMorACT:

BYJRK:
same error

pratikpc:
Can you share the full logs?

What is printed on the console?

Screenshots would be helpful

issaclin32:
**Format Factory** changed their installer. Unfortunately I can't find a way to extract it properly.

I have [downgraded to the last working version, and disabled update](https://github.com/ScoopInstaller/Extras/commit/001a67f10c3b89dc84148a726f1edc913993fa2d).

Will work on that if we can find a way to extract it.

```
FORMAT-FACTORY#5.9.0.0#HTTP_PUBLIC.PCFREETIME.COM_FFSETUP5.9.0.0.EXE_DL
│  DynActsBLL.dll
│  GenericSetup.exe  -> This file cannot be further extracted.
│  GenericSetup.exe.config
│
├─de
│      GenericSetup.resources.dll
│
├─es
│      GenericSetup.resources.dll
│
├─fr
│      GenericSetup.resources.dll
├─it
│      GenericSetup.resources.dll
│
├─pt
│      GenericSetup.resources.dll
│
└─ru
        GenericSetup.resources.dll
```

PR #16088, opened 2025-09-02
format-factory: Update to version 5.12.0.0, fix installation

ciaranchen:
Updated the software version to 5.12.0.0, modified the download link and hash value, rewrote the installation script to use silent installation, and added an uninstall script and automatic update configuration.

1. The download URL has been moved to http://down.pcgeshi.com/FormatFactory_5.12.0.0_setup.exe . Its server doesn't support the Referer header, so `#sourceforge.net.exe` has been added to the URL. `exe` ensures the downloaded file is an exe file.
2. The installer can't be decompressed using 7-zip because it switches to NSIS Installer. I modified the `installer` and other fields of manifest based on the [`anaconda3` manifest](https://github.com/ScoopInstaller/Extras/blob/master/bucket/anaconda3.json).

Relates to #13289 and #7371 .

- [*] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [*] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md) 

## Summary by CodeRabbit

* **New Features**
  * Enabled automatic version detection and autoupdate for FormatFactory.
  * Added silent installer and silent uninstaller flows with post-install cleanup.

* **Bug Fixes**
  * Updated download source and checksum to address installation issues.
  * Replaced legacy pre-install extraction with a reliable installer-based process.

* **Chores**
  * Bumped FormatFactory to version 5.12.0.0.

### DIFF
--- a/bucket/format-factory.json
+++ b/bucket/format-factory.json
@@ -26,7 +26,7 @@
         ]
     ],
     "checkver": {
-        "url": "http://www.pcfreetime.com/formatfactory/index.php",
+        "url": "http://www.pcfreetime.com/formatfactory/CN/index.html",
         "regex": "版本：([\\d+.]+)"
     },
     "autoupdate": {


z-Fng:
Scoop does not usually run installer directly, so that users get the [benefits Scoop promised](https://github.com/ScoopInstaller/Scoop?tab=readme-ov-file#what-does-scoop-do) when installing it via Scoop.

That's probably why we decided to [downgraded to the last working version, and disabled update](https://github.com/ScoopInstaller/Extras/issues/7371#issuecomment-976839197).

Consider adding it to [Nonportable](https://github.com/ScoopInstaller/Nonportable) bucket instead.

PR #16266, opened 2025-10-04
discord@1.0.9210: switch to official version

### LINKED ISSUE #8383: Discord updated to portable edition, now opens multiple instances when invoked multiple times
## Bug Report

**Package Name:** discord

### Current Behaviour

If I invoke `discord-portable.exe` more than once, I get a new instance of discord per invocation, deviating from the prior behavior and breaking the simple AHK script I was using:
```
!+d::
if WinExist("ahk_exe DiscordPTB.exe")
    WinActivate ahk_exe DiscordPTB.exe
else
    EnvGet, UserProfile, UserProfile
    Run "%UserProfile%\scoop\apps\discord\current\discord-portable.exe"
return
```
- if I ever close the discord window, AHK can't find it (of course), so it it re-invokes the entry binary, and that doesn't seem to guard against multiple instances any more (it used to work, and that was when the binary in there was just called `discord.exe`

### Expected Behaviour
re-invoking the entry binary should re-surface the single instance of discord, like it used to and like it does on other platforms where I use it
I have tried re-invoking the Discord.exe in the app folder, but that does the same thing, so it appears as if the portable version doesn't do single-instance.

For the mean-time, I've switched to DiscordPTB, which behaves as expected.

### Possible Solution

go back to whatever method was used before to distribute discord, instead of the portable version
_or_
raise an issue with the maintainer of the portable version

the prior method worked fine and is within control of this repo, so it seems like the easier solution

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:** 7.2.2

#### Scoop Configuration

```json
{
  "lastUpdate": "2022-04-25T11:04:06.1953022+02:00",
  "SCOOP_REPO": "https://github.com/ScoopInstaller/Scoop",
  "SCOOP_BRANCH": "master"
}
```

zeldrisho:
Closes #8383

Closes portapps/discord-portable/issues/91

This version deals with the issues of portapps version (autostart, universal linking), it can autoupdate at "current" folder

### DIFF
--- a/bucket/discord.json
+++ b/bucket/discord.json
@@ -1,39 +1,42 @@
 {
-    "version": "1.0.9208-23",
-    "description": "Free Voice and Text Chat",
-    "homepage": "https://discord.com/",
+    "version": "1.0.9210",
+    "description": "Group Chat That's All Fun & Games",
+    "homepage": "https://discord.com",
     "license": {
         "identifier": "Freeware",
         "url": "https://discord.com/terms"
     },
+    "notes": "If it doesn't autostart, re-enable app startup setting",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/portapps/discord-portable/releases/download/1.0.9208-23/discord-portable-win64-1.0.9208-23.7z",
-            "hash": "3f865d535bccfbf437904211f9532eee0c3d5c5fa766cdf5447aa9ac9d2e361d"
+            "url": "https://stable.dl2.discordapp.net/distro/app/stable/win/x64/1.0.9210/DiscordSetup.exe#/dl.7z",
+            "hash": "97d9ed582ec7efe45a0eb76765b41972483012ebf526708568d5e3d4701ba083"
         }
     },
+    "pre_install": [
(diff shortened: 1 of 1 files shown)


SorYoshino:
- You can use `x64/([\\d.]+)` to extract the version number, as `([\\d.]+)` is the regular expression commonly used in the [official wiki](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate) for version detection.

- After testing, the current manifest works fine during normal usage. However, when a new version is detected, the software automatically downloads and installs the update. The solution in commit [07a1343](https://github.com/ScoopInstaller/Extras/commit/07a1343e3e4dd1e1b996d40ba5377bc67776576d) no longer works.

- The relevant information should be provided in the `notes` field; otherwise, users might mistakenly believe that their data has been lost.

zeldrisho:
@SorYoshino app can be allowed to autoupdate if it updates only inside the "current" folder, like some manifests (telegram, obsidian)

zeldrisho:
@aliesbelik @HUMORCE could you kindly review this PR? Also my previous PR (users request new packages for a long time ago)

SorYoshino:
> app can be allowed to autoupdate if it updates only inside the "current" folder, like some manifests (telegram, obsidian)

Thank you for your explanation, and I agree with your point of view.

PR #16334, opened 2025-10-14
flclash: Update to version 0.8.90, drop arm64 support, add hash extraction

BaoPaper:
The automatic update process was failing because the upstream repository has stopped compiling arm64 packages, but our update script still included arm64 architecture detection. This caused the update to fail when attempting to fetch non-existent arm64 binaries.

This commit removes the arm64 architecture check from the auto-update process to restore functionality.

## Summary by CodeRabbit

* **Chores**
  * Upgraded to version 0.8.90.
  * Updated 64‑bit x86 download link and checksum to match the new release.
  * Simplified version check behavior and adjusted auto-update to rely on the downloaded file checksum URL.
  * Removed ARM64 entries from releases and auto-update; ARM64 users should use an alternative distribution.

### DIFF
--- a/bucket/flclash.json
+++ b/bucket/flclash.json
@@ -15,14 +15,15 @@
             "FlClash"
         ]
     ],
-    "checkver": {
-        "github": "https://github.com/chen08209/FlClash"
-    },
+    "checkver": "github",
     "autoupdate": {
         "architecture": {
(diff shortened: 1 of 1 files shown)


z-Fng:
/verify. Update checkver, add hash extraction.

PR #16373, opened 2025-10-19
fvim: Update to version 0.3.549, fix checkver & autoupdate

o-l-a-v:
Changes:

* Scripted custom checkver logic because author does a lot non-standard stuff with versioning, tag name of releases, and naming release artifacts.

## Summary by CodeRabbit

* **Release**
  * Version bumped to 0.3.549+ee4316c with updated 64‑bit and arm64 assets and checksums.
* **Updates**
  * Improved update detection and delivery to reliably fetch the latest release and provide the correct 64‑bit and arm64 downloads.

### DIFF
--- a/bucket/fvim.json
+++ b/bucket/fvim.json
@@ -1,5 +1,5 @@
 {
-    "version": "0.3.549+ee4316c",
+    "version": "0.3.549",
     "description": "Neovim front-end UI",
     "homepage": "https://github.com/yatli/fvim",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


SorYoshino:
I believe we don't need to keep `+ee4316c` in the `version` field. Using just `0.3.549` is sufficient, as previous releases show that the `[\d.]+` part has never been duplicated, so no additional information is needed to distinguish versions.

o-l-a-v:
I removed `+ee4316c` as suggested by @SorYoshino. I kept the old version format in the checkver logic as `$matchTagname`, in case it's needed in the future.

Issue #16379, opened 2025-10-19
[Bug] Various packages stopped updating due to checkver errors


z-Fng:
### Description
Relates to:
- #2308

The issue comes from:
- #16349

There are lots of error logs during excavator execution. Various packages stopped updating due to checkver errors.
See: https://github.com/ScoopInstaller/Extras/actions/runs/18627830520/job/53108310389

Contributors are warmly welcome to help fix one or more of these broken manifest. Please note:
- Make sure there are no existing fix PRs.
- Create a separate PR or a separate commit for each manifest, with a title following the naming rules.
- Please do not attempt to close this issue in your PR.
- For manifests that are still available but will no longer receive updates, remove their checkver fields instead of deprecating them.
- For autoupdate filed, if we can still install a specific version via `scoop install app@version` (i.e., autoupdate is version-specific), removing it is also not recommended. https://github.com/ScoopInstaller/Extras/pull/16392#issuecomment-3438955656
> IMO, if we keep the autoupdate field, users can still install a specific version via `scoop install app@version`. Otherwise:
> ```
> PS> scoop install fbflipper@0.238.0
> WARN  Given version (0.238.0) does not match manifest (0.239.0)
> WARN  Attempting to generate manifest for 'fbflipper' (0.238.0)
> 'fbflipper' does not have autoupdate capability
> ```
> P.S. If https://github.com/ScoopInstaller/Scoop/pull/6370 gets merged, this will no longer be necessary and can be removed later.

### Package List
Here are all the packages stopped updating due to checkver errors.

This result was obtained using a simple matching script and may contain false positives.

- [x] adventuregamestudio: #16436 
- [x] anaconda3: #16678
- [x] anki: #16115
- [x] anythingllm: #16385 
- [x] authy: #16386 
- [x] autohotkey: #16384
- [x] b1: Deprecate manifest https://github.com/ScoopInstaller/Extras/commit/77302c5e4a3d30445d22d1f5e806e8c0ad3fc6ce
- [ ] basecamp
- [x] betaflight-blackbox-explorer: #16387
- [x] biblioteq: #16394 
- [ ] bitmessage
- [x] blender: #16563
- [x] capframex: https://github.com/ScoopInstaller/Extras/commit/d6f5a8eeb8a331943e054db95b7308586571d099
- [ ] ccleaner
- [x] chef-workstation: #16071 #16383
- [x] cider: #16388
- [x] classic-volume-mixer: #16382
- [x] clementine: #16381
[...]

PR #16394, opened 2025-10-19
biblioteq@2025.05.17: Checkver exclude versions with no build artifacts

o-l-a-v:
Very few new versions of biblioteq actually have build artifacts added. Seems it's a manually process to build and attach those:

* <https://github.com/textbrowser/biblioteq/issues/300>

Thus, the manifest should only look for new versions with at least one build artifact added to it, to avoid errors on each run of Excavator.

## Summary by CodeRabbit

* **Documentation**
  * Clarified version-checking behavior and noted that some releases may lack build artifacts.

* **Chores**
  * Switched version checks to use the GitHub Releases API and refined release-tag detection for more accurate update info.

### DIFF
--- a/bucket/biblioteq.json
+++ b/bucket/biblioteq.json
@@ -1,4 +1,5 @@
 {
+    "##": "Not all releases include build artifacts, hence the checkver.jsonpath. See: https://github.com/textbrowser/biblioteq/issues/300.",
     "version": "2025.05.17",
     "description": "Professional archiving, cataloging, and library management suite providing connectivity to PostgreSQL and SQLite.",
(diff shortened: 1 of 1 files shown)


o-l-a-v:
Related: <https://github.com/ScoopInstaller/Extras/issues/16379>.

o-l-a-v:
Why not have the comment? It's not a note that will be visible during install/upgrade.

z-Fng:
> Why not have the comment? It's not a note that will be visible during install/upgrade.

Yeah, I know. I thought removing it would make the manifest clearer, but now I see it was actually necessary.

PR #16449, opened 2025-10-23
youtrack@2025.1.82518: Removed checkver

Vixb1122:
Changes:

- Removed checkver and autoupdate

Because:

- ZIP releases are no longer created for new versions, now only Docker is supported.
     - https://www.jetbrains.com/help/youtrack/server/youtrack-zip.html
     - https://www.jetbrains.com/youtrack/download/previous.html

Relates to #16379

## Summary by CodeRabbit

* **Chores**
  * Updated distribution config so ZIP packages are no longer generated for new releases.
  * Disabled automatic remote version check and related update extraction logic.

### DIFF
--- a/bucket/youtrack.json
+++ b/bucket/youtrack.json
@@ -28,5 +28,12 @@
         "conf",
         "data",
         "logs"
-    ]
+    ],
+    "autoupdate": {
+        "url": "https://download.jetbrains.com/charisma/youtrack-$version.zip",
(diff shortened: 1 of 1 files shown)


z-Fng:
> - For autoupdate filed, if we can still install a specific version via `scoop install app@version` (i.e., autoupdate is version-specific), removing it is also not recommended.

PR #16452, opened 2025-10-24
hwinfo: Migrate package source to SourceForge, refactor manifest

### LINKED ISSUE #16361: [Bug]: hwinfo stopped updating
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

hwinfo

### Expected/Current Behaviour

The current version of hwinfo is 8.32, but scoop is not detecting it since 8.30 update. Additionally 8.30 gives 404 on install which will probably be moot once the update is fixed.

Run scoop install hwinfo
Observe that version 8.30 is trying to be installed (and fails)
Expected 8.32 (as of the time of writing)

### Steps to Reproduce

```console
scoop install hwinfo
Installing 'hwinfo' (8.30-5800) [64bit] from 'extras' bucket
URL https://www.sac.sk/download/utildiag/hwi_830.zip is not valid
```

### Possible Solution

It looks like something changed upsteam with hwinfo that require updating hwinfo.json here to match. In order to sovle this:

1. Find out why hwinfo version is no longer updating
2. Fix hwinfo.json applying information found in step 1 to resolve the issue
3. Verify that after the fix an installation of scoop that is on hwinfo 8.30 currently successfully updates to 8.32 or later
4. Verify that a fresh installation can install actual latest hwinfo successfully

### Scoop and Buckets Version

```console
scoop --version
Current Scoop version:
b588a06e (HEAD -> master, tag: v0.5.3, origin/master, origin/HEAD) chore(release): Bump to version 0.5.3 (resync) (#6436)

'main' bucket:
9c918ba0d (HEAD -> master, origin/master, origin/HEAD) sqldef: Update to version 3.1.16

'extras' bucket:
3815e15d27 (HEAD -> master, origin/master, origin/HEAD) furmark@2.10.0.0: Fix hash (Closes #16360)
```

### Scoop Config

```console
scoop config

shim                  : kiennq
aria2-warning-enabled : False
debug                 : False
alias                 :
last_update           : 17-Oct-25 17:29:44
scoop_repo            : https://github.com/lukesampson/scoop
scoop_branch          : master
```

### PowerShell Version

```console
$PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.5.3
PSEdition                      Core
GitCommitId                    7.5.3
OS                             Microsoft Windows 10.0.26200
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Additional Softwares

_No response_

brian6932:
~~hwinfo's website should be avoided as a `checkver` source, the site uses strong Cloudflare bot protections~~ Discussed with the developer in https://www.hwinfo.com/forum/threads/10829/.
Closes #16366
Fixes #16361
Relates to #16379

## Summary by CodeRabbit

* **Chores**
  * Updated to version 8.32.
  * Switched downloads and autoupdate to SourceForge.
  * License now includes an official reference URL.
  * Unified executable and shortcut configuration for easier access.
  * Replaced prior persistence declarations with an installer-based persistence workflow.
  * Installer and uninstall behavior improved to better preserve and restore user data.
  * Architecture-specific install scripts added for 64-bit, 32-bit, and ARM64 builds.


SorYoshino:
When writing the manifest in #16366, I actually considered using **SourceForge**.  

However, the `$majorVersion` and `$minorVersion` values retrieved from there are concatenated together, making it difficult to distinguish between them when `$majorVersion` is greater than or equal to 10.  

For that reason, I decided not to adopt this approach.

---

Furthermore, I don’t think the `-patch` information you mentioned in the forum is a real concern.  

Since the `hwinfo` manifest was first added to the **Extras** bucket in 2018, the `-patch` field has **only had an actual effect** in two cases — in commits https://github.com/ScoopInstaller/Extras/commit/7bcb67ea35796b0e35e134d387bd6eef209eb3a9 and https://github.com/ScoopInstaller/Extras/commit/498241361647ab2f9297461c90c5d78178f04d0a. Even then, there was a suspicious detail: when updating from **7.22-4730** to **7.22-4731**, the corresponding hash value didn’t change. 
 
Moreover, Scoop’s manifest update mechanism appears to be **forced**, meaning the hash can still be corrected even without relying on this `-patch` part.

brian6932:
With regex, it's not too hard to segment the majorVer yourself (as long as some version doesn't hit 100 minor vers, since there's no zero, I'd hope not), ofc it would be better to have the /ver.txt endpoint, but I think this is the safest/best option for now.

SorYoshino:
> With regex, it's not too hard to segment the majorVer yourself (as long as some version doesn't hit 100 minor vers, since there's no zero, I'd hope not),

Yes, if `$majorVersion` reaches 10 or a higher number, it can be matched correctly by simply changing the pattern to `\d{2}`. Similarly, if `$minorVersion` reaches 100 or more, a similar adjustment can be made.  

However, this would inevitably increase maintenance frequency and could introduce other hidden issues — which is exactly why I chose not to use **SourceForge** as the `checkver` source.  

Hopefully, the author of `hwinfo` will allow us to use `ver.txt`, since both of the current `checkver` approaches have certain limitations.

brian6932:
> Yes, if `$majorVersion` reaches 10 or a higher number, it can be matched correctly by simply changing the pattern to `\d{2}`.

This is an inevitability, and there's no reason to alter it in the future, the `+` is enough. The minor version getting to `100` I'm very doubtful of, as there's no zero to indicate the placement, it's a possibility, but extremely small.

SorYoshino:
Oh, sorry, I didn’t notice before that you had already updated the regex to `hwi_(\d+)(\d{2})\.zip`.

I hadn’t given this issue much thought earlier and hadn’t considered that this approach could solve the problem. Thank you for your response. 

As for the possibility of `$minorVersion` having more than two digits, since the manifest was added to the Extras bucket in 2018, such a case has never occurred and can be considered extremely unlikely. On this basis, this method indeed doesn’t pose any significant issues.

SorYoshino:
Hello, upstream has relaxed the rules, so we can now use `ver.txt`. There's no need for us to modify the `checkver` logic at this point—just keep the refactoring to the manifest.

If you agree, you can close this PR. I'll revert to the original `checkver` approach in #16366. Alternatively, you can directly restore the original `checkver` method in this PR—either way is fine with me.

brian6932:
So I think that the `autoupdate` bins, and `hash` should still use Sourceforge, but `checkver` should be using /ver.txt. This way we get hash checking, and don't have to use a mirror. I'll add it onto this PR since most of that's already here.

SorYoshino:
I have no problem.

z-Fng:
>  https://www.hwinfo.com/forum/threads/10829/

Thank you so much for your effort. @brian6932

z-Fng:
So is #16452 actually a fix based on #16366? Why do it this way? And what if changes are needed?

A pull request should focus on a single change IMO.

bucket/hwinfo.json
```diff
@@ -1,81 +1,60 @@
...
-    "persist": [
-        "HWiNFO64.INI",
-        "HWiNFO32.INI"
+    "installer": {
+        "script": [
+            "$setting_name = if ($architecture -match '\\d{2}(?=bit)') { 'HWiNFO{0}.INI' -f $matches[0] } else { 'HWiNFO_ARM64.INI' }",
+            "if (-not (Test-Path \"$persist_dir\\$setting_name\")) {",
+            "    Set-Content -Path \"$dir\\$setting_name\" -Value '[Settings]', 'AutoUpdate=0' -Encoding ASCII",
+            "}",
+            "persist_data @{ persist = \"$setting_name\" } $original_dir $persist_dir",
```
z-Fng, line 37: Avoid using functions from an external scope, as renaming may break this logic.
For commonly used functions, refactoring usually assumes they won't change, or if they do, all manifests using them will be updated accordingly. However, functions like `persist_data` are easily overlooked.

May need a fix after merging.

z-Fng (approved):
Thank you both for your contributions!

bucket/hwinfo.json
SorYoshino: Yes, currently only the `cygwin` and `hwinfo` manifests in the buckets use `persist_data`.

When refactoring `hwinfo`, I referenced the `cygwin` manifest because `hwinfo` is somewhat special: the archive contains only three binaries for different architectures. Aside from their respective configuration files and the `*64_KEY.txt`, it does not generate any additional files in `$dir` after execution. This makes it highly suitable for scenarios where each architecture requires only its specific files. Moreover, since `architecture..persist` does not function correctly at the moment, I adopted `persist_data` to ensure the intended behavior.

As for maintainability, I believe there is no cause for concern. On one hand, I performed thorough testing before submitting the PR, so the likelihood of issues is very low. On the other hand, for most manifests I maintain, I continue tracking their status afterward and will allocate time to fix issues if they arise. Additionally, I use this application myself, so I naturally would not allow errors to affect my own experience.

---

If there is any potential issue, it would concern the `arm64`-specific components—for example, whether the configuration file should be `HWiNFO_ARM64.INI` and whether the key file follows the `*64_KEY.txt` naming pattern. I do not have the corresponding hardware to perform verification, so I could only refine the relevant parts by referencing the other two architectures. However, as mentioned above, I will continue monitoring any related issues, so the overall risk remains low.

brian6932:
>So is #16452 actually a fix based on #16366? Why do it this way? And what if changes are needed?

>A pull request should focus on a single change IMO.

Originally not, it was just sourceforge, but then I noticed that PR, so I merged the changes into my branch.

z-Fng:
Haha, I think next time we can avoid doing it this way. There aren't many overlapping changes between the two PRs, so any merge conflicts would be easy to resolve. Submitting separate PRs also makes it easier to make adjustments if changes are needed. But it's fine -- it's already merged.

Thanks again for your hard work.

bucket/hwinfo.json
z-Fng: > Additionally, I use this application myself, so I naturally would not allow errors to affect my own experience.

Thank you for taking the time to look into this manifest.

Consider this: if the `Cygwin` manifest did not contain this pattern, you probably wouldn't use it this way either. Once this appears in a manifest, it tends to 'spread', and more manifests will require special tracking over time -- which is something I want to avoid. I'll allow it this time, but I would still reject this type of change next time.

PR #16508, opened 2025-10-31
zed: Add notes regarding disabling the built-in updater

Vixb1122:

## Summary by CodeRabbit

* **Chores**
  * Installer now disables automatic updates by default in the app's settings.
  * Existing user settings are preserved (including comments) while ensuring the update preference is set to off.

### DIFF
--- a/bucket/zed.json
+++ b/bucket/zed.json
@@ -3,6 +3,7 @@
     "description": "Zed is a high-performance, multiplayer code editor from the creators of Atom and Tree-sitter. It's also open source.",
     "homepage": "https://zed.dev/",
     "license": "AGPL-3.0, Apache-2.0, GPL-3.0",
+    "notes": "Please manually disable Zed's auto-update feature in the settings to prevent the updater from installing the program in unintended locations.",
     "architecture": {
         "64bit": {
             "url": "https://github.com/zed-industries/zed/releases/download/v0.220.3/Zed-x86_64.exe",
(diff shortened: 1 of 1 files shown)


bucket/zed.json
```diff
@@ -16,7 +16,29 @@
     "installer": {
         "script": [
             "Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir '{app}'",
-            "Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir '{code_GetInstallDir}' -Removal"
+            "Expand-InnoArchive -Path \"$dir\\$fname\" -ExtractDir '{code_GetInstallDir}' -Removal",
+            "# Turn off Auto Update",
+            "$zedSettingsPath = Join-Path $env:APPDATA 'Zed' 'settings.json'",
```
z-Fng: 1. Please avoid using multiple arguments in `Join-Path` for compatibility with Windows PowerShell 5.1, as it only supports a single ChildPath.
2. Would it be better to move this logic to a separate post_install script?
3. Is it necessary to keep `$commentLines` in the JSON file?

z-Fng:
Since [zed-industries/zed@b71ef54](https://github.com/zed-industries/zed/commit/b71ef540fc0289cb59cdf25fe9733b36ae71c8cf), [`initial_user_settings.json`](https://github.com/zed-industries/zed/blob/main/assets/settings/initial_user_settings.json) contains trailing commas.

PowerShell's native `ConvertFrom-Json` and `ConvertTo-Json` follows strict JSON standards and cannot parse files with trailing commas. Because of this standard mismatch, we cannot reliably read or write this configuration file using built-in tools.

In light of this, I suggest we do not add the script to disable the built-in updater.

Vixb1122:
> Since [zed-industries/zed@b71ef54](https://github.com/zed-industries/zed/commit/b71ef540fc0289cb59cdf25fe9733b36ae71c8cf), [`initial_user_settings.json`](https://github.com/zed-industries/zed/blob/main/assets/settings/initial_user_settings.json) contains trailing commas.
> 
> PowerShell's native `ConvertFrom-Json` and `ConvertTo-Json` follows strict JSON standards and cannot parse files with trailing commas. Because of this standard mismatch, we cannot reliably read or write this configuration file using built-in tools.
> 
> In light of this, I suggest we do not add the script to disable the built-in updater.

replace with notes @z-Fng

PR #16529, opened 2025-11-04
helium@0.5.8.1: Add data migration

fahim-ahmed05:
### Summary
- Migrate userdata from previous non scoop installation.
- Create a junction link to userdata in localappdata, so that if the browser is started from its exe it don't create new userdata folder.
- Removes the junction link when browser is uninstalled.

## Summary by CodeRabbit

* **New Features**
  * Automated user-data migration during installation to preserve existing settings and files by relocating data to the app’s persistent directory, ensuring continuity after upgrade and handling migration errors gracefully.
  * Automatic cleanup during uninstallation to remove legacy filesystem links and reduce residual artifacts, resulting in a cleaner system state after removal.

### DIFF
--- a/bucket/helium.json
+++ b/bucket/helium.json
@@ -56,7 +56,8 @@
     ],
     "post_uninstall": [
         "$appdata = \"$env:LOCALAPPDATA\\imput\\Helium\\User Data\"",
-        "if ((Get-Item $appdata -ErrorAction SilentlyContinue)?.LinkType -eq 'Junction') {",
+        "$item = Get-Item $appdata -ErrorAction SilentlyContinue",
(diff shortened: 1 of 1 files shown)


SorYoshino:
To be honest, I don’t think such logic should be added to the official buckets. Although persistence is one of Scoop’s key features, we should not forcibly persist user data that is not stored within `$dir`. Doing so would greatly increase the complexity of the manifest and could introduce potential issues.

Even though there are manifests in the Extras bucket that contain similar scripts, I still do not support adding this kind of logic to the official buckets. We should only add persistence logic for applications whose maintainers explicitly allow users to change the data storage location—such as through configuration files or environment variables—and that continue to function properly after such modifications.

**The above reflects my personal opinion; others may hold different views.**

PR #16542, opened 2025-11-06
sozi: Update to version 23.7.25, fix checkver & autoupdate

o-l-a-v:
Changes:

* Update to latest stable version.
* Fix checkver, autoupdate and install logix to handle new installer
* Inform of leftover directory `%APPDATA%\sozi` if `uninstall --purge`

Because:

* There is no longer a ZIP included, only an Electron installer.

## Summary by CodeRabbit

* **Chores**
  * Updated to version 23.7.25.
  * Switched to a new 64-bit installer format with updated checksum for streamlined installation.
  * Improved auto-update and release detection by pointing to the official releases endpoint for more reliable updates.

* **New Features**
  * Added post-uninstall cleanup to remove leftover directories when performing a purge.

### DIFF
--- a/bucket/sozi.json
+++ b/bucket/sozi.json
@@ -23,23 +23,19 @@
         ]
     ],
     "post_uninstall": [
-        "if ($purge -and -not $global) {",
-        "    $LeftoverDirectories = [string[]](",
-        "        [System.IO.Path]::Combine($env:APPDATA, 'sozi') | Where-Object -FilterScript {[System.IO.Directory]::Exists($_)}",
+        "if ($purge) {",
+        "    $Directories = [string[]](",
+        "        [System.IO.Path]::Combine($env:APPDATA, 'sozi')",
         "    )",
-        "    if ($LeftoverDirectories.Count -gt 0) {",
-        "        $Message = [string] [System.Environment]::NewLine",
-        "        $Message += 'Found following leftover directories outside Scoop''s control:'",
-        "        $LeftoverDirectories.ForEach{",
-        "            $Message += [System.Environment]::NewLine",
-        "            $Message += '* {0}' -f $_",
+        "    $Directories.ForEach{",
+        "        if ([System.IO.Directory]::Exists($_)) {",
+        "            $null = [System.IO.Directory]::Delete($_,$true)",
         "        }",
-        "        info $Message",
         "    }",
         "}"
     ],
     "checkver": {
-        "url": "https://api.github.com/repos/sozi-projects/Sozi/releases",
+        "url": "https://api.github.com/repositories/940219/releases",
         "jsonpath": "$[?(@.prerelease == false)]",
         "regex": "/v(?<tag>[\\d.]+)/Sozi.Setup.([\\d.]+)-(?<build>[\\d]+).exe"
     },


o-l-a-v:
Relates to:

* <https://github.com/ScoopInstaller/Extras/issues/16379>
* <https://github.com/ScoopInstaller/Scoop/issues/6542>
* <https://github.com/ScoopInstaller/Scoop/issues/6543>

o-l-a-v:
@coderabbitai generate docstrings

bucket/sozi.json
```diff
@@ -17,15 +22,27 @@
...
+        "    $Directories = [string[]](",
+        "        [System.IO.Path]::Combine($env:APPDATA, 'sozi')",
+        "    )",
+        "    $Directories.ForEach{",
+        "        if ([System.IO.Directory]::Exists($_)) {",
+        "            $null = [System.IO.Directory]::Delete($_,$true)",
+        "        }",
+        "    }",
+        "}"
+    ],
```
z-Fng, line 36: Personally, I'd advise against adding any operations that modify files outside the Scoop root directory. Scoop isn't supposed to manage those.

And for global installs, we wouldn't be able to handle them properly anyway.

o-l-a-v: If one allows Scoop to install apps that will add/modify files outside Scoop's directory, I think it's warranted that Scoop cleans it up too.

SorYoshino: > Personally, I'd advise against adding any operations that modify files outside the Scoop root directory. Scoop isn't supposed to manage those.

I agree with this point.

> I think it's warranted that Scoop cleans it up too.

In fact, many software uninstallers do not remove residual files in `$env:APPDATA`.  

Scoop is merely a package manager, and I believe its responsibility should be equal to or less than that of the software’s own uninstaller — that is, it should only operate on files it can actually manage.

o-l-a-v: > In fact, many software uninstallers do not remove residual files in $env:APPDATA.

Yup, unfortunately. An uninstaller should offer the option to clean up everything though, so not having this in an uninstaller is negligence, laziness.

As a user, if I uninstall using `--purge` I expect everything to be removed, so that a reinstall of the app in question will be like a fresh install; no previous config or data. I agree that it should not do it by default, but I think `--purge` should be able to. If Scoop maintainers do not I agree I suggest a new feature to be added to Scoop, maybe as a setting or as a new uninstall option. It's valuable to have a package manager that also handles this IMO.

As a minimum, there should be guidelines that highlight how manifests should handle this.

Edit: Maybe paths outside `%USERPROFILE%\scoop` can be printed to the terminal when `--purge`?

SorYoshino: > It's valuable to have a package manager that also handles this IMO.

I agree with this perspective, but it seems somewhat impractical.

If such functionality were to be introduced into Scoop Core, aside from matching keywords within `$env:APPDATA`, I can hardly think of any feasible approach. However, this method carries the risk of accidentally deleting data belonging to other software or even the system itself — a risk that cannot be entirely eliminated.

Moreover, achieving a truly “clean” reinstall of an application is far from simple. In addition to `$env:APPDATA`, applications may store data or other files in `$env:ProgramData`, `$env:windir`, or even in the Windows Registry. Addressing all of these would open a Pandora’s box of complications. As a package manager, I don’t believe Scoop should pursue such functionality before resolving more fundamental issues.

o-l-a-v:
I changed the manifest to inform about leftovers instead. In my testing it looked like this:

```pwsh
PS > scoop uninstall --purge sozi

Uninstalling 'sozi' (23.7.25).
Removing shortcut ~\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Scoop Apps\Sozi.lnk
Unlinking ~\scoop\apps\sozi\current
Running post_uninstall script...INFO
Found following leftover directories outside Scoop's control:
* C:\Users\olavb\AppData\Roaming\sozi
done.
Removing persisted data.
'sozi' was uninstalled.

PS >
```

Until Scoop eventually get a built in future to do something with leftovers, is this an acceptable approach?

bucket/sozi.json
o-l-a-v: I only thought it'd be feasible for Scoop to look for leftovers outside `%scoop%` if paths were specified in the manifest.

Not that Scoop should automatically try to search for leftovers in file system and registry, that is a new can of worms.

SorYoshino: Alright, I misunderstood your point earlier.

However, as I mentioned before, there are many possible locations where software may store its data. If we intend to output or delete those files, users would need to explicitly list them in the manifest.

This approach makes it possible to completely remove all data generated after installing software via Scoop, not just user data. But whether the goal is to remove only user data or all data, it inevitably increases the manifest’s complexity. Maintainers would have to spend more time verifying these entries and might face more frequent maintenance in the future. This is not good for the already slow-moving community. I remain cautious about this approach.

z-Fng: https://github.com/ScoopInstaller/Scoop/issues/6542#issuecomment-3552065949

PR #16758, opened 2025-12-09
hemmelig: Add version 1.0.0

ltguillaume:
Adds Hemmelig CLI v1.0.0
https://github.com/HemmeligOrg/Hemmelig.app/releases/tag/cli-v1.0.0

## Summary by CodeRabbit

* **Chores**
  * Added a Windows distribution manifest to enable official Windows installs and automatic updates from releases, with versioned downloads and integrity checks to ensure safe, updatable installations.

### DIFF
--- a/bucket/hemmelig.json
+++ b/bucket/hemmelig.json
@@ -1,7 +1,7 @@
 {
     "version": "1.0.0",
     "description": "Share sensitive information securely with client-side encryption and self-destructing messages.",
-    "homepage": "https://github.com/HemmeligOrg/Hemmelig.app",
+    "homepage": "https://hemmelig.app",
     "license": {
         "identifier": "O'Saasy",
         "url": "https://github.com/HemmeligOrg/Hemmelig.app/blob/v7/LICENSE"


z-Fng:
PRs labeled `package-request-needed` are less likely to be reviewed and may be closed if there are no updates for a long time.

Please create a [package request](https://github.com/ScoopInstaller/Extras/issues/new?assignees=&labels=package-request&projects=&template=package-request.yml&title=%5BRequest%5D%3A+) before submitting a PR to add a new package to this bucket. When creating a package request, you can check the **criteria** for a package to be accepted in this bucket, which can help you determine whether your PR is likely to be approved. The information provided in the issue can also help maintainers quickly get the necessary details.

> Extra manifests for [Scoop](https://scoop.sh/), the Windows command-line installer. For manifests that don't fit the [Main criteria](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket).

> ### For a package to be acceptable in this bucket, it should be:
> 1. **Reasonably well-known and widely used. e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks** *
> 2. **English interface (or at least English documentation)** *
> 3. **Latest stable version** *
> 4. Full version (i.e. not a trial version)
> 5. Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

ltguillaume:
Done: #16759

I have made the request in Extras, not Main, because the (only) criterium it does not meet is that it's not a developer tool (in the narrowest sense of the word).

ltguillaume:
@aliesbelik I'm trying to understand the changes made:
1. The license is probably because it's not (yet) a "default" license type, I guess?
2. The architecture wrapper, I can understand. Using just `url:` seems perfectly valid, since there is only one architecture available, but it wouldn't communicate to Scoop _which_ architecture is available.
3. Is there any particular reason to use the repository ID instead of its name? Using the ID is less transparent and _if_ the repo is ever renamed, GitHub should just redirect without any issues.
4. The `hash` seems superfluous to me, because AFAIK Scoop will still calculate and include it in the manifest without it.

aliesbelik:
> 1. The license is probably because it's not (yet) a "default" license type, I guess?
> 2. The architecture wrapper, I can understand. Using just `url:` seems perfectly valid, since there is only one architecture available, but it wouldn't communicate to Scoop _which_ architecture is available.

Exactly. 
For details you could use [manifest properties](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests) description.

> 3. Is there any particular reason to use the repository ID instead of its name? Using the ID is less transparent and _if_ the repo is ever renamed, GitHub should just redirect without any issues.

Redirect works for the repo GH link but not for the API endpoint (wonder if it was already changed/fixed): https://github.com/ScoopInstaller/Main/pull/5215#issuecomment-1778915170.

> 4. The `hash` seems superfluous to me, because AFAIK Scoop will still calculate and include it in the manifest without it.

[Adding hash to autoupdate](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#adding-hash-to-autoupdate).
First `autoupdate.hash` section (if specified), then github-generated digests for release assets (if available), only then - downloading file(s) and calculating hash(es).

ltguillaume:
@aliesbelik Thanks for taking the time for explaining this. I'll go over [my own bucket](https://github.com/ltguillaume/schep) and see if it needs adjustments and I'll make sure to implement this in any future PRs.

bucket/hemmelig.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "1.0.0",
+    "description": "Share sensitive information securely with client-side encryption and self-destructing messages.",
+    "homepage": "https://github.com/HemmeligOrg/Hemmelig.app",
```
z-Fng: 1. Should the homepage be updated to https://hemmelig.app ?
```suggestion
    "homepage": "https://hemmelig.app",
```
~~2. Would it make more sense to rename the manifest to `hemmelig-cli.json`?~~

ltguillaume: I don't know, should it link to the project's homepage with a description/help/contact/issues (i.e. GitHub), or a deployed instance of the software (i.e. hemmelig.app)?

I thought the first would make more sense, but I'm not sure now about what the convention is for webapps.

z-Fng: I'd lean more toward linking to the deployed instance, as it provides immediate access to the actual service for users.

ltguillaume: All right, if most/more manifests do this, then let's change it.

z-Fng:
> Redirect works for the repo GH link but not for the API endpoint (wonder if it was already changed/fixed): [ScoopInstaller/Main#5215 (comment)](https://github.com/ScoopInstaller/Main/pull/5215#issuecomment-1778915170).

It looks like redirects work for the GitHub API as well now.
e.g.,
- http://api.github.com/repos/th-ch/youtube-music/releases/latest

ltguillaume:
> > Redirect works for the repo GH link but not for the API endpoint (wonder if it was already changed/fixed): [ScoopInstaller/Main#5215 (comment)](https://github.com/ScoopInstaller/Main/pull/5215#issuecomment-1778915170).
> 
> It looks like redirects work for the GitHub API as well now. e.g.,

Good to know! Personally, I think using the repo ID is less user friendly/transparent, so I'd opt for changing this policy in Scoop manifests.

z-Fng:
Exactly. I'm also wondering if we can use the repo name directly in the GitHub API now, instead of the repository ID. It would be much more intuitive. We could even go a step further and use CI to fix/correct the repo links.

ltguillaume:
Related to this PR, I've also been looking into how to deprecate the manifest in my own bucket, ideally making it "transition" to using the Extras bucket. I can't find anything actionable about this in the documentation, other than that I can move my manifest into the `deprecated` folder and Scoop will notify the user _only if they manually check via `scoop status` or `scoop list`_. There is no warning during an update and as far as I can see, no update will be performed if another bucket does supply an update for the same package if the bucket of origin has deprecated the package (of course this can cause issues).

1. Would there be a notification if there's a new version in the Extras manifest, even if it was originally installed from another bucket? (I'm afraid not)
2. I'm guessing the only way to transition to the Extras bucket is to uninstall, then reinstall, or to change the bucket in `apps\hemmelig\install.json`?

I'm asking because this may always happen, in any bucket, without the user really noticing.

z-Fng:
> 1. Would there be a notification if there's a new version in the Extras manifest, even if it was originally installed from another bucket? (I'm afraid not)

No, there won't be.

> 2\. I'm guessing the only way to transition to the Extras bucket is to uninstall, then reinstall, or to change the bucket in `apps\hemmelig\install.json`?

Exactly. We are lacking the mechanisms to remove or replace a manifest now. See: https://github.com/ScoopInstaller/Main/pull/7524#issuecomment-3753899493 As a workaround, we typically add a deprecation notice to the existing manifest and then deprecate it in a few months. e.g., https://github.com/ScoopInstaller/Extras/pull/17040

ltguillaume:
Got it. Maybe the smallest change to really improve things would be to list deprecated packages (just like with `status`) on every `scoop update` underneath the commits listing. It would at least alert users properly, because `status` and `list` simply won't be triggered (regularly) by most users.

PR #16190, opened 2025-09-21
servy: Add version 1.7

aelassas:
Add Servy version 1.7.

## Summary by CodeRabbit

* **New Features**
  * Adds a Windows installer manifest for Servy with versioned downloads and SHA256 integrity checks.
  * Installs desktop and Start Menu shortcuts for Servy and Servy Manager.
  * Enables automatic update checks and streamlined autoupdates for 64‑bit installations.
  * Supports Inno Setup–based installation and maps the command-line tool for immediate use.

PR #16795, opened 2025-12-16
fix(servy): switch to installer-based manifest to avoid innounp incompatibility

aelassas:
This PR switches Servy to an installer-based Scoop manifest to avoid incompatibility with `innounp` and modern Inno Setup versions.

After releasing Servy 4.0, the current manifest fails during installation with the following error:
```
innounp - the Inno Setup Unpacker, Version 2.65.2 (11/09/2025)

Inno Setup archive:           servy-4.0-x64-installer.exe
=> Signature detected: Inno Setup Setup Data (6.6.1)
This is not directly supported, but i'll try to unpack it as version 6502
Inno Setup version detected:  6.6.1 (Unicode)
Critical error: The setup files are corrupted. Please obtain a new copy of the program.
```

This error is caused by `innounp` not fully supporting Inno Setup 6.6.x.

The installer itself is valid and signed.

This PR resolves the issue by:

* Switching from `innounp` extraction to an installer-based manifest
* Using silent install and uninstall flags
* Supporting multiple Inno Setup uninstaller filenames (`unins*.exe`)

This PR restores successful installation via Scoop. Without it, users cannot install Servy.

Related issue: https://github.com/aelassas/servy/issues/28

## Summary by CodeRabbit

* **New Features**
  * Added structured installer and uninstaller support with pre-install checks, silent-install arguments, and an automated uninstaller flow.
  * Added user-visible notes describing installer behavior and uninstallation.

* **Improvements**
  * Replaced legacy installer flag with a clearer manifest layout for installer/uninstaller configuration.
  * Updated product description wording to reference pre-launch and post-launch scripts and parameters.
  * Retained existing update/version retrieval behavior.

### DIFF
--- a/bucket/servy.json
+++ b/bucket/servy.json
@@ -9,8 +9,16 @@
             "hash": "1126aaecb151fe91140c39e247631c1043700bc8733b7a3b48e18525e124c578"
         }
     },
+    "notes": [
+        "Installer requires admin privileges.",
+        "Installs into Scoop's managed folder ($dir).",
+        "Installer updates system PATH for global CLI access.",
+        "Supports clean uninstallation via unins*.exe."
+    ],
+    "pre_install": "if (-not (is_admin)) { throw 'This package requires admin privileges to install' }",
     "installer": {
         "args": [
+            "/DIR=\"$dir\"",
             "/VERYSILENT",
             "/SUPPRESSMSGBOXES",
             "/NORESTART"
@@ -18,7 +26,7 @@
     },
     "uninstaller": {
         "script": [
-            "$installDir = Join-Path $env:ProgramFiles 'Servy'",
+            "$installDir = $dir",
             "if (Test-Path $installDir) {",
             "    $uninstaller = Get-ChildItem $installDir -Filter 'unins*.exe' -ErrorAction SilentlyContinue |",
             "                                 Sort-Object LastWriteTime -Descending |",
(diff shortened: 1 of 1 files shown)


z-Fng:
https://github.com/ScoopInstaller/Extras/pull/16190#discussion_r2385189260

aelassas:
@z-Fng Thank you for the clarification.

The problem is that **Servy 4.0 installer** is not fully compatible with the InnoSetup extractor used by Scoop. Servy 4.0 installer is signed, the extractor produces the following error:

```
=> Signature detected: Inno Setup Setup Data (6.6.1)
This is not directly supported, but i'll try to unpack it as version 6502
Inno Setup version detected:  6.6.1 (Unicode)
Critical error: The setup files are corrupted. Please obtain a new copy of the program.
```

This causes the installation to **fail completely** when Scoop tries to extract it without running the full installer.

Previous installers (not signed) work.

Because of this, the **only reliable way to install Servy is to run the full installer**, which supports silent flags (`/VERYSILENT`, `/SUPPRESSMSGBOXES`, `/NORESTART`) and allows us to:

* Respect `$dir` for Scoop-managed installation
* Handle multiple uninstaller filenames (`unins*.exe`)
* Ensure shortcuts and CLI (`bin`) are correctly registered
* Require admin as needed for system-wide installation

In short, using the extractor is not an option due to **the error of signed inno setup installer**, and running the installer is the only way to guarantee a successful, fully functional installation.

aelassas:
@z-Fng After running `scoop update innounp` then `scoop install servy`, It worked. I've updated the Scoop instructions in the README.md and the wiki. 

Do you know if Scoop plans to include an updated innounp in a future release, so users don't have to update it manually?

z-Fng:
>  I've updated the Scoop instructions in the README.md and the wiki.

Thank you for your understanding.

> Do you know if Scoop plans to include an updated innounp in a future release, so users don't have to update it manually?

I'm currently working on implementing this feature and may include it in the next version. See also: https://github.com/ScoopInstaller/Scoop/issues/6551

PR #17004, opened 2026-01-14
mu-editor@1.2.0: Improve shortcut

### LINKED ISSUE #17002: [Improvement] mu-editor: Bypass VBS Script for Start Menu Shortcut
The start menu shortcut created by the `mu-editor` manifest currently points to a VBS script that simply runs the command `"[Directory of script]\Python\pythonw.exe" -I -m mu`. We can therefore change the shortcut to run this command directly if we want to avoid adding an unnecessary step to each launch of the program (in fact, the shortcut that is created when installing Mu traditionally through the MSI installer is set up to run this specific command directly rather than point to the VBS script). This will also remove or reduce dependency on file associations relating to `.vbs` scripts.

bc-185:
- The Start Menu Shortcut that is created when installing the `mu-editor` manifest points to a VBS script that is set up to run a python command
- By changing it to run this command directly (like the shortcut created when running the MSI installer), we can improve startup times by bypassing the script entirely, and make the install less dependent on `.vbs` file associations.

Closes #17002

## Summary by CodeRabbit

* **Chores**
  * Updated Mu Editor launcher configuration to execute through Python directly with optimized parameters, replacing the previous batch file approach.

### DIFF
--- a/bucket/mu-editor.json
+++ b/bucket/mu-editor.json
@@ -19,9 +19,9 @@
     "extract_dir": "Mu Editor",
     "shortcuts": [
         [
-            "Mu Editor.vbs",
+            "Python\\pythonw.exe",
             "Mu Editor",
-            "",
+            "-I -m mu",
             "icon.ico"
         ]
     ],


z-Fng:
> * By changing it to run this command directly (like the shortcut created when running the MSI installer), we can improve startup times by bypassing the script entirely, and make the install less dependent on `.vbs` file associations.

The downside of not using the provided .vbs file is that the manifest has to manage the launch arguments. If they change, we'll have to update the manifest accordingly. However, since [`mu-editor`](https://github.com/ScoopInstaller/Extras/blob/master/bucket/mu-editor.json) hasn't been updated in a long time, this approach should be fine.

z-Fng (approved):
Thanks for your contribution!

PR #17292, opened 2026-02-26
aida64business: Add version 8.25.8200

### LINKED ISSUE #17271: [Request]: aida64business
### Prerequisites

- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [x] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 100 stars and/or 50 forks)
- [x] English interface (or at least English documentation)
- [x] Latest stable version
- [ ] Full version (i.e. not a trial version)
- [ ] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

aida64business

### Description

System information, diagnostics, and auditing application with extra scripting capabilities.

### Homepage

https://www.aida64.com/products/aida64-business

### Download Link(s)

https://download.aida64.com/aida64business825.zip

A lot of pro-users leverage AIDA64 to monitor their computers.

Compared to the Extreme version, 
https://github.com/ScoopInstaller/Extras/blob/master/bucket/aida64extreme.json
the Business version can be used in scripting scenarios.

Let's say start a FPU stress test for 30min
`aida64 /sst fpu /sstdur 30`
or run a memory & cache bandwith test
`aida64 /CMBENCH memory-cache`

Detailed comparison can be found here, the aida64business is basically **the FULL FAT version of AIDA64** (of course you need a license but that is the same for aida64extreme).
https://www.aida64.com/compare-aida64-features
And the CLI documentation is also available for those interested.
https://aida64.co.uk/user-manual/help/command-line-options?p=business

deuteros-gex:
Closes #17271

The manifest is basically the same as the extreme version, where the major difference is the following line so that the binary also gets its own shim.
`    "bin": "aida64.exe",`

## Summary by CodeRabbit

* **New Features**
  * AIDA64 Business can be installed and managed via the package system with automatic version detection and autoupdates.
  * Installer creates missing configuration, license/key, and auxiliary files during setup.
  * Application settings and user files are preserved across installs and updates.
  * Desktop shortcut and executable are configured automatically during installation.

### DIFF
--- a/bucket/aida64business.json
+++ b/bucket/aida64business.json
@@ -6,8 +6,19 @@
         "identifier": "Shareware",
         "url": "https://www.aida64.com/licensing"
     },
-    "url": "https://download.aida64.com/aida64business825.zip",
-    "hash": "03cb63dbd707896f7c1c221c0e1c0b50ab7a55c7bb8ead865f90932f73b87b5a",
+    "architecture": {
+        "64bit": {
+            "url": "https://download.aida64.com/aida64business825.zip",
+            "hash": "03cb63dbd707896f7c1c221c0e1c0b50ab7a55c7bb8ead865f90932f73b87b5a"
+        }
+    },
+    "bin": "aida64.exe",
+    "shortcuts": [
+        [
+            "aida64.exe",
+            "AIDA64 Business"
+        ]
+    ],
     "pre_install": [
         "$FILE = 'aida64.ini'",
         "if (!(Test-Path \"$persist_dir\\$FILE\")) {",
@@ -24,18 +35,15 @@
         "pkey.txt",
         "aida64.ini"
     ],
-    "bin": "aida64.exe",
-    "shortcuts": [
-        [
-            "aida64.exe",
-            "AIDA64 Business"
-        ]
-    ],
     "checkver": {
         "url": "https://www.aida64.com/downloads/latesta64be",
         "regex": "AIDA64 Business\\s+([0-9]+\\.[0-9]+\\.[0-9]+)"
     },
     "autoupdate": {
-        "url": "https://download.aida64.com/aida64business$majorVersion$minorVersion.zip"
+        "architecture": {
+            "64bit": {
+                "url": "https://download.aida64.com/aida64business$majorVersion$minorVersion.zip"
+            }
+        }
     }
 }


bucket/aida64business.json
```diff
@@ -0,0 +1,41 @@
...
+        "}"
+    ],
+    "persist": [
+        "Reports",
+        "aida64.reg.ini",
+        "aida64.sst.ini",
+        "pkey.txt",
+        "aida64.ini"
+    ],
+    "bin": "aida64.exe",
```
z-Fng: > where the major difference is the following line so that the binary also gets its own shim.
> ` "bin": "aida64.exe",`

https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md#for-scoop-buckets

> If the program file is a GUI application and it doesn't accept any commandline arguments, no need to add it in bin.

Does the GUI binary `aida64.exe` accept command-line arguments? Generally, we don't think it's necessary to create a shim if the binary doesn't accept any commandline arguments.

deuteros-gex: Actually the GUI binary does support CLI arguments, that is a major difference between aida64**Business** and aida64**Extreme**

Detailed comparison can be found here.
https://www.aida64.com/compare-aida64-features
And the CLI documentation is also available for those interested.
https://aida64.co.uk/user-manual/help/command-line-options?p=business

One can easily leverage the CLI capabilities in aida64business. Examples can be found #17271 

Yes, once the GUI binary gets CLI arguments from a terminal, it will still launch its GUI ,but behave as the CLI ordered it to, totally manipulable and usable for scripting scenarios.

bucket/aida64business.json
```diff
@@ -0,0 +1,49 @@
...
+    "homepage": "https://www.aida64.com/products/aida64-business",
+    "license": {
+        "identifier": "Shareware",
+        "url": "https://www.aida64.com/licensing"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://download.aida64.com/aida64business825.zip",
+            "hash": "03cb63dbd707896f7c1c221c0e1c0b50ab7a55c7bb8ead865f90932f73b87b5a"
+        }
```
z-Fng, line 13: https://www.aida64.com/downloads/latesta64xe

> AIDA64 v8.xx officially supports only 64-bit Windows 10 or newer operating systems.

z-Fng (approved):
Thanks for your contribution!

The manifest will update automatically as the package updates, as long as the upstream releases new packages that follow the same pattern defined in the manifest. If there are any other changes, feel free to submit a PR.

PR #17705, opened 2026-04-29
aegisub: Persist `config.json`

emotion3459:
# Summary
Aegisub uses config.json to store misc settings. The manifest needs to persist this for most settings to carry over versions. We do not need to create the file ourselves as the portable zip comes with a pre-configured config.

### DIFF
--- a/bucket/aegisub.json
+++ b/bucket/aegisub.json
@@ -2,7 +2,13 @@
     "version": "3.4.2",
     "description": "Cross-platform advanced subtitle editor",
     "homepage": "https://www.aegisub.org/",
-    "license": "https://github.com/TypesettingTools/Aegisub/blob/master/LICENCE",
+    "license": {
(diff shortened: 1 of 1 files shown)


bucket/aegisub.json
```diff
@@ -20,7 +20,8 @@
         "autoback",
         "autosave",
         "catalog",
-        "config"
+        "config",
+        "config.json"
```
z-Fng: In my tests, hard links don't work. When Aegisub modifies the configuration, it appears to recreate the file rather than writing to the existing configuration file.

z-Fng (changes requested):

bucket/aegisub.json
emotion3459: I copied the approach from [ddu](https://github.com/ScoopInstaller/Extras/blob/master/bucket/ddu.json) and quick testing shows it seems to work.

z-Fng (approved):
Thanks for your contribution!

PR #17732, opened 2026-05-01
azuredatastudio: Deprecate manifest

notyusufkhalifa:
This PR removes Azure Data Studio as it was retired by Microsoft in [here](https://learn.microsoft.com/en-us/sql/tools/whats-happening-azure-data-studio?view=sql-server-ver17&tabs=dev)

- [X] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`


z-Fng:
We generally do not deprecate a manifest as long as the binaries are still available to download.

PR #17830, opened 2026-05-19
lenovolegiontoolkit: Switch upstream to `SSC-STUDIO/LenovoLegionToolkit`

SSC-STUDIO:
## Summary
- update lenovolegiontoolkit to 3.6.16
- switch homepage and download source to SSC-STUDIO/LenovoLegionToolkit
- update package hash to the 3.6.16 release asset
- align suggested runtime to .NET Desktop Runtime 10.0


z-Fng:
Please do not modify the upstream of an existing package unless the original upstream explicitly specifies a new repository.

PR #17865, opened 2026-05-23
soundswitch@7.1.0: Fix file naming issue, add arm64 support

AndeFriday:
The SoundSwitch v7.1.0 installer uses suffixes for filenames (`,1` for x64 and `,2` for ARM64).

This PR optimizes the manifest by:
1. Moving common `url` and `hash` to the root (as both architectures use the same installer).
2. Adding explicit `64bit` and `arm64` architecture support.
3. Implementing `pre_install` scripts to clean up incompatible binaries and rename the correct ones by removing architecture suffixes.

This ensures proper installation and shortcut creation on both x64 and ARM64 Windows systems.

- [x] Use conventional PR title: `soundswitch: fix file naming issue and add arm64 support in v7.1.0`

### DIFF
--- a/bucket/soundswitch.json
+++ b/bucket/soundswitch.json
@@ -3,10 +3,20 @@
     "description": "Switch your default playback devices and/or recording devices using simple hotkeys",
     "homepage": "https://soundswitch.aaflalo.me/",
     "license": "GPL-2.0-or-later",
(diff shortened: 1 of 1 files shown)


z-Fng (approved):
Thanks for your contribution!

PR #17868, opened 2026-05-23
clashtui@0.3.0: Add arm64 support

JohanChane:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 on:
-  pull_request:
+  pull_request: read
   push:
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 on:
-  pull_request:
+  pull_request: read
```
z-Fng: This is clearly wrong. Why was this change made?

z-Fng (approved):
Thanks for your contribution!

.github/workflows/ci.yml
JohanChane: I don't know why this error occurred. So I clicked the `Explain error` button, and it suggested making changes this way. I found it very weird at the time, but since it was only on my forked repository, I decided to give it a try.

PR #18014, opened 2026-06-10
gui-for-singbox@1.25.1: Add libcronet.dll at installation

### LINKED ISSUE #17911: gui-for-singbox: Automatically download libcronet.dll during installation for NaiveProxy support
**Title: gui-for-singbox: Automatically download libcronet.dll during installation for NaiveProxy support**

## Description

When installing `gui-for-singbox` via Scoop and using the **NaiveProxy** outbound, sing-box fails to start with the following fatal error:

```
FATAL[0000] create service: initialize outbound[8]: cronet: library not found. Place libcronet.dll in the executable directory or PATH
```

NaiveProxy requires `libcronet.dll` to be present in the same directory as `sing-box.exe`. Currently, users must manually download it from an external source and place it in the correct directory — and repeat this process after every `scoop update`.

## Feature Request

I would like the Scoop manifest for `gui-for-singbox` to handle `libcronet.dll` automatically, ideally in one of the following ways:

1. **Auto-download during install** — include `libcronet.dll` as an additional download in the manifest (architecture-aware: `amd64` or `arm64`) and place it next to `sing-box.exe`
2. **Optional download via post-install script** — prompt the user during installation asking whether they want to download `libcronet.dll` for NaiveProxy support
3. **Persist across updates** — if bundling is not preferred, at minimum use Scoop's `persist` feature so the file is not wiped on `scoop update`

## Why This Matters

- NaiveProxy is a commonly used outbound in sing-box for traffic obfuscation
- There is no indication during installation that this dependency is missing until the user tries to use NaiveProxy and the core crashes
- The correct `libcronet.dll` file must match the user's architecture, which Scoop already handles well via `architecture` fields in manifests

## libcronet.dll Source

The library can be downloaded from the official sing-box-related releases:
- `libcronet-windows-amd64.dll` → rename to `libcronet.dll`
- `libcronet-windows-arm64.dll` → rename to `libcronet.dll`

Available at: https://github.com/SagerNet/cronet-go/releases/

## Environment

- OS: Windows (amd64)
- Installed via: Scoop (Extras bucket)
- App: `gui-for-singbox` (latest)

## References

- sing-box NaiveProxy docs: https://sing-box.sagernet.org/configuration/outbound/naive/

arvdk:
Closes #17911

Uses second download link for the download of libcronet.dll. This link is updated using checkver & autoupdate (but probably only when the version of gui-for-singbox itself changes). 

### DIFF
--- a/bucket/gui-for-singbox.json
+++ b/bucket/gui-for-singbox.json
@@ -5,16 +5,28 @@
     "license": "GPL-3.0-or-later",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/GUI-for-Cores/GUI.for.SingBox/releases/download/v1.25.1/GUI.for.SingBox-windows-amd64.zip",
-            "hash": "614da937a831d98ca08dd26ea10f8f0bd6218df74198f19183de2ba675999a27"
+            "url": [
+                "https://github.com/GUI-for-Cores/GUI.for.SingBox/releases/download/v1.25.1/GUI.for.SingBox-windows-amd64.zip",
+                "https://github.com/SagerNet/cronet-go/releases/download/v148.0.7778.96-1/libcronet-windows-amd64.dll#/libcronet.dll"
+            ],
+            "hash": [
+                "614da937a831d98ca08dd26ea10f8f0bd6218df74198f19183de2ba675999a27",
+                "c7434cfa93c3041321dd19111c4de6c52b8a9531a65661ba45425d3c51ec69e2"
+            ]
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


z-Fng:
> Windows: `libcronet.dll` must be in the same directory as sing-box.exe or in a directory listed in PATH

I guess this won't work. The [documentation](https://sing-box.sagernet.org/configuration/outbound/naive/#structure) states that libcronet.dll must be in the same directory as `sing-box.exe` or in a directory listed in the PATH. However, this change would download libcronet.dll to the same directory as `GUI.for.SingBox.exe` instead.

Furthermore, this approach is quite unreasonable: adding cross-repository download links for `libcronet.dll` breaks maintenance guidelines, introduces security risks, and forces an optional dependency on all users.

PR #17288, opened 2026-02-26
office-tool-plus: Update to version 11.0.27.0, drop 32bit support

b4imetu:
- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md) 

## Summary by CodeRabbit

* **Updates**
  * Software version upgraded to 11.0.27.0
  * Homepage URL updated to the new official site
  * Removed the "suggest" section from the package
  * 64‑bit and ARM64 builds refreshed with new packages and hashes
  * 32‑bit (x86) support removed
  * Autoupdate metadata cleaned to reflect current architectures only

Issue #16577, opened 2025-11-11
[Improvement]: Specify exact version when suggesting .NET runtime


z-Fng:
Currently, manifests that recommend a .NET runtime LTS, or tags like current or preview, point to moving targets. These targets change whenever a new .NET LTS or preview release is published.

As a result, once the referenced version advances, the original recommendation may no longer be valid. Applications built against a previous runtime are not guaranteed to be forward compatible with the newer runtime versions.

> Some manifests that depend on `windowsdesktop-runtime-lts` are actually referring to version 6.0, since they weren't updated when the LTS version changed last time.

To ensure consistency and prevent unexpected compatibility issues, it's recommended to specify an exact version (e.g., windowsdesktop-runtime-8.0) instead of using rolling tags such as 'lts'.

If there are no specific reasons for not doing this previously, I plan to start migrating to fixed-version references.

Lutra-Fs:
I agree with your point. However, would this potentially introduce additional bucket dependencies? I recall being advised previously not to depend on manifests outside the `main` and `extras` buckets (though I can't remember which PR or issue that was discussed in).

SorYoshino:
If the upstream provides installation packages that include the corresponding dependencies, modifying the manifest’s download link to use those packages may be the most compatible approach, even though it would consume more storage space.

If the upstream does not provide such installation packages, using specific dependency versions is undoubtedly the most compatible method. This inevitably introduces dependencies on other buckets, but I believe this is acceptable, as those dependencies are essential runtime environments. It should be regarded as a special case. For example, `office-tool-plus` requires `.NET Desktop Runtime 8.0` to run. When the `.NET Desktop Runtime` version on my computer was 9.0, attempting to launch `office-tool-plus` prompted me to install `.NET Desktop Runtime 8.0`.

Currently, the `windowsdesktop-runtime-lts` manifest in the Extras bucket has not been automatically updated to version 10.0 for the following reason. However, once it is updated to the latest version, continuing with the current approach would cause newly installed users to encounter the same issue I mentioned above — even if they install the suggested dependency. This is precisely the situation we should aim to prevent.

```powershell
┏[ D:\Software\Scoop\Local\apps\scoop\current\bin][ master ≡]
└─> .\checkver.ps1 -App windowsdesktop-runtime-lts -Dir "D:\Software\Scoop\Local\buckets\Extras\bucket"
[...]

PR #18173, opened 2026-06-28
office-tool-plus@11.5.7.0: Switch to framework-dependent version

### LINKED ISSUE #18172: [Enhancement]: office-tool-plus: Use Non-Runtime Package and Suggest Windows Desktop Runtime
## Package

`office-tool-plus`

## Bucket

`extras`

## Motivation

The current `office-tool-plus` manifest uses a package variant that includes the .NET Windows Desktop Runtime. Since Scoop already provides runtime packages separately, it would be better to use the Office Tool Plus package variant that does not bundle the runtime.

This would reduce duplicated files and better align with Scoop’s philosophy of reusing shared packages instead of bundling dependencies inside every app package.

## Proposed change

Update the `office-tool-plus` manifest to:

- Use the package variant without the bundled .NET runtime.
- Add `versions/windowsdesktop-runtime-10.0` as a suggested package.

ENDlezZenith:
Switch `office-tool-plus` to the package variant that does not bundle the .NET Windows Desktop Runtime, and add `versions/windowsdesktop-runtime-10.0` as a suggested runtime package, as this reuses runtime and aligns with the philosophy of other manifests in scoop.

Closes #18172

### DIFF
--- a/bucket/office-tool-plus.json
+++ b/bucket/office-tool-plus.json
@@ -30,10 +30,10 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/YerongAI/Office-Tool/releases/download/v$version/Office_Tool_with_runtime_v$version_x64.7z"
+                "url": "https://github.com/YerongAI/Office-Tool/releases/download/v$version/Office_Tool_v$version_x64.zip"
             },
             "arm64": {
-                "url": "https://github.com/YerongAI/Office-Tool/releases/download/v$version/Office_Tool_with_runtime_v$version_arm64.7z"
+                "url": "https://github.com/YerongAI/Office-Tool/releases/download/v$version/Office_Tool_v$version_arm64.zip"
             }
         }
     }
--- a/bucket/office-tool-plus.json
+++ b/bucket/office-tool-plus.json
@@ -3,13 +3,16 @@
     "description": "A powerful and useful tool for Office deployments",
     "homepage": "https://www.officetool.plus",
     "license": "GPL-3.0-or-later",
+    "suggest": {
+        ".NET Desktop Runtime 10.0": "versions/windowsdesktop-runtime-10.0"
+    },
     "architecture": {
         "64bit": {
-            "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_with_runtime_v11.5.7.0_x64.7z",
+            "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_v11.5.7.0_x64.zip",
             "hash": "0ea722c9a646006cfe645ae464f3f9a052110d5f444a0e1166996c2fef791a85"
         },
         "arm64": {
-            "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_with_runtime_v11.5.7.0_arm64.7z",
+            "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_v11.5.7.0_arm64.zip",
             "hash": "457f41128110509c474b81bd464d4db9c6a14d580f84779a0cb2c178a41df969"
         }
     },
--- a/bucket/office-tool-plus.json
+++ b/bucket/office-tool-plus.json
@@ -9,11 +9,11 @@
     "architecture": {
         "64bit": {
             "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_v11.5.7.0_x64.zip",
-            "hash": "0ea722c9a646006cfe645ae464f3f9a052110d5f444a0e1166996c2fef791a85"
+            "hash": "cb0fd9db95d0ffd73c642eb29ac24964e51dc651ac3762ae56f2521bad8480d8"
         },
         "arm64": {
             "url": "https://github.com/YerongAI/Office-Tool/releases/download/v11.5.7.0/Office_Tool_v11.5.7.0_arm64.zip",
-            "hash": "457f41128110509c474b81bd464d4db9c6a14d580f84779a0cb2c178a41df969"
+            "hash": "65e6184857edb01365cd42b64f134cc71e29195286ccd615d79d45d99cc34a30"
         }
     },
     "extract_dir": "Office Tool",


SorYoshino:
> I suggest either specifying the required .NET version, or switching to .NET self-contained releases.
> 
> * [[Improvement]: Specify exact version when suggesting .NET runtime #16577](https://github.com/ScoopInstaller/Extras/issues/16577)
> 
> Edit: https://github.com/YerongAI/Office-Tool/releases/tag/v11.0.27.0
> 
> > We recommend that you download the package that includes the runtime for easier use.

- https://github.com/ScoopInstaller/Extras/pull/17288#discussion_r2857109073

---

> This would reduce duplicated files and better align with Scoop’s philosophy of reusing shared packages instead of bundling dependencies inside every app package.

[The official Scoop wiki](https://github.com/ScoopInstaller/Scoop/wiki/Dependencies) doesn't seem to mention this.

In my opinion, using the self-contained version directly is more in line with the wiki.

> Apps are self-contained units
> 
> They should keep their own copies of any libraries they need to run, and not rely on or interfere with any libraries outside their own install path.

If upstream provides a self-contained package with the required dependencies included, I believe we should prioritize that option. The package size of self-contained builds is usually not significantly larger than builds without bundled dependencies, which is generally not a major concern on modern systems. This approach also avoids relying on the `depends` or `suggest` fields in manifests and helps prevent either forced installation or requiring users to install dependencies manually.

If upstream does not provide a self-contained package for specific dependencies, I believe we should prioritize using the `suggest` field in the manifest. Since most manifests outside the Nonportable bucket are installed by extraction rather than through a traditional installer, their dependencies are generally runtime dependencies rather than installation-time dependencies. In most cases, the `suggest` field is sufficient to address this, and this is also a [commonly adopted approach across the buckets](https://github.com/search?q=org%3AScoopInstaller+%22suggest%22%3A&type=code).

Finally, there is the `depends` field, which is generally intended for [installation-time dependencies](https://github.com/ScoopInstaller/Nonportable/pull/536#discussion_r2706218240). I believe this corresponds to what you referred to as “bundling dependencies inside every app package”.

ENDlezZenith:
> > * [[Improvement]: Specify exact version when suggesting .NET runtime #16577](https://github.com/ScoopInstaller/Extras/issues/16577)
> 
> In my opinion, using the self-contained version directly is more in line with the wiki.
> 
> > Apps are self-contained units
> > They should keep their own copies of any libraries they need to run, and not rely on or interfere with any libraries outside their own install path.

While dotnet runtime is also advised as a special case in the same issue you mentioned and I quote:

> However, windowsdesktop-runtime(-lts) is typically only suggested and is not listed as a direct dependency, as it may not be present on the user's system via Scoop.

dotnet runtimes are just too common on users’ system, you may even have tens of them on your system, not to mention that users who are using scoop are more likely to be familiar with their computers, just advising prioritizing packages with runtime might just be an overkill. Like you would not assume a user that is new to computers to use Linux as their first operating system. Linux itself is a barrier to newcomers and Scoop itself is also a barrier for users who cannot manage their own environment.

> > This would reduce duplicated files and better align with Scoop’s philosophy of reusing shared packages instead of bundling dependencies inside every app package.
> 
> [The official Scoop wiki](https://github.com/ScoopInstaller/Scoop/wiki/Dependencies) doesn't seem to mention this.

This is not just restricted to Scoop, do you see any package manager solution that distribute a package that includes another package in itself rather than list it in the dependency? Package manger itself is designed for users to **clearly manage every single package on their system**. Where is the versatile nature of package manager now that you leave runtimes everywhere on the system and you have no way to manage and monitor them? Are you prioritizing using Scoop just because the packages installed this way is easy to use?

> Finally, there is the `depends` field, which is generally intended for [installation-time dependencies](https://github.com/ScoopInstaller/Nonportable/pull/536#discussion_r2706218240). I believe this corresponds to what you referred to as “bundling dependencies inside every app package”.

What I refer to as "bundling dependencies inside every app package" just means using the package with runtime.

SorYoshino:
> do you see any package manager solution that distribute a package that includes another package in itself rather than list it in the dependency?

First, it should be clarified that whether a package manager selects a package with bundled runtimes as the download source depends entirely on what upstream provides. If upstream does not offer a package without bundled runtimes, then the self-contained package becomes the only available option.

Returning to the case where upstream provides both variants, I would still recommend using the self-contained version. I understand your point, but as you mentioned, some users choose Scoop simply for convenience, including those who may not manage their environment well. I believe this group represents the majority of users, and using packages with bundled runtimes directly can avoid unnecessary issues.

> Are you prioritizing using Scoop just because the packages installed this way is easy to use?

I use Scoop because it stores application data under `$persist_dir` and does not write unnecessary entries into the registry. This allows me to quickly restore my working environment after reinstalling the system and manage installed software more effectively. At the same time, I use Winget to manage applications that are less suitable for installation through Scoop, such as the runtimes mentioned in this PR.

The runtimes included in self-contained packages do not conflict with separately installed versions, and I personally do not mind using a few hundred additional megabytes of disk space.

ENDlezZenith:
> First, it should be clarified that whether a package manager selects a package with bundled runtimes as the download source depends entirely on what upstream provides. If upstream does not offer a package without bundled runtimes, then the self-contained package becomes the only available option.

Sure, there is no doubt on this, and vice versa. For example, `g-helper` only provides a package without bundled runtimes, then the non-runtime package becomes the only available option, so we also suggests dotnet runtime 8 in that manifest.

> Returning to the case where upstream provides both variants, I would still recommend using the self-contained version. I understand your point, but as you mentioned, some users choose Scoop simply for convenience, including those who may not manage their environment well. I believe this group represents the majority of users, and using packages with bundled runtimes directly can avoid unnecessary issues.

> The runtimes included in self-contained packages do not conflict with separately installed versions, and I personally do not mind using a few hundred additional megabytes of disk space.

Interesting that now you changed your objective view and I quote your own words in the previous issue:

> For software unrelated to runtime libraries, I believe we should not use “more widely used” as the criterion for inclusion in the Extras bucket, since the concept of “wider usage” is difficult to define objectively. For example, if the user base for two versions is roughly equal, this standard becomes meaningless, and other criteria should be used instead. Purely relying on user numbers is also difficult or impractical to evaluate.

You now believe the majority prefers the packages with bundled runtimes.

> When both versions (with and without the runtime) are available, the preferred approach is to place the more commonly used or feature-complete version in the extras bucket, and the other in the versions bucket. However, it can be difficult to determine which one is more widely used. https://github.com/ScoopInstaller/Versions/issues/2515#issuecomment-3387483037

> We need a new rule to decide where self-contained versions should be placed.

So clearly the discussion in the previous issue doesn't come to an end, and we won't have a mutuality satisfactory solution until a maintainer could come here and give a further indication on this. As you see we still do have this new rule on which to place, where to place, and how to determine the wider and common usage. And neither you nor me here can provide an authoritative indication or statistic on which side overweighs.

z-Fng (approved):

z-Fng:
> I suggest either specifying the required .NET version, or switching to .NET self-contained releases.

I guess I didn't fully consider the implications when I merged #17288.

> We need a new rule to decide where self-contained versions should be placed.

This is still a highly controversial issue. Until a new consensus is reached, it's better to follow the historical practice of past maintainers. https://github.com/ScoopInstaller/Extras/pull/11132#issuecomment-1520138240 The self-contained one will be placed in the versions bucket instead.

PR #18229, opened 2026-07-05
mpv-git: Adopt timestamp-based versioning for intra-day builds

emotion3459:
# Summary
zhongfly releases new mpv-git builds twice per day. Because they are tagged as YYYY-MM-DD-SHA, versions produced within the same day are not monotonically increasing, causing Scoop's updater to sometimes ignore the second daily build.
This PR ensures a monotonically increasing version string to eliminate this edge case. Additionally, it should enable users on older shinchiro builds to auto-update to the new zhongfly builds without using `--force`.

### DIFF
--- a/bucket/mpv-git.json
+++ b/bucket/mpv-git.json
@@ -1,5 +1,5 @@
 {
-    "version": "2026-07-04-33111f3212",
+    "version": "202607041246",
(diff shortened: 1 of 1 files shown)


z-Fng (approved):
Thank you for your contribution!

PR #18231, opened 2026-07-05
kari: add version 0.0.0-test

Dhairya3391:
Terminal-based media browser and streamer with multi-provider search and external player support. See https://github.com/Dhairya3391/kari


Dhairya3391:
Closing until a production release is available. Will resubmit after v1.3.5+ is tagged.

PR #18327, opened 2026-07-17
vnote: Persist user data

### LINKED ISSUE #18227: [Improve]: VNote 4 portable
- VNote 4 supports portable mode. (VNote 3.x doesn't)
- If there is a folder named `config` next to the vnote.exe, VNote will start in portable mode.

Add the following lines in bucket/vnote.json
```
"persist": [
    "config"
]
```

l46983284-cpu:
Closes #18227

VNote 4 starts in portable mode when a `config` folder sits next to `vnote.exe` (confirmed by the author in vnotex/vnote#2699). The current Extras manifest has no `persist`, so upgrades would wipe that portable config.

Changes:
- `"persist": "config"` to keep the portable config across updates.
- `notes` with the author-confirmed one-time migration mapping: contents of `%APPDATA%\VNote` go to `config\app`, contents of `%LOCALAPPDATA%\VNote` go to `config\local`.

No version/hash/url change.


SorYoshino:
Will VNote automatically migrate user data? If not, please [implement a migration script](https://github.com/ScoopInstaller/Extras/blob/269e95e200d61a1e023d8150dfef94371554979b/bucket/qbittorrent.json#L19-L32), or inform users in the [`notes` field](https://github.com/ScoopInstaller/Extras/blob/269e95e200d61a1e023d8150dfef94371554979b/bucket/quicklook.json#L6-L9) to migrate their data manually.

alexchen-sys:
Yeah — VNote itself doesn’t auto-migrate non-portable settings into the portable config folder.

I added notes with the common AppData paths, plus a best-effort pre_install migrate (qbittorrent-style) that copies from those locations into `$persist_dir\config` only when that folder is still empty. If config is already present, it leaves it alone.

`persist` stays `"config"`. /verify is green on the latest head.

bucket/vnote.json
```diff
@@ -3,6 +3,12 @@
     "description": "A Vim-inspired note-taking platform",
     "homepage": "https://app.vnote.fun",
     "license": "LGPL-3.0-only",
+    "notes": [
+        "VNote does not auto-migrate non-portable settings into the portable `config` folder.",
+        "If you used VNote before this change, copy your old settings into the persist folder once:",
+        "from `%APPDATA%\\VNoteX\\VNote`, `%LOCALAPPDATA%\\VNoteX\\VNote`, `%APPDATA%\\vnote`, or `%LOCALAPPDATA%\\VNote`",
+        "to `$persist_dir\\config` (or let the install script migrate when that folder is empty)."
+    ],
```
SorYoshino, line 9: You only need to keep one of the `notes` fields and migration script.

Also, does VNote use all four of these paths to store user data, or does it store user data in only one of them?

I noticed that you used `break` in the migration script, which means that as long as there are any items in one of the folders, the script will stop migrating data from the other folders after migrating all the data from that folder.

alexchen-sys:
Thanks @SorYoshino — good catches.

I dropped the migration script and kept only the `notes` path (quicklook-style). VNote does not auto-migrate into the portable `config` folder next to `vnote.exe`.

On the paths: for a normal non-portable profile it is one Qt user config tree, typically `%APPDATA%\VNote\VNote` or `%LOCALAPPDATA%\VNote\VNote` (seen in upstream issues), not four simultaneous stores. I had over-listed guessed variants earlier, including a bad `VNoteX` path — notes now only mention those two common locations and say it is not both at once for one profile.

`persist` stays the single string `"config"`.

seaspeak:
From my testing here, the files under `%localappdata%/VNote` need to be moved to `config/local`, while `%appdata%/VNote` corresponds to `config/app`.

alexchen-sys:
Thanks for testing this. I updated the notes to map the two locations separately: `%LOCALAPPDATA%\VNote` to `config\local` and `%APPDATA%\VNote` to `config\app`.

seaspeak:
The VNote author confirmed the paths. see https://github.com/vnotex/vnote/issues/2699#issuecomment-5069893514

alexchen-sys:
Thanks for confirming with upstream — the notes already map `%LOCALAPPDATA%\VNote` → `config\local` and `%APPDATA%\VNote` → `config\app`.

alexchen-sys:
Cleaned up the notes to match the paths confirmed by the author: contents of `%APPDATA%\VNote` go to `config\app`, contents of `%LOCALAPPDATA%\VNote` go to `config\local`. Thanks @seaspeak for testing and confirming with upstream, and @SorYoshino for the review.

z-Fng (approved):

PR #18345, opened 2026-07-20
installerclean: Update to version 2.2.0, fix autoupdate, update license

FarmLox:
Two changes, both things Excavator cannot do on its own.

**Versioned download filenames.** From v2.2.0 the portable download carries the version: `InstallerClean-2.2.0-portable.exe`. The autoupdate template now tracks the versioned name, with a `#/InstallerClean-portable.exe` fragment so the download is saved under the same name as before and the `shortcuts` entry keeps working; the hash still comes from the `$url.sha256` sidecar published with each release. The concrete `version`, `url` and `hash` are bumped to 2.2.0 (released today), since Excavator cannot cross the rename by itself.

**License.** InstallerClean relicensed from MIT to Apache 2.0 in v2.1.0 (https://github.com/no-faff/InstallerClean/blob/main/LICENSE). Excavator keeps `version`, `url` and `hash` current but does not touch `license`, so this needs doing by hand.

### DIFF
--- a/bucket/installerclean.json
+++ b/bucket/installerclean.json
@@ -1,13 +1,13 @@
 {
-    "version": "2.1.0",
+    "version": "2.2.0",
     "description": "Open source tool to safely clean orphaned Windows Installer files and reclaim disk space.",
     "homepage": "https://github.com/no-faff/InstallerClean",
-    "license": "MIT",
+    "license": "Apache-2.0",
     "notes": "Requires administrator privileges to access C:\\Windows\\Installer.",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


FarmLox:
CodeRabbit has this backwards. InstallerClean relicensed from MIT to Apache 2.0, and GitHub's own licence detection for the repo now reports Apache-2.0. The LICENSE file is the Apache 2.0 text as of the v2.1.0 tag: https://github.com/no-faff/InstallerClean/blob/v2.1.0/LICENSE. The manifest saying MIT is the stale value, which is what this PR fixes.

PR #18371, opened 2026-07-23
glabels: Add version 3.99-master639

ibsorn:
Adds **glabels** — a designer for labels, business cards and barcodes, with a template database covering 35+ stationery vendors and document-merge from CSV.

gLabels is GPLv3 software by Jaye Evins and contributors. The Qt version (`glabels-qt`) builds fine on Windows but upstream publishes no Windows binaries, so this manifest points at an unofficial community build. The manifest's description and `notes` both state that plainly and send application bugs upstream rather than to the packager. Upstream has been notified the builds exist: https://github.com/j-evins/glabels-qt/issues/31#issuecomment-5055774938

The archive is produced by a public GitHub Actions workflow from unmodified upstream sources — [one workflow file](https://github.com/ibsorn/glabels-windows/blob/main/.github/workflows/build-windows.yml), no vendored code. Every build runs the upstream unit tests and then renders a PDF from the *packaged* tree, so an archive that does not actually run never gets published.

Manifest notes:

- Uses the portable ZIP, which needs no installer. The application locates its templates as `../share/glabels-qt` relative to the executable, so `extract_dir` keeps the archive's own `bin/` + `share/` layout intact.
- Only the CLI renderer `glabels-batch-qt` is shimmed into `bin`; the GUI gets a shortcut.
- `checkver` reads the release tag from the GitHub API, and `autoupdate` takes hashes from the `SHA256SUMS.txt` published alongside every release.
- Validated against `schema.json`; `scoop install`/`scoop uninstall` conventions follow the existing portable-archive manifests in this bucket.

### DIFF
--- a/bucket/glabels.json
+++ b/bucket/glabels.json
@@ -1,0 +1,45 @@
+{
+    "version": "3.99-master639",
+    "description": "Designer for labels, business cards and barcodes, with a large template database (unofficial Windows build of gLabels 4)",
+    "homepage": "https://github.com/ibsorn/glabels-windows",
+    "license": "GPL-3.0-or-later",
+    "notes": [
+        "These are unofficial community builds of gLabels 4 (glabels-qt), which upstream does not distribute for Windows.",
+        "gLabels 4 is still in development: treat it as pre-release software and keep backups of your projects.",
+        "Report application bugs at https://github.com/j-evins/glabels-qt/issues and packaging problems at https://github.com/ibsorn/glabels-windows/issues"
+    ],
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


ibsorn:
Closing this myself — the `not-meet-criteria` label is right, and I should have checked the package-request criteria before opening a PR rather than after.

gLabels clears the popularity bar comfortably (the upstream project has ~470 stars and the application ships in the Debian, Fedora, Ubuntu and Arch repositories), but it plainly fails **"Latest stable version"**: gLabels 4 has no stable release yet, and what this manifest packages is a snapshot of upstream's `master`. That criterion is marked required, so there is nothing to argue about here and no reason to spend a maintainer's time on it.

I'll open a package request and submit again once upstream tags 4.0. Apologies for the noise, and thanks for the clearly written criteria — they made it easy to work out that this did not belong here.

PR #18378, opened 2026-07-25
gohide: Add version 1.1.10

zxwdev:
Adds `gohide` 1.1.10 to the Extras bucket.

gohide is a Windows utility that hides the running apps you choose — their windows, taskbar buttons
and system-tray icons — with a single global hotkey, then restores everything exactly where it was.
It can also mute a hidden app or suspend it so a hidden game stops using CPU and GPU. Freeware for
up to 3 managed apps (no time limit, no account, no telemetry, fully offline); an optional Pro tier
is a one-time purchase. I am the developer.

- Homepage: https://gohide.net
- Source of binaries: https://github.com/gohide/GoHide/releases

Notes:

- The app ships as a Velopack package, so the manifest installs the portable `.nupkg` payload
  (`extract_dir: lib\app`) rather than running the Setup.exe bootstrapper — this keeps the install
  fully contained in the Scoop app directory, same approach as `dyad`, `folo` and `noi`.
- `autoupdate.hash` reads the SHA-1 from the Velopack/Squirrel `RELEASES` file published alongside
  each release.
- Tested end to end on Windows 11: `scoop install` downloads, passes the hash check, extracts, and
  creates a working `gohide.exe` shim/shortcut.

### DIFF
--- a/bucket/gohide.json
+++ b/bucket/gohide.json
@@ -1,0 +1,39 @@
+{
+    "version": "1.1.10",
+    "description": "Hides app windows, taskbar buttons and system-tray icons with a global hotkey, then restores them exactly where they were.",
+    "homepage": "https://gohide.net",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://gohide.net/en/legal/terms"
+    },
+    "notes": [
+        "gohide needs administrator rights to hide other apps' windows and tray icons.",
+        "Free for up to 3 managed apps; Pro is a one-time purchase (https://gohide.net/en/pricing)."
+    ],
+    "architecture": {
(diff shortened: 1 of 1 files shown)


zxwdev:
Closing this myself — I should have checked the [package-request criteria](https://github.com/ScoopInstaller/Extras/issues/new?template=package-request.yml) before opening a PR rather than after. gohide's repo is at 0 stars/0 forks, well under the "reasonably well-known and widely used" bar, so this is out of scope for Extras today.

Apologies for the noise. I'll ship it in my own bucket for now and revisit if the project ever clears the popularity bar.

PR #18388, opened 2026-07-26
subtitleedit@5.1.0: Persist more files

SiriosDev:
Persisting Dictionaries and All other user customizable folders


SiriosDev:
ok, now should be all, I decided not to mimic `persist` by making a bulk copy of all the folders using pre- and post-scripts, since they can contain several gigabytes of files that would be duplicated this way.
Maybe I could do a bulk shim via script, but I’m not sure if that goes against Scoop’s best practices—let me know if you think that might be a good idea.)

z-Fng (approved):

PR #18007, opened 2026-06-08
aegisub(-arch1t3cht): Persist `automation` folder

arvdk:
Closes #17957

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #18490, opened 2026-08-09
aegisub@3.4.2: Persist more files

SiriosDev:
Dynamically add to the persist all JSON configuration files that are placed by automations directly in the app's root directory

### DIFF
--- a/bucket/aegisub.json
+++ b/bucket/aegisub.json
@@ -18,7 +18,7 @@
     "extract_dir": "aegisub-portable",
     "pre_install": [
         "# Hardlinks won't work properly here to persist files, because this app creates a new file instead of writing to the existing one.",
-        "Get-ChildItem -LiteralPath \"$persist_dir\" -File | Where-Object { $_.Name -notin @('manifest.json', 'install.json') } | ForEach-Object {",
+        "Get-ChildItem -LiteralPath \"$persist_dir\" -File -Filter '*.json' | Where-Object { $_.Name -notin @('manifest.json', 'install.json') } | ForEach-Object {",
         "    if (Test-Path -LiteralPath \"$persist_dir\\$($_.Name)\" -PathType Leaf) {",
         "        Copy-Item -Path \"$persist_dir\\$($_.Name)\" -Destination \"$dir\\$($_.Name)\" -Force",
         "    }",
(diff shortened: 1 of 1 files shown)


bucket/aegisub.json
```diff
@@ -18,9 +18,11 @@
     "extract_dir": "aegisub-portable",
     "pre_install": [
         "# Hardlinks won't work properly here to persist files, because this app creates a new file instead of writing to the existing one.",
-        "'config.json' | ForEach-Object {",
-        "    if (Test-Path -LiteralPath \"$persist_dir\\$_\" -PathType Leaf) {",
-        "        Copy-Item -Path \"$persist_dir\\$_\" -Destination \"$dir\\$_\" -Force",
+        "Test-Path -LiteralPath \"$persist_dir\" {",
+        "    Get-ChildItem -LiteralPath \"$persist_dir\" -File -Filter '*.json' | ForEach-Object {",
```
z-Fng, line 22: 1. Have you tested this manifest? To my knowledge, there seems to be a syntax error.
2. I'd recommend explicitly listing the files that need to be persisted, although both explicit listing and using wildcards have their own pros and cons. https://github.com/ScoopInstaller/Extras/blob/48ef736f7bcb43f0eb57a7a8076052308f4d09a7/bucket/beyondcompare.json#L73
3. Please check if [`aegisub-arch1t3cht`](https://github.com/ScoopInstaller/Extras/blob/master/bucket/aegisub-arch1t3cht.json) also needs a similar update like, as it is a fork. See: #18007

z-Fng (changes requested):

bucket/aegisub.json
SiriosDev: 1. Oh yeah, I don't know why I was convinced the test path included the “if” statement 🤦
2. I could, but these JSON files are the configuration files for automations that don't use the designated path and instead place their config files in the root of the package. I could include them manually one by one, but there would be a risk of filename changes across different versions of the automations, or even just the risk of including only the ones I need and not all possible ones.
3. I definitely wanted to try it, but the current build crashes on my setup

SiriosDev: done

z-Fng: > 2\. I could, but these JSON files are the configuration files for automations that don't use the designated path and instead place their config files in the root of the package. I could include them manually one by one, but there would be a risk of filename changes across different versions of the automations, or even just the risk of including only the ones I need and not all possible ones.

Oh, I see.

> Dynamically add to the persist all JSON configuration files that are placed by automations directly in the app's root directory

I didn't fully understood what that meant at that time.

SiriosDev:
Wait, 

Edit:

okay, never mind, the email notification was formatted incorrectly and I noticed a problem, but everything looks fine in the web UI

z-Fng (approved):

PR #18542, opened 2026-08-17
psysonic: Update to version 1.51.0, fix checkver

Psychotoxical:
The upstream repository moved from `Psychotoxical/psysonic` to `Psysonic/psysonic` (organization transfer). All URLs updated; the old ones still redirect, so nothing was broken.

Also fixes autoupdate, which is why this manifest sat at 1.50.0: `checkver` scraped the repository landing page for `Psysonic v([\d.]+)`, and that string is no longer present in its HTML. It now reads `/releases/latest` and matches the project's `app-v` tag prefix, which also keeps release candidates out.

Checksum is GitHub's own digest for the published installer.

### DIFF
--- a/bucket/psysonic.json
+++ b/bucket/psysonic.json
@@ -24,7 +24,7 @@
     ],
     "checkver": {
         "github": "https://github.com/Psysonic/psysonic",
-        "regex": "releases/tag/app-v([\\d.]+)"
+        "regex": "app-v([\\d.]+)"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


Psychotoxical:
/verify

The regex was written against the releases page, but GitHub mode sets `jsonpath: $.tag_name` in API mode (cbd7c64), so it now matches the tag name directly. `app-v([\d.]+)` resolves to `1.51.0` on both the API and the HTML path.

PR #18545, opened 2026-08-18
raspberry-pi-imager: Fix registry path

brian6932:
I'm unsure of whether or not I made a mistake when I wrote this in the past, or if the actual registry path changed.

### DIFF
--- a/bucket/raspberry-pi-imager.json
+++ b/bucket/raspberry-pi-imager.json
@@ -5,7 +5,7 @@
     "license": "Apache-2.0",
     "notes": [
         "To opt out of telemetry, run:",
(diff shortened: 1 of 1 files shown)


z-Fng:
> I'm unsure of whether or not I made a mistake when I wrote this in the past, or if the actual registry path changed.

- https://github.com/ScoopInstaller/Extras/pull/16988

Due to an upstream ApplicationName change (raspberrypi/rpi-imager@3b4052a), the 2.x registry path differs from 1.x. #16988 specifically fixed this. The existing registry path is already correct. 

image

brian6932:
Ah alright, didn't know it was changed, should've ran a blame first.

PR #18431, opened 2026-07-31
krokiet@12.0.1: Switch to skia_opengl variant

julianrendell:
Krokiet stopped working for me, and on checking the source repo releases page, the developer recommends the `windows_krokiet_on_windows_skia_opengl.exe` variant over the `windows_krokiet_on_linux.exe` one.

The skia variant works for me 😄

PR #18622, opened 2026-08-27
krokiet@12.0.1: Switch to all backends asset

Vinfall:
Follow-up of https://github.com/ScoopInstaller/Extras/pull/18431#pullrequestreview-4875641601.

Please note that the quoted FAQ section itself is outdated. Currently the exe in `windows_krokiet_on_windows_all_backends.zip` can detect backend automatically, thus having better compatibility than the recommended `windows_krokiet_on_linux.exe`. Because of this, I think it's fine to not create multiple shortcuts.

Supersedes #18431
Relates to https://github.com/qarmin/czkawka/issues/2055

### DIFF
--- a/bucket/krokiet.json
+++ b/bucket/krokiet.json
@@ -5,13 +5,13 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/qarmin/czkawka/releases/download/12.0.1/windows_krokiet_on_linux.exe",
-            "hash": "084fa6f9d97574704a5104d207b8805e510f3f1c9907e7828607e7fce6a67abf"
+            "url": "https://github.com/qarmin/czkawka/releases/download/12.0.1/windows_krokiet_on_windows_all_backends.zip",
+            "hash": "13f68c9b29b7667caf913c1223dbcdb58e9e51b8a050e00ed1f1fd31508d17b5"
         }
     },
     "shortcuts": [
         [
-            "windows_krokiet_on_linux.exe",
+            "krokiet.exe",
             "Krokiet"
         ]
     ],
@@ -19,7 +19,7 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/qarmin/czkawka/releases/download/$version/windows_krokiet_on_linux.exe"
+                "url": "https://github.com/qarmin/czkawka/releases/download/$version/windows_krokiet_on_windows_all_backends.zip"
             }
         }
     }
--- a/bucket/krokiet.json
+++ b/bucket/krokiet.json
@@ -3,6 +3,9 @@
     "description": "Krokiet is the Slint frontend for Czkawka, helping you remove unnecessary files from your computer.",
     "homepage": "https://github.com/qarmin/czkawka",
     "license": "MIT",
+    "suggest": {
+        "vcredist": "extras/vcredist2022"
+    },
     "architecture": {
         "64bit": {
             "url": "https://github.com/qarmin/czkawka/releases/download/12.0.1/windows_krokiet_on_windows_all_backends.zip",


bucket/krokiet.json
```diff
@@ -5,21 +5,22 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/qarmin/czkawka/releases/download/12.0.1/windows_krokiet_on_linux.exe",
-            "hash": "084fa6f9d97574704a5104d207b8805e510f3f1c9907e7828607e7fce6a67abf"
+            "url": "https://github.com/qarmin/czkawka/releases/download/12.0.1/windows_krokiet_on_windows_all_backends.zip",
+            "hash": "13f68c9b29b7667caf913c1223dbcdb58e9e51b8a050e00ed1f1fd31508d17b5"
```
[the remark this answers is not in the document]
Vinfall: Looks like false positive, https://github.com/qarmin/czkawka/releases/download/12.0.1/deps_windows.bat only suggests ffmpeg and libheif/libraw/libavif as optional dependencies, which users can install on their own and is not required to use most functions.

z-Fng:
> Currently the exe in `windows_krokiet_on_windows_all_backends.zip` can detect backend automatically, thus having better compatibility than the recommended `windows_krokiet_on_linux.exe`.

Are you sure about this? The backend seems to be controlled by the `SLINT_BACKEND` environment variable, with `winit-femtovg` being the default, which is also the only difference among all these backend `.bat` files.

> https://github.com/qarmin/czkawka/blob/master/instructions/FAQ.md#q-krokiet-window-is-completely-black--blank--invisible
Try a different build / rendering backend. The default backend is meant to be the most compatible, but no single backend works with every system and GPU driver. Download the all_backends build and start it through its per-backend launcher scripts (or set SLINT_BACKEND yourself) until one renders correctly.

Has this FAQ guide become outdated as well?

Edit: I still stand by my original point: for better compatibility, we should keep the existing `windows_krokiet_on_linux.exe`, add a windows_krokiet_on_windows_all_backends asset, and create shortcuts for the various backends.

Gitoffthelawn:
At this time, I'm not sure of the best solution, but I can provide additional data points:  
 - Until recently, Krokiet (installed via Scoop) worked fine for me.
 - Like what @julianrendell experienced (as reported in #18431), Krokiet recently stopped working. I'm not sure of the exact date, but I think it was with the v12.0.0 or v12.0.1 update.
 - By "stopped working", I mean that it crashes on launch. No window appears and the Windows Event Log confirms crash on launch.
 - Like what @julianrendell reported, I found that manually installing `windows_krokiet_on_windows_skia_opengl.exe` fixes that crash.
 - The current (v12.0.1) release notes lists `windows_krokiet_on_windows_skia_opengl.exe` first:  
    > If you are a beginner or don't know which version to use, start with one of these depending on your OS (if they don’t work, try different builds):
    >    Windows – windows_krokiet_on_windows_skia_opengl.exe, windows_krokiet_on_linux.exe, windows_czkawka_gui_gtk_412.zip

HTH

Vinfall:
> Are you sure about this?

Yes, it's a feature of Slint (the underlying GUI framework), not krokiet-specific, see [Winit Backend](https://docs.slint.dev/latest/docs/slint/guide/backends-and-renderers/backend_winit/) and [Backends & Renderers](https://docs.slint.dev/latest/docs/slint/guide/backends-and-renderers/backends_and_renderers/).

> Has this FAQ guide become outdated as well?

Nope, I mean the claim in release notes that `windows_krokiet_on_linux.exe` is preferred over `windows_krokiet_on_windows_all_backends.zip` is outdated.

I'm open to creating multiple shortcuts (although it's kinda redundant due to the fallback feature above), but what's the point of keeping `windows_krokiet_on_linux.exe`? It does NOT have better compatibility. `windows_krokiet_on_windows_all_backends.zip` is the most compatible one out there because it, as the name suggests, includes everything.

> I found that manually installing windows_krokiet_on_windows_skia_opengl.exe fixes that crash.

Yeah, but if you try the exe in `windows_krokiet_on_windows_all_backends.zip`, it would work as well. And it also works inside VM, while `windows_krokiet_on_windows_skia_opengl.exe` doesn't. The only down side of switching to all backends asset is it takes up 20M more space, that's it. And even that can be justified because, ironically, the _download size_ is actually smaller...

> The current (v12.0.1) release notes lists windows_krokiet_on_windows_skia_opengl.exe first

Like I said in the PR desc, this is outdated. The developer suggests you to try that first because it's mostly supposed to work so you don't have to choose between different backends. But it does not work sometimes (e.g. inside VM) while all backends always work. Thus, in the scope of scoop manifest, _the_ better option is all backends one.

Gitoffthelawn:
@Vinfall To get the `all_backends` version working in a VM, are you using one of the included batch files, or running `krokiet.exe` (from `windows_krokiet_on_windows_all_backends.zip`) directly?  What about outside of a VM?

z-Fng:
> > Are you sure about this?
> 
> Yes, it's a feature of Slint (the underlying GUI framework), not krokiet-specific, see [Winit Backend](https://docs.slint.dev/latest/docs/slint/guide/backends-and-renderers/backend_winit/) and [Backends & Renderers](https://docs.slint.dev/latest/docs/slint/guide/backends-and-renderers/backends_and_renderers/).

Got it. I'm not a Krokiet user myself and was just going off the docs, so I don't know the exact inner workings.

> I'm open to creating multiple shortcuts (although it's kinda redundant due to the fallback feature above), but what's the point of keeping `windows_krokiet_on_linux.exe`? It does NOT have better compatibility. `windows_krokiet_on_windows_all_backends.zip` is the most compatible one out there because it, as the name suggests, includes everything.

Suggesting we keep the original `windows_krokiet_on_linux.exe` and add multiple shortcuts was just to be safe with compatibility in case switching to all_backends broke someone's setup.

If you're sure it won't be an issue, sounds good to me!

bucket/krokiet.json
```diff
@@ -5,21 +5,22 @@
...
         }
     },
+    "bin": "krokiet.exe",
```
z-Fng: One more thing, does it accept any command-line arguments?

> If the program file is a GUI application and it doesn't accept any commandline arguments, no need to add it in bin.

Vinfall:
@Gitoffthelawn Just run it directly. If this sounds too hard to understand, you may think this way: this is a C#/.NET project, `windows_krokiet_on_windows_skia_opengl.exe` is the minimal executable that would work if you have installed the runtime, and `windows_krokiet_on_windows_all_backends.zip` is the portable zip that bundles executable and runtime altogether so it always works.

bucket/krokiet.json
Vinfall: Nope, fixed now.

Gitoffthelawn:
@Vinfall Thanks. That's a great explanation. In which VM did `windows_krokiet_on_windows_skia_opengl.exe` fail for you?

Based on what you've written, I agree that `windows_krokiet_on_windows_all_backends.zip` is likely the best option for the Scoop Extras bucket.  I'm not sure if I'll use it or create a private manifest to use `windows_krokiet_on_windows_skia_opengl.exe`, but like you said, if `windows_krokiet_on_windows_all_backends.zip` offers the greatest compatibility (albeit with a 20MB penalty), I think it's the best choice for the Scoop general audience.

If we want, we can also publish a manifest for `windows_krokiet_on_windows_skia_opengl.exe` to https://github.com/ScoopInstaller/Versions

Vinfall:
> In which VM did windows_krokiet_on_windows_skia_opengl.exe fail for you?

Windows Server 2025 inside VirtualBox with the VBoxSVGA graphics controller. This is actually expected due to the limiteed OpenGL support. I think it's possible to use it after installing mesa3d-mingw but this VM does not have that. Either way, this is going to be off-topic :(

z-Fng (approved):

Gitoffthelawn:
For those interested, I performed some research on the topic:  https://github.com/qarmin/czkawka/issues/2056

Vinfall:
I'm puzzled now that the variants that should work do not work but the variant that should not work actually works. What to do? By all means, the previous `windows_krokiet_on_linux.exe` surely no longer works, so we have these options:
1. Switch to `windows_krokiet_on_windows_skia_opengl.exe`, but honestly speaking I don't know why it _sometimes works_
2. Keep `windows_krokiet_on_windows_all_backends.zip`, add multiple shortcuts along with a note suggesting users to test all variants. This works for me but not for @ Gitoffthelawn, and I think users might be puzzled as well.
3. Keep `windows_krokiet_on_windows_all_backends.zip` and add `windows_krokiet_on_windows_skia_opengl.exe`, add multiples shortcuts, defaults to `windows_krokiet_on_windows_skia_opengl.exe`. This makes everyone happy by bundling everything, but it's quite chaotic.
4. Wait for the upstream to debug this, and discuss it later.

PR #18665, opened 2026-09-03
javascriptobfuscator: Add version 3.4.5

### LINKED ISSUE #18666: [Request]: JavaScript Obfuscator
### Name

JavaScript Obfuscator

### Description

A JavaScript source-protection tool: it renames identifiers, encodes strings,
flattens control flow and compresses output so shipped JavaScript is hard to read
or reuse. The Windows app protects source locally — files never leave the machine
— and bundles `jso-local`, a CLI front end to the same engine for build scripts
and CI.

### Homepage

https://javascriptobfuscator.com/

### Download Link

https://javascriptobfuscator.com/download/javascriptobfuscator-3.4.5.zip

This is a permanent, versioned URL published specifically so a manifest can pin a
hash. The vendor also serves a mutable `/download/javascriptobfuscator.zip` that
always holds the current release; the manifest deliberately does not use it.
Checksums for every published archive are at
https://javascriptobfuscator.com/download/SHA256SUMS.txt

This is a commercial product, not a GitHub project, so the stars/forks guideline
does not apply. What I can offer instead is externally verifiable:

- **Continuously published since 2004.** The earliest Internet Archive capture of
  the domain is 30 November 2004, and the site has 213 monthly-deduplicated
  captures across 22 years:
  https://web.archive.org/web/20041130013750/http://javascriptobfuscator.com/
- The two user-facing binaries are Authenticode signed as `richscripts inc`.
- The same publisher ships the engine on npm (`jso-protector`,
  `javascriptobfuscator-com`, and four bundler plugins) under the npm account
  `richscripts`.
- A winget submission for the same build is in review and has passed Microsoft's
  URL, policy, catalog and installer-scan validation:
  https://github.com/microsoft/winget-pkgs/pull/428655
- A Chocolatey submission for the same build is in moderation.

I want to be straightforward about the weak spot: I do not have a public download
counter or a store rating to point at, and npm installs for the CLI are currently
small. If long-running commercial distribution plus signed binaries plus a
hash-pinnable permanent URL is not enough for Extras, I would rather you say so
and close this than leave it sitting.

### Criteria

- [x] English interface and English documentation
- [x] Latest stable version (3.4.5)
- [ ] Full version — free to install and usable on a permanent free tier (not a
      time-limited trial), but stronger presets and larger inputs are paid.
- [x] Fairly standard install — a zip at a version-specific URL, extracted with
      no pre/post install scripting beyond suppressing shims for three runtime
      helper executables.

### Related PR

https://github.com/ScoopInstaller/Extras/pull/18665

richtexteditor:
Adds `javascriptobfuscator` 3.4.5 — a commercial JavaScript source-protection
tool with a Windows desktop client (shipping since 2004) and a bundled CLI.

- `javascriptobfuscator.exe` — desktop app (x64), also registered as a shortcut
- `cli\jso-local.exe` — CLI, exposed as `jso-local`

Verified before opening this PR:

- `hash` was produced by downloading the URL and hashing the bytes. It matches
  the vendor's own published manifest at
  https://javascriptobfuscator.com/download/SHA256SUMS.txt
- The download URL is a permanent versioned path. The vendor publishes it
  specifically so a pinned hash stays valid; the mutable
  `/download/javascriptobfuscator.zip` was deliberately not used.
- `checkver` was run against the live page and returns exactly `3.4.5` (it does
  not pick up the separate legacy WinForms archive on the same page).
- The `autoupdate` hash regex was run against the live `SHA256SUMS.txt` and
  returns the same hash as the one in this manifest.
- Both `bin` paths were confirmed present inside the archive.
- Binaries are Authenticode signed as `CN=richscripts inc`.

Note: the app is proprietary and free to install, with paid tiers for larger
inputs and stronger protection. It runs without an account on a free tier.

### DIFF
--- a/bucket/javascriptobfuscator.json
+++ b/bucket/javascriptobfuscator.json
@@ -1,0 +1,43 @@
+{
+    "version": "3.4.5",
+    "description": "Obfuscate and protect JavaScript source locally, with a desktop app and a CLI.",
+    "homepage": "https://javascriptobfuscator.com/",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://javascriptobfuscator.com/terms.aspx"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://javascriptobfuscator.com/download/javascriptobfuscator-3.4.5.zip",
+            "hash": "e693523c9bc8d84becc67820404dc166915d24b23e3c12dd504669353c84d8bd"
+        }
+    },
+    "bin": [
+        "javascriptobfuscator.exe",
+        [
+            "cli\\jso-local.exe",
+            "jso-local"
+        ]
+    ],
+    "shortcuts": [
+        [
+            "javascriptobfuscator.exe",
+            "JavaScript Obfuscator"
+        ]
+    ],
+    "checkver": {
+        "url": "https://javascriptobfuscator.com/downloads.aspx",
+        "regex": "javascriptobfuscator-([\\d.]+)\\.zip"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://javascriptobfuscator.com/download/javascriptobfuscator-$version.zip"
+            }
+        },
+        "hash": {
+            "url": "https://javascriptobfuscator.com/download/SHA256SUMS.txt",
+            "regex": "$sha256\\s+javascriptobfuscator-$version\\.zip"
+        }
+    }
+}


richtexteditor:
Package request opened as requested: #18666

It sets out what I can actually evidence (continuous publication since 2004 with 213 Internet Archive captures, Authenticode-signed binaries, a hash-pinnable permanent download URL, and parallel winget/Chocolatey submissions), and is equally direct about what I cannot: there is no public download counter or store rating to point at, and npm installs for the CLI are currently small. If that does not clear the Extras bar, please say so and close both — I would rather have a clear no than an open PR.

PR #18753, opened 2026-09-14
deskstitch: Add version 0.7.7

jonseoane:
Adds DeskStitch, a software KVM for two Windows PCs. One keyboard and mouse drive both machines and the cursor crosses between their screens. Unlike the other tools in this category it places monitors by their physical size in millimetres rather than by pixels, so a 4K panel next to a 1440p panel of the same size lines up exactly.

- Homepage: https://deskstitch.com
- Release: https://github.com/jonseoane/deskstitch-releases/releases/tag/v0.7.7

Notes on the manifest:

- Uses the portable zip, which contains `DeskStitch.exe` and a README at the root. I downloaded it and listed the archive rather than assuming the layout.
- Both hashes were verified by downloading the archives and hashing them; they match the publisher's published `SHA256SUMS.txt`.
- `persist` covers `config.json` and `layout.json`, which the portable build keeps next to the executable, so settings and the saved desk layout survive an update.
- `autoupdate` reads hashes from the `SHA256SUMS.txt` asset attached to each release, so it is version-pinned rather than pointing at a moving file.
- Proprietary and paid: free for seven days on the PC with the keyboard and mouse, then a one-time licence. The second PC never needs a key. Flagged in `notes` so nobody is surprised after installing.

I am the publisher. Happy to adjust anything to fit the bucket's conventions.


z-Fng:
> See the [package request form](https://github.com/ScoopInstaller/Extras/issues/new?labels=package-request&template=package-request.yml&title=%5BRequest%5D%3A+), [contribution guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md), and [manifest wiki](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests).

This package doesn't fully meet the [essential criteria](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket). Please consider creating [your own bucket](https://github.com/ScoopInstaller/Scoop/wiki/Buckets#creating-your-own-bucket) instead.

PR #18768, opened 2026-09-17
qgroundcontrol: Update to version 5.1.4, add arm64 support

mdiamantino:
Updates `qgroundcontrol` from 5.0.8 to 5.1.4.

Upstream release: https://github.com/mavlink/qgroundcontrol/releases/tag/v5.1.4

Hashes were computed from the files downloaded from the URLs in this manifest.


aliesbelik (changes requested):
- Please fix `autoupdate` template as well.
- Add `arm64` option along with 64bit.

z-Fng (approved):

z-Fng:
> * Please fix `autoupdate` template as well.
> * Add `arm64` option along with 64bit.

@aliesbelik  Sorry, didn't notice there was already an existing review.

@mdiamantino Scoop relies on [automation](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate) to keep manifests up to date. Simply bumping the version is only a partial fix and would still require manual updates down the road, so please make sure to fix both checkver and autoupdate as well. Thanks for taking the time to contribute anyway!

PR #18781, opened 2026-09-19
pscompletions: Move to main bucket

abgox:
See https://github.com/ScoopInstaller/Main/pull/8528
For the migration: https://github.com/ScoopInstaller/Main/issues/7270#issuecomment-5741616115


abgox:
- @z-Fng 
-  I delete it directly based on https://github.com/ScoopInstaller/Main/issues/7270#issuecomment-5741616115
- Or move it to the deprecated directory instead?

z-Fng:
> * I delete it directly based on [[Request]: PSCompletions Main#7270 (comment)](https://github.com/ScoopInstaller/Main/issues/7270#issuecomment-5741616115)
> * Or move it to the deprecated directory instead?

> If you are confident that this is feasible, most users can be migrated smoothly, and it's worth the effort, please feel free to open a PR following the process outlined above.
> 
> ...
> > * [opencode: Deprecate manifest](https://github.com/ScoopInstaller/Extras/commit/ce96a0a47163c77ecf365685b46654d53c20145a)

Following existing conventions, it would be better to move this to the deprecated directory.

z-Fng (approved):

z-Fng:
Note: 
> https://github.com/ScoopInstaller/Main/issues/7270#issuecomment-5741616115
> * If users cannot get updates via Scoop while the module keeps telling them a new version exists, they'll naturally assume the manifest was migrated.

Per the package owner's preference, we are dropping this manifest immediately instead of following the usual deprecation period.

